PR #28, opened 2014-03-11
Fixed some level 4 warnings in VS2013

jdale88:
Added casts for the simple stuff like implicit type narrowing.
Disabled the constant conditional expression warning.

### DIFF
--- a/format.cc
+++ b/format.cc
@@ -633,7 +633,7 @@ void fmt::BasicFormatter<Char>::DoFormat() {
       } else {
         out = writer.GrowBuffer(1);
       }
-      *out = static_cast<char>(arg.int_value);
+      *out = static_cast<Char>(arg.int_value);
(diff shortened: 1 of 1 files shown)


vitaut:
@jdale88, could you please review the updated [CONTIBUTING](https://github.com/fmtlib/fmt/blob/ebec00138bd10e949ff34955bcb096387cd32acc/CONTRIBUTING.md) document, particularly the part about licensing, and let me know if you agree with it being applied to your contributions to {fmt}? The library is likely to be relicensed (#1073) so I'm collecting approval from all earlier contributors. Thanks!

jdale88:
I'm fine with this, as long as the contribution was still legally made back in 2014 and my acceptance of your new license doesn't re-do that contribution in 2019. If it does, then no, you should remove it.

Issue #86, opened 2014-12-31
Overload `operator <<` for `std::wostream` doesn't work


aruhan:
Overload `operator <<` for `std::wostream` doesn't work.

````cpp
class Date {
  int year_, month_, day_;
 public:
  Date(int year, int month, int day) : year_(year), month_(month), day_(day) {}

  int year() const { return year_; }
  int month() const { return month_; }
  int day() const { return day_; }

  friend std::wostream &operator<<(std::wostream &os, const Date &d) {
    os << d.year_ << L'-' << d.month_ << L'-' << d.day_;
    return os;
  }
};

std::wstring s = format(L"The date is {0}", Date(2012, 12, 9));
EXPECT_EQ(L"The date is 2012-12-9", s);
````

RESULT:

````sh
>>> error: Value of: s
>>>   Actual: L"The date is 2\00\01\02\0-"
>>>   Expected: L"The date is 2012-12-9"
````

PR #87, opened 2015-01-05
Fix format using wostream produces wrong results

aruhan:
A fix for #86


vitaut:
Thanks for the fix! Merged and added a test case (https://github.com/cppformat/cppformat/commit/cd828a869165240bd2afd56482692b16cbe9f84c).

aruhan:
I'm glad to contribute to this nice library.

Thanks.

PR #89, opened 2015-01-06
Add format linking to tests

mparaniak:
On my system (Arch Linux) attempt to compile tests ends up in linkage failure. The proposed patch adds proper library to CMakeLists.txt.


vitaut:
Merged, thanks!

PR #91, opened 2015-01-21
fix breathe submodule ref

stev47:
the ref vitaut/breathe@66ebb90f6a6cbe3c7a58fca983bf5804695ab064 was missing when cloning the repo and trying to init/update the submodules.
used vitaut/breathe@505cd3fd0c87b1fce38cf05401ea2e0a8e8b0deb instead which exists in the master branch.


vitaut:
Thanks for catching this. https://github.com/vitaut/breathe is a temporary fork which will be deleted once its content is merged into mainline, so I switched to the main repo https://github.com/michaeljones/breathe instead: https://github.com/cppformat/cppformat/commit/531ee615af27e349ccbcc25f65b0c27f713126b9.

Issue #96, opened 2015-02-06
warnings under Xcode 6.1.1 (probably with -Weverything)


dixlorenz:
commit  ba17cca432416c6df7c862fc38c4c7a5905d0bc7, 2/6/2015:

format.h:
Unknown Command Tag Name:
line #142: /rst
line #164: \endrst
line #1567: out.write("Current point:\n");
the last one ist about "\n"; this might well be an Xcode bug

format.cpp:
Semantic issue, 'return' will never be executed
line #254, line #276
these returns are after an unconditional FMT_THROW

warning: comparison of integers of different signs: 'const int' and 'std::size_t' (aka 'unsigned long')
line #628: if (spec.precision_ >= 0 && spec.precision_ < str_size)
spec.precision_ and str_size are the problem.

dixlorenz:
one more: in Result visit(const Arg &arg) clang complains about the fallthrough in release mode and suggests inserting [[clang::fallthrough]]. This gives a warning in debug mode though, because after the assert(false) it is unreachable code.

vitaut:
Reopening until the fallthrough warning is resolved.

vitaut:
@dixlorenz Could you please check if the fix in branch [fallthrough](https://github.com/cppformat/cppformat/tree/fallthrough) resolves the fallthrough warning? I tried detecting the availability of `[[clang::fallthrough]]` based on clang version there (you might need to change `600` to your version in https://github.com/cppformat/cppformat/blob/fallthrough/format.h#L124 if it is lower).

dixlorenz:
Almost perfect. It resolves the fallthrough warning, but in debug build I now get

warning: fallthrough annotation in unreachable code

because of the assert(false)... Somewhat interestingly (to me): this code

    default:
      assert(false);
      return FMT_DISPATCH(visit_int(arg.int_value));
    case Arg::INT:
      return FMT_DISPATCH(visit_int(arg.int_value));

does not give a warning about unreachable code (or any other).

That might be the easiest/cleanest solution. It does have the same semantics as the fallthrough and doesn't need macros, #ifdefs or rely on the compiler or its version at all.

vitaut:
So does [this commit](https://github.com/cppformat/cppformat/commit/a92c179a6dfc5037d0c191ddfabb1c3fe12f1439) fixes the warnings? (I used your suggestion but simplified it a bit since there is no need to dispatch there.) If it works I'll merge it into master and remove the fallthrough attribute detection.
Thanks!

[...]

PR #97, opened 2015-02-06
fix compile warnings for Xcode

dixlorenz:
for me issue #96 is resolved with these changes, no more warnings

### DIFF
--- a/format.cc
+++ b/format.cc
@@ -66,8 +66,10 @@ using fmt::internal::Arg;
 #ifndef FMT_THROW
 # if FMT_EXCEPTIONS
 #  define FMT_THROW(x) throw x
+#  define FMT_RETURNAFTERTHROW(x)
 # else
 #  define FMT_THROW(x) assert(false)
+#  define FMT_RETURNAFTERTHROW(x) return x
 # endif
 #endif
 
@@ -251,7 +253,7 @@ class WidthHandler : public fmt::internal::ArgVisitor<WidthHandler, unsigned> {
 
   unsigned visit_unhandled_arg() {
     FMT_THROW(fmt::FormatError("width is not integer"));
-    return 0;
+    FMT_RETURNAFTERTHROW(0);
   }
 
   template <typename T>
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR, but unfortunately it breaks the OS X release build: https://travis-ci.org/cppformat/cppformat/jobs/49730384.

dixlorenz:
It seems it simply doesn't recognise the attribute; I am using "Apple LLVM version 6.0 (clang-600.0.56) (based on LLVM 3.5svn)", I would assume "AppleClang 6.0.0.6000054" is newer. I cannot reproduce the compile error; the only idea I have is to take out the attribute and just turn off the warning...

vitaut:
I've merged your fixes for the documentation and return-after-throw warnings as separate commits https://github.com/cppformat/cppformat/commit/45d70b71ec227698fa92e1d6fb2f29f77af4fb93 and https://github.com/cppformat/cppformat/commit/37ad436068595a0722331510c8bfdf51c6df5749. Signed/unsigned comparison warning was fixed a bit earlier in https://github.com/cppformat/cppformat/commit/59254410daed335ba1e267a4606e666cf4a68391. The issue with `[[clang::fallthrough]]` needs further investigation though.
Thanks!

dixlorenz:
Thanks for the merge; about the fallthrough, the easy way would be

# pragma clang diagnostic ignored "-Wimplicit-fallthrough"

vitaut:
I think the `fallthrough` attribute is kind of nice, so I tried detecting it with [__has_cpp_attribute](http://clang.llvm.org/docs/LanguageExtensions.html#has-cpp-attribute) in https://github.com/cppformat/cppformat/commit/d7fea1ba0c387be568236fc4bd5e945eaa0145ae. Hope it works for you, otherwise feel free to (re)open an issue or PR.

dixlorenz:
I do prefer the attribute way too, this compiles for me but the warning is back; took me a bit to figure it out, __has_cpp_attribute has only been added to clang on 11/14/2014:

http://blog.llvm.org/2014/11/llvm-weekly-46-nov-17th-2014.html
http://reviews.llvm.org/rL221991 

The fix will probably work with the next version of Xcode (assuming that's based on clang 3.6), but not with the current one.

PR #102, opened 2015-02-07
added color support for windows

Ryuuke:


vitaut:
Thanks a lot for the PR. A few comments/requests if you don't mind:

* Could you leave the include of `windows.h` in `format.cc` and instead of conditionally defining the `Color` enum, add an array that maps standard ANSI colors to Windows-specific ones? Something like:
  ```
// somewhere in format.cc:
unsigned char COLORS[] = {
  0,
  FOREGROUND_RED | FOREGROUND_INTENSITY,
  FOREGROUND_GREEN | FOREGROUND_INTENSITY,
  ...
};
  ```

* `GetStdHandle(STD_OUTPUT_HANDLE)` can be called once in `print_colored`.

* Could you please add checks if the calls to `GetStdHandle` and `SetConsoleTextAttribute` succeeded? Something like:

  ```
HANDLE handle = GetStdHandle(STD_OUTPUT_HANDLE);
if (handle == INVALID_HANDLE_VALUE)
      FMT_THROW(WindowsError(GetLastError(), "cannot get output handle"));
  ```

* And finally, is there a way to unit test this as it is done for ANSI version in https://github.com/cppformat/cppformat/blob/master/test/format-test.cc#L1373?

Ryuuke:
Yeah no problem !

vitaut:
Thanks! This mostly looks good and my only concern now is the use of mutable global state (`RESET_COLOR` and `reset_color_flag`) which is not thread-safe. I suggest calling `GetConsoleScreenBufferInfo` to get the current color unconditionally and getting rid of these global variables.

Ryuuke:
As you like, In C++11 we could use std::call_once in this case for thread safety, right?

vitaut:
Yeah, but the library should be compatible with pre-C++11 compilers. Thanks updating the PR.

vitaut:
I've merged it with minor changes (https://github.com/cppformat/cppformat/commit/7004d1edf6021bcbe09814e973c68fb3c0287a26 & https://github.com/cppformat/cppformat/commit/9368b6a4ee499f5246851b7e0af8ebd2dee26ea5). However, the `PrintColored` test is failing: https://ci.appveyor.com/project/vitaut/cppformat/build/1.0.683.

Ryuuke:
The error in the test is coming from the codes below
```c++
  if(!GetConsoleScreenBufferInfo(handle, &infoCon))
    FMT_THROW(GetLastError(), "cannot get console informations");
```
```c++
 if(!SetConsoleTextAttribute(handle, color))
    FMT_THROW(GetLastError(), "cannot set console color");
```
it seems like it always throw with gtest, but I've tested in a project and it works...

vitaut:
Looks like the `GetStdHandle` doesn't work when you redirect the output from a console to a pipe with `EXPECT_WRITE`. Is there another way to test this, perhaps by accessing the console text directly via the Windows API?

vitaut:
I've moved the Windows implementation of `print_colored` to a separate branch, [print-colored](https://github.com/cppformat/cppformat/tree/print-colored) for now as it needs further testing. I'm happy to look into it myself, but it may take some time.

vitaut:
@Ryuuke, could you please review the updated [CONTIBUTING](https://github.com/fmtlib/fmt/blob/ebec00138bd10e949ff34955bcb096387cd32acc/CONTRIBUTING.md) document, particularly the part about licensing, and let me know if you agree with it being applied to your contributions to {fmt}? The library is likely to be relicensed (#1073) so I'm collecting approval from all earlier contributors. Thanks!

Ryuuke:
Looks good to me  !

PR #108, opened 2015-02-13
(WIP) Small fixes and improvements

CarterLi:
Good project, very fast and portable! When I was using cppformat as a tiny logging library for my project, I found some improvements. Not finished yet, but I decided to create PR earlier for everyone's reviewing.
### Changes: ###
1. Added some methods to BufferedFile to write raw texts without formating ( useful when writing logs )
2. Use more amesome C++11 features
 * _Deleted functions_
 * Use _static assertion_ actively
 * Change _0_ to _nullptr_ where appropriate
3. Improve Windows / MSVC support
 * Use _throw()_ in MSVC
 * Link with _format.lib_ automatically in MSVC
 * Silence some warnings
4. CMakeLists.txt improvements
 * Enable C++11 support when building main lib too ( why we didn't enable it before? )
 * Try to support clang: use libc++ when building with clang for better C++11 support ( doesn't work yet )

### TODOs: ###
- [ ] One warning remaining: ```format.cc:491:55: warning: declaration of 'error_code' shadows a member of 'this' [-Wshadow]```
- [ ] UT for extra methods added for BufferedFile. I'm not familier with gmock, investigating
- [ ] Fix clang support. Currently cmake complains ```Could NOT find Threads (missing: Threads_FOUND) ```. Don't know what's happening:(

### Tested platforms ###
1. Windows MSVC 2013
2. Windows MSYS2 g++ 4.9.2
3. Archlinux g++ 4.9.2
4. Archlinux clang++ 3.5.1 with libc++

Thanks
Carter

### DIFF
--- a/posix.cc
+++ b/posix.cc
@@ -148,13 +148,13 @@ void fmt::File::close() {
 
 fmt::LongLong fmt::File::size() const {
 #ifdef _WIN32
-  LARGE_INTEGER size = {};
+  LARGE_INTEGER filesize;
   HANDLE handle = reinterpret_cast<HANDLE>(_get_osfhandle(fd_));
-  if (!FMT_SYSTEM(GetFileSizeEx(handle, &size)))
+  if (!FMT_SYSTEM(GetFileSizeEx(handle, &filesize)))
     throw WindowsError(GetLastError(), "cannot get file size");
-  FMT_STATIC_ASSERT(sizeof(fmt::LongLong) >= sizeof(size.QuadPart),
+  FMT_STATIC_ASSERT(sizeof(fmt::LongLong) >= sizeof(filesize.QuadPart),
       "return type of File::size is not large enough");
-  return size.QuadPart;
+  return filesize.QuadPart;
 #else
   typedef struct stat Stat;
   Stat file_stat = Stat();
@@ -243,7 +243,7 @@ fmt::BufferedFile fmt::File::fdopen(const char *mode) {
 
 long fmt::getpagesize() {
 #ifdef _WIN32
-  SYSTEM_INFO si = {};
+  SYSTEM_INFO si;
   GetSystemInfo(&si);
   return si.dwPageSize;
 #else
--- a/posix.cc
+++ b/posix.cc
@@ -105,12 +105,29 @@ int fmt::BufferedFile::fileno() const {
 }
 
 void fmt::BufferedFile::flush() {
-    if (!file_)
-        return;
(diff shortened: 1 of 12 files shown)


posix.cc
```diff
@@ -134,13 +165,13 @@ void fmt::File::close() {
 
 fmt::LongLong fmt::File::size() const {
 #ifdef _WIN32
-  LARGE_INTEGER size = {};
+  LARGE_INTEGER filesize;
```
CarterLi: Silence -Wshadow

CarterLi:
g++ supports nullptr before change __cplusplus to 201103, will fix it

vitaut:
Thanks for your interest in C++ Format and a huge contribution.

A few comments:

1. New `BufferedFile` methods overall look good, but
    * I'd rename `write_raw` to `write`. The idea is to mostly follow the naming of the underlying methods, `fwrite` in this cases.
    * The following checks shouldn't be there, because calling a method on a closed file should fail in one way or another, not silently do nothing:

     ```c++
    if (!file_)
          return;
     ```
2. `FMT_NOEXCEPT` without an argument: great idea, can merge it any time (if you submit it in a separate PR).

3. `nullptr`: I'm afraid this change can't go in, because the library shouldn't define any unprefixed macro  to avoid potential conflicts and warnings in the client code (however unlikely they are). It is possible to use FMT_NULLPTR instead, but it's probably make the code to ugly and not worth it. So I suggest keeping `0` for now. I have plans to write a small program that will do automatic translation of some C++11 constructs into C++98 that will address this issue.

In general, it would be nice to split it in several PRs if it is not too difficult. As I said the `FMT_NOEXCEPT

CarterLi:
Thanks for your reply.

1. ```I'd rename write_raw to write. The idea is to mostly follow the naming of the underlying methods, fwrite in this cases.```
 * Have thought before. Concern: we already have __fmt::MemoryWriter::write__, which looks like __fmt::BufferedFile::write__. But they have different behavior: It's ok to say somefile.write("{}{}{}"), but you will get a fmt::FormatError exception when you run the same code.
 * I agree. Will throw an exception.

3. Well, nullptr is one of the most clean feature in C++11 in my opinion. Too bad we can't use it for supporting ancient compilers. Ok I'll revert the change by some simple %s/nullptr/0/g

As there are some FIXMEs left, I don't expect a quick merge. Other comments are welcomed.

CarterLi:
Will split this into several smaller PRs, closing

PR #110, opened 2015-02-14
Use throw() when noexcept is not supported

CarterLi:
I decide to split my old [PR](https://github.com/cppformat/cppformat/pull/108) into smaller ones as project owner's request.
I'll create another one after this gets merged.


vitaut:
Thanks for clarification. Merged.

CarterLi:
Thanks for merging @vitaut
We should do more CMake checks in the future because code like ```__cplusplus>=201103``` may incorrectly disable C++11 features before g++-4.7 due to the gcc [bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=1773)

PR #111, opened 2015-02-14
Use explicitly deleted functions to make types non-copyable

CarterLi:
Note ```(FMT_GCC_VERSION >= 404 && __cplusplus >= 201103L)``` means ```(FMT_GCC_VERSION >= 407 && __cplusplus >= 201103L)``` actually because of the gcc [bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=1773). CMake checks are suggested.
I'll create another PR after this gets merged. As for nullptr, I think it's acceptable as a [monkey patch](http://en.wikipedia.org/wiki/Monkey_patch). Such things have been done by [libc++](http://llvm.org/viewvc/llvm-project/libcxx/trunk/include/cstddef?view=markup#l87).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,6 +117,17 @@ endif ()
 #  add_definitions(-DFMT_USE_NOEXCEPT=1)
 #endif ()
 
+#check_cxx_source_compiles("
+#  struct C{
+#    C()=delete;
+#    C(const C&)=delete;
+#    C& operator=(const C&)=delete;
+#  };
+#  int main(){}" FMT_DELETED_FUNCTIONS)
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for another great PR. I wouldn't worry too much about older versions of GCC as it is covered b
by the fallback definition of `FMT_DISALLOW_COPY_AND_ASSIGN`:

```c++
# define FMT_DISALLOW_COPY_AND_ASSIGN(TypeName) \
    TypeName(const TypeName&); \
    TypeName& operator=(const TypeName&)
```

As for the monkey patch, if I understand correctly this would mean adding another required header to the library which I'd rather avoid.

PR #112, opened 2015-02-15
Detect static assertion support

CarterLi:
In addition, we use [__GXX_EXPERIMENTAL_CXX0X__](http://stackoverflow.com/questions/2958398/gnu-c-how-to-check-when-std-c0x-is-in-effect#answer-2958415) for better C++11 detection now


vitaut:
Merged, thanks.

PR #113, opened 2015-02-16
Silences warnings & adds BufferedFile methods

CarterLi:
1. Silences more warnings when building on Windows
 * Use ```GCC diagnostic ignored``` because I couldn't come up with a better name
2. Adds more ```BufferedFile``` methods which are useful when writing logs
 * Use assertion to check closed file
 * Notable FIXME & still missing unit tests

### DIFF
--- a/format.cc
+++ b/format.cc
@@ -487,6 +487,11 @@ FMT_FUNC int fmt::internal::UTF16ToUTF8::convert(fmt::WStringRef s) {
   return 0;
 }
 
+# ifdef __GNUC__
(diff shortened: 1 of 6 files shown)


vitaut:
The commits that silence warnings look fine, I had only one minor comment there. However, could you submit this as two separate PRs, one for warnings and another for extra `BufferedFile` methods? They are unrelated and the latter probably needs more discussion. Also while you at it, perhaps you could get rid of `#pragma once` commits that cancel each other to keep the history clean (although it is not critical). Thanks!

PR #116, opened 2015-02-20
Use MSVC intrinsics for better performance

CarterLi:
Try to fix https://github.com/cppformat/cppformat/issues/115
Not tested yet due to lacking of MSVC compiler

### DIFF
--- a/format.h
+++ b/format.h
@@ -43,6 +43,13 @@
 #if _SECURE_SCL
 # include <iterator>
 #endif
+#ifdef _MSC_VER
+# include <intrin.h> // _BitScanForward, _BitScanForward64
+# pragma intrinsic(_BitScanForward)
+# ifdef _WIN64
+#  pragma intrinsic(_BitScanForward64)
+# endif
+#endif
 
 #ifdef __GNUC__
 # define FMT_GCC_VERSION (__GNUC__ * 100 + __GNUC_MINOR__)
@@ -539,7 +546,25 @@ struct BasicData {
 
 typedef BasicData<> Data;
 
-#if FMT_GCC_VERSION >= 400 || FMT_HAS_BUILTIN(__builtin_clzll)
+#if _MSC_VER
+# ifdef _WIN64
+# define FMT_EFFICIENT_COUNT_DIGITS
+inline unsigned count_digits(uint64_t n) {
+  uint32_t index;
+  assert(__BitScanForward64(&index, n | 1));
+  unsigned t = (64 - index) * 1233 >> 12;
+  return t - (n < Data::POWERS_OF_10_64[t]) + 1;
+}
+# endif
+inline unsigned count_digits(uint32_t n) {
+  uint32_t index;
+  assert(__BitScanForward(&index, n | 1));
+  unsigned t = (32 - index) * 1233 >> 12;
+  return t - (n < Data::POWERS_OF_10_32[t]) + 1;
+}
+#elif FMT_GCC_VERSION >= 400 || defined __clang__
+# if FMT_HAS_BUILTIN(__builtin_clzll)
+# define FMT_EFFICIENT_COUNT_DIGITS
 // Returns the number of decimal digits in n. Leading zeros are not counted
 // except for n == 0 in which case count_digits returns 1.
 inline unsigned count_digits(uint64_t n) {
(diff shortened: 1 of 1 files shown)


vitaut:
Merged, thanks!

vitaut:
Applied some changes in https://github.com/cppformat/cppformat/commit/27e0bf72c02e830d5346738e59ea9848c32c2de9 as I think the correct intrinsic to use is `_BitScanReverse`. Also the intrinsic was not called in release version due to `assert`.

CarterLi:
You're correct thanks.
A question: Why do you [disable](https://github.com/cppformat/cppformat/blob/master/format.h#L584) ```count_digits(uint32_t n)``` conditionally when ```FMT_BUILTIN_CLZLL``` is not avaliable? I think ```count_digits(uint32_t n)``` can coexist with the fallback version of count_digits perfectly since ```count_digits(uint32_t n)``` is only an optimization in my opinion.
Note: ```_BitScanReverse64``` is only avaliable on __WIN64__. Provide ```count_digits(uint32_t n)``` should provide much better performance when compiling 32bit version.
In addition, we may implement ```__builtin_ctzll``` using ```__builtin_ctz``` when ```__builtin_ctzll``` when ```__builtin_ctzll``` is not avaliable ([for example](http://llvm.org/viewvc/llvm-project/libcxx/trunk/include/support/win32/support.h?view=markup#l134)), since 64bit operation is much more slower than 32bit operation on x86 and int (ie int32_t) is much more widely used than long long (ie int64_t).

vitaut:
> Why do you disable count_digits(uint32_t n) conditionally when FMT_BUILTIN_CLZLL is not avaliable?

Good point, there is no reason to disable optimized `count_digits(uint32_t n)` when `FMT_BUILTIN_CLZLL` is not avaliable. I've fixed this in https://github.com/cppformat/cppformat/commit/cbc46d9e78f6935b83ca233f6f2c2fff533ab1ef.

As for implementing `__builtin_clzll` in terms of `__builtin_clz` on 32-bit platforms, this is reasonable although not critical, because, as you mentioned yourself, `int` is much more widely used than `long long` and `int` is already optimized on 32-bit platforms. Besides, 32-bit systems are becoming increasingly rare. But patches are welcome =).

CarterLi:
well, what I want to say is that we only provide count_digits(uint64_t n) when __builtin_clz(ll) is not available, which means: if we call count_digits with an int on 32bit platform, we convert it into 64bit long long, do 64bit divisions

vitaut:
Right, but `__builtin_clz` or equivalent should be available on most platforms. Do you know any popular platform where `__builtin_clz` is not provided? Not that I'm completely opposed to adding a fallback `count_digits(uint32_t)` variant (perhaps, by making `count_digits` a template), just want to make sure that we don't add anything that is almost never used.

patlecat:
Ah I didn't know you're were using intrinsics already for GCC, RapidJson is also big on SSE usage to speed up their lib. MSVC coverage is very welcome. Cool :D

patlecat:
Here is a way to check which intrinsics are available for MSVC (in case you don't know already):
https://msdn.microsoft.com/en-us/library/hskdteyh.aspx

CarterLi:
According to the [Intel Instruction Set page](http://web.itu.edu.tr/kesgin/mul06/intel/instr/bsr.html), Bit Scan Reverse (BSR) is supported on i386+, should be supported in all modern CPUs.
Runtime CPUID check may not be used in this case since CPUID check itself takes time

vitaut:
I agree with @CarterLi that it's better to avoid runtime checks, but thanks for the suggestion @patlecat. I was thinking of adding SSE support, but it [adds very little in terms of performance](https://github.com/miloyip/itoa-benchmark#results) while making implementation much more complicated.

patlecat:
@vitaut  Then why did @miloyip rely on SSE for RapidJson?

You have a library here, so leave it to the developer on what nifty features to enable for his project, then you don't need CPUID, or offer a CPUID tool seperately to make the decision easier for your users.

vitaut:
@patlecat Maybe I'm missing something, but it looks like RapidJson is not using SSE for integer to string conversion: https://github.com/miloyip/rapidjson/blob/master/include/rapidjson/internal/itoa.h

Anyway, patches are welcome =)

miloyip:
Yes, @vitaut . From my implementations, SSE2 does not improve the performance much in integer-to-string conversion so I didn't put that in RapidJSON. However, RapidJSON got performance boost by using SSE2/4 for skipping whitespaces during parsing. It also uses bit scan intrinsics in Grisu implementation (float-to-string conversion) as well.

RapidJSON does not use dynamic dispatching according to CPUID. It simply uses static binding. Some discussions about this can be found [here](https://code.google.com/p/rapidjson/issues/detail?id=53).

vitaut:
Thanks, @miloyip, that's what I thought. You've done a great job with RapidJSON BTW.

PR #118, opened 2015-02-22
Emulate _BitScanForward64 on WIN32

CarterLi:
Implementation used: http://llvm.org/viewvc/llvm-project/libcxx/trunk/include/support/win32/support.h?view=markup#l135
Note: __builtin_ctzll which targets -m32 does similar thing already

__NOT__ tested due to lacking of MSVC compiler


vitaut:
Merged, thanks!

PR #119, opened 2015-02-24
Add an install command to the cmake file (to allow make install)

TrentHouliston:
Currently you can't make install cppformat?
This fixes that


vitaut:
Merged, thanks!

Issue #129, opened 2015-03-10
Doesn't compile on Windows under MinGW (TDM-gcc)


cstamford:
Compilation fails on Windows when using the TDM-gcc (http://tdm-gcc.tdragon.net/) MinGW distribution (and possibly others, though I did not test) with the flag -std=c++14 due to the following error:

```
error: 'swprintf' was not declared in this scope
```

When removing that flag (or undefining \__STRICT_ANSI__) compilation then fails with this error:

```
C:\CppFormat\format.cc: In instantiation of 'static int fmt::internal::CharTraits<wchar_t>::format_float(wchar_t*, std::size_t, const wchar_t*, unsigned int, int, T) [with T = double; std::size_t = unsigned int]':
C:\CppFormat\format.cc:1159:48:   required from here
C:\CppFormat\format.cc:391:45: error: invalid conversion from 'std::size_t {aka unsigned int}' to 'const wchar_t*' [-fpermissive]
         swprintf(buffer, size, format, value) :
                                             ^
In file included from C:\MinGW/lib/gcc/mingw32/4.9.2/include/c++/cstdio:42:0,
                 from C:\CppFormat\format.h:36,
                 from C:\CppFormat\format.cc:28:
C:\MinGW/include/stdio.h:579:37: note: initializing argument 2 of 'int swprintf(wchar_t*, const wchar_t*, ...)'
 _CRTIMP int __cdecl __MINGW_NOTHROW swprintf (wchar_t*, const wchar_t*, ...);
```

This is because swprintf is defined without the maxlen parameter in stdio.h as such:

```cpp
/* These differ from the ISO C prototypes, which have a maxlen parameter (like snprintf).  */
#ifndef __STRICT_ANSI__
_CRTIMP int __cdecl __MINGW_NOTHROW	swprintf (wchar_t*, const wchar_t*, ...);
_CRTIMP int __cdecl __MINGW_NOTHROW	vswprintf (wchar_t*, const wchar_t*, __VALIST);
#endif
```

vitaut:
Thanks for catching this and providing a fix.

PR #130, opened 2015-03-10
Added the define FMT_SWPRINTF to allow compilation under TDM-gcc (and po...

cstamford:
...ssibly other MinGW distributions).

Resolves issue #129.


vitaut:
Merged, thanks!

PR #134, opened 2015-03-14
Improve behaviors when disabling exceptions

CarterLi:
Functions which unconditionally throw exceptions should never return, with or without exception enabled

Tested compiler: AppleClang 6.1, brew g++-4.9 on OS X 10.10.2

### DIFF
--- a/format.cc
+++ b/format.cc
@@ -451,6 +451,7 @@ const uint64_t fmt::internal::BasicData<T>::POWERS_OF_10_64[] = {
 };
 
 FMT_FUNC void fmt::internal::report_unknown_type(char code, const char *type) {
+  (void)type;
   if (std::isprint(static_cast<unsigned char>(code))) {
     FMT_THROW(fmt::FormatError(
         fmt::format("unknown format code '{}' for {}", code, type)));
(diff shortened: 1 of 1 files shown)


format.cc
```diff
@@ -66,10 +66,24 @@ using fmt::internal::Arg;
 #ifndef FMT_THROW
 # if FMT_EXCEPTIONS
 #  define FMT_THROW(x) throw x
-#  define FMT_RETURN_AFTER_THROW(x)
 # else
-#  define FMT_THROW(x) assert(false)
-#  define FMT_RETURN_AFTER_THROW(x) return x
+#  ifdef DEBUG
```
vitaut: I think it should be `#ifndef NDEBUG` instead of `#ifdef DEBUG`.

format.cc
```diff
@@ -66,10 +66,24 @@ using fmt::internal::Arg;
...
+#   define FMT_THROW(x) std::abort()
+#  endif
+# endif
+#endif
+
+#ifndef FMT_NORETURN
+# if defined __GNUC__ || defined __clang__
+#  define FMT_NORETURN __attribute__((__noreturn__))
+# elif defined _MSC_VER
+#  define FMT_NORETURN __declspec(noreturn)
```
vitaut: I think there should be an `#else` clause with the fallback definition of `FMT_NORETURN`:
```c++
#else 
#  define FMT_NORETURN
```

vitaut:
Looks great, thanks. I have only two minor comments.

CarterLi:
Fixed, thanks

vitaut:
Thanks!

vitaut:
@CarterLi, could you please review the updated [CONTIBUTING](https://github.com/fmtlib/fmt/blob/ebec00138bd10e949ff34955bcb096387cd32acc/CONTRIBUTING.md) document, particularly the part about licensing, and let me know if you agree with it being applied to your contributions to {fmt}? The library is likely to be relicensed (#1073) so I'm collecting approval from all earlier contributors. Thanks!

PR #137, opened 2015-03-17
Fix CMake CPP11_FLAG not being added to compile flags when FMT_EXTRA_TESTS is OFF.

Gachapen:
I might have misunderstood something, but it seems wrong that you have to enable the extra tests for C++11 support.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,8 +75,10 @@ if (CMAKE_COMPILER_IS_GNUCXX)
   set_target_properties(format PROPERTIES COMPILE_FLAGS
     "-Wall -Wextra -Wshadow -pedantic")
 endif ()
-if (CPP11_FLAG AND FMT_EXTRA_TESTS)
+if (CPP11_FLAG)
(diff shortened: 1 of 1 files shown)


vitaut:
The idea here is to test compilation with both the default flags and -std=c++11. Otherwise the library is compiled with the default flags. However, this doesn't really matter as (1) all the code that relies on C++11 features is in the header (mostly inline variadic function templates) and (2) the library is distributed in the source code and you can specify your own options.

Since this can be a bit confusing, I've added a comment in https://github.com/cppformat/cppformat/commit/8ed603145722d1266b3a30a1c0b2d71f08722c39.

BTW your PR revealed fragility of the macro-test which I fixed in https://github.com/cppformat/cppformat/commit/538c2610cb829f4ad05153b0d6b8484797a5523f.

Thanks!

Gachapen:
OK, I understand. It was indeed confusing, but the comment helps a bit, so thanks for that.

Normally I let my submodules only use their own flags so that they are set up properly for my system, with no modification from my application project. But this results into format.cc not being compiled with the -std=c++11 flag. But if all the C++11 code is inline, then this doesn't matter.

PR #141, opened 2015-03-20
Support formatting more than 16 arguments with printf API when using variadic templates

Shauren:
Helper class TypeList holds a reference to array with packed type data (living on the stack)


Naios:
:+1:

vitaut:
Great stuff, thanks for working on this! I'll need to take a closer look before merging though (hopefully later today or tomorrow).

vitaut:
Merged, thanks.

PR #154, opened 2015-04-27
fixed std::min and std::max problem

DevO2012:
fixed std::min and std::max problem because of min and max macros in win...


vitaut:
Thanks for the PR. `windows.h` is a quite a mess, but I don't think that C++ Format should unconditionally define global macros that can affect it. The reason is that such macros can break other library's or user code when `windows.h` is included elsewhere. I think it should be up to the user to decide whether these macros are defined or not and they can be defined before including `format.h` or on a project level.

PR #161, opened 2015-05-15
Correct misspellings

syohex:


vitaut:
Thanks for the corrections.

PR #189, opened 2015-07-25
Fixing doc building with newer pip versions

xentec:
``CalledProcessError`` should only be raised, if ``'No command ...'`` is in ``stderr``, not otherwise.

### DIFF
--- a/doc/build.py
+++ b/doc/build.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 # Build the documentation.
 
 from __future__ import print_function
--- a/doc/build.py
+++ b/doc/build.py
@@ -17,7 +17,7 @@ def pip_install(package, commit=None):
       # Old versions of pip such as the one installed on Travis don't support
       # the show command - continue installation in this case.
       # Otherwise throw CalledProcessError.
-      if 'No command by the name pip show' not in stderr:
+      if p.returncode > 1 and 'No command by the name pip show' not in stderr:
         raise CalledProcessError(p.returncode, cmd)
     package = 'git+git://github.com/{0}.git@{1}'.format(package, commit)
   check_call(['pip', 'install', '-q', package])


xentec:
Well, this is interesting: Travis somehow fails on the opposite, indicating that his ``pip`` version does not support ``show``.

Is the check for ``show`` support even necessary?

vitaut:
Thanks for the PR, but the logic looks correct (although the comment could be better). `pip show` is used to check if the package is already installed. If the `show` command is not supported pip gives error "No command by the name pip show" which is ignored and the installation continues, otherwise a `CalledProcessError` exception is thrown.

> Is the check for show support even necessary?

Yes, the check is necessary because the old version of pip installed on travis doesn't support the show command.

Is there any problem when you run unchanged `build.py`?

xentec:
Yes, I get the ``CalledProcessError`` when building it on Arch Linux, because ``p.returnCode`` is 1 but ``stderr`` is empty.
Using ``virtualenv 13.1.0`` and ``pip 7.1.0``.

vitaut:
Could be a change in pip behavior. What happens when you run `pip show somenoninstalledpackage` from a command line, does it return nonzero error code and doesn't print anything?

xentec:
Pretty much.
```
» doc/virtualenv/bin/pip show breathe; echo $?
1
» doc/virtualenv/bin/pip show somenoninstalledpackage; echo $?
1
» doc/virtualenv/bin/pip show sphinx; echo $?
---
Metadata-Version: 2.0
Name: Sphinx
Version: 1.3.1
Summary: Python documentation generator
Home-page: http://sphinx-doc.org/
Author: Georg Brandl
Author-email: georg@python.org
License: BSD
Location: /home/xentec/build/cppformat-git/src/cppformat-git/build/doc/virtualenv/lib/python3.4/site-packages
Requires: Jinja2, alabaster, babel, Pygments, six, sphinx-rtd-theme, snowballstemmer, docutils
0
```

vitaut:
I've tried to modify the logic in `build.py` to make it work with the new version of pip: https://github.com/cppformat/cppformat/commit/3b224e18136013f3765b8df407cc537f32c93885

Does it work for you?

xentec:
I updated my fix, which is not pretty nor very future proof.
Let's just hope this won't need as much tinkering as you needed with MinGW.

vitaut:
Merged, thanks!

One way to improve this is to ignore the error from pip show altogether but the current approach is OK too even if it is not particularly pretty.

> Let's just hope this won't need as much tinkering as you needed with MinGW.

Haha yeah, MinGW32 was a bit of pain.

PR #191, opened 2015-08-04
fixed #190 Argument shadowing and Clang pedantic support

### LINKED ISSUE #190: Argument shadowing and Clang pedantic support
Hi,

Pedantic warnings do not work on Clang (cmake 3.2.3, OSX 10.10.4, Apple LLVM version 6.1.0 (clang-602.0.53) (based on LLVM 3.6.0svn)).

I've made some changes to the main CMakeLists.txt to support that, and also fixed the shadowing warnings that come up with FMT_PEDANTIC set to ON.

I'm issuing a pull request now.

rpopescu:


rpopescu:
That's actually the first thing I tried, and it doesn't work - the warning flags are not being set.
I have no idea why that is.
Cmake since 3.0.2 at least [says that COMPILE_FLAGS is deprecated](http://www.cmake.org/cmake/help/v3.0/prop_tgt/COMPILE_FLAGS.html?highlight=compile_flags), so I don't think that the current way it's being done is the best one.

vitaut:
It turned out the compile flags where overwritten later. This is fixed in https://github.com/cppformat/cppformat/commit/a8d12ade1c98f0d64d38cb1d4b55026f3f9fe274. `COMPILE_FLAGS` are used for compatibility with older CMake such as the one installed on Travis CI. At some point these will be replaced with `target_compile_options`.

Thanks!

PR #196, opened 2015-08-17
Fix warnings

jkflying:
Child attribute was being instantiated before parent attribute, gives warnings under GCC 4.9.1


vitaut:
Merged, thanks!

PR #197, opened 2015-08-17
new CMake option to disable generating the install target

maddinat0r:
The install target is interfering with CPack when using cppformat as a dependency in another CMake project. When creating a package for said project, CPack also packages all cppformat library files and includes (which is not what we want since cppformat is already statically linked into our project).


vitaut:
Merged, thanks!

PR #198, opened 2015-08-19
add CMake option to toggle tests (on by default)

maddinat0r:
This will allow to skip tests in case we don't need them (no tests -> faster compiling).


vitaut:
LGTM. Merged and, while at it, restructured a bit moving most test-specific stuff to `test/CMakeLists.txt` in https://github.com/cppformat/cppformat/commit/4274f15f0da920285e489b8954adbf544ea108ca.

Thanks!

PR #200, opened 2015-08-30
add CMake option to toggle doxygen documentation

maddinat0r:
Yet another PR, this time with the ability to toggle creation of doxygen documentation (on by default).

By the way, keep up the good work! :+1:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,7 @@ endif ()
 option(FMT_PEDANTIC "Enable extra warnings and expensive tests." OFF)
 option(FMT_INSTALL "Generate install target." ON)
 option(FMT_TESTS "Generate tests." ON)
+option(FMT_DOCS "Generate doxygen documentation." ON)
 
 project(FORMAT)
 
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR, but what is the purpose of this option? The documentation is not generated by CMake, but only when you build the doc target which is not built by default.

maddinat0r:
You are right, but this might (in my case) confuse people when they compile a project (which depends on cppformat), since the main project will have a target named "doc". This target however will generate cppformat documentation, not the main project's documentation.  
I could also push another commit to mark this option as advanced, so that it's invisible by default for the "normal" user (people who directly setup and compile cppformat as main project, instead of creating CMake dependencies).

vitaut:
Fair enough. I think there is no need to mark the option advanced as the number of options is small, but I've changed the naming of this option and `FMT_TESTS` to be consistent with the names of the targets they controls, `FMT_DOC` for the doc target and `FMT_TEST` for the test target. `FMT_INSTALL` is already in this form (https://github.com/cppformat/cppformat/commit/2541dab11729c5f8be78980a7c1813226155a338). Hope you don't mind.

maddinat0r:
Sounds logical to me. Thanks for your work! :+1:

PR #204, opened 2015-09-27
User-defined literals for format and named arguments

dean0x7d:
This is a bit of fun with C++11 user-defined literals. This brings C++ Format a little closer to Python's `str.format` for literals:

```Python
# Python
"{0}{1}{0}".format("abra", "cad")
```
```C++
// C++11 with user-defined literals
"{0}{1}{0}"_format("abra", "cad");
```
The literal `_format` just passes the arguments to `fmt::format` and is equivalent to `fmt::format("{0}{1}{0}", "abra", "cad");`

Literals can be used with named arguments as well:
```Python
# Python
"Hello, {name}! Your number is {number}".format(name="World", number=1);
```
```C++
// C++11 with user-defined literals
"Hello, {name}! Your number is {number}"_format("name"_a="World", "number"_a=1);
```

The `_a` literal just constructs a `fmt::arg` under the hood.

Compiler support for UDLs is required. Minimum Clang 3.1, GCC 4.7 or VS2015.
To use the literals `using namespace fmt::literals;` will import just `_format` and `_a`. But `using namespace fmt;` is also a possibility.

### DIFF
--- a/format.h
+++ b/format.h
@@ -177,6 +177,16 @@ inline uint32_t clzll(uint64_t x) {
     TypeName& operator=(const TypeName&)
 #endif
 
+#ifndef FMT_USE_USER_DEFINED_LITERALS
+// All compilers which support UDLs also support variadic templates. This
+// makes the fmt::literals implementation easier. However, an explicit check
+// for variadic templates is added here just in case.
+# define FMT_USE_USER_DEFINED_LITERALS \
+   FMT_USE_VARIADIC_TEMPLATES && \
+   (FMT_HAS_FEATURE(cxx_user_literals) || \
+       (FMT_GCC_VERSION >= 407 && FMT_HAS_GXX_CXX11) || _MSC_VER >= 1900)
+#endif
+
 #ifndef FMT_ASSERT
 # define FMT_ASSERT(condition, message) assert((condition) && message)
 #endif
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the contribution! This looks like a cool even if a bit unconventional use of user-defined literals. I'd like to experiment with this a bit more before merging in, hopefully in the beginning of the next week when I return from the trip I'm currently in.

vitaut:
@dean0x7d, I get some warnings and errors on GCC 4.8.4 after merging this:

```
cppformat/test/format-test.cc:1615:427: warning: invalid suffix on literal; C++11 requires a space between literal and identifier [-Wliteral-suffix]
   EXPECT_EQ(format("{first}{second}{first}{third}",
                                                                                                                                                                                                                                                                                                                                                                                                                                           ^
...
cppformat/test/format-test.cc: In member function ‘virtual void LiteralsTest_Format_Test::TestBody()’:
/home/viz/Downloads/3/cppformat/test/format-test.cc:1610:237: error: expected ‘)’ before ‘{’ token
   EXPECT_EQ(format("{}c{}", "ab", 1), "{}c{}"_format("ab", 1));
                                                                                                                                                                                                                                             ^
...
```

Do you know what might be causing these?

vitaut:
I've temporarily moved the changes to the [udl](https://github.com/cppformat/cppformat/tree/udl) branch until the above issue is resolved.

dean0x7d:
This is really strange. I tested this originally with GCC 4.8.4, Clang 3.7 and VS2015. I just tried again with GCC 4.8.4 and I can't reproduce the error. Everything is working fine.

Is it possible that it's some system/library specific preprocessor definition messing with `_format`?

vitaut:
Interestingly, the error only occurs in tests. I can compile and run a small standalone example successfully:

```c++
#include "format.h"
int main() {
  fmt::print("The answer is {answer}.", "answer"_a=42);
}
```

And `_format` is not defined there - I've just checked with an `#ifdef` and `#error`.

vitaut:
Looks like a preprocessor problem, because changing the test to

```c++
  std::string str = "{}c{}"_format("ab", 1);
  EXPECT_EQ(format("{}c{}", "ab", 1), str);
```
compiles.

dean0x7d:
The gtest macros pass their arguments both as code and as strings using the macro stringize operator (#). It's possible that's where it trips over the literal syntax. But that's still pretty weird. I haven't been able to reproduce the error either in tests or standalone. I tried with GCC 4.8.4 and 5.2.0.

vitaut:
Yeah, looks like a stringification issue. The following code

```c++
#define stringify(x) #x
stringify("{}c{}"_format("ab", 1));
```

is preprocessed to

```c++
""{}c{}"_format(\"ab\", 1)";
```
Note that the double quotes in `"{}c{}"` are not escaped.

Here's a relevant bug report: https://gcc.gnu.org/ml/gcc-bugs/2015-04/msg02027.html

I guess a simple solution is to assign the result of `"..."_format(...)` to a string variable and pass it to `EXPECT_EQ` as I did in my previous comment.

dean0x7d:
I made the required changes to the tests. I haven't been able to reproduce the stringification bug on my end, so I can't actually test the workaround but I think this should be fine.

Is there any way to add the relevant commit to this pull request or should I open a new one?

vitaut:
Normally the PR is updated automatically when you push to correspondent branch, but I guess not for merged PRs. So I just pulled the commit from your branch. As expected the tests compile correctly now (why the issue only reproduces on my machine is a mystery). Thanks!

vitaut:
@dean0x7d BTW, could you by any chance add some docs describing the new functionality?

PR #206, opened 2015-10-11
Document user-defined literals

dean0x7d:
I noticed that named arguments were not covered on the "Overview" page so I added that before the literals. There's also a little bit extra about variadic templates in the "Portability" section.

The new text which describes the UDL functionality is all on the "Overview" page. I'm not really sure what, if anything, should be added to the "API" page. The actual signature of the literal operators
```c++
ImplDetail operator"" _udl(const char *s, std::size_t)
```
is not really useful for the end user. The return type is an implementation detail and the parameter list only matters to the compiler. The final syntax is pretty easy to understand
```c++
"format_str"_format(args...)
"arg_name"_a=value
```
but it doesn't correspond to the literal operator signature.

So I see three options here:
1) Regular doxygen docs with the `operator""` signature.
2) Custom write the docs with final syntax as the signature.
3) Add a note inside the existing API docs for `fmt::format` and `fmt::arg` mentioning `_format` and `_a` as alternative syntax.

Let me know which of those, if any, would be preferable.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -65,6 +65,34 @@ The Format API also supports positional arguments useful for localization:
 
   fmt::print("I'd rather be {1} than {0}.", "right", "happy");
 
+Named arguments can be created with ``fmt::arg``. This makes it easier to track 
+what goes where when multiple values are being inserted:
+
+.. code:: c++
+
+  fmt::print("Hello, {name}! The answer is {number}. Goodbye, {name}.",
+             fmt::arg("name", "World"), fmt::arg("number", 42));
+
+If your compiler supports C++11 user-defined literals, the suffix ``_a`` offers 
+an alternative, slightly terser syntax for named arguments:
+
+.. code:: c++
+
+  fmt::print("Hello, {name}! The answer is {number}. Goodbye, {name}.",
+             "name"_a="World", "number"_a=42);
+
+The ``_format`` suffix may be used to format string literals similar to Python:
+
+.. code:: c++
+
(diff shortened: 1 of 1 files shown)


vitaut:
Looks great, thanks!

I agree that the signatures of literal operators are not particularly useful for the users, but it might be a good idea to document them anyway in the API section at least for the sake of introducing the names `_format` and `_a` that reside in a separate namespace `fmt::literal` and give a location to link to. As in other cases, the API docs could give examples of use which would help users less familiar withe UDL. It would probably be a good idea to keep the API docs of `_format` and `_a` close to that of `format` and `arg` respectively.

dean0x7d:
I ran into a bit of trouble. It looks like Sphinx can't parse user-definel literals:
```
cppformat/doc/api.rst:29: WARNING: Error when parsing function declaration:
If no return type {
Invalid definition: Expecting "(" in parameters_and_qualifiers. [error at 26]
  internal::UdlFormat<char> fmt::literals::operator""_format(const char *, std::size_t)
  --------------------------^
} else if return type {
Invalid definition: Expected identifier in nested name. [error at 49]
  internal::UdlFormat<char> fmt::literals::operator""_format(const char *, std::size_t)
  -------------------------------------------------^
}
```

That's after running `make doc` with this setup: c12e3d6e14df6d7613f6d6fae35dfb0fc7999418.
I took a look at the Sphinx C++ parser and it wouldn't be too difficult to add UDL support. I'll see about submitting a PR to Sphinx. But it would take time for the changes to trickle down to a stable version.

PR #214, opened 2015-10-25
MS Visual Studio "/Zc:wchar_t-" compiler parameter support added

slavanap:
If ```/Zc:wchar_t-``` compiler parameter is specified, then ```wchar_t``` is defined as
```typedef unsigned short wchar_t;```
Thus, this leads to conflict in template definition in ```format.h```

You can verify it in any Visual Studio project with spdlog with 
```
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /Zc:wchar_t-")
```

reference: https://msdn.microsoft.com/ru-ru/library/dh8che7s.aspx

### DIFF
--- a/format.h
+++ b/format.h
@@ -940,7 +940,9 @@ class MakeValue : public Arg {
   // characters and strings into narrow strings as in
   //   fmt::format("{}", L"test");
   // To fix this, use a wide format string: fmt::format(L"{}", L"test").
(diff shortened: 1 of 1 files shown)


vitaut:
Merged with a minor change in https://github.com/cppformat/cppformat/commit/acfa858e3c420a880d7ff128ef57233aa7d4c76b (without this change `unsigned short` values would be formatted as characters). Thanks!

PR #215, opened 2015-10-26
MSVC 2015 supports noexcept

dmkrepo:
Use noexcept in MSVC too. Supported since 1900. (https://msdn.microsoft.com/en-us/library/dn956976.aspx)


vitaut:
Merged, thanks!

patlecat:
Previous version of MSVC supported it too as I had to point out on several occasions, at least MSVC2013 has it with the macro: _NOEXCEPT

vitaut:
@patlecat Good to know, thanks. However, `noexcept` is optional and even if it is not supported everything will just work (tm) anyway.

PR #218, opened 2015-11-02
Fix warnings

inguin:
These two commits fix a number of warnings when compiling with the gcc options -Wundef and -Wfloat-equal.

The first commit is pretty straight-forward: Use #ifdef instead of #if when checking _MSC_VER, and set a default value for options like FMT_USE_NOEXCEPT when undefined.

The second commit replaces a (value != value) expression used to test whether a floating point number is NaN with an explicit call to isnan(). I added a set of compatibility functions similar to those for isinf(), but I have not tested them with any compiler other than gcc.

### DIFF
--- a/format.cc
+++ b/format.cc
@@ -55,7 +55,7 @@ using fmt::internal::Arg;
 #if __GNUC__ && !__EXCEPTIONS
 # define FMT_EXCEPTIONS 0
 #endif
-#if _MSC_VER && !_HAS_EXCEPTIONS
+#if defined(_MSC_VER) && !_HAS_EXCEPTIONS
 # define FMT_EXCEPTIONS 0
 #endif
 #ifndef FMT_EXCEPTIONS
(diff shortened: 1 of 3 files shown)


vitaut:
Merged with a minor change in https://github.com/cppformat/cppformat/commit/8af9bf1d06b414fc61ccc7fa31b31098a24cbb30 to avoid defining system macro `_SECURE_SCL`. Thanks!

vitaut:
@inguin, could you please review the updated [CONTIBUTING](https://github.com/fmtlib/fmt/blob/ebec00138bd10e949ff34955bcb096387cd32acc/CONTRIBUTING.md) document, particularly the part about licensing, and let me know if you agree with it being applied to your contributions to {fmt}? The library is likely to be relicensed (#1073) so I'm collecting approval from all earlier contributors. Thanks!

inguin:
@vitaut: Fine for me, use my contribution under any license you like.
BTW, I'm very excited about std::format() being accepted for C++20. Thanks for all your work!

PR #220, opened 2015-11-03
Fix warnings

inguin:
Even more fixes for optional GCC warnings:

* When building with -Wctor-dtor-privacy, gcc mistakenly assumes that the IsConvertibleToInt template class is unusable. Method visibility changed from private to protected as a workaround.
* Add a few explicit casts to silence -Wconversion warnings.
* Disable sign-compare warning in the header. Silencing the warning otherwise would require many explicit casts due to the C/C++ integer promotion rules.
* Test GCC macros __GNUC__ and __EXCEPTIONS for presence instead of value.

### DIFF
--- a/format.cc
+++ b/format.cc
@@ -1265,7 +1265,7 @@ FMT_FUNC void fmt::print(std::ostream &os, CStringRef format_str, ArgList args) 
 
 FMT_FUNC void fmt::print_colored(Color c, CStringRef format, ArgList args) {
   char escape[] = "\x1b[30m";
-  escape[3] = '0' + static_cast<char>(c);
+  escape[3] = static_cast<char>('0' + c);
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for another great PR. Just one question: could you post the warnings that you get for `IsConvertibleToInt` with `-Wctor-dtor-privacy` and the compiler version? I was not able to reproduce this on gcc 4.8.4.

inguin:
I'm getting this warning with gcc 5.1.1 (from Fedora Linux 21):

cppformat/format.h:916:7: warning: all member functions in class ‘fmt::internal::IsConvertibleToInt<T>’ are private [-Wctor-dtor-privacy]

vitaut:
Merged, thanks!

PR #221, opened 2015-11-04
switch from writing two separate digits to writing the digits as a uint16

jinxidoru:
Here's a very minor optimization that you might want to take a look at.  Instead of writing two separate characters in FormatInt, I am writing it once as a uint16.  In my tests, this tended to be 5-10% faster, though it bounces around a bit.  If you want to use this, feel free.

### DIFF
--- a/format.cc
+++ b/format.cc
@@ -582,6 +582,9 @@ const char fmt::internal::BasicData<T>::DIGITS[] =
     "6061626364656667686970717273747576777879"
     "8081828384858687888990919293949596979899";
 
+template <typename T>
+const uint16_t *fmt::internal::BasicData<T>::U16_DIGITS = (const uint16_t*)DIGITS;
+
 #define FMT_POWERS_OF_10(factor) \
   factor * 10, \
   factor * 100, \
--- a/format.h
+++ b/format.h
@@ -726,6 +726,7 @@ struct BasicData {
   static const uint32_t POWERS_OF_10_32[];
   static const uint64_t POWERS_OF_10_64[];
   static const char DIGITS[];
+  static const uint16_t* U16_DIGITS;
 };
 
 typedef BasicData<> Data;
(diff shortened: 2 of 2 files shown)


newnon:
On old arms CPU data must be aligned on it size, and i'm not sure about is it always true for this code, because all int16 values must be on even memory adress

jinxidoru:
That's why I adjusted the buffer size, so that it would be properly aligned, though it would probably be wise to call that out more clearly.

dean0x7d:
The buffer size will not affect the alignment, only the element type is considered. In this case, `buffer_` has an alignment requirement of 1 byte, but you're accessing `uint16_t` which has an alignment requirement of 2. This will be architecture dependent behavior: x86 is pretty tolerant of alignment, however other architectures are not as forgiving and will generate an alignment fault. Even on x86, there may be varying performance (2 bytes being split across cache lines comes to mind, which is a pretty big penalty).

The member array `buffer_` actually turns out to be OK because it's in the class with a pointer so it will share the pointer's more strict alignment of 4 or 8 bytes. However, this could change accidentally simply by changing the composition of the class. It would be best to explicitly state the alignment: make `buffer_` an array of `uint16_t` and cast to `char` as needed. This is safe because it's casting from stricter alignment requirements (2 bytes) to more relaxed (1 byte), i.e. the opposite of what's happening now.

The static `DIGITS` array isn't a class member so it doesn't have any alignment restrictions except `char`. It'll have a 50/50 chance of being on an even byte. This should also be explicitly aligned.

C-style casts can be misleading. It can either be a `static_cast` or `reinterpret_cast`. For a `static_cast` the compiler makes sure everything is correct, including alignment, or it will refuse to compile. For a `reinterpret_cast` it will just use the big hammer and trust the user to point it at the right nail, or it'll just make whatever is there look like a nail anyway. So a `reinterpret_cast` is a pretty good signal to the user to be careful. A C-style cast may or may not be a `reinterpret_cast` so it obscures things.

jinxidoru:
The buffer size needed to be adjusted because of the way you start from one byte before the end.  Without an odd-numbered buffer size, the positioning would be off for the writes.  That said, I think your recommendation of using a uint16_t for the buffer makes a lot more sense since it is clear as to what is happening rather than voodoo magic behind the curtain.

The DIGITS array on the other hand, that's a good point.  I hadn't thought about that one.  I suspect the likelihood of it being aligned is actually more like 90% because of the way the compiler typically arranges structs, but it'd be better to be explicit.  The better option would probably be to switch the references so that the U16_DIGITS is constructed directly and the DIGITS array points at the U16_DIGITS memory.  That would ensure alignment.

Let me know if you would like me to make the changes and update the pull request.  I'm going to implement these changes into my other project regardless.  Thanks for the input!

dean0x7d:
I think it's a pretty interesting optimization, so I just put my two cents in regarding the alignment. 
@vitaut will give you the official word on the pull request.

vitaut:
Thanks @jinxidoru for the PR and @dean0x7d for reviewing it. I completely agree with all @dean0x7d wrote and would be definitely interested in seeing the updated PR.

vitaut:
Also this seems to be undefined under the strict aliasing rule.

vitaut:
Closing due to inactivity. Also I don't think controlling buffer alignment is practical, because it can be externally provided as in `FixedBuffer`.

PR #228, opened 2015-11-12
Fix static_assert feature detection on clang

dean0x7d:
I had an unused typedef warning as shown below. Looks like `static_assert` couldn't be found on clang
This PR fixes it.

```
cppformat/posix.cc:162:3: warning: unused typedef 'Assert163'
      [-Wunused-local-typedef]
  FMT_STATIC_ASSERT(sizeof(fmt::LongLong) >= sizeof(file_stat.st_size),
  ^
```

### DIFF
--- a/posix.h
+++ b/posix.h
@@ -73,7 +73,7 @@
 # define FMT_USE_STATIC_ASSERT 0
 #endif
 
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks!

PR #229, opened 2015-11-13
Fix warning when using Intel C++ on OS X

LogicalKnight:
Unfortunately, Intel's C++ compiler defines `__clang__` which means that
some of the pragmas in use that Intel C++ doesn't use will cause
warnings to be generated.


vitaut:
Thanks!

Issue #92, opened 2015-01-28
fmt::format<Char, T> inefficiency


seanmiddleditch:
Around format.h:2066 is the function:

```
template <typename Char, typename T>
void format(BasicFormatter<Char> &f, const Char *&format_str, const T &value) {
  std::basic_ostringstream<Char> os;
  os << value;
```

The use of `basic_ostringstream` is unnecessary here. The use of this template creates a new string stream - including a new string buffer - for each value without an explicit formatter.

`basic_ostringstream` is essentially just a specialized `basic_ostream` that binds to a `basic_stringbuf`. It's this string buffer that's the major unnecessary overhead. `basic_stringbuf` is an implementation of `basic_streambuf` that writes into a `basic_string`. The stream buffer is the object that does all of the actual writing and real work for IOStreams; the other classes that most users are familiar with are just a more usable facade.

cppformat could include its own `basic_formatbuf` that directly writes into the buffer space used by cppformat, directly applying the format parameters. This would remove any need for excess allocations or temporary string buffers. It would continue allowing the use of cppformat with any type that supports IOStreams output without (very much) overhead.

vitaut:
Thanks for the suggestion. I did plan to implement a custom stream buffer, but didn't get around to it yet.

patlecat:
Yeah stringstreams with a custom allocator class would be awesome and really make sense, because they are kinda slowish in STL.

NotImplemented:
I have experimented with custom basic_formatbuf which is derived from basic_streambuf and I made it possible to write to Formatter's writer().buffer_. This solution is done to preserve ability to work with types, which support output to IO streams, without any changes. However Writer is unaware of what is written by basic_formatbuf.

vitaut:
Awesome. Could you submit a pull request? BTW your nick looks familiar, Mike is it you? =)

vitaut:
Fixed in https://github.com/cppformat/cppformat/commit/d266adf80559d2cfe1f7ef8f2462e3e7ffdd7c6b.

PR #230, opened 2015-11-17
fmt::format<Char, T> inefficiency #92

NotImplemented:
Should we write some tests for this item? What is a proper way to run tests locally?

### DIFF
--- a/format.h
+++ b/format.h
@@ -32,7 +32,6 @@
 
 #include <cassert>
 #include <cmath>
-#include <cstddef>  // for std::ptrdiff_t
 #include <cstdio>
 #include <algorithm>
 #include <limits>
@@ -484,11 +483,12 @@ class Buffer {
 template <typename T>
 template <typename U>
 void Buffer<T>::append(const U *begin, const U *end) {
-  std::ptrdiff_t num_elements = end - begin;
-  if (size_ + num_elements > capacity_)
-    grow(size_ + num_elements);
+  assert(begin <= end);
+  std::size_t new_size = size_ + (end - begin);
+  if (new_size > capacity_)
+    grow(new_size);
   std::copy(begin, end, internal::make_ptr(ptr_, capacity_) + size_);
-  size_ += num_elements;
+  size_ = new_size;
 }
 
 namespace internal {
--- a/format.h
+++ b/format.h
@@ -2029,11 +2029,6 @@ class BasicWriter {
   std::size_t size() const { return buffer_.size(); }
 
   /**
-    Returns underlying buffer.
-   */
-  Buffer<Char>& buffer() const { return buffer_; }
-
-  /**
     Returns a pointer to the output buffer content. No terminating null
     character is appended.
    */
--- a/format.h
+++ b/format.h
@@ -115,7 +115,7 @@ inline uint32_t clzll(uint64_t x) {
 # define FMT_GCC_EXTENSION
 #endif
 
-#ifdef __clang__
+#if defined(__clang__) && !defined(__INTEL_COMPILER)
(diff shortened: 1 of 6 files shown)


format.h
```diff
@@ -2628,16 +2639,39 @@ class BasicArrayWriter : public BasicWriter<Char> {
...
 // Formats a value.
 template <typename Char, typename T>
-void format(BasicFormatter<Char> &f, const Char *&format_str, const T &value) {
-  std::basic_ostringstream<Char> os;
-  os << value;
-  std::basic_string<Char> str = os.str();
-  internal::Arg arg = internal::MakeValue<Char>(str);
-  arg.type = static_cast<internal::Arg::Type>(
-        internal::MakeValue<Char>::type(str));
-  format_str = f.format(format_str, arg);
```
mwinterb: While I haven't looked in depth at all the errors that are occurring, based on some of the error messages, it seems likely that it's because of the removal of this line. It seems like fixing this may require using a temporary internal::MemoryBuffer that's used by basic_formatbuf, and then format a BasicStringRef<Char> that's extracted from the MemoryBuffer. Since MemoryBuffer can avoid memory allocations for short strings, it seems like this would meet part of the goals for this issue, even if there would still be some unnecessary copies in the "empty format string" case.

mwinterb: To be a bit more helpful on my suggestion, keeping the majority of your change the same, except for changing basic_format's constructor to accept a Buffer<Elem> directly:
```c++
// Formats a value.
template <typename Char, typename T>
void format(BasicFormatter<Char> &f, const Char *&format_str, const T &value) {
  internal::MemoryBuffer<Char, internal::INLINE_BUFFER_SIZE> buffer;

  basic_formatbuf<Char> format_buf(buffer);
  std::basic_ostream<Char> output(&format_buf);
  output << value;
  format_buf.flush();

  BasicStringRef<Char> str(buffer.size() > 0 ? &buffer[0] : NULL, buffer.size());
  internal::Arg arg = internal::MakeValue<Char>(str);
  arg.type = static_cast<internal::Arg::Type>(
        internal::MakeValue<Char>::type(str));
  format_str = f.format(format_str, arg);
}
```
This passes all of the tests, based on an an older commit, though.

NotImplemented:
Yes, that will fix. Failing tests were connected with format check.

vitaut:
@NotImplemented Thanks for the PR, Mike! Just two minor comments in addition to what @mwinterb wrote. The `format` function should be covered by existing tests, but it might be useful to test `basic_formatbuf`. You can run the tests locally via the `test` target (`RUN_TESTS` in MSVC), e.g.

```
> make test
```

vitaut:
@NotImplemented LGTM and happy to merge. But could you [squash the commits](http://stackoverflow.com/a/5201642/471164) first to get rid of the ones that were failing on CI?

format.h
```diff
@@ -2628,12 +2634,56 @@ class BasicArrayWriter : public BasicWriter<Char> {
...
 
+template<class Elem, class Traits = std::char_traits<Elem> >
+class basic_formatbuf : public std::basic_streambuf<Elem, Traits> {
+
+  typedef typename std::basic_streambuf<Elem, Traits>::int_type int_type;
+  typedef typename std::basic_streambuf<Elem, Traits>::traits_type traits_type;
+
+  using std::basic_streambuf<Elem, Traits>::setp;
+  using	std::basic_streambuf<Elem, Traits>::pptr;
+  using	std::basic_streambuf<Elem, Traits>::pbase;
```
danilaml: no space?

vitaut:
Merged, thanks!

PR #236, opened 2015-12-01
Silence some performance warnings in Visual Studio

Jopie64:
Visual studio issues performance warnings when converting implicitly from bool to int


vitaut:
Thanks!

PR #239, opened 2015-12-02
Update README URLs based on HTTP redirects

ReadmeCritic:
Created with https://github.com/dkhamsing/frankenstein

### GitHub Corrected URLs 
Was | Now 
--- | --- 
https://github.com/HarpyWar/pvpgn | https://github.com/pvpgn/pvpgn-server 

### HTTPS Corrected URLs 
Was | Now 
--- | --- 
http://docs.python.org/2/library/stdtypes.html#str.format | https://docs.python.org/2/library/stdtypes.html#str.format 
http://docs.python.org/3/library/string.html#module-string | https://docs.python.org/3/library/string.html#module-string 
http://en.wikipedia.org/wiki/Internationalization_and_localization | https://en.wikipedia.org/wiki/Internationalization_and_localization 

### Other Corrected URLs 
Was | Now 
--- | --- 
http://www.kbengine.org/ | http://kbengine.org/ 
https://code.google.com/p/saddy/ | https://github.com/mamontov-cpp/saddy-graphics-engine-2d

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -28,7 +28,7 @@ Features
 * Format API with `format string syntax
   <http://cppformat.github.io/latest/syntax.html>`_
   similar to the one used by `str.format
-  <http://docs.python.org/2/library/stdtypes.html#str.format>`_ in Python.
+  <https://docs.python.org/2/library/stdtypes.html#str.format>`_ in Python.
 * Safe `printf implementation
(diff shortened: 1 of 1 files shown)


vitaut:
Merged, thanks!

ReadmeCritic:
:wink:

PR #240, opened 2015-12-03
Fix fmt::BufferedFile::fileno not compiling with Android NDK.

Gachapen:
With Android NDK arm7 gcc 4.9, `fileno` is defined as a macro, so calling `fileno` with the scope operators will fail. Therefore `FMT_POSIX_CALL` is replaced with `FMT_SYSTEM`. `BufferedFile::fileno` has also been renamed to `fileno_` to avoid conflicts with the global `fileno` macro.

I'm a bit unsure about using `FMT_SYSTEM`, but at least it solves the problem for me.

### DIFF
--- a/posix.cc
+++ b/posix.cc
@@ -102,8 +102,8 @@ void fmt::BufferedFile::close() {
 // A macro used to prevent expansion of fileno on broken versions of MinGW.
 #define FMT_ARGS
 
-int fmt::BufferedFile::fileno() const {
-  int fd = FMT_POSIX_CALL(fileno FMT_ARGS(file_));
+int fmt::BufferedFile::fileno_() const {
+  int fd = FMT_SYSTEM(fileno FMT_ARGS(file_));
   if (fd == -1)
     throw SystemError(errno, "cannot get file descriptor");
   return fd;
--- a/posix.h
+++ b/posix.h
@@ -195,7 +195,7 @@ public:
 
   // We place parentheses around fileno to workaround a bug in some versions
   // of MinGW that define fileno as a macro.
-  int (fileno)() const;
+  int (fileno_)() const;
 
   void print(CStringRef format_str, const ArgList &args) {
     fmt::print(file_, format_str, args);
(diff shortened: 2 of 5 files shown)


Gachapen:
Looks like I should rebase on the master tip and fix the new issues. Still, do you think this is an OK solution?

vitaut:
Thanks for the PR. Before deciding whether this is a good fix, could you give more details, in particular, what exactly error do you get and how is `fileno` defined in Android NDK?

Gachapen:
Thanks for the quick response. I just rebased to fix the issues. I'll run a build with the current master and provide the error messages. I also noticed that you undef `fileno` in `posix-test.cc` if `__MINGW32__` is defined. Maybe something like this can be done for Android as well?

Gachapen:
Here's the exact error. Notice that this is part of a larger project. I can make a minimal working example if you wish.
```
In file included from /home/magnus/coding/suttung/nox-engine-sample/source/nox-engine/third_party/cppformat/posix.h:38:0,
                 from /home/magnus/coding/suttung/nox-engine-sample/source/nox-engine/third_party/cppformat/posix.cc:33:
/home/magnus/coding/suttung/nox-engine-sample/source/nox-engine/third_party/cppformat/posix.cc:105:24: error: expected unqualified-id before '(' token
 int fmt::BufferedFile::fileno() const {
                        ^
```

I also tried adding `#undef fileno` to posix.cc and it solved the problem.

Gachapen:
Actually, a minimal working example shouldn't be necessary as the tests also fail to build.

I also tested undeffing `fileno` if `__ANDROID__` is defined. It solves the problem, probably in a better way than this PR. Should I create a new PR?

```
#ifdef __ANDROID__
#  undef fileno
#endif
```
This is the exact same as for mingw32.

vitaut:
I think the most robust (and easiest) solution is to replace

```c++
#ifdef __MINGW32__
# undef fileno
#endif
```
with 
```c++
#ifdef fileno
# undef fileno
#endif
```
Feel free to open a PR (otherwise I can make the change myself later) and thanks for testing this.

Gachapen:
Great! I'll make a PR right away. Glad to contribute :)

PR #241, opened 2015-12-03
Fix fileno causing compile error when #defined

Gachapen:
Error:
expected unqualified-id before '(' token
 int fmt::BufferedFile::fileno() const {

This is an issue with Android NDK and mingw32.


vitaut:
Merged and made similar fix in posix-test (https://github.com/cppformat/cppformat/commit/aa7bb101ed0c4dbf3b737acec9f8fa1bc0c6f1c4). Thanks!

PR #245, opened 2015-12-09
Declarations for shared library in Windows.

macdems:
When trying to use cppformat in a shared library I have found that it does not compile in Windows (tested with MinGW32-w64). The reason for this is a lack fo __declspec(dllimport) and __declspec(dllexport) declaration. This PR solves this issue: when compiling shared library, one needs to add definition FMT_EXPORTS and skip it when using the library.

I hope this small improvement can help to make this great library even better.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,12 +94,17 @@ if (BIICODE)
   return()
 endif ()
 
-add_library(cppformat ${FMT_SOURCES})
-if (BUILD_SHARED_LIBS AND UNIX AND NOT APPLE)
-  # Fix rpmlint warning:
-  # unused-direct-shlib-dependency /usr/lib/libformat.so.1.1.0 /lib/libm.so.6.
-  target_link_libraries(cppformat -Wl,--as-needed)
-endif ()
+if (BUILD_SHARED_LIBS)
+    add_library(cppformat SHARED ${FMT_SOURCES})
+    if (UNIX AND NOT APPLE)
+      # Fix rpmlint warning:
+      # unused-direct-shlib-dependency /usr/lib/libformat.so.1.1.0 /lib/libm.so.6.
+      target_link_libraries(cppformat -Wl,--as-needed)
+    endif ()
+else ()
+    add_library(cppformat ${FMT_SOURCES})
+endif()
+
 if (FMT_PEDANTIC AND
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -115,7 +115,7 @@ if (CPP11_FLAG AND FMT_PEDANTIC)
 endif ()
 
 set_target_properties(cppformat
-  PROPERTIES COMPILE_FLAGS "${FMT_EXTRA_COMPILE_FLAGS}")
+  PROPERTIES COMPILE_FLAGS "-DFMT_EXPORTS ${FMT_EXTRA_COMPILE_FLAGS}")
```
vitaut: `FMT_EXPORTS` shouldn't be defined unconditionally, only when the `BUILD_SHARED_LIBS` CMake variable is set, because it breaks building the static library.

format.h
```diff
@@ -2773,7 +2789,7 @@ void BasicWriter<Char>::write_double(
           spec.width() > static_cast<unsigned>(n)) {
         width = spec.width();
         CharPtr p = grow_buffer(width);
-        std::memmove(get(p) + (width - n) / 2, get(p), n * sizeof(Char));
+        std::uninitialized_copy(p, p + n, p + (width - n) / 2);
```
vitaut: This overwrites the latest changes. Please rebase your branch to the current master.

format.cc
```diff
@@ -425,7 +419,7 @@ FMT_FUNC void fmt::SystemError::init(
 }
 
 template <typename T>
-int fmt::internal::CharTraits<char>::format_float(
+FMT_EXPORT int fmt::internal::CharTraits<char>::format_float(
```
vitaut: Is it necessary to use `__declspec(dllexport)` both on declarations and definitions? If not I suggest using it on the header declarations only.

vitaut:
Thanks for the PR. I think this could be useful, but I have some comments/questions (above).

macdems:
Concerning your comments, I have made several improvements to the patch. I hope it will now allow to use cppformat as dynamic library in Windows.

vitaut:
Merged with very minor adjustments in https://github.com/cppformat/cppformat/commit/1cba0aea27915c06192b47f2d4429c8ddaad3aa8. Thanks for contributing to the library!

PR #251, opened 2015-12-23
Add fprintf overload that writes to a std::ostream

nickhutchinson:
For completeness, add an overload for fprintf that writes to a std::ostream.


vitaut:
Merged, thanks!

PR #258, opened 2016-01-05
Ignore breathe and /doc/sphinx-bootstrap-theme directories in git

Gachapen:
These are created when building the library and makes the directory dirty. This is annoying when using cppformat as a submodule since the submodule always will show up in the diff as dirty.

Personally I'd prefer to not modify the source directory at all when configuring and building. It should rather create files in the build directory, but I don't know if that is possible in this case.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,8 +3,10 @@ bin/
 /doc/doxyxml
 /doc/html
 /doc/virtualenv
+/doc/sphinx-bootstrap-theme
 /Testing
 /install_manifest.txt
+/breathe
 *~
 *.a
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR but `sphinx-bootstrap-theme` is no longer used and `breathe` is installed in `virtualenv` which is already ignored in the current version. Which version do you use?

Gachapen:
You're right. I can't reproduce it anymore on 2.0.0 or master. I was pretty sure it was 2.0.0, but I'm starting to doubt that now. I'll close the issue.

Issue #261, opened 2016-01-10
MSVC performance issue with ArgMap


mwinterb:
While looking into #193, I was surprised to see std::map memory allocation related operations high up in the profile. It turned out that this was because ArgMap is a member of BasicFormatter, ArgMap has an std::map<> member that's default constructed, and in MSVC's std library implementation, all of map's constructors allocate a root node<sup>1</sup>. libc++ and libstdc++ do not, at first glance, which is probably why this hasn't been seen already (I don't think anyway, no closed issues with ArgMap popped up in a search).

Options I see for addressing this:
1. Defer construction of the std::map in ArgMap until it is initialized. Having something like boost::optional would be nice here, but it's not necessary.
2. Instead of using a map for the backing store, use a vector. I think all three major std library implementations have a non-allocating default constructor. The vector may not even need to be sorted, as it doesn't look like anything depends on order or even uniqueness and there probably isn't going to be a large number of named arguments. If it does, it's trivial to maintain that constraint since the map is effectively immutable after init() is called.

However, both of those options would break binary compatibility.Option 2 seems like the cleanest option.

I've done a prototype with option 2, and it seemed to improve #193 from ~1.5s to ~0.9s, roughly in line with the results you're currently getting.

<sup>1</sup> The rationale for why it does this is described in this paper, just search for Lavavej.
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4055.html
While it's in the context of a move constructor, and list, the point being made applies equally to the default constructor and other node-based containers.

vitaut:
Replacing a map with a vector looks like a good idea. In fact, I was planning to do something along these lines, but didn't have time for it yet. The binary compatibility will be broken in both options, but I think it's worth it. We'll just have to bump the major version number and it might take a while until this is released.

Since you've done some work in this direction already, could you by any chance submit a PR?

vitaut:
Fixed by #262. Thanks!

PR #262, opened 2016-01-10
Changed ArgMap to be backed by a vector instead of a map.

mwinterb:
The main reason for this is to avoid a dynamic memory allocation in every format() call with Visual Studio if there are no named arguments. Addresses #261.

### DIFF
--- a/format.h
+++ b/format.h
@@ -36,7 +36,8 @@
 #include <memory>
 #include <stdexcept>
 #include <string>
-#include <map>
+#include <vector>
+#include <utility>
 
 #ifndef FMT_USE_IOSTREAMS
 # define FMT_USE_IOSTREAMS 1
@@ -1671,7 +1672,7 @@ namespace internal {
 template <typename Char>
 class ArgMap {
  private:
-  typedef std::map<fmt::BasicStringRef<Char>, internal::Arg> MapType;
+  typedef std::vector<std::pair<fmt::BasicStringRef<Char>, internal::Arg> > MapType;
   typedef typename MapType::value_type Pair;
 
   MapType map_;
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR. One obvious improvement for large number of named arguments is to sort the vector and use binary search, but it can be done separately.

@jamboree, do you have any comments regarding this?

jamboree:
I'd have had used `boost::flat_map` if I could. That said, the source code is usually written by human, I think in practice there won't be too many arguments, so linear search is probably enough.

format.h
```diff
@@ -1680,7 +1681,12 @@ class ArgMap {
   FMT_API void init(const ArgList &args);
 
   const internal::Arg* find(const fmt::BasicStringRef<Char> &name) const {
-    typename MapType::const_iterator it = map_.find(name);
+    typename MapType::const_iterator it = map_.begin();
+    // the list is unsorted, so just return the first matching name.
+    for (; it != map_.end(); ++it) {
+      if (it->first == name)
+        break;
```
jamboree: I'd prefer return early here and prevent the unnecessary check in the end.

mwinterb:
My general thought on sorted or not matches @jamboree's, but if sorting is desired I'd rather get that out of the way now rather than need to break binary compat again later.

Also the handling of duplicate names is basically the same as it was before (ignoring them), but it doesn't match what Pythons str.format does. Should duplicated named arguments throw an exception or should they remain ignored?

jamboree:
@mwinterb, if you decide to check duplicate names, I'd suggest using assertions since it's definitely a programmer mistake that should be found in testing.

vitaut:
Another option is to use some kind of a hash table, but unless someone is really interested in investing time in benchmarking different options, I wouldn't worry about it and just used linear search.

I agree with @jamboree that the check for duplicate names, if introduced, better be done as an assertion.

vitaut:
Merged with a minor change in https://github.com/cppformat/cppformat/commit/06f3abe26dd1f6a410e5270b94696a9f40c6a2c6 to return early from `ArgMap::find` as suggested by @jamboree. Any improvements are welcome as separate PRs. Thanks!

mwinterb:
Thanks @vitaut!

And on throw vs assert, I was thinking that it'd be like if the format string was wrong, but since the format strings may come from an external source and argument names won't, assert is the right choice.

If the need for catching duplicate argument names comes up at some point I'll make a PR with that.

PR #267, opened 2016-01-31
Update project layout

niosHD:
This PR implements the change which has been discussed in #264.

It basically moves the source files into a cppformat subdirectory and performs a major update on the cmake scripts. This was necessary in order to properly test if the includes and defines are properly propagated with the library targets. I think that I could considerably simplify the build system which should also benefit the future development.

Please have a look at my changes and tell me what you think. :)


vitaut:
Thanks for your work on improving the build system! I'll look into this in more details, but for now just a quick comment that Appveyor build is failing: https://ci.appveyor.com/project/vitaut/cppformat/build/1.0.1521

niosHD:
I know, but fixing the windows build is non trivial on a Linux machine without any usable hints from the build log. Anyway, I think I figured it out now. ;)

vitaut:
> I know, but fixing the windows build is non trivial on a Linux machine without any usable hints from the build log. Anyway, I think I figured it out now. ;)

Thanks! I often break Windows build myself, and then have to fix it based on CI logs =). I've started looking at the commits and everything looks fine so far (nice job), just a few minor comments.

niosHD:
Thank you for the detailed review and the great support! It is very much appreciated! I will address your comments as soon as possible.

vitaut:
I've finally finished reviewing the PR (that was a lot of changes =)) and I should say your CMake-fu is much better than mine. Hope the amount of comments is not intimidating, these are mostly minor things. In general I think this is a great improvement and thanks again for working on it.

niosHD:
Not at all, you are welcome! I think I addressed most of your comments now. The only thing I have seen just now is that Travis finds some warnings which are not reported by my local compiler. I need to have a look at those next.

vitaut:
Merged, thanks!

niosHD:
Nice. Thank you!

daveisfera:
Is there a reason that the cppformat.cc is installed? Usually, only header files are installed.

Also, should the .cmake files be included in the packaging for Fedora? If so where?

One last thing is that in the .zip file, the .cmake files are in the the build directory, but that conflicts with the [build instructions](http://cppformat.github.io/latest/usage.html#building-the-library).

niosHD:
The `cppformat.cc` file is installed to make it possible to use the library in header-only mode (see `FMT_HEADER_ONLY` in the [README](https://github.com/cppformat/cppformat/blob/master/README.rst)). In this mode the code of the cc file is directly included into the h file (see [code](https://github.com/cppformat/cppformat/blob/52f89065e1843f4123198df326b480380d993312/cppformat/format.h#L3790)).

Yes, including the generated cmake files would be important. Otherwise the find "magic" does not work. Note that you have to install the cmake files which are installed when you call `make install`, or those that are part of the generated archive when you call `make package`. The files in the build directory can not be used for installation!

According to the [cmake documentation](https://cmake.org/cmake/help/v3.0/command/find_package.html) the following paths are searched by cmake on unix:
```
<prefix>/(lib/<arch>|lib|share)/cmake/<name>*/          (U)
<prefix>/(lib/<arch>|lib|share)/<name>*/                (U)
<prefix>/(lib/<arch>|lib|share)/<name>*/(cmake|CMake)/  (U)
```
I currently install to `<prefix>/lib/cppformat/` which corresponds to the second pattern. However, it would be possible to make this configurable if this path contradicts which some packaging guidelines of Fedora.

Could you elaborate on your last point? Exactly what .zip file are you talking about?

daveisfera:
Ok, header-only mode sounds like an interesting feature, but it's akin to static linking, which is [prohibited in Fedora](https://fedoraproject.org/wiki/Packaging:Guidelines#Packaging_Static_Libraries) (and many distributions), so I won't package the files to support it.

I'll include the .cmake files, but in Fedora, the path is /usr/share/<name>/cmake/, so can that be made configurable?

I was referring to the [.zip file for the release](https://github.com/cppformat/cppformat/archive/2.1.0.zip). In the .zip, there's a directory called build that has the .cmake files in it and so when you try to follow the build instructions, the mkdir fails.

niosHD:
Fair enough.

Making it configurable is no big deal, I'll follow up with a PR.

I have no idea why the build directory with the cmake files is in the release zip file but I am sure @vitaut can shed some light here.

PR #277, opened 2016-02-23
Fix switch fall-through warning

pweiskircher:
Clang with `-Wimplicit-fallthrough` enabled shows a warning here without
the break. See http://clang.llvm.org/docs/AttributeReference.html#fallthrough-clang-fallthrough


vitaut:
Thanks!

Issue #280, opened 2016-02-28
WinU Build error


VikingExplorer:
Has anyone tried to build this library in a Windows Universal context?

I'm getting:
1>c:\dev\couloir\3plibs\cppformat-2.0.0\format.cpp(740): error C3861: 'LocalFree': identifier not found
1>c:\dev\couloir\3plibs\cppformat-2.0.0\format.cpp(746): error C2065: 'FORMAT_MESSAGE_ALLOCATE_BUFFER': undeclared identifier

Steps to reproduce:
1) Open Visual Studio 2015 Community Edition, update 1
2) Create new project, Visual C++ ->Windows -> Windows Runtime Component (Universal Windows)
3) Add CppFormat source files
4) Build

vitaut:
I have no idea what Universal Windows is, but since targeting it gives an error for `LocalFree` which is defined in `windows.h`, you can probably fix it by defining `FMT_USE_WINDOWS_H` to 0.

VikingExplorer:
Thanks. The solution:
```
#ifdef WIN_U
#ifndef WIN32_LEAN_AND_MEAN
#define WIN32_LEAN_AND_MEAN
#endif
#define FMT_USE_WINDOWS_H 0
#define NOMINMAX						// this section needed to accomodate CppFormat
#include <algorithm>
namespace Gdiplus
{
	using std::min;
	using std::max;
}															// end CppFormat
#include <ppltasks.h>
#include <windows.h>
#endif
```

I'm surprised that anyone could be unaware of Windows 10. Already with 7x the Linux market share, there are about 200 million active monthly users. It's projected that 1 billion devices will be running Windows 10 within 2-3 years.

https://www.netmarketshare.com/operating-system-market-share.aspx?qprid=10&qpcustomd=0
http://www.extremetech.com/gaming/218039-steamos-ubuntu-or-windows-10-which-delivers-the-fastest-game-performance
http://winsupersite.com/windows-10/microsoft-confirms-windows-10-now-active-over-200-million-devices-monthly

vitaut:
I know what Windows 10 is, I don't know what Universal Window is and why it breaks WinAPI. Anyway, I'm glad that `FMT_USE_WINDOWS_H` works.

VikingExplorer:
WinU = WinAPI

amc522:
[...]

PR #285, opened 2016-03-03
Changed format_windows_error to not need LocalFree

mwinterb:
This is for non-'Desktop' applications that have a more limited collection of functions.
Addresses #280.

### DIFF
--- a/cppformat/format.cc
+++ b/cppformat/format.cc
@@ -588,27 +588,25 @@ FMT_FUNC void fmt::WindowsError::init(
 FMT_FUNC void fmt::internal::format_windows_error(
     fmt::Writer &out, int error_code,
     fmt::StringRef message) FMT_NOEXCEPT {
-  class String {
-   private:
-    LPWSTR str_;
-
-   public:
-    String() : str_() {}
-    ~String() { LocalFree(str_); }
-    LPWSTR *ptr() { return &str_; }
-    LPCWSTR c_str() const { return str_; }
-  };
   FMT_TRY {
-    String system_message;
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR, but why not free the buffer with `HeapFree (GetProcessHeap(), allocatedMessage)` as suggested in https://msdn.microsoft.com/en-us/library/windows/desktop/ms679351(v=vs.85).aspx instead of increasing the buffer size in a loop?

mwinterb:
The documentation for the define indirectly calls out that it wasn't available for "Windows Store apps" for Windows 8 and 8.1 and the latest Windows SDK only defines it for "desktop family" applications. #280 also pointed out that FORMAT_MESSAGE_ALLOCATE_BUFFER was not defined. It not being defined for non-desktop family applications if Windows 10 is being targeted is probably a Windows header file bug, but cppformat would still have to deal with old SDKs even if that gets fixed.

So, basically, just to maximize compatibility with as many different Windows "versions" as possible.

I need to also amend this PR with a test to make sure that TBS_E_PROVISIONING_NOT_ALLOWED works consistently with directly calling FormatMessageW(ALLOCATE) since that error code forces the retry, I just wasn't aware of that case when I started.

vitaut:
Fair enough. I'm not quite sure what `TBS_E_PROVISIONING_NOT_ALLOWED` is about but I'll wait for you to update the PR.

mwinterb:
The message text for `TBS_E_PROVISIONING_NOT_ALLOWED` is more than 500 characters long, so it'll just be to verify that the buffer is too small case is handled correctly.

vitaut:
Merged, thanks!

PR #286, opened 2016-03-03
Define FMT_NOEXCEPT empty when exceptions are disabled

Naios:
-> Fixes warnings about disabled exception support in MSVC

```
2>...\cppformat\cppformat\cppformat\posix.h(112): warning C4577: 'noexcept' used with no exception handling mode specified; termination on exception is not guaranteed. Specify /EHsc (compiling source file ...\cppformat\cppformat\cppformat\posix.cc)
2>...\cppformat\cppformat\cppformat\format.cc(143): warning C4577: 'noexcept' used with no exception handling mode specified; termination on exception is not guaranteed. Specify /EHsc
```

### DIFF
--- a/cppformat/format.h
+++ b/cppformat/format.h
@@ -150,21 +150,6 @@ typedef __int64          intmax_t;
 # include <utility>  // for std::move
 #endif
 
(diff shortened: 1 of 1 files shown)


vitaut:
Merged, thanks!

PR #312, opened 2016-05-03
Fix user-defined literal detection for Intel C++ compiler (#311)

### LINKED ISSUE #311: User-defined literals check for Intel C++ compiler
When compiling with the Intel C++ compiler version 14 using c++11 mode, the check for whether user-defined literals can be used does not work. Including `format.h` leads to the error:
```
include/cantera/ext/format.h(3223): error: expected an operator
  operator"" _format(const char *s, std::size_t) { return {s}; }
```
This compiler version does not support user-defined literals (support is introduced in version 15, per https://software.intel.com/en-us/articles/c0x-features-supported-by-intel-c-compiler). However, the macro `FMT_USE_USER_DEFINED_LITERALS` ends up being defined only by the GCC version.

The Intel compiler version can be checked using the macros `__ICC` on Linux/OSX and `__ICL` on Windows, which returns the number `1400` for my compiler version. There is apparently no macro that is actually cross-platform.

I've only checked this on version 2.1, but it looks like the test for this hasn't changed in master.

dean0x7d:
Fixes #311.

The `FMT_USE_USER_DEFINED_LITERALS` check is beginning to resemble a dumpster fire, but as per Intel's [C++11 feature notes](https://software.intel.com/en-us/articles/c0x-features-supported-by-intel-c-compiler) both the system compiler (gcc/msc) and icc itself must support the given feature:

> On Windows: when using Intel C++ compiler with Visual Studio 2010* or 2012*, the C++11 features supported by Visual C++ 2010/2012 are enabled by default. Use "/Qstd=c++11" to turn on the support for all other cases.
> On Linux or Mac OS X: the C++11 features supported by gcc on the path are enabled by default. Use "-std=c++11" to turn on the support for all other cases.

To make sure the UDL detection still works on any compiler I added a new compile test which checks that the feature detection in the header matches the one in CMake. `CMAKE_CXX_COMPILER` and `CPP11_FLAG` are passed to `compile-test` to ensure the same configuration.

Tested and working with Intel C++ 14.0-16.0 on Linux and 16.0 on Windows.

### DIFF
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -105,7 +105,7 @@ typedef __int64          intmax_t;
 # define FMT_ICC_VERSION __ICL
 #endif
 
-#if defined(__clang__) && !defined(__INTEL_COMPILER)
+#if defined(__clang__) && !defined(FMT_ICC_VERSION)
 # pragma clang diagnostic push
 # pragma clang diagnostic ignored "-Wdocumentation"
 #endif
@@ -3915,7 +3915,7 @@ operator"" _a(const wchar_t *s, std::size_t) { return {s}; }
 # pragma GCC diagnostic pop
 #endif
 
-#if defined(__clang__) && !defined(__INTEL_COMPILER)
+#if defined(__clang__) && !defined(FMT_ICC_VERSION)
 # pragma clang diagnostic pop
 #endif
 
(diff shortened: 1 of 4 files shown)


vitaut:
Merged, thanks a lot for fixing this!

PR #313, opened 2016-05-03
Fix a few bugs related to compile tests

dean0x7d:
I noticed some compile test bugs while working on #312 (but these bugs aren't actually specific to Intel's compiler).

1. In `expect_compile` and `expect_compile_error` the cache variable was not always unset because `message(FATAL_ERROR)` exits immediately. Even if a test was fixed it would appear broken until the cache was reset manually.

2.  The expected-to-fail `FMT_STATIC_ASSERT` test was failing for the wrong reason: missing macro instead of the actual static assert.

3.  If the main project was configured with a non-default compiler, `find-package-test` failed with linker errors because `ctest --build-and-test` would still use the system default.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -137,8 +137,10 @@ if (FMT_PEDANTIC)
     "${CMAKE_CURRENT_BINARY_DIR}/find-package-test"
     --build-generator ${CMAKE_GENERATOR}
     --build-makeprogram ${CMAKE_MAKE_PROGRAM}
-    --build-options "-DFMT_DIR=${PROJECT_BINARY_DIR}"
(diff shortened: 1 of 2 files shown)


vitaut:
Great fixes, thanks!

vitaut:
@dean0x7d, could you please review the updated [CONTIBUTING](https://github.com/fmtlib/fmt/blob/ebec00138bd10e949ff34955bcb096387cd32acc/CONTRIBUTING.md) document, particularly the part about licensing, and let me know if you agree with it being applied both to your contributions to {fmt}? The library is likely to be relicensed (#1073) so I'm collecting approval from all earlier contributors.

Issue #327, opened 2016-05-17
Android build is failing because of localeconv


dpantele:
Locale support added the following issue with the android build:

```
/tmp/asdk/android-ndk-r10e/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/bin/arm-linux-androideabi-g++ -MMD -MP -MF ./obj/local/armeabi-v7a/objs-debug/hello-jni/hello-jni.o.d -fpic -ffunction-sections -funwind-tables -fstack-protector-strong -no-canonical-prefixes -march=armv7-a -mfpu=vfpv3-d16 -mfloat-abi=softfp -fno-exceptions -fno-rtti -mthumb -Os -g -DNDEBUG -fomit-frame-pointer -fno-strict-aliasing -finline-limit=64 -O0 -UNDEBUG -marm -fno-omit-frame-pointer -I/tmp/asdk/android-ndk-r10e/sources/cxx-stl/gnu-libstdc++/4.9/include -I/tmp/asdk/android-ndk-r10e/sources/cxx-stl/gnu-libstdc++/4.9/libs/armeabi-v7a/include -I/tmp/asdk/android-ndk-r10e/sources/cxx-stl/gnu-libstdc++/4.9/include/backward -Ijni/../../fmt-3.0.0 -Ijni -DANDROID  -Wa,--noexecstack -Wformat -Werror=format-security  -fexceptions     -I/tmp/asdk/android-ndk-r10e/platforms/android-19/arch-arm/usr/include -c  jni/hello-jni.cpp -o ./obj/local/armeabi-v7a/objs-debug/hello-jni/hello-jni.o 
In file included from jni/hello-jni.cpp:19:0:
jni/../../fmt-3.0.0/fmt/format.h: In member function 'void fmt::BasicWriter<Char>::write_int(T, Spec)':
jni/../../fmt-3.0.0/fmt/format.h:2756:45: error: 'struct lconv' has no member named 'thousands_sep'
     fmt::StringRef sep = std::localeconv()->thousands_sep;
```

The problem is that in some implementation the lconv structed is just stubbed:
```
#if 1 /* MISSING FROM BIONIC - DEFINED TO MAKE libstdc++-v3 happy */
struct lconv { };
struct lconv *localeconv(void);
#endif /* MISSING */
```

However, c++ api exists, and everything compiles after the patch: https://gist.github.com/dpantele/d2e2aec8ff23b0208245c8a6e882f7fe

[...]

PR #328, opened 2016-05-17
Workaround for missing lconv support in android

dpantele:
Fixing https://github.com/fmtlib/fmt/issues/327

Better to check if this API really works on Android. 
Also please check if that is an OK use for the StringRef.

### DIFF
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -910,6 +910,21 @@ class ThousandsSep {
   }
 };
 
+// Returns the thousands separator for the current locale.
+// On android the lconv structure is stubbed using an empty structure
+// The test is for the size only, not for the presense of 
+// thousands_sep in std::lconv, because if one would add thousands_sep
+// at some point, the size of structure would be at least sizeof(char*)
+template<typename Lconv, bool=(sizeof(Lconv) >= sizeof(char*))>
+struct Locale {
+  static fmt::StringRef thousands_sep() { return ""; }
+};
+
+template<typename Lconv>
+struct Locale<Lconv, true> {
+  static fmt::StringRef thousands_sep() { return static_cast<Lconv*>(std::localeconv())->thousands_sep; }
+};
+
 // Formats a decimal unsigned integer value writing into buffer.
 // thousands_sep is a functor that is called after writing each char to
 // add a thousands separator if necessary.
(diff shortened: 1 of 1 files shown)


dpantele:
@vitaut , do you have any idea why mingw fails and VS is fine? I don't have mingw environment at the moment

dpantele:
What I could do blindly is to catch `std::runtime_error` and use std::locale("") in that case.

vitaut:
Thanks for the PR. According to [this thread](https://sourceforge.net/p/mingw/mailman/message/31768770/):

> The GCC's libstdc++ does not support locales other than the C locale on MinGW. Somebody has to implement it first.

So I changed it back to C locale but added a detection for empty `lconv` in https://github.com/fmtlib/fmt/commit/3cefa72f2019616cc3610df9554743f74e4ade92. It would be even better to detect `lconv::thousands_sep` is there with SFINAE, but this simple solution should do for now. Does it work for you?

Gachapen:
Hi, I have the same issue, but your fix (3cefa72) does not work. It's choosing the default `thousands_sep`:
```
In file included from fmt/posix.h:48:0,
                 from fmt/posix.cc:33:
fmt/format.h: In function 'fmt::StringRef fmt::internal::thousands_sep()':
fmt/format.h:915:60: error: 'struct lconv' has no member named 'thousands_sep'
 fmt::StringRef thousands_sep() { return std::localeconv()->thousands_sep; }
```

Anything I can do to help?

vitaut:
@Gachapen Thanks for testing it. Could you check what is `sizeof(lconv)` on Android? I guess the size check is not very reliable, maybe I should use a proper SFINAE detection for `thousands_sep`.

Gachapen:
@vitaut Just checked, it is 1.

vitaut:
@Gachapen Thanks. Just realized that my "fix" is rubbish because the `thousands_sep` template is compiled anyway.

Gachapen:
@vitaut Hehe, yes I see that now too :) 

This is a bit too complex for me (I'd just `#ifdef` it). I'll happily test any fix you can come up with.

dpantele:
I will do the proper SFINAE version right now.

dpantele:
SFINAE with c++98 looks really messy... What do you think about this approach? https://github.com/dpantele/fmt/commit/998bd2c40628128abbfa9807ce73d8f2a2308c5d

vitaut:
@dpantele Definitely looks better but I wonder why the tests are failing on mingw.

dpantele:
@vitaut I haven't pushed changes to the pull request branch, now all builds pass.

Gachapen:
I can confirm that this now builds. Though is it safe to assume that the size should be less or equal to 1 when it has no members?

vitaut:
@dpantele Got it. Could you squash the commits and rebase to current master?

@Gachapen You are right, the size check is not very robust. A better option is to detect `lconv::thousands_sep` specifically using something like http://stackoverflow.com/a/1007175/471164, but it will require more work.

dpantele:
Ok, reabsed and cleaned-up in https://github.com/fmtlib/fmt/pull/328/commits/45a150945597b95102656c8f7a5819e08cf244ae

vitaut:
Merged, thanks!

PR #333, opened 2016-05-24
Create doc/html directory

jeffs:
The html directory must exist before the documentation can be built, or
else `make` reports the following error:

    CMake Error at doc/cmake_install.cmake:31 (file):
      file INSTALL cannot find "/Users/jeff/pkg/fmtlib/fmt/doc/html".
    Call Stack (most recent call first):
      cmake_install.cmake:33 (include)

    make: *** [install] Error 1

This commit is just a work-around; I don't know enough CMake to update the build, but it probably should just create the html directory automatically if necessary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ bin/
 /_CPack_Packages
 /doc/doxyxml
 /doc/html
+!/doc/html
 virtualenv
 /Testing
 /install_manifest.txt


.gitignore
```diff
@@ -2,6 +2,7 @@ bin/
 /_CPack_Packages
 /doc/doxyxml
 /doc/html
```
Spartan322: why not just remove this instead?

jeffs: Maybe that is a better approach.  Why was the html directory ignored in the first place?  It seemed like it might contain generated files (from *.rst) that should not be committed to version control.

vitaut:
Thanks for the PR, but the `html` directory should be created automatically when running `make doc`. What exactly are exact cmake and make commands that you run?

jeffs:
cmake -DCMAKE_INSTALL_PREFIX:PATH="$top/usr" .
make -j8 all install

This morning, the build work fines for me, though none of the code has
changed.  Is it possible the "-j" flag to `make` just caused a race
condition?

On Tue, May 24, 2016 at 9:43 AM, Victor Zverovich <notifications@github.com>
wrote:

> Thanks for the PR, but the html directory should be created automatically
> when running make doc. What exactly are exact cmake and make commands
> that you run?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/fmtlib/fmt/pull/333#issuecomment-221273118>
>



-- 

[image: Orchard Platform] <http://www.orchardplatform.com/>

Jeffrey Schwab / Senior Software Engineer
(617) 407-2542 / jeff@orchardplatform.com

Orchard Platform
101 5th Avenue, 4th Floor, New York, NY 10003
http://www.orchardplatform.com

[image: Twitter]  <https://twitter.com/orchardplatform>[image: Linkedin]
<https://www.linkedin.com/company/orchard-app>

Orchard Blog <http://www.orchardplatform.com/blog/> | Marketplace Lending
Meetup <http://www.meetup.com/Peer-to-Peer-Lending-P2P/>

-- 
*NOTICE TO RECIPIENTS*: This communication is confidential and intended for 
the use of the addressee only. If you are not an intended recipient of this 
communication, please delete it immediately and notify the sender by return 
email. Unauthorized reading, dissemination, distribution or copying of this 
communication is prohibited. This communication does not constitute an 
offer to sell or a solicitation of an indication of interest to purchase 
any loan, security or any other financial product or instrument, nor is it 
an offer to sell or a solicitation of an indication of interest to purchase 
any products or services to any persons who are prohibited from receiving 
such information under applicable law. The contents of this communication 
may not be accurate or complete and are subject to change without notice. 
As such, Orchard App, Inc. (including its subsidiaries and affiliates, 
"Orchard") makes no representation regarding the accuracy or completeness 
of the information contained herein. The intended recipient is advised to 
consult its own professional advisors, including those specializing in 
legal, tax and accounting matters. Orchard does not provide legal, tax or 
accounting advice.

vitaut:
The problem is that the `install` target tries to install docs but they are only built with `make doc`. Should be fixed in https://github.com/fmtlib/fmt/commit/dc1e36fa0e4e3d69f11cb028a0b56910a6fa3c3f. Thanks for catching this!

PR #348, opened 2016-06-25
Fix building with clang in header-only mode

LogicalKnight:
Building under a recent clang compiler causes the use of an extern
template for the BasicData struct. However, the only instantiation of
that structure in format.cc is only done if FMT_HEADER_ONLY is not
defined. This causes the build to fail in C++11 or C++14 mode.
Therefore, only declare the BasicData template as extern if
FMT_HEADER_ONLY is not defined.


LogicalKnight:
Quick note: I'm working with Apple's Clang version 7.3.0. This change allows the code to compile in header-only mode under clang when compiling in either C++11 or C++14 mode. Thanks!

vitaut:
Good catch, thanks!

Issue #147, opened 2015-04-08
Faster float format


newnon:
Have you seen this project? They have fast float to string conversions
https://code.google.com/p/stringencoders/

vitaut:
Thanks for the link, @newnon. I haven't seen this before but I have plans to implement faster floating-point formatting, so it might be worth looking at the methods used in stringencoders too.

vitaut:
There is also relevant discussion here: https://github.com/cppformat/cppformat/pull/116

vitaut:
A custom implementation of floating-point formatting will also allow locale-independent formatting as in Python.

NotImplemented:
@vitaut, what will be the fastest way to ramp-up and contribute?

vitaut:
@NotImplemented Awesome that you interested in contributing! I think for this particular issue we need an implementation of one of the Grisu algorithms described in [Printing Floating-Point Numbers Quickly and Accurately with Integers](http://www.cs.tufts.edu/~nr/cs257/archive/florian-loitsch/printf.pdf) to be used instead of `printf` for floating-point formatting in https://github.com/cppformat/cppformat/blob/96d518f2f88c0c594e29f1603b89049694b3b6b0/format.cc#L549-L575. This might be a lot of work though. A somewhat simpler thing would be to implement a custom stream buffer (#92).

NotImplemented:
@newnon, conversion in stringencoders is not correct. Have you seen lines in modp_dtoa()?
```cpp
    if (prec < 0) {
        prec = 0;
    } else if (prec > 9) {
        /* precision of >= 10 can lead to overflow errors */
        prec = 9;
    }
```
Here is benchmark of several conversion algorithms I have done:

Algorithm | Speed
------------ | -------------
fmt::sprintf("%17g") 	| 17.326
sprintf("%17g") 	| 14.321
grisu2 			| 2.145
grisu 			| 8.872

newnon:
No i haven't look inside may be you are right and grisu2 is the fastest algorithm. Unfortunately i'm not to strong in math. Look on https://github.com/miloyip/dtoa-benchmark may be it possible to take some code from there

manylegged:
FWIW googles double-conversion library is the fastest according to milo's benchmarks (the "milo" implementation he references is pretty much double-conversion with some special cases removed) and the interface looks like it would ammenable to formatting requirements. https://github.com/google/double-conversion

NotImplemented:
[...]

PR #364, opened 2016-07-21
Allow use of "," format specifier as in Python

benesch:
f68771a helpfully enabled the "n" typecode for locale-aware printing of
integers, but not for floats. This commit enables the use of "n" for
locale-aware printing of floats, as in Python, and additionally
introduces the "," format specifier, which can be applied to "d", "f",
and "g" (and their upper-case variants, when present) to force
locale-aware grouping.

"," is particularly useful for floats, as "n" will use exponential
notation for large floats, while ",f" never will.

Unlike in Python, there is no distinction between ",d" and "n" for
integers, or ",g" and "n" for floats. (In Python, "," will always output
a literal comma, while "n" is locale-aware.) In fmtlib, both "," and "n"
are locale-aware.

### DIFF
--- a/doc/syntax.rst
+++ b/doc/syntax.rst
@@ -74,14 +74,14 @@ although some of the formatting options are only supported by the numeric types.
 The general form of a *standard format specifier* is:
 
 .. productionlist:: sf
-   format_spec: [[`fill`]`align`][`sign`]["#"]["0"][`width`]["." `precision`][`type`]
+   format_spec: [[`fill`]`align`][`sign`]["#"]["0"][`width`][","]["." `precision`][`type`]
    fill: <a character other than '{' or '}'>
    align: "<" | ">" | "=" | "^"
    sign: "+" | "-" | " "
    width: `integer` | "{" `arg_id` "}"
    precision: `integer` | "{" `arg_id` "}"
    type: `int_type` | "c" | "e" | "E" | "f" | "F" | "g" | "G" | "p" | "s"
-   int_type: "b" | "B" | "d" | "o" | "x" | "X"
+   int_type: "b" | "B" | "d" | "n" | "o" | "x" | "X"
 
 The *fill* character can be any character other than '{' or '}'.  The presence
 of a fill character is signaled by the character following it, which must be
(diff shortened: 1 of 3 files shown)


benesch:
The fact that `write_int` does its own string conversion but `write_double` defers to `snprintf` makes testing this annoying, but it does the job on every `snprintf` implementation I could find!

Let me know what you think!

doc/syntax.rst
```diff
@@ -144,11 +144,12 @@ decimal-point character appears in the result of these conversions
...
 
-.. ifconfig:: False
-
-   The ``','`` option signals the use of a comma for a thousands separator.
-   For a locale aware separator, use the ``'n'`` integer presentation type
-   instead.
+The ``","`` option signals the use of a locale-aware thousands
+separator. In many locales, this may not be an actual `,` character.
+This option is supported for types ``'d'``, ``'f'``, ``'g'``, ``'F'``,
+``'G'``, and none. Type ``'n'`` always implies the use of a local-aware
```
vitaut: local -> locale

vitaut:
Looks great, thanks for contributing! A few comments:

1. Appveyor builds fail with "Incorrect format specifier" assert: https://ci.appveyor.com/project/vitaut/fmt/build/build%20211/job/4i5fbgjupuyd54sd#L732

2. Do I understand correctly that `,` is only locale-aware because floating-point formatting relies on `sprintf`? I'd rather follow Python's convention and make `,` locale-unaware in the future, but it's OK for now.

Regarding `write_double`, the long-term plan is to have our own implementation of floating-point formatting (#147).

benesch:
Ah, shoot! AppVeyor's failure suggests that Visual Studio doesn't support the `%'f` format code for formatting floats with locale-aware grouping. Looks like we'll need to defer this PR until #147 is resolved. On the bright side, it's `snprintf` that's preventing a distinction between a locale-aware `n` and a non-locale-aware `,` specifier.

Thanks for a great library, by the way!

vitaut:
Closing this PR for now then =(.

PR #366, opened 2016-08-04
FMT_USE_CPP11 option pollutes build flags with '-std=c++11'

avalchev:
When included (with add_subdirectory), `FMT_USE_CPP11` option pollutes build flags with `-std=c++11`.  If master project has `set(CMAKE_CXX_STANDARD 14)` , then compilation flags become `-std=c++11 -std=gnu++14`.

Enable `FMT_USE_CPP11` by default only when `MASTER_PROJECT` is `ON`, otherwise inherit from master project.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,7 +23,9 @@ option(FMT_PEDANTIC "Enable extra warnings and expensive tests." OFF)
 option(FMT_DOC "Generate the doc target." ${MASTER_PROJECT})
 option(FMT_INSTALL "Generate the install target." ${MASTER_PROJECT})
 option(FMT_TEST "Generate the test target." ${MASTER_PROJECT})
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR! I agree that the library shouldn't add `-std=c++11` to compile flags if the parent project already specifies the standard. However, I suggest checking `CMAKE_CXX_STANDARD` instead of `MASTER_PROJECT`. Then `-std=c++11` will be defined if the parent project doesn't set the standard itself.

vitaut:
Does https://github.com/fmtlib/fmt/commit/12a6027b40ee0fa9e744d1d25333a2650fd9aed7 work for you?

vitaut:
Closing this in favor of https://github.com/fmtlib/fmt/commit/12a6027b40ee0fa9e744d1d25333a2650fd9aed7, but comments are welcome.

Issue #379, opened 2016-09-07
Format arguments serialization/deserialization


chronoxor:
Hello! I'd like to use your format library in a project where we need to store format strings with arguments in a queue and process them later in a background thread. Is it theoretically possible to serialize somehow ArgList in memory buffer and then restore in order to call fmt::format() with string pattern later?

vitaut:
It is not possible to serialize `ArgList` in general, because it may contain pointers to arbitrary objects the types of which are only known in the formatting functions. However, you can define a variadic function that takes a format string and arguments and copies them into a queue. Then you can construct `ArgList` in another thread and populate it with these arguments.

chronoxor:
Yes of course I'm going to support format messages and plain data parameters that could be serialize only by value (e.g. int, double). For strings maybe it'll be possible to serialize char buffer + size directly into the buffer.  

Could you please briefly describe the roadmap or provide a code fragment how such methods might be implemented with variadic function to catch an idea?

void Serialize(Buffer& buffer, ArgList args)
ArgList Deserialize(const Buffer& buffer)

vitaut:
You can use `ArgMap::init` as an example of how to iterate over `ArgList`: https://github.com/fmtlib/fmt/blob/ab054532ce255c6aed60f5427ceda8748db21dee/fmt/format.cc#L407. For each argument you can serialize type and its value. Note that will require using some internal APIs.

chronoxor:
https://github.com/fmtlib/fmt/pull/382

chronoxor:
Travis-ci build still not pass for OSX in release mode for my serialize/deserialize unit test. I think its compiler performs huge optimization and deletes buffer before fmt::format() is called.

std::string serialize_deserialize(const char *format_str, fmt::ArgList args) {
  std::vector<uint8_t> buffer;
  args.serialize<char>(buffer);
  args = args.deserialize<char>(buffer);
  std::string result = fmt::format(format_str, args);
  buffer.clear();
  return result;
}

As I have no OSX I cannot catch and debug it in a proper way. Linux and Windows builds works properly in both debug and release modes!

vitaut:
Thanks for the PR, but since this functionality is not directly related to formatting, I don't think this needs to be in the core library. Also serialization of `CustomValue` won't work in general because it may refer to an arbitrary user-defined object.

chronoxor:
[...]

PR #382, opened 2016-09-19
Format arguments serialization/deserialization

chronoxor:
I've implemented cross-thread ArgList serialization/deserialization with support of base types (char, int, double), strings (c-string, std::string, std::wstring), named arguments and custom types (serialize all plain fields by value - as sizeof(Date) class to serialize year_, month_, day_ fields).

Also I provided my implementation with some unit tests which covers the most common cases.


vitaut:
Closed as discussed in #379.

PR #384, opened 2016-09-22

chronoxor:


vitaut:
Could you squash the commits? Also why do you need to add `size` to `CustomValue`?

PR #386, opened 2016-09-25

chronoxor:
…ty components


chronoxor:
I need CustomValue.size to be able to serialize a custom value in buffer by its size, then in deserialize phase I just update CustomValue.value with a pointer inside buffer. For plain structs it works well:

class Date {
  int year_, month_, day_;
};

PR #387, opened 2016-09-25

chronoxor:
I need CustomValue.size to be able to serialize a custom value in buffer by its size, then in deserialize phase I just update CustomValue.value with a pointer inside buffer. For plain structs it works well:

class Date {
int year_, month_, day_;
};

### DIFF
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -2812,7 +2812,10 @@ void BasicWriter<Char>::write_int(T value, Spec spec) {
   }
   case 'n': {
     unsigned num_digits = internal::count_digits(abs_value);
-    fmt::StringRef sep = internal::thousands_sep(std::localeconv());
+    fmt::StringRef sep = "";
+#ifndef ANDROID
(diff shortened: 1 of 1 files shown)


vitaut:
There are two problems with `CustomValue::size`:

1. It is unsafe in general because it only works for trivially copyable types and no checks are performed to guarantee this.
2. It adds an overhead for the common case when serialization is not used.

Therefore, I suggest using variadic functions and serialize arguments directly instead of going through `ArgList`:

```c++
void serialize(Buffer &buffer, const char *s) {
  // serialize string
}

void serialize(Buffer &buffer, int n) {
  // serialize int
}

// ...

template <typename T, typename... Args>
void serialize(Buffer &buffer, const T &first, const Args &...args) {
  serialize(buffer, first);
  serialize(buffer, args);
}

template <typename... Args>
void serialized_format(const char *fmt, const Args &...args) {
  Buffer buffer;
  serialize(buffer, fmt);
  serialize(buffer, args...);
  // ...
}
```

PR #389, opened 2016-09-27

chronoxor:
I removed CustomValue.size overhead in my PR and keep only types() accessors in ArgList


vitaut:
Merged, thanks!

PR #390, opened 2016-09-27
Introduced NamedArgWithType<Char, T> : NamedArg<Char>

chronoxor:
Hello!

I found that if I put custom type as a named argument it is impossible to get its size in variadic functions:

template < typename Char >
void serialize(Buffer &buffer, const fmt::internal::NamedArg< Char >& arg) {
  // serialize named argument
}

So I extend fmt::internal::NamedArg with no overhead:

template < typename Char, typename T >
struct NamedArgWithType : NamedArg< Char > {
  NamedArgWithType(BasicStringRef< Char > argname, const T &value)
  : NamedArg< Char >(argname, value) {}
};

With this fix I was able to cover all args serializaton/deserializaton for base and trivially copyable types. 

Hope you'll accept this PR as well! Thanks in advance!

### DIFF
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -1041,6 +1041,8 @@ struct Arg : Value {
 
 template <typename Char>
 struct NamedArg;
+template <typename Char, typename T>
+struct NamedArgWithType;
 
 template <typename T = void>
 struct Null {};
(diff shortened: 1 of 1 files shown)


vitaut:
Merged, thanks!

Issue #392, opened 2016-10-04
Simple example for how to format a custom type (without a stream output operator)


pwm1234:
I have read the documentation, looked over the unit tests, and searched around for an example, but I am still not sure exactly how to implement custom formatting for a user defined type. Suppose I have a structure

    struct MyStruct { double a, b, c, d; }

What do I need to do so that the following code works:

    MyStruct s;
    std::cout << fmt::format("s={}", s) << std::endl;

where the custom formatted output will use something similar to:

    fmt::format("[MyStruct: a={:.1f}, b={:.2f}, c={:.3f}, d={:.4f}]", s.a, s.b, s.c, s.d)

I know about how fmt will use a custom stream output operator, but I do not have one (and do not want one) for the structure type I want to format. I have looked over the ``ArgVisitorTest`` in ``util-test.cc``, but it does not really help me. What is a simple way to have custom formatting (without resorting to a custom stream output operator <<)?

By the way, I REALLY like fmt, and I will gladly provide a pull request with documentation updates once I understand how to do this.

Thanks!

vitaut:
If you don't want to overload `operator<<`, you can provide an overload of `format_arg` for yout type instead:

```c++
struct MyStruct { double a, b, c, d; };

template <typename ArgFormatter>
void format_arg(fmt::BasicFormatter<char, ArgFormatter> &f,
                const char *&format_str, const MyStruct &s) {
  f.writer().write("[MyStruct: a={:.1f}, b={:.2f}, c={:.3f}, d={:.4f}]",
                   s.a, s.b, s.c, s.d);
}

MyStruct s = {1, 2, 3, 4};
std::cout << fmt::format("s={}", s) << std::endl;
```

pwm1234:
Thanks! This did the trick for me. (I am however, using a slightly older version of fmt, from 8/10/2016, so the required function name is ``format``; it looks like it was renamed to ``format_arg`` on 8/14.) Is now a good time to update to the latest commit? Any idea when you will make another release?

In reviewing the documentation, it looks like a good place to mention how to do this is right after the "Date and Time formatting" section.

vitaut:
Right, `format` are was renamed to `format_arg` in this context to avoid ambiguity with the main `format` function and improve error reporting.

> Is now a good time to update to the latest commit?

Yes, the master is in good state right now.

> Any idea when you will make another release?

[...]

PR #393, opened 2016-10-05
Document use of format_arg for user-defined type

pwm1234:
This is my attempt to provide documentation for how to use format_arg so one may use a user-defined type in a call to format. This documents the discussion around issue #392. Thanks for your help. I hope you find this contribution useful.


vitaut:
Merged. Thanks for contributing!

PR #397, opened 2016-10-11
Remove unnecessary "fmt/" prefix inside fmt directory

chronoxor:
I suggest removing "fmt/" prefix in *.h, *.cc files inside fmt directory. Consider someone using fmt library as a git submodule or third-party library and has format.h helper file with direct referencing: 

#include "../../modules/fmt/fmt/format.h"
#include "../../modules/fmt/fmt/ostream.h"

Nevertheless I put explicit relative path to format.h and ostream.h I need to add include_directories("modules/fmt") anywhere I use my format.h helper, because ostream.h contins the following inside:

#include "fmt/format.h"

and I believe it could be easily replaced without any side effects with  

#include "format.h"

### DIFF
--- a/fmt/format.cc
+++ b/fmt/format.cc
@@ -25,8 +25,8 @@
  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include "fmt/format.h"
-#include "fmt/printf.h"
+#include "format.h"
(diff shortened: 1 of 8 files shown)


vitaut:
The `fmt` prefix does seem unnecessary because includes in the same directory take precedence over the ones on the search paths. Merged, thanks!

PR #402, opened 2016-10-29
Fix Linux compilation with -fno-exceptions

JanHellwig:
In some places `throw` was used instead of `FMT_THROW` which caused the compilation to fail when exceptions are disabled.
I didn't change this in the tests but only in the library itself which might be used by projects that are compiled without exceptions.


vitaut:
Merged, thanks!

PR #403, opened 2016-10-30
fix clang warnings

Lectem:
When building with c++14 and -Weverything, I encountered the following warnings which are now fixed : 
-  dynamic exception specifications are deprecated [-Wdeprecated]
- 'FMT_USE_OVERRIDE' is not defined, evaluates to 0 [-Wundef]
- definition of implicit copy constructor for 'FormatError' is deprecated because it has a user-declared destructor [-Wdeprecated]

### DIFF
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -194,8 +194,23 @@ typedef __int64          intmax_t;
 # endif
 #endif
 
+// This is needed because GCC still uses throw() in its headers when exceptions are disabled
+#ifndef FMT_DESTRUCTOR_NOEXCEPT
+# if FMT_EXCEPTIONS || FMT_GCC_VERSION
+#  if FMT_USE_NOEXCEPT || FMT_HAS_FEATURE(cxx_noexcept) || \
+    (FMT_GCC_VERSION >= 408 && FMT_HAS_GXX_CXX11) || \
+    FMT_MSC_VER >= 1900
+#   define FMT_DESTRUCTOR_NOEXCEPT noexcept
+#  else
+#   define FMT_DESTRUCTOR_NOEXCEPT throw()
+#  endif
+# else
+#  define FMT_DESTRUCTOR_NOEXCEPT
+# endif
+#endif
+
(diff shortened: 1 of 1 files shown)


Lectem:
To make GCC compile while removing the deprecated warning of throw() for clang, I had to introduce a new define FMT_DESTRUCTOR_NOEXCEPT as the headers would still use throw() even when exceptions are disabled. It has the same behaviour as FMT_NOEXCEPT for compilers other than GCC.
It seems to be more of a libstdc++11 problem though, I'm not sure if it could pose problems for some versions of clang.

foonathan:
I am personally not happy with the destructor noexcept workaround (but @vitaut opinion counts).
The problem could be solved more elegantly if we remove the destructor in the exception classes, because the implicitly generated one *should* be `throw()` in those cases automatically (IIRC). I'm guessing the destructors are there to avoid the weak vtable warning?

Anyways, I'd merge that PR (and would include it in the 3.0.1 I was just about to release), but I'm waiting on @vitaut opinion here.

Lectem:
I wanted to remove the destructors at first too but I suspected they were added for a reason, it would be nice to confirm why.

vitaut:
Merged with minor changes in https://github.com/fmtlib/fmt/commit/a5b9611e48f5a0e48c21c9c4c0aa88d3ddd8f61a. Thanks, @Lectem!

@foonathan, I'd prefer to remove dtors too, but as you correctly guessed they were added to prevent `-Wweak-vtables` warning in clang (https://github.com/fmtlib/fmt/commit/b26e76efe9417ea773bb1e6c6ce875646c71a346). Alternative suggestions are welcome.

vitaut:
Hmm, the fix caused failures on Travis, so I moved it to the https://github.com/fmtlib/fmt/tree/warnings branch for further investigation.

vitaut:
Merged to `master`: https://github.com/fmtlib/fmt/commit/8f455c10b0f1aee898f71313b21c43b098dadb34.

PR #407, opened 2016-11-01
Support Unicode filenames on Windows

codicodi:
This change allows to open files with UTF-8 encoded strings on Windows (no codepage nonsense). 
Although I am not quite familiar with testing methods used here and therefore have no clue why it fails checks.

### DIFF
--- a/fmt/posix.cc
+++ b/fmt/posix.cc
@@ -64,8 +64,9 @@ inline std::size_t convert_rwcount(std::size_t count) { return count; }
 
 std::FILE* u8fopen(fmt::CStringRef filename, fmt::CStringRef mode) {
 #ifdef _WIN32
-  fmt::UTF8ToUTF16 conv(filename.c_str());
-  return FMT_SYSTEM(_wfopen(conv.c_str(), modebuff));
+  fmt::UTF8ToUTF16 nconv(filename.c_str());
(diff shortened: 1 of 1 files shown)


vitaut:
I think using UTF-8 for file names is a good idea, but please use [`UTF8ToUTF16`](https://github.com/fmtlib/fmt/blob/master/fmt/format.h#L980) instead of `MultiByteToWideChar`. It will do all the buffer management and error handling automatically.

fmt/posix.cc
```diff
@@ -61,6 +61,27 @@ typedef ssize_t RWResult;
...
+}
+
+int u8open(fmt::CStringRef filename, int flags, int mode) {
+#ifdef _WIN32
+  fmt::UTF8ToUTF16 conv(filename.c_str());
+  int fd = -1;
+  FMT_POSIX(wsopen_s(&fd, conv.c_str(), flags, _SH_DENYNO, mode));
+  return fd;
+#else
+  return FMT_POSIX_CALL(open(filename, flags, mode));
```
vitaut: This doesn't compile because `open` expects a C string. BTW why not pass `const char *` instead of `CStringRef` here and in `u8fopen` to avoid calls to `c_str`?

vitaut:
Closing for now because of compilation errors.

Issue #408, opened 2016-11-02
Consider being "Wzero-as-null-pointer-constant"-clean


alabuzhev:
gcc 4.7 added a new command-line option `-Wzero-as-null-pointer-constant` to warn when a literal '0' is used as null pointer constant, which is useful to catch subtle errors as early as possible.

If this diagnostic is enabled project-wide, fmt produces a gazillion warnings because of things like `Char *start = 0;` in templates in the format.h. And since templates are instantiated in the user code, it's impossible to suppress this warning for fmt headers only.

Either nullptr or NULL should make gcc happy.
`#define nullptr NULL` is perfectly fine for pre-C++11 compilers.

vitaut:
Using `NULL` sounds reasonable. Would you be willing to submit a PR?

PR #409, opened 2016-11-02
Using FMT_NULLPTR instead of literal 0

alabuzhev:
Based on [issue 408](https://github.com/fmtlib/fmt/issues/408)

I've added FMT_NULLPTR macro (similar to other emulated language features there) that would expand to NULL or nullptr if available, as gcc isn't happy even with NULL in some cases.

Also added a couple of FMT_OVERRIDE where necessary.

### DIFF
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -214,6 +214,16 @@ typedef __int64          intmax_t;
 # endif
 #endif
 
+#ifndef FMT_NULLPTR
+# if FMT_HAS_FEATURE(cxx_nullptr) || \
+   (FMT_GCC_VERSION >= 408 && FMT_HAS_GXX_CXX11) || \
+   FMT_MSC_VER >= 1600
+#  define FMT_NULLPTR nullptr
+# else
+#  define FMT_NULLPTR NULL
+# endif
+#endif
+
 // A macro to disallow the copy constructor and operator= functions
 // This should be used in the private: declarations for a class
 #ifndef FMT_USE_DELETED_FUNCTIONS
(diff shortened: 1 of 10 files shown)


fmt/format.h
```diff
@@ -748,7 +758,7 @@ void MemoryBuffer<T, SIZE, Allocator>::grow(std::size_t size) {
   std::size_t new_capacity = this->capacity_ + this->capacity_ / 2;
   if (size > new_capacity)
       new_capacity = size;
-  T *new_ptr = this->allocate(new_capacity);
```
foonathan: You've added a second argument here for some reason that causes compilation errors.

alabuzhev: See http://en.cppreference.com/w/cpp/memory/allocator/allocate
Second argument is defaulted to literal 0 and causes a warning when -Wzero-as-null-pointer-constant diagnostic enabled.

I presume compilation error happens in mock-allocator.h?
If so, just adding a dummy parameter to its allocate() will fix that (and also will make that mock more conformant).

alabuzhev: See http://en.cppreference.com/w/cpp/memory/allocator/allocate
Second argument is defaulted to literal 0 and causes a warning when -Wzero-as-null-pointer-constant diagnostic enabled.
I presume compilation error happens in mock-allocator.h?
If so, just adding a dummy parameter to its allocate() will fix that (and also will make that mock more conformant).

vitaut:
Nice! Thanks for working on it (and @foonathan for reviewing).

PR #410, opened 2016-11-03
Fix Klockwork compilation warning

chenhayat:


fmt/format.h
```diff
@@ -1912,7 +1913,10 @@ class ArgFormatterBase : public ArgVisitor<Impl, void> {
 
   void visit_bool(bool value) {
     if (spec_.type_)
-      return visit_any_int(value);
+    {
+        visit_any_int(value);
+        return;
+    }
```
vitaut: Please follow Google style guide: https://google.github.io/styleguide/cppguide.html

foonathan:
@vitaut Could you perhaps add a .clang-format file? It would make contributing much easier.

vitaut:
@foonathan What will be the benefit of having `.clang-format` in the project itself compared to say passing `-style=google` to `clang-format`?

foonathan:
Oh, so you're using the google style without modification? Did not know that.

vitaut:
Yes, with the exception of snake_case instead of CapitalizedWords for method names, but this doesn't affect formatting. I should probably document it somewhere.

foonathan:
Yes, you should document it.

But running clang-format over the source files will also change some other stuff like the indentation of your `#defines` when in `#ifdef`, break lines differently, etc.

vitaut:
@chenhayat Merged in https://github.com/fmtlib/fmt/commit/05ba3e788846957ddb88277ed73b54073e595907, thanks!

@foonathan I've documented the coding style in https://github.com/fmtlib/fmt/blob/master/CONTRIBUTING.rst. `clang-format` does mess up preprocessor indentation, so I am a bit reluctant of applying it globally. Hopefully they'll fix https://llvm.org/bugs/show_bug.cgi?id=17362 and until then I suggest limiting it to formatting patches.

Issue #418, opened 2016-11-09
Minimal supported library subset


alabuzhev:
[Documentation](http://fmtlib.net/latest/usage.html) says that including format.h and format.cc should be enough, but format.cc also includes printf.h and printf.h includes ostream.h.
I've experimented a little, looks like #include "printf.h" in format.cc required only for explicit instantiations of PrintfFormatter for char and wchar_t.
Maybe it's possible to move them to printf.cc?
Do you have any plans to make format.h/cc self-contained?
If no - it's probably worth to update the docs.

vitaut:
Good catch, there indeed should be `printf.cc` and `format.cc` shouldn't be including `printf.h`.

PR #419, opened 2016-11-10
Issue #418 - Minimal supported library subset

alabuzhev:
As per #418.

### DIFF
--- a/fmt/format.cc
+++ b/fmt/format.cc
@@ -26,7 +26,6 @@
  */
 
 #include "format.h"
-#include "printf.h"
 
 #include <string.h>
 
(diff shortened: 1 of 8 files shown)


foonathan:
Thank, you, merged.

fmt/format.cc
```diff
@@ -486,34 +483,6 @@ FMT_FUNC void report_windows_error(
 }
 #endif
 
-FMT_FUNC void print(std::FILE *f, CStringRef format_str, ArgList args) {
```
vitaut: This and the following two methods are actually part of the core library and not `printf` extension

alabuzhev: Apologies, they looked quite similar. Created #420.

PR #445, opened 2016-12-29
Don't export the -std=c++11 flag from the fmt target

EricWF:
Currently the `fmt` CMake target exports `-std=c++11` as part of its interface; Meaning any project that links to `fmt` has the `-std=c++11` added to its compilation. However this will override the `-std=<ver>` flag provided by that project. This breaks projects that use C++14 or C++1z. In general I think it's bad style for projects to force a specific dialect, or any compile flags, upon its users. 

This patch fixes the issue by making `-std=c++11` a private flag of `fmt`. This prevent it from infecting downstream users.


vitaut:
Good catch and thanks for the fix! Merged.

Issue #290, opened 2016-03-08
static library cannot be linked against a pie executable


cinghiale:
how to reproduce:
```cpp
#include <format/format.h>
int main() {
    fmt::print("hello world");
}
```

```bash
$ g++ -I ./support/deps/include/ -pie test.cpp ./support/deps/lib/libcppformat.a
/usr/bin/ld: /tmp/ccPx6StM.o: relocation R_X86_64_32 against `.rodata' can not be used when making a shared object; recompile with -fPIC
/tmp/ccPx6StM.o: error adding symbols: Bad value
collect2: error: ld returned 1 exit status
```
Following the compiler's advice results in a working library but I don't know how to patch the cmake recipe

vitaut:
To enable compilation with `-fPIC` use the following CMake command:

```
cmake -DCMAKE_POSITION_INDEPENDENT_CODE=TRUE ...
```

EricWF:
Thanks for the correction!

PR #446, opened 2016-12-29
Build with -fPIC so libfmt.a can be linked into shared libraries

EricWF:
Linking `libfmt.a` when linking a shared library fails because it is not built as "position independent code". This patch fixes the issue by adding `-fPIC` to the build whenever the flag is supported.

### DIFF
--- a/fmt/CMakeLists.txt
+++ b/fmt/CMakeLists.txt
@@ -1,3 +1,5 @@
+include(CheckCXXCompilerFlag)
+
 # Define the fmt library, its includes and the needed defines.
 # *.cc are added to FMT_HEADERS for the header-only configuration.
 set(FMT_HEADERS format.h format.cc ostream.h ostream.cc printf.h printf.cc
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for another PR, but I'd rather not set `-fPIC` unconditionally because it may add slight overhead for clients that don't need it. I think a better solution is to use `CMAKE_POSITION_INDEPENDENT_CODE` which can be set from command line or a client project:
https://github.com/fmtlib/fmt/issues/290#issuecomment-193873137.

gabyx:
What is the overhead? Are there more assembler instructions? Thanks resolving this issue!

Cheaterman:
Can this be fixed? The infinitesimal amount of overhead does not justify in ANY way why this would be entirely broken when linked against dynamic libraries?

foonathan:
You just need to build it with the CMake option `CMAKE_POSITION_INDEPENDENT_CODE` set to true...

PR #448, opened 2017-01-03
Allow %i as integer format specifier

mojoBrendan:
%i is a valid integer specifier, behaves the same as %d


vitaut:
Thanks for the PR, but `%i` specifier is already supported:

```c++
#include <fmt/printf.h>

int main() {
  fmt::printf("%i", 42);
}
```

mojoBrendan:
Sorry about that. Totally missed the part where they are combined in format.cc

Issue #449, opened 2017-01-03
Allow %s as generic format specifier


mojoBrendan:
Porting from another library (tinyformat), which allows %s to be used to get the default format for any parameter. At the moment this causes fmt to throw.

It would be nice (and I think straightforward) if %s could be supported for all types in the printf style functions.

vitaut:
Sounds like a reasonable suggestion. This is already supported for `bool` and should be straightforward to do for other types. However it may take some time till I get to this since I'm busy with other stuff. Perhaps someone could contribute?

mojoBrendan:
Sounds great. I have something working at the moment. Let me clean it up a bit. I'll get a pull request happening next week.

PR #453, opened 2017-01-10
Allow %s as generic format specifier in printf

mojoBrendan:
Signed integers are formatted as %d
Unsigned integers are formatted as %u
Doubles are formatted as %f
Chars are formatted as %c
Void Pointers are formatted as %p

### DIFF
--- a/fmt/printf.h
+++ b/fmt/printf.h
@@ -61,6 +61,24 @@ class IsZeroInt : public ArgVisitor<IsZeroInt, bool> {
   bool visit_any_int(T value) { return value == 0; }
 };
 
+// returns the default type for format specific "%s"
+class DefaultType : public ArgVisitor<DefaultType, wchar_t> {
+ public:
+  wchar_t visit_char(int) { return 'c'; }
+
+  wchar_t visit_bool(bool) { return 's'; }
+
+  wchar_t visit_pointer(const void *) { return 'p'; }
+
+  template <typename T>
+  wchar_t visit_any_int(T) { return 'd'; }
+
+  template <typename T>
+  wchar_t visit_any_double(T) { return 'f'; }
+
+  wchar_t visit_unhandled_arg() { return 's'; }
+};
+
 template <typename T, typename U>
 struct is_same {
   enum { value = 0 };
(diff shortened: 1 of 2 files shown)


mojoBrendan:
For https://github.com/fmtlib/fmt/issues/449

fmt/printf.h
```diff
@@ -92,9 +92,18 @@ class ArgConverter : public ArgVisitor<ArgConverter<T>, void> {
       visit_any_int(value);
   }
 
+  void visit_char(char value) {
+    if (type_ != 's' && type_ != 'S')
```
vitaut: I think that having uppercase `S` is redundant and inconsistent with other specifiers. Please leave only `s` here and below.

fmt/printf.h
```diff
@@ -463,6 +472,23 @@ void PrintfFormatter<Char, AF>::format(BasicCStringRef<Char> format_str) {
...
+        } else if (arg.type == Arg::BOOL) {
+            // leave bool alone
+        } else if (arg.type <= Arg::LAST_INTEGER_TYPE) {
+            // all other integer types
+            spec.type_ = 'd';
+        } else if (arg.type <= Arg::LAST_NUMERIC_TYPE) {
+            // all floating point types
+            spec.type_ = 'f';
+        }
+    }
```
vitaut: Please convert this if-else chain into a visitor and merge with the similar logic under for integers right below.

fmt/printf.h
```diff
@@ -92,9 +110,18 @@ class ArgConverter : public ArgVisitor<ArgConverter<T>, void> {
       visit_any_int(value);
   }
 
+  void visit_char(char value) {
+    if (type_ != 's')
+      visit_any_int(value);
+  }
```
mojoBrendan: ArgConverter runs before DefaultType, and I couldn't see a clean way to switch it around.

vitaut:
Merged, thanks!

mojoBrendan:
Thankyou :)

foonathan:
@mojoBrendan @vitaut #504 reported warnings because the return type of your `DefaultType` is `wchar_t`, which is later truncated to `char`. Is there any special reason why it is `wchar_t`?

foonathan:
Nevermind, it was done in PR #495.

PR #457, opened 2017-01-17
Fix missing intrinsic when included from C++/CLI

calumr:
Managed C++ code doesn't have the _BitScanReverse* intrinsics, we must use the fallback code for count_digits etc.


vitaut:
Merged, thanks!

PR #458, opened 2017-01-19
Fix Android not being detected with latest NDK toolchain

Gachapen:
When using the NDK 13b toolchain standalone or with CMake, ANDROID is not defined,
but __ANDROID__ is defined instead.

I can't say for sure that is and will always be the case, but it works for me, and I have some unofficial sources: http://stackoverflow.com/a/15335544 and http://markmail.org/message/5ekhfztchs45lz3n#query:+page:1+mid:jfqgzrmwm37uyrt6+state:results

### DIFF
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -2916,7 +2916,7 @@ void BasicWriter<Char>::write_int(T value, Spec spec) {
   case 'n': {
     unsigned num_digits = internal::count_digits(abs_value);
     fmt::StringRef sep = "";
-#ifndef ANDROID
+#if !(defined(ANDROID) || defined(__ANDROID__))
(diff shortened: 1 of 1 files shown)


foonathan:
Thanks, merged.

enh:
Android maintainer here: yes, `__ANDROID__` is the correct macro for "you're building for an Android target". you might want to remove the test for `ANDROID` because that can be set in some surprising cases (such as "you're using an Android toolchain/inside the AOSP build system, but targeting the host").

vitaut:
Removed the `ANDROID` check in https://github.com/fmtlib/fmt/commit/867b330966f7082d196852f3a7f68c8d1cba83f0. Thanks for clarification, @enh.

Issue #460, opened 2017-01-30
MSW: Unable to build / test as dll


richardeakin:
Hi there and thanks for the wonderful formatting library. I've been using it for the last couple years and it has been great, my goto for string formatting.

Now I'm in need of building fmtlib as part of a larger shared library, specifically with Visual Studio 2015 / v140, which I do by adding format.cc to its sources and enabling `FMT_EXPORT`. To get this to work however, I had to add some extra `FMT_API` declarations, at minimum [these](https://github.com/richardeakin/fmt/commit/ea45a410f5846d69407b6f8d01b97346b6913c1e). After doing this, I went to see how fmtlib's tests run as is. So I cloned the master branch and generated Visual Studio 2015 projects with `-DBUILD_SHARED_LIBS=1`, then proceeded to build the solution and hit a number of other compiler errors. I fixed a few others in [this branch](https://github.com/richardeakin/fmt/commits/dllexport_additions) but stopped when I hit one that would require restructing a test in order to gain some feedback first on how to proceed. What is the goal of the `FMT_API` support?

I'm not so familiar with the entirety of the codebase as I've only ever used format.h/cpp (I've ignored posix.h/cpp, printf.h/cpp, etc - are those only for testing and benchmarks?), so I'm not sure if `FMT_API` declarations belong there for the sole purpose of building all of the tests with a shared lib configuration.

By the way, the error message I stopped on was due to [this line](https://github.com/fmtlib/fmt/blob/bc6af7548b4202f3c188282a6020b8365252af85/test/format-impl-test.cc#L32), which causes a ` __declspec(dllimport)` to end up in that file while it wasn't actually being imported from a DLL. Instead I think that test could just use the header only version in order to gain extra access to format.cc.

vitaut:
Thanks for the feedback.

> What is the goal of the FMT_API support?

The original goal is to export all necessary symbols to be able to use the fmt library itself DLL on Windows. However, it probably hasn't been kept up to date because the default configuration is a static library.

If you want to use fmt as a part of a larger DLL, it might still better to link it a static library and set `DBUILD_SHARED_LIBS` to 0 (or use the default).

Another options is to use the header-only mode.

richardeakin:
Would you accept a PR with the FMT_API additions I needed to add if I throw one up? Perhaps only those that affect format.h/cc, as the rest of the code seems to be for tests that weren't designed with DLL support in mind.

[...]

PR #469, opened 2017-02-07
Additional FMT_API declarations where needed

richardeakin:
To dll export from within a larger dll. See #460 for details.

Note that couple `internal::CharTraits` had to be exported because it is used from within a function defined in the header (that wasn't exported).


vitaut:
Merged, thanks!

PR #473, opened 2017-02-22
Added missing FMT_OVERRIDE specifier to FormatBuf::overflow

0x414c:
Add `override` specifier to `FormatBuf::overflow` virtual function.


foonathan:
Thank you, merged.

PR #475, opened 2017-02-23
Add default copy constructor to SystemError

dschmidt:
This PR fixes 
```fmt/format.h:2387:3: error: definition of implicit copy constructor for 'SystemError' is deprecated because it has a user-declared destructor [-Werror,-Wdeprecated]   ~SystemError() FMT_DTOR_NOEXCEPT;``` 
in clang-3.9

### DIFF
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -2405,6 +2405,7 @@ class SystemError : public internal::RuntimeError {
   SystemError(int error_code, CStringRef message) {
     init(error_code, message, ArgList());
   }
+  SystemError(const SystemError&) = default;
   FMT_VARIADIC_CTOR(SystemError, init, int, CStringRef)
 
   FMT_API ~SystemError() FMT_DTOR_NOEXCEPT;
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -253,13 +253,10 @@ typedef __int64          intmax_t;
 #ifndef FMT_DEFAULTED_COPY_CTOR
 # if FMT_USE_DEFAULTED_FUNCTIONS || FMT_HAS_FEATURE(cxx_defaulted_functions) || \
    (FMT_GCC_VERSION >= 404 && FMT_HAS_GXX_CXX11) || FMT_MSC_VER >= 1800
-#  define FMT_DEFAULTED  = default
 #  define FMT_DEFAULTED_COPY_CTOR(TypeName) \
-    TypeName(const TypeName&) = default
+    TypeName(const TypeName&) = default;
 # else
-#  define FMT_DEFAULTED  {}
-#  define FMT_DEFAULTED_COPY_CTOR(TypeName) \
-    TypeName(const TypeName&) {}
+#  define FMT_DEFAULTED_COPY_CTOR(TypeName)
 # endif
 #endif
 
@@ -2422,7 +2419,7 @@ class SystemError : public internal::RuntimeError {
   SystemError(int error_code, CStringRef message) {
     init(error_code, message, ArgList());
   }
-  FMT_DEFAULTED_COPY_CTOR(SystemError);
+  FMT_DEFAULTED_COPY_CTOR(SystemError)
   FMT_VARIADIC_CTOR(SystemError, init, int, CStringRef)
 
   FMT_API ~SystemError() FMT_DTOR_NOEXCEPT;
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -246,6 +246,23 @@ typedef __int64          intmax_t;
     TypeName& operator=(const TypeName&)
 #endif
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by foonathan** [comment]
@vitaut Why does it have a destructor anyway?

**Comment by dschmidt** [comment]
So ... are you going to fix this otherwise? Want me to do anything?
Not sure what to conclude from a closed-with-no-path-forward-PR.

**Comment by foonathan** [comment]
Sorry, I hit the wrong button accidentally :D

**Comment by dschmidt** [comment]
I'd assume to mark the destructor explicitly noexcept.

**Comment by foonathan** [comment]
Yes, but it could also be defaulted then.

**Comment by vitaut** [comment]
@dschmidt, thanks for the PR. Could you hide default behind a macro similar to `FMT_OVERRIDE` for C++-98 compatibility?

@foonathan, unfortunately the dtor is there for a reason. I don't remember the details but they can be dug out from the version control history.

**Comment by dschmidt** [comment]
Yes, most probably. Checking.

**Comment by dschmidt** [comment]
AFAICT defaulted and deleted functions are supported from the very same compiler versions and thus I used the same version macros as for the deleted functions

File: fmt/format.h [hunk 1]
```diff
@@ -246,6 +246,23 @@ typedef __int64          intmax_t;
...
+#ifndef FMT_DEFAULTED_COPY_CTOR
+# if FMT_USE_DEFAULTED_FUNCTIONS || FMT_HAS_FEATURE(cxx_defaulted_functions) || \
+   (FMT_GCC_VERSION >= 404 && FMT_HAS_GXX_CXX11) || FMT_MSC_VER >= 1800
+#  define FMT_DEFAULTED  = default
+#  define FMT_DEFAULTED_COPY_CTOR(TypeName) \
+    TypeName(const TypeName&) = default
+# else
+#  define FMT_DEFAULTED  {}
+#  define FMT_DEFAULTED_COPY_CTOR(TypeName) \
+    TypeName(const TypeName&) {}
```
vitaut: I think the defaulted copy ctor is equivalent to not having a user-defined copy ctor. Having one with an empty body will not work if the class has members. Therefore I suggest defining FMT_DEFAULTED_COPY_CTOR to nothing here:

```c++
#  define FMT_DEFAULTED_COPY_CTOR(TypeName)
```

`FMT_DEFAULTED` is not used and can probably be removed.

fmt/format.h
```diff
@@ -2405,6 +2422,7 @@ class SystemError : public internal::RuntimeError {
   SystemError(int error_code, CStringRef message) {
     init(error_code, message, ArgList());
   }
+  FMT_DEFAULTED_COPY_CTOR(SystemError);
```
vitaut: I suggest moving semicolon to the definition `FMT_DEFAULTED_COPY_CTOR` to prevent it dangling if the latter is defined to empty.

fmt/format.h
dschmidt: Good catch. Will define it to empty.

I added `FMT_DEFAULTED` analogously to `FMT_DELETED` because I thought it might come in handy at some point, but I certainly don't insist. Will remove if you prefer.

fmt/format.h
dschmidt: Makes sense.

dschmidt:
As soon as you're happy with it, I would like to squash the commits.

foonathan:
Commits can be squashed when merging the PR automatically, no need for that.

fmt/format.h
foonathan: > I think the defaulted copy ctor is equivalent to not having a user-defined copy ctor.

Note that this is not the case if the class has a user-defined move constructor. If no copy ctor is declared, it won't have a copy constructor. But by writing = default it will have one.

dschmidt: There are no move semantics in C++98 though...

foonathan: Yes, it's just something that should be kept in mind, just in case.

dschmidt:
Any idea what's up with the mingw builds?

foonathan:
Master's also broken so not your fault.

vitaut:
Appveyor updated MinGW - gcc is 5.3.0 now. Typical.

vitaut:
Clarified that formatting functions are defined in `fmt/format.h`: http://fmtlib.net/dev/api.html#format-api. Other headers are already documented.

vitaut:
Oops, wrong thread.

dschmidt:
So are you happy with this now or do you want me to change anything still?

vitaut:
Looks good to me, thanks for addressing the comments. I'll fix the master and merge this in.

dschmidt:
Wooohoo, sounds good. Thanks for the quick feedback cycle!

vitaut:
Merged, thanks!

PR #481, opened 2017-03-05
Don't bring Arg struct into global namespace

LogicalKnight:
This fixes compiling fmtlib in header-only mode when user code also has
something called 'Arg' defined: the using in format.cc brings fmt::internal::Arg into the global namespace. This change means that the Arg struct is referred to with its full namespace.


LogicalKnight:
Oops. Made a mistake. Sorry for the noise. Will recreate and resubmit shortly!

PR #482, opened 2017-03-05
Don't bring Arg struct into global namespace

LogicalKnight:
Take two!

This fixes compiling fmtlib in header-only mode when user code also has
something called 'Arg' defined. Now qualifying Arg struct with internal
namespace.


vitaut:
Thanks for the fix, merged!

Issue #484, opened 2017-03-07
Build error when using compiler wrapper nvcc_wrapper


pkestene:
i've already mention this problem on spdlog, issue tracker, for the record here is the problem:

When building fmt with nvcc_wrapper, which wraps nvcc (Nvidia compiler), one gets the following error:

nvcc_wrapper    -I/home/pkestene/install/spdlog/github/fmt  -O3 -DNDEBUG   -std=c++11 -o CMakeFiles/fmt.dir/ostream.cc.o -c /home/pkestene/install/spdlog/github/fmt/fmt/ostream.cc
/home/pkestene/install/spdlog/github/fmt/fmt/ostream.h: In function ‘void fmt::format_arg(fmt::BasicFormatter<Char, ArgFormatter>&, const Char*&, const T&)’:
/home/pkestene/install/spdlog/github/fmt/fmt/ostream.h:103:50: **error: ‘ArgFormatter’ is not a template**
   typedef internal::MakeArg< BasicFormatter<Char> > MakeArg;
                                                  ^
Please notice that when using regular nvcc, the code build fine, but when building with nvcc_wrapper, nvcc_wrapper replaces option "-c" by "-x cu" and goes on as if source code might contain actual cuda code. This wrapper is really useful for using performance portability tool kokkos.

There is a simple fix which consists in replacing ArgFormatter by ArgFormatter_ in template argument.

vitaut:
Thank you for reporting this and providing the fix.

PR #485, opened 2017-03-07
fix for nvcc_wrapper compiler

pkestene:
This pull request adresses issue #484


vitaut:
Thanks for the fix, merged!

PR #490, opened 2017-03-30
printf("%.f") should use zero precision, not default precision

manylegged:
Really minor change to make fmt::printf better conform to the man page when omitting the optional precision digit

An optional period, `.', followed by an optional digit string giving a precision which specifies the number of digits to appear after the decimal point, for e and f formats, or the maximum number of bytes to be printed from a string; if the digit string is missing, the precision is treated as zero;

### DIFF
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -1179,10 +1179,31 @@ struct ConvertToIntImpl2<T, true> {
   };
 };
 
+template <typename T, bool ENABLE_CONVERSION>
+struct ConvertToIntImpl3 {
+  enum {
+    value = false
+  };
+};
+
+template <typename T>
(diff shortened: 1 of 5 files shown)


manylegged:
I added a test for this zero-precision case, and also for the related case with non-zero precision.

vitaut:
The new tests seem to be failing.

manylegged:
Whoops, forgot about rounding... Should be passing now. I actually installed cmake etc. and ran it this time...

vitaut:
Merged the first 3 commits (squashed) in https://github.com/fmtlib/fmt/commit/cbac016cced41bdb885fd3b51cecac73a431d966, thanks!

The last two commits seem unrelated, please submit as a separate PR.

PR #494, opened 2017-04-10
Fix Visual Studio 2017 new warning (C4244: 'argument': conversion fro…

chronoxor:
…m 'int' to 'const char', possible loss of data)


vitaut:
Merged, thanks!

Issue #462, opened 2017-02-05
Complete the new write API


vitaut:
Complete the new named-argument based write API introduced in #452. One problem to solve is how to dispatch to different methods based on type specifier:

```c++
w.write(42, type='x');
```

In the old API it was `w << hex(42)` and handled more efficiently than the general case. One possible solution is to use:

```c++
w.write(42, hex);
```
instead.

vitaut:
The new write API may not be needed because, as mentioned in https://github.com/fmtlib/fmt/issues/546#issuecomment-337450603, it should be possible to generate optimal code at compile time using the format-based API.

PR #495, opened 2017-04-11
Consistent behavior of operator<< for known types

effzeh:
Consistent behavior of operator<< for known types.

The (templated) operator<< defined in ostream.h currently prevents efficient
handling of known types in BasicWriter. E.g.:

```
  MemoryWriter w;
  w << std::string("hello");
```

calls this operator (if visible). This will unneccessarily construct a
std::ostream object using a FormatBuf as its streambuf and insert the string
into the std::ostream object. (The same is true for `float` and some other
built-in types.) The write_str method in BasicWriter basically does the same
using a cheap memcpy.

This patch attempts to fix this performance problem.

But this patch might also break some code, e.g.:

```
    std::string s = (MemoryWriter() << 3.14).str();
```

only works in C++11 now. Note that

```
    std::string s = (MemoryWriter() << UserDefinedType()).str();
```

didn't work before, but should work now.

I haven't found another solution which works and is still safe (neither in C++03
nor in C++11), so I'm just posting this here for discussion...

### DIFF
--- a/fmt/format.cc
+++ b/fmt/format.cc
@@ -368,7 +368,7 @@ FMT_FUNC void internal::format_windows_error(
       if (result != 0) {
         UTF16ToUTF8 utf8_message;
         if (utf8_message.convert(system_message) == ERROR_SUCCESS) {
-          out << message << ": " << utf8_message;
+          out << message << ": " << StringRef(utf8_message);
           return;
         }
         break;
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -155,6 +155,24 @@ typedef __int64          intmax_t;
 # endif
 #endif
 
+#ifndef FMT_USE_OSTREAM_RVALUE
+// If non-zero, enables operator<< for BasicWriter&&, such that things like
+//  std::string s = (MemoryWriter() << "i = " << i).str();
+// are possible.
+//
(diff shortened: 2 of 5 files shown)


vitaut:
Thanks a lot for the PR!

> But this patch might also break some code, e.g.:
>    std::string s = (MemoryWriter() << 3.14).str();

This is a bit unfortunate. I'll try to come up with an alternative solution that doesn't break existing code. If not then I guess we'll have to go with the backward-incompatible one.

effzeh:
Should be backwards compatible now. Turned out the only difficulty were some inconsistences between MSVC and other compilers. If this gets accepted, the commits should probably be squashed... :-)

vitaut:
Thanks for working on this. Unfortunately there is still an issue with [implicit conversions](https://github.com/fmtlib/fmt/pull/495/files#diff-40b396c6c173a86eedec78943ffa2728L371), so I decided to revert (dcfd40a1b8aa1a101b025b41bde92b67126a8bb8) the commit that introduced this problem in the first place because it not only interferes with known types but is also not compatible with older compilers which is one of the goals of this library. `operator<<` is likely to go away eventually in favor of a more expressive API based on named arguments: #462.

effzeh:
Yes, thats clearly the easiest solution. Thanks!

PR #497, opened 2017-04-12
add SOURCELINK_SUFFIX for compatibility with Sphinx 1.5

ginggs:
With Sphinx 1.5, this is needed by searchtools.js to display the source
snippets (see sphinx-doc/sphinx#2454).

With earlier Sphinx versions, this is a no-op because the undefined variable
will evaluate to an empty string.


vitaut:
Merged, thanks!

PR #499, opened 2017-04-16
fmt::internal::is_streamable works on gcc 4.7

zhaohuaxishi:
i test the the code on https://gcc.godbolt.org/, and only gcc 4.7 works,
gcc 4.6 fail to complie


vitaut:
Merged, thanks!

PR #502, opened 2017-04-27
Allow compiling and using as DLL in windows

AndreasSchoenle:
Removes warnings and errors when compiling in VS 2015 with BUILD_SHARED_LIBS set to "ON".

### DIFF
--- a/fmt/printf.h
+++ b/fmt/printf.h
@@ -62,21 +62,21 @@ class IsZeroInt : public ArgVisitor<IsZeroInt, bool> {
 };
 
 // returns the default type for format specific "%s"
-class DefaultType : public ArgVisitor<DefaultType, wchar_t> {
+class DefaultType : public ArgVisitor<DefaultType, char> {
  public:
-  wchar_t visit_char(int) { return 'c'; }
+  char visit_char(int) { return 'c'; }
 
-  wchar_t visit_bool(bool) { return 's'; }
+  char visit_bool(bool) { return 's'; }
 
-  wchar_t visit_pointer(const void *) { return 'p'; }
-
-  template <typename T>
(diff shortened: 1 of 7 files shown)


fmt/printf.h
```diff
@@ -495,7 +495,9 @@ void PrintfFormatter<Char, AF>::format(BasicCStringRef<Char> format_str) {
 
     if (spec.type_ == 's') {
       // set the format type to the default if 's' is specified
-      spec.type_ = internal::DefaultType().visit(arg);
+
+      // Avoid warning for wchar_t
+      spec.type_ = char(internal::DefaultType().visit(arg));
```
vitaut: Could you change the return values in `DefaultType` instead of using a cast here?

AndreasSchoenle:
Changed the return type of DefaultType to 'char' as requested. Also I needed to change the assertion in `void Buffer<T>::append(const U *begin, const U *end)` as utiltest did not succeed in the DLL version in DEBUG builds. 

The reason was that `internal::to_unsigned` failed when requesting a buffer of size `std::numeric_limits<size_t>::max()` in the FormatSystemError test as `int(end-begin)` is then converted to -1. In the static lib version the assertion was caught by the exception handler in format_system_error, producing the expected behavior also in Debug builds. (In Release builds the memory allocation fails, producing the same result). In the DLL version this was no longer the case. 

Hopefully I did not misinterpret the intention of the test.

AndreasSchoenle:
Maybe it is possible to add appveyor tests with BUILD_SHARED_LIBS=ON?

foonathan:
@AndreasSchoenle What do you mean by `BUILD_SHARED_LIBS=ON`? Could you do a PR?

Issue #297, opened 2016-03-21
format.cc (in header only mode) includes windows.h without defining WIN32_LEAN_AND_MEAN


gabime:
and this for example [breaks](https://github.com/gabime/spdlog/issues/179#issuecomment-199079702) boost and other libs

vitaut:
I don't think it's a good idea to define `WIN32_LEAN_AND_MEAN` there because it may break users who include `windows.h` without this macro and use these rare features. There are two alternative solutions:

1. Define `FMT_USE_WINDOWS_H` to 0 if you don't need Windows error reporting functionality and Unicode converters provided by C++ Format.
2. Define `WIN32_LEAN_AND_MEAN` yourself if you are OK with potential issues.

This can be done in the same place where you define `FMT_HEADER_ONLY`.

gabime:
In this case I might defineFMT_USE_WINDOWS_H to 0 in spdlog..

 What is the Windows error reporting functionality?

vitaut:
It's [WindowsError](http://cppformat.github.io/latest/api.html#_CPPv2N3fmt12WindowsErrorE) exception and `report_windows_error` function that automatically translate the error code returned by `GetLastError()` into a text message. The formatting functions don't use this, so if you don't use this either, spdlog won't be affected.

PR #503, opened 2017-05-01
Don't include the world with WIN32_LEAN_AND_MEAN

Quentin01:
This PR reduces the size of what `windows.h` will include by using the define `WIN32_LEAN_AND_MEAN`.

In fact it also help to prevent bad order of inclusion because some header files in Windows need to be include in a good order and that can cause some problems when you use fmt without looking for what it is including.

### DIFF
--- a/fmt/format.cc
+++ b/fmt/format.cc
@@ -41,6 +41,9 @@
 #endif
 
 #if FMT_USE_WINDOWS_H
+# ifndef WIN32_LEAN_AND_MEAN
+#  define WIN32_LEAN_AND_MEAN
+# endif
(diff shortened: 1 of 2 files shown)


foonathan:
Thanks for the PR.

vitaut:
It's OK to define `WIN32_LEAN_AND_MEAN` when compiling a library, but not when using fmt in the header-only mode for reasons discussed in #297, so I've adjusted one of the checks in https://github.com/fmtlib/fmt/commit/c56d345985c5f717e8f63dd3ec3da8765fd07748.

PR #510, opened 2017-05-15
Changing ArgArray template argument from unsigned to size_t

trozen:
Each instantiation of ArgArray template uses sizeof operator, which returns a std::size_t value. GCC 7.1 warns about invalid conversion (error: conversion to ‘unsigned int’ from ‘long unsigned int’ may alter its value [-Werror=conversion]).


vitaut:
Merged, thanks!

trozen:
Thanks to you for this great project :-)

2017-05-20 7:19 GMT+02:00 Victor Zverovich <notifications@github.com>:

PR #511, opened 2017-05-19
CMake imported targets should be namespaced

bjoernthiel:
see https://cmake.org/cmake/help/v3.8/manual/cmake-developer.7.html


vitaut:
Merged in https://github.com/fmtlib/fmt/commit/048d2aec277810372919bd6b324fb562601ba9d9 with the namespace changed from `Fmt` to `fmt` for consistency with the library name. Thanks!

PR #513, opened 2017-05-23
cmake namespace fixes

niosHD:
Hi,

this PR addresses two small issues which have been missed in PR #511. First, the second target in the find-package-test has not been updated (see 9f48bb8). Second, in my opinion it should not matter how (`add_subdirectory` or `find_package`) the library has been added into a consuming project. I therefore added aliases with the prefix which makes it possible to always use the namespace version (`fmt::fmt`) consistently (see a68c7a8).

Regards,
niosHD


vitaut:
Seems reasonable. Thanks for the PR!

wieland400:
Hi,

what happened to this change? The alias is not defined in the current version. As a result

> target_link_libraries(foo PRIVATE fmt::fmt-header-only)

from the documentation http://fmtlib.net/latest/usage.html#header-only-usage-with-cmake yields a

> Target "foo" links to target "fmt::fmt-header-only" but the target was
>   not found.  Perhaps a find_package() call is missing for an IMPORTED
>   target, or an ALIAS target is missing?
>

niosHD:
It seems it was lost when the [std branch was merged](https://github.com/fmtlib/fmt/commit/b00053247ceb7fc3d95bbdc9b250ddaa4436313d). In the [4.x branch it is still present](https://github.com/fmtlib/fmt/blob/4.x/fmt/CMakeLists.txt#L42). @vitaut would you mind to apply the patch again or should I send a new PR?

vitaut:
Hmm, indeed looks like I missed it while merging.

vitaut:
On a second thought, I'm leaning towards killing the CMake namespace altogether (or, rather, not migrating it from the 4.x branch) because there are just two exported targets and both already have `fmt` in their names so adding a namespace seems redundant.

niosHD:
FWIW, I would keep the namespace. Like mentioned in the original PR #511, having the namespace is recommended by CMake's develop documentation and influences how errors are reported when something is wrong with the imported target.

vitaut:
Good point, re-merged and added a comment why the namespace is useful.

vitaut:
@niosHD, could you please review the updated [CONTIBUTING](https://github.com/fmtlib/fmt/blob/ebec00138bd10e949ff34955bcb096387cd32acc/CONTRIBUTING.md) document, particularly the part about licensing, and let me know if you agree with it being applied both to your contributions to {fmt}? The library is likely to be relicensed (#1073) so I'm collecting approval from all earlier contributors.

PR #515, opened 2017-05-29
Describe cmake use of header-only target

ibell:


vitaut:
Thanks a lot, merged!

PR #520, opened 2017-06-16
Fixed builtin CLZ being used incorrectly

### LINKED ISSUE #519: Random data on formatting integers with Clang with Microsoft CodeGen
Put simply, when I write this:

	cout << fmt::format("{}", 100u) << "\n";
	cout << fmt::format("{}", 100) << "\n";
	cout << fmt::format("{}", "aoeu") << "\n";
	cout << fmt::format("{}", 2.5) << "\n";
	cout << fmt::format("{} - {} - {}", 2.5, 3, "a") << "\n";

I get:

	♀õôw100
	0¸+
	aoeu
	2.5
	2.5 - Í×©O/☺P§3 - a

The characters seem to be random, varying every time I run the program.
I have ostreams included, if that matters.

Compiler: Visual Studio 2017 - Clang with Microsoft CodeGen (v141_clang_c2)

Cleroth:
Fixes #519

### DIFF
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -288,11 +288,11 @@ typedef __int64          intmax_t;
 #endif
 
 #if FMT_GCC_VERSION >= 400 || FMT_HAS_BUILTIN(__builtin_clz)
-# define FMT_BUILTIN_CLZ(n) __builtin_clz(n)
+# define FMT_BUILTIN_CLZ(n) (31 - __builtin_clz(n))
 #endif
 
 #if FMT_GCC_VERSION >= 400 || FMT_HAS_BUILTIN(__builtin_clzll)
-# define FMT_BUILTIN_CLZLL(n) __builtin_clzll(n)
+# define FMT_BUILTIN_CLZLL(n) (63 - __builtin_clzll(n))
 #endif
 
 // Some compilers masquerade as both MSVC and GCC-likes or


vitaut:
Thanks a lot for the fix but `FMT_BUILTIN_CLZ*` shouldn't be changed unconditionally. Moreover, the implementation is likely to be fixed in the future which will break the workaround. Perhaps it should be wrapped in `#ifdef _MSC_VER`:

```
#ifndef _MSC_VER
# if FMT_GCC_VERSION >= 400 || FMT_HAS_BUILTIN(__builtin_clz)
#  define FMT_BUILTIN_CLZ(n) __builtin_clz(n)
# endif
...
#endif
```
Will this work for you?

Cleroth:
That would work, because it would avoid the broken code path...
I'm testing the result of __builtin_clz on different machines, and I'm getting the 'expected' value on one, and the 'inverted' value on the other. I think this actually may be due to the instruction not being supported on all CPUs.

According to [this page](http://www.felixcloutier.com/x86/LZCNT.html):
> It should be noted that on processors that do not support LZCNT, the instruction byte encoding is executed as BSR

And according to [this](https://stackoverflow.com/a/25683798/2483904):
> LZCNT == (31 - BSR)

Thus, if you want to use LZCNT, you really need to check for whether it's supported. The compiler does not do this for you. See [here](https://godbolt.org/g/Ttiras) (when you specify -march=native, clang will use lzcnt without checking for support, and so will GCC).

vitaut:
Fixed in https://github.com/fmtlib/fmt/commit/569c5bdbf166e7227f2933ed83c7322263659bd9, thanks!

Note that in https://godbolt.org/g/Ttiras the compiler emits the correct code for `__builtin_clz` both when using `BSR` and `LZCNT`, no manual adjustment is necessary. And it's not specific to `LZCNT` instruction, the same applies to SSE and similar. Whoever compiles the library should specify the correct target architecture.

As for this issue, it's definitely a bug in clang with Microsoft CodeGen because it violates the semantics of `__builtin_clz`.

PR #526, opened 2017-06-25
remove unnecessary method

maddinat0r:
Since the 4.0 release is almost there, that function should be removed as described in the comment.


vitaut:
Good catch, thanks!

PR #527, opened 2017-06-25
remove 'FMT_CPPFORMAT' CMake option

maddinat0r:
Another backward compatibility thing removed before 4.0 release.


vitaut:
Merged, thanks!

PR #534, opened 2017-07-03
added thread safety

wind85:

### DIFF
--- a/fmt/printf.h
+++ b/fmt/printf.h
@@ -23,78 +23,77 @@ namespace internal {
 // signed and unsigned integers.
 template <bool IsSigned>
 struct IntChecker {
-	template <typename T>
-	static bool fits_in_int(T value) {
-		unsigned max = std::numeric_limits<int>::max();
-		return value <= max;
-	}
-	static bool fits_in_int(bool) { return true; }
+  template <typename T>
+  static bool fits_in_int(T value) {
+    unsigned max = std::numeric_limits<int>::max();
+    return value <= max;
+  }
+  static bool fits_in_int(bool) { return true; }
 };
 
 template <>
 struct IntChecker<true> {
-	template <typename T>
-	static bool fits_in_int(T value) {
-		return value >= std::numeric_limits<int>::min() &&
-		       value <= std::numeric_limits<int>::max();
-	}
-	static bool fits_in_int(int) { return true; }
+  template <typename T>
+  static bool fits_in_int(T value) {
+    return value >= std::numeric_limits<int>::min() &&
+           value <= std::numeric_limits<int>::max();
+  }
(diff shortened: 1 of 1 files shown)


wind85:
Hello I didn't see that you provide support for pre C++11, would you agree to conditionally compile the thread safe support only if C++11 is supported?

foonathan:
Could you please keep the indentation as-is? ːD

wind85:
Sorry about that I forgot I have clang-format configured to format on save...

fmt/printf.h
```diff
@@ -588,6 +600,8 @@ FMT_VARIADIC(int, printf, CStringRef)
   \endrst
  */
 inline int fprintf(std::ostream &os, CStringRef format_str, ArgList args) {
+  std::mutex mtx;
+  std::unique_lock<std::mutex> lck(mtx);
```
vitaut: I don't think this is necessary because the writer is not accessible from multiple threads and it doesn't use any global mutable state.

wind85: Yes you are right the writer is local instance though std::ostream is not thread safe...

wind85: 27.1.3 Thread safety [iostreams.thread-safety]

Concurrent access to a stream object [string.streams, file.streams], stream buffer object [stream.buffers], or C Library stream [c.files] by multiple threads may result in a data race [intro.multithread] unless otherwise specified [iostream.objects]. [Note: Data races result in undefined behavior [intro.multithread].

vitaut: Sure but if you want to synchronize access to `std::ostream` then the mutex should be placed elsewhere. If you put it in this function, you'll create a new one on every call. Moreover, I think it belongs in the client code and not in the library.

wind85: Different school of thought, I think thread safety should be provided by the library in 2017, it could be implemented differently and more efficiently. I have one last question is it a thing of not wanting to add thread safety or a matter of implementation?

vitaut: I think the library shouldn't provide weaker thread-safety guarantees than the underlying I/O API, that's all. If `ostream` is thread-safe (as in http://open-std.org/JTC1/SC22/WG21/docs/papers/2015/p0053r0.html) then the function that uses it should be too. But I strongly believe that explicitly synchronizing access to `ostream` is out of scope of the library and if anyone is interested, it provides a simple way to write a custom formatting function in just a few lines of code. This function can take a mutex, or even better, one can put `ostream` and `mutex` in a class and provide synchronized formatting methods.

vitaut: But thanks for raising this question, I didn't really think about thread-safety of iostreams before.

wind85: Thanks for answering...

PR #539, opened 2017-07-06
Modification to format.h to suppress C4574 on MSVC

alexanderbock:
In Visual Studio 2017, C4574: 'identifier' is defined to be '0': did you mean to use '#if identifier'?

### DIFF
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -99,7 +99,9 @@ typedef __int64          intmax_t;
 #  define FMT_HAS_GXX_CXX11 1
 # endif
 #else
+# define FMT_GCC_VERSION 0
 # define FMT_GCC_EXTENSION
+# define FMT_HAS_GXX_CXX11 0
 #endif
 
 #if defined(__INTEL_COMPILER)
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -120,7 +120,7 @@ typedef __int64          intmax_t;
 #endif
 
 // GLM defines __has_feature to '0' which causes MSVC to throw C4574 
-#if defined(__has_feature) && __has_feature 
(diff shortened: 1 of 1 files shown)


fmt/format.h
```diff
@@ -119,7 +119,8 @@ typedef __int64          intmax_t;
 # define FMT_GNUC_LIBSTD_VERSION (__GNUC_LIBSTD__ * 100 + __GNUC_LIBSTD_MINOR__)
 #endif
 
-#ifdef __has_feature
+// GLM defines __has_feature to '0' which causes MSVC to throw C4574
```
vitaut: What is GLM?

vitaut:
Comments in commits look weird in GitHub, so I'll ask it here: what is GLM?

alexanderbock:
A very popular header-only math library: https://github.com/g-truc/glm

At least from my experience, it is pretty well used

vitaut:
Merged the https://github.com/fmtlib/fmt/pull/539/commits/7eff0809a4dd292517e6f2762666104c6cc68171 commit. However, defining `__has_feature` to 0 looks like a bug in GLM and should be reported there instead of trying to implement a workaround here.

alexanderbock:
According to https://github.com/g-truc/glm/blob/master/glm/detail/setup.hpp#L768:
```
#ifndef __has_feature
#	define __has_feature(x) 0 // Compatibility with non-clang compilers.
#endif
```
So I assume that they have a good reason for it

vitaut:
Sure, but GSL shouldn't be defining a system macro. It should define a macro in its own "namespace", e.g. `GLM_HAS_FEATURE`. The fmt library uses `FMT_HAS_FEATURE` specifically for this reason.

PR #542, opened 2017-07-12
Fix undefined behavior in UDL macro

xylosper:
`FMT_USE_USER_DEFINED_LITERALS` macro expands to `defined()` which is undefined behavior.


vitaut:
Merged, thanks!

PR #545, opened 2017-07-14
Update format.h to remove C4574 error on MSVC 14.2

alexanderbock:
Similar to the Pullrequest #539, `_SECURE_SCL` caused the same  `warning C4574: '_SECURE_SCL' is defined to be '0': did you mean to use '#if _SECURE_SCL'`.  `_SECURE_SCL` is defined in the `MSVC/14.10.25017/include/yvals.h` by Microsoft itself


vitaut:
Merged, thanks!

PR #547, opened 2017-07-17
Make format_arg() accept class hierarchies

rollbear:
If a base class can be formatted, it is assumed
that the hierarchy can be formatted from the base
class. The idiom is not uncommon with `ostream`s.

See the document example.


vitaut:
Merged with minor tweaks (mostly to follow the coding style) in https://github.com/fmtlib/fmt/commit/9c56a8ce5c4f91686f80810c4559103bca00e34e. Thanks a lot for the PR!

PR #550, opened 2017-07-22
printf: Add support for glibc's %m format

mikecrowe:
glibc's printf family of functions support '%m' to emit the error string
that corresponds to the current value of errno. It has no corresponding
argument.

In order to take advantage of safe_strerror, PrintfFormatter::format calls
format_system_error, which has been modified to not emit the separator if
the message prefix is empty.

'm' needs to be added to the list of special characters in format-test.cc's
check_unknown_types since it would otherwise always be valid in the format
string.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -162,7 +162,9 @@ The following functions use `printf format string syntax
 <http://pubs.opengroup.org/onlinepubs/009695399/functions/fprintf.html>`_ with
 the POSIX extension for positional arguments. Unlike their standard
 counterparts, the ``fmt`` functions are type-safe and throw an exception if an
-argument type doesn't match its format specification.
+argument type doesn't match its format specification. They also support the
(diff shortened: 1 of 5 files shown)


vitaut:
Thanks a lot for the PR. The implementation looks great but I'm not sure if this feature is generally useful enough to justify extending the grammar with a dedicated specifier (even thought that's what glibc does). I'd prefer to have it handled via the extension mechanism instead, e.g. something like

```
printf("%s", system_error());
```

where `system_error` returns an object that captures the current `errno` and gives the string representation of it when formatted. It should be possible to do by providing an appropriate overload of `format_arg ` and doesn't require adding a new specifier.

mikecrowe:
Fair enough.

I'm incorporating fmt::sprintf into an existing set of code with many uses of %m that was previous using functions that called std::vsprintf internally. We can just patch fmtlib locally until we can modify all the call sites.

Thanks.

Mike.

Issue #552, opened 2017-08-11
noreturn function throwing exception gives "noreturn func returns"


Manu343726:
The function `internal::report_unknown_type()` declared at `format.h:935` as `noreturn` is throwing an exception on exceptions-enabled through the `FMT_THROW` macro. I've caugh this by building a dependent project with `-Werror -Wall -Wextra -pedantic` enabled on GCC 7.1.1. There's an unconditional throw at the end of the function, so as far as I'm concerned the function should not be triggering the error.

I've tried to write a minimal reproducible case (http://coliru.stacked-crooked.com/a/a7b0ef79054ac456) but as you can see there's no error there.
As far I can see the warning should not be triggered in that function, my output is more or less (could not make the clipboard copy work...):

```
fmt/format.cc:296:1: error: the 'noreturn' function returns
```

vitaut:
Thanks for the bug report. Could you give the full compiler invocation command with all options as given e.g. by `make VERBOSE=1`? I wasn't able to repro this with just `-Werror -Wall -Wextra -pedantic`.

Manu343726:
I've added a job to the CI build with the same original config and after a custom patch to fix integer conversion warnings the error is no longer reproduced: https://travis-ci.org/Manu343726/cppast/jobs/263927793 (Same in my local build of fmt with [that patch](https://github.com/Manu343726/fmt/commit/29a7e2c8a88294be181c58278b4d1a0c8e443870)).

I'm quite annoyed, definitely the purpose of `[[noreturn]]` is to troll me. I will close this as no longer reproducible, but I will ping you here if it happens again... Sorry :(

BTW, tomorrow I will try to PR that patch to your master branch.

Thanks for all.

PR #553, opened 2017-08-14
Explicitly cast int -> char to prevent conversion warnings

Manu343726:
This prevents integer conversion warnings so the library can be compiled with GCC and `-Wconversion` enabled (also `-Werror -Wall -Wextra -pedantic`, but I think the offender was `-Wconversion`) as "discussed" in #552.

Note this change makes possible to build fmt with these flags, but **the unit tests cannot be compiled, googletest triggers a lot of conversion warnings**.

### DIFF
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -742,7 +742,7 @@ template <typename T>
 template <typename U>
 void Buffer<T>::append(const U *begin, const U *end) {
   FMT_ASSERT(end >= begin, "negative value");
-  std::size_t new_size = size_ + (end - begin);
+  std::size_t new_size = size_ + static_cast<std::size_t>(end - begin);
   if (new_size > capacity_)
     grow(new_size);
   std::uninitialized_copy(begin, end,
--- a/fmt/printf.h
+++ b/fmt/printf.h
@@ -110,7 +110,7 @@ class ArgConverter : public ArgVisitor<ArgConverter<T>, void> {
       visit_any_int(value);
   }
 
-  void visit_char(char value) {
+  void visit_char(int value) {
     if (type_ != 's')
       visit_any_int(value);
   }


fmt/format.h
```diff
@@ -1671,7 +1671,7 @@ class ArgVisitor {
       FMT_ASSERT(false, "invalid argument type");
       break;
     case Arg::INT:
-      return FMT_DISPATCH(visit_int(arg.int_value));
+      return FMT_DISPATCH(visit_int(static_cast<char>(arg.int_value)));
```
vitaut: I don't think this is correct.

Manu343726:
Hi,
Sorry for the delay,

You were right, that change was wrong. I've changed the commit ton only include an explicit cast in `visit_char` case. But it seems that the test that is failing:

```
4: [ RUN      ] FormatterTest.Examples
4: /home/manu343726/Documentos/fmt/test/format-test.cc:1494: Failure
4: Value of: format(L"Cyrillic letter {}", L'\x42e')
4:   Actual: L"Cyrillic letter ."
4: Expected: L"Cyrillic letter \x42e"
4: Which is: L"Cyrillic letter \x42E"
4: [  FAILED  ] FormatterTest.Examples (0 ms
```

is related to the `visit_char` cast instead of the `visit_int` one (I've pushed the commit to fire your build again). I'm not sure what `visit_char` is being called in that context, since there are multiple `visit_char` functions:

```
 ack visit_char
test/util-test.cc
614:  Result visit_char(int value) { return static_cast<char>(value); }

fmt/printf.h
67:  char visit_char(int) { return 'c'; }
113:  void visit_char(char value) {
251:  void visit_char(int value) {

fmt/format.h
1608:  Result visit_char(int value) {
1684:      return FMT_DISPATCH(visit_char(static_cast<char>(arg.int_value)));
2005:  void visit_char(int value) {
```

vitaut:
The reason why the test is failing is that the character type can be `wchar_t` in which case it shouldn't be cast to `char`. I think a proper fix is to change
```c++
void visit_char(char value)
```
to 
```c++
void visit_char(int value)
```
in line 113 of `printf.h`. Could you check if it fixes the warning? (I wasn't able to repro it for some reason.)

Manu343726:
Great, that fixes the warnings, all tests passing. Thanks a lot for your time.

Manu343726:
Ooook fixed https://travis-ci.org/foonathan/cppast/builds/266489096

vitaut:
Merged, thanks!

PR #556, opened 2017-08-18
Add one more CMake warning fix

henryiii:
This will remove one more warning that can't be stopped from a calling CMakeLists. This possibly should be set to new at some point (CMake 3.3 behavior would be different / better), as well as CMP0048.


vitaut:
Merged, thanks!

Issue #560, opened 2017-08-28
Move ArgMap::init into format.h


Yelnats321:
Currently, `internal::ArgMap<Char>::init` is defined inside `format.cc` and explicitly instantiated for `char` and `wchar_t`, even though the function itself is char type agnostic. This leads to issues using a character type other than these two. The easiest solution is to simply move the function definition into `format.h`. I am unsure on the compile time impact of such a change, although it would seem to be negligible, especially considering that most other functions are not explicitly instantiated (with the exception of `format_float` for understandable reasons and `FixedBuffer::grow`).

vitaut:
Seems reasonable. Could you by any chance submit a PR?

Yelnats321:
Yeah sure just give me a day or two.

PR #563, opened 2017-09-02
Make ArgMap::init not explicitly instantiated

Yelnats321:
PR for issue #560.


Yelnats321:
The diff is kind of hard to read but I just moved it out and also removed FMT_API since it's a template function that's defined in the header too.

vitaut:
Merged, thanks!

PR #571, opened 2017-09-17
Make format work with C++17 std::string_view

thelostt:
Enable `std::string_view` to work with format when compiling with C++17.

Change log:

+ `fmt::StringRef` is constructible from `std::string_view`.
+ `fmt::StringRef` is convertible to `std::string_view` (only explicitly)[1].
+ `fmt::format` accepts `std::string_view` as an argument for the argument list.

[1] This is open to debate. Should it be an implicit conversion, or should it not exist at all? The argument in favor of it is that `std::string_view` is now the common string reference, and all other string implementations could take advantage of it by providing a conversion to it. A counter-argument would be that fmtlib prefers `fmt::StringRef` as the common string reference.

### DIFF
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -45,6 +45,19 @@
 // The fmt library version in the form major * 10000 + minor * 100 + patch.
 #define FMT_VERSION 40001
 
+#if defined(__has_include)
+# define FMT_HAS_INCLUDE(x) __has_include(x)
+#else
+# define FMT_HAS_INCLUDE(x) 0
+#endif
+
+#if FMT_HAS_INCLUDE(<string_view>) && __cplusplus > 201402L
+# include <string_view>
+# define FMT_HAS_STRING_VIEW 1
+#else
+# define FMT_HAS_STRING_VIEW 0
+#endif
+
 #if defined _SECURE_SCL && _SECURE_SCL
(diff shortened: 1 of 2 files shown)


vitaut:
Looks great, thanks for the PR!

> Should it be an implicit conversion, or should it not exist at all?

Conversion seems fine unless it causes some issues for existing code such as overload ambiguity. Eventually `StringRef` will be replace with `std::string_view` anyway.

mridgewell:
Thanks for this, but it doesn't work in Visual Studio because the `__cplusplus` definition is out of date in Microsoft's compiler. I suggest either removing that check or adding a specific one for `_MSC_VER`.

vitaut:
@mridgewell If you are interested in `string_view` support on MSVC, could you submit a PR adding a version check specific to this compiler to

```c++
#if FMT_HAS_INCLUDE(<string_view>) && __cplusplus > 201402L
```
?

thelostt:
Removing the check for `__cplusplus` prevents GCC from compiling, even if you're not making use of string_view in your own code. There's no issue with Clang though, as it accepts <string_view> in any standard version. I'm not sure about MSVC.

mridgewell:
I wouldn't mind, but looks like @mwinterb has already done the work, which solves my problem :)
Thanks!

PR #578, opened 2017-10-08
Added MSVC checking for support for string_view.

mwinterb:
Microsoft allows customizing the standard version the compiler will target, so a simple _MSC_VER check is inadequate, but __has_include support was added after string_view, so _MSC_VER is still needed.

Also, adding "override" to destructors that are virtual by inheritance.

### DIFF
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -51,7 +51,8 @@
 # define FMT_HAS_INCLUDE(x) 0
 #endif
 
-#if FMT_HAS_INCLUDE(<string_view>) && __cplusplus > 201402L
+#if (FMT_HAS_INCLUDE(<string_view>) && __cplusplus > 201402L) || \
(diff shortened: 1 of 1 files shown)


mwinterb:
I've fixed that, but I've noticed that the CStringRefTest.Ctor test does not compile with MSVC now. https://github.com/fmtlib/fmt/blob/master/test/format-test.cc#L177
But I don't think that it should compile regardless? If it is compiling, it would appear it's converting from:
string_view -> string -> CStringRef. But the string_view -> string conversion is supposed to be explicit:
 [string.cons](http://eel.is/c++draft/string.cons).
In fact, both gcc and clang also do not compile this: https://godbolt.org/g/dEVCQV.

vitaut:
Merged, thanks!

Issue #579, opened 2017-10-08
Warning using "allocate" directly


mwinterb:
MSVC's library implementers are [aggressively](https://twitter.com/StephanTLavavej/status/872904135980466176) following the deprecations from the standard library in C++17. Using their "daily" releases, this warning occurs:
```
fmt/format.h(850,1): warning C4996: 'std::allocator<wchar_t>::allocate': warning STL4010: Various members of std::allocator are deprecated in C++17. Use std::allocator_traits instead of accessing these members directly. You can define _SILENCE_CXX17_OLD_ALLOCATOR_MEMBERS_DEPRECATION_WARNING or _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS to acknowledge that you have received this warning.
  T *new_ptr = this->allocate(new_capacity, FMT_NULL);
^
<PATH>\packages\VisualCppTools.Community.Daily.VS2017Layout.14.11.25719-Pre\lib\native\include\xmemory0(992): note: see declaration of 'std::allocator<wchar_t>::allocate'
	_CXX17_DEPRECATE_OLD_ALLOCATOR_MEMBERS _DECLSPEC_ALLOCATOR _Ty * allocate(
fmt/format.h(846): note: while compiling class template member function 'void fmt::internal::MemoryBuffer<wchar_t,500,std::allocator<wchar_t>>::grow(::size_t)'
void MemoryBuffer<T, SIZE, Allocator>::grow(std::size_t size) {
fmt/format.h(1095): note: see reference to class template instantiation 'fmt::internal::MemoryBuffer<wchar_t,500,std::allocator<wchar_t>>' being compiled
  MemoryBuffer<wchar_t, INLINE_BUFFER_SIZE> buffer_;
```

The fix is easy,  `T *new_ptr = std::allocator_traits<Allocator>::allocate(*this, new_capacity, FMT_NULL);` but that breaks C++98 compat, and I don't know if a small preprocessor mess is worth it for this one case. And I also don't know if Microsoft is going to fully commit to the deprecations in official releases "soon".

For now, locally I just replace the this->allocate with the allocator_traits indirection, but that needs another change to actually work, so I'll submit a PR for that shortly.

mwinterb:
This is obviously really low priority, especially since there's an opt-out, I just wanted there to be "visibility" for when others eventually get this warning.

vitaut:
>  I don't know if a small preprocessor mess is worth it for this one case

[...]

PR #583, opened 2017-10-16
Use allocator_traits if available.

mwinterb:
This is to avoid using functionality deprecated in C++17. Addresses #579.


vitaut:
Merged, thanks!

PR #595, opened 2017-11-02
Add concat and concat_separated

Shakti213:
Concat simply concatenates all arguments and will replace a single placeholder
with the concatenation. Example:
format("({})", concat("hello ", "world")) will result in "(hello world)"

concat_separated concatenates all arguments with a separator between them.
Example: format("({})", concat(", ", 1, 2, 3)) will result in "(1, 2, 3)"

### DIFF
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -4087,11 +4087,8 @@ void format_arg(fmt::BasicFormatter<Char, ArgFormatter> &f,
 }
 
 struct ArgConcatNoSeparator {
-    template<typename ArgFormatter>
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR, but I'm not sure if this functionality should be in the core library. It covers a fairly narrow use case (static list of values) and can be easily implemented in terms of existing `fmt::join` which is a bit more general:

```c++
template <typename T>
auto join(std::initializer_list<T> args, fmt::CStringRef sep) {
  return fmt::join(args.begin(), args.end(), sep);
}

fmt::print("{}", join({"hello", "world"}, ","))
```

PR #598, opened 2017-11-05
Removed warnings on MSVC 2017

virgiliofornazin:
Silence warnings of constant if expression in Microsoft Visual C++ compiler used in Visual Studio 2017

### DIFF
--- a/fmt/printf.h
+++ b/fmt/printf.h
@@ -125,7 +125,14 @@ class ArgConverter : public ArgVisitor<ArgConverter<T>, void> {
     using internal::Arg;
     typedef typename internal::Conditional<
         is_same<T, void>::value, U, T>::type TargetType;
+#if defined(_MSC_VER)
+#pragma warning(push)
+#pragma warning(disable: 4127)
+#endif
     if (sizeof(TargetType) <= sizeof(int)) {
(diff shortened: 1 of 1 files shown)


fmt/printf.h
```diff
@@ -125,7 +125,14 @@ class ArgConverter : public ArgVisitor<ArgConverter<T>, void> {
     using internal::Arg;
     typedef typename internal::Conditional<
         is_same<T, void>::value, U, T>::type TargetType;
+#if defined(_MSC_VER)
+#pragma warning(push)
+#pragma warning(disable: 4127)
+#endif
     if (sizeof(TargetType) <= sizeof(int)) {
```
vitaut: There is a function `const_check` specifically to workaround this warning. Does it work if you replace the explicit warning suppression with

```c++
if (const_check(sizeof(TargetType) <= sizeof(int))) {
```
?

virgiliofornazin: Yes, it does! Better that #ifdef guarding!

May I submit another pull request, or you could fix it by yourself?

vitaut: > May I submit another pull request, or you could fix it by yourself?

Please do. Thanks!

PR #604, opened 2017-11-06
Added support for "namespace encapsulation"

virgiliofornazin:
Added support for "namespace encapsulation" (better explained as a custom namespace prefix) for fmt library, using the new header file custom_namespace.h

This file adds support for using fmt library defining a custom namespace in your code
to avoid conflicting with other fmt library versions or 'encapsulate' the implementation
you are using inside another namespace.

Those four macros below provide this feature:
- **FMT_CUSTOM_NAMESPACE** - 1 if you want a custom namespace prefix, 0 otherwise
- **FMT_CUSTOM_NAMESPACE_USING_NAMESPACE** - the 'using namespace _CUSTOM_NAMESPACE_;' used into fmt source code, and
MUST terminate with a semicolon. It also could be used by your application if needed (but it's not recommended);
- **FMT_CUSTOM_NAMESPACE_PREFIX** - the '_CUSTOM_NAMESPACE_' namespace prefix for the encapsulation namespace;
- **FMT_CUSTOM_NAMESPACE_BEGIN** - the 'namespace _CUSTOM_NAMESPACE_ {' namespace begin for the encapsulation namespace;
- **FMT_CUSTOM_NAMESPACE_END** - the '} // namespace _CUSTOM_NAMESPACE_' namespace end for the encapsulation namespace;

Please note that you could use more than one namespace nested, as the example below:

```
#define FMT_CUSTOM_NAMESPACE                      1
#define FMT_CUSTOM_NAMESPACE_USING_NAMESPACE      using namespace my_cpp_library::formatting_libraries;
#define FMT_CUSTOM_NAMESPACE_PREFIX               my_cpp_library::formatting_libraries
#define FMT_CUSTOM_NAMESPACE_BEGIN                namespace my_cpp_library { namespace formatting_libraries {
#define FMT_CUSTOM_NAMESPACE_END                  } }

```
Then, in your code, you use this fmt in following way:

```
#include <fmt/format.h>
#include <iostream>

template <typename ... args>
std::string format_some_string(char const * mask, args && ... argsv)
{
	my_cpp_library::formatting_libraries::fmt::MemoryWriter writer;

	writer.write(mask, std::forward<args>(argsv) ...);
	
	return writer.str();
}

int main(int argc, char ** argv)
{
	std::cout << format_some_string("{} from somewhere", "Hello") << std::endl;

	return 0;
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ CMakeCache.txt
 CMakeFiles
 Makefile
 run-msbuild.bat
+.vs
+.vscode
--- a/fmt/CMakeLists.txt
+++ b/fmt/CMakeLists.txt
@@ -1,6 +1,6 @@
 # Define the fmt library, its includes and the needed defines.
 # *.cc are added to FMT_HEADERS for the header-only configuration.
-set(FMT_HEADERS container.h format.h format.cc ostream.h ostream.cc printf.h
+set(FMT_HEADERS container.h custom_namespace.h format.h format.cc ostream.h ostream.cc printf.h
                 printf.cc string.h time.h)
 if (HAVE_OPEN)
   set(FMT_HEADERS ${FMT_HEADERS} posix.h)
--- a/fmt/container.h
+++ b/fmt/container.h
@@ -1,4 +1,4 @@
-/*
+﻿/*
  Formatting library for C++ - standard container utilities
 
  Copyright (c) 2012 - 2016, Victor Zverovich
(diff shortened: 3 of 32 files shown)


vitaut:
Thanks for the PR, but this change is very intrusive. If the main purpose is to allow using multiple versions of the library then it's better be done with inline namespaces.

vitaut:
Sorry, meant to request changes rather than close.

virgiliofornazin:
Inline namespaces sound good for versioning.

But they can't solve the 'issue' that a users could change the namespace from fmt to my_library::some_deep_namespace::detail::fmt, for example.

the custom_namespace.h was added because some unit tests reference directly fmt:: namespace
before including format.h or another. I will think in something that could  be done to make it better and less intrusive.

PR #605, opened 2017-11-09
Fix for warning C5030 in VS2015

alabuzhev:
Fix for "warning C5030: attribute 'maybe_unused' is not recognized" in VS2015.


vitaut:
Merged, thanks!

Issue #572, opened 2017-09-24
printf %ls support - C++ standard compatibility


vgalka-sl:
Hi,

The current library version does not allow mixing wchar_t* arguments into char* format strings. The following code does not compile (at least not on MS VS2017):

`fmt::printf("%ls", L"foo");`

However, according to the C++11 standard, it should be valid for printf-like functions.
> C++11 includes the C library as described by the 1999 ISO C standard and its Technical Corrigenda 1, 2 and 3 (ISO/IEC 9899:1999 and ISO/IEC 9899:1999/Cor.1,2,3), plus <cuchar> (as by ISO/IEC 19769:2004).

Looking into [ISO/IEC 9899:TC2](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1124.pdf), it describes the %ls format specifier as following (page 279):

> If an l length modifier is present, the argument shall be a pointer to the initial element of an array of wchar_t type. Wide characters from the array are converted to multibyte characters (each as if by a call to the wcrtomb function, with the conversion state described by an mbstate_t object initialized to zero before the first wide character is converted) up to and including a terminating null wide character. The resulting multibyte characters are written up to (but not including) the terminating null character (byte).

It would be nice to have this standard compatibility :-)

Best regards,
Vasili

vitaut:
Thanks for raising this, I agree that it should be supported. Would you be willing to contribute a fix by any chance?

vgalka-sl:
I haven't dived into the internals yet. Made a workaround for now...
But I'll try when I find time :-)

pulsa:
This sounds very useful.  I started looking at how to do it.
My ideal implementation would ignore the 'l' in "%ls" and automatically convert all string arguments to match the format string encoding, for both (s)printf() and format().  This simplifies the format string requirements and avoids a source of run-time exceptions, which is important because...

If narrow format strings can accept wide arguments (and vice versa), we lose compile-time safety checks like #606.  But if we can always do the right thing with the input, this is not a problem.

However, conversion errors can happen if the input is invalid UTF-8, or if it contains characters that aren't in the output code set (non-Unicode only).  We could throw an exception, replace with some other character or truncate the string.  Should this be configurable somehow?

[...]

PR #606, opened 2017-11-09
Disallow strings as input for wstring templates

smat:
This change will prevent mixing character encodings by accident, by disallowing using strings as input for wstring templates.

The change will cause the following to be a compile error:

```
EXPECT_EQ(L"abc", format(L"{}", "abc"));
```

Use case: Using a wstring as a template with an UTF-8 string will mix UTF-32/UTF-16 and UTF-8 in the resulting string. When you try to convert the string to UTF-8, `to_bytes` will throw an exception.

```
fmt::WMemoryWriter out;
out.write(L"{}", "\u00F6");
const auto utf8_string = std::wstring_convert<std::codecvt_utf8<wchar_t>>().to_bytes(out.str());
std::cout << utf8_string << std::endl;
```

### DIFF
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -1227,6 +1227,20 @@ struct WCharHelper<T, wchar_t> {
   typedef Null<T> Unsupported;
 };
 
+// A helper class template to enable or disable overloads taking
+// characters and strings in MakeValue.
+template <typename T, typename Char>
+struct CharHelper {
+    typedef T Supported;
+    typedef Null<T> Unsupported;
+};
+
+template <typename T>
+struct CharHelper<T, wchar_t> {
+    typedef Null<T> Supported;
+    typedef T Unsupported;
+};
+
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR!

This is somewhat related to #572. Instead of giving a compile-time error it might be better to  convert a multibyte string into a wide one as if by calling [mbrtowc](http://en.cppreference.com/w/cpp/string/multibyte/mbrtowc). What do you think?

vitaut:
Implemented a slightly different fix in https://github.com/fmtlib/fmt/commit/ba01a5db32db389e785129f7f52a315911c8056e. This gives a compiler error now:
```c++
fmt::format(L"{}", "\u00F6");
```

Thanks for catching the issue!

hestad:
When will the next release of fmt be? We are waiting for this commit to find its way into spdlog :)

PR #607, opened 2017-11-10
Added support for pre-c++17 experimental string_view

virgiliofornazin:
Also, added _POSIX_C_SOURCE to make library compile and pass on all tests on Cygwin environment.

### DIFF
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -57,12 +57,11 @@
 # define FMT_HAS_STRING_VIEW 1
 # define FMT_HAS_EXPERIMENTAL_STRING_VIEW 0
 #else
+# define FMT_HAS_STRING_VIEW 0
 # if (FMT_HAS_INCLUDE(<experimental/string_view>) && __cplusplus > 201402L)
 #  include <experimental/string_view>
-#  define FMT_HAS_STRING_VIEW 0
 #  define FMT_HAS_EXPERIMENTAL_STRING_VIEW 1
 # else
-#  define FMT_HAS_STRING_VIEW 0
 #  define FMT_HAS_EXPERIMENTAL_STRING_VIEW 0
 # endif
 #endif
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -1521,7 +1521,7 @@ class MakeValue : public Arg {
 #if FMT_HAS_STRING_VIEW
   FMT_MAKE_WSTR_VALUE(const std::wstring_view &, WSTRING)
 #endif
-#if FMT_HAS_STRING_VIEW
+#if FMT_HAS_EXPERIMENTAL_STRING_VIEW
   FMT_MAKE_WSTR_VALUE(const std::experimental::wstring_view &, WSTRING)
 #endif
   FMT_MAKE_WSTR_VALUE(WStringRef, WSTRING)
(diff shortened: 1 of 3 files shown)


virgiliofornazin:
FIxed one wrong #if FMT_HAS_(EXPERIMENTAL_)STRING_VIEW  added in commit before

fmt/format.h
```diff
@@ -55,8 +55,16 @@
     (defined(_MSVC_LANG) && _MSVC_LANG > 201402L && _MSC_VER >= 1910)
 # include <string_view>
 # define FMT_HAS_STRING_VIEW 1
+# define FMT_HAS_EXPERIMENTAL_STRING_VIEW 0
 #else
-# define FMT_HAS_STRING_VIEW 0
```
vitaut: I suggest keeping this and removing 2 identical defines below since they don't depend on whether experimental string_view exists.

fmt/format.h
```diff
@@ -582,6 +590,26 @@ class BasicStringRef {
   }
 #endif
 
+#if FMT_HAS_EXPERIMENTAL_STRING_VIEW
+  /**
+  \rst
```
vitaut: nit: Please indent the rst block for consistency with the rest of the code.

test/CMakeLists.txt
```diff
@@ -11,6 +11,10 @@ target_compile_options(gmock PUBLIC ${CPP11_FLAG})
 target_compile_definitions(gmock PUBLIC GTEST_HAS_STD_WSTRING=1)
 target_include_directories(gmock PUBLIC .)
 
+if (CYGWIN)
+  target_compile_definitions(gmock PUBLIC _POSIX_C_SOURCE=200809)
```
vitaut: What is this for?

fmt/format.h
virgiliofornazin: Can't get your point, can you paste here how you want the final code to look like?

fmt/format.h
virgiliofornazin: ok

test/CMakeLists.txt
virgiliofornazin: This is because google test fails to run on Cygwin because they doesn't define _POSIX_C_SOURCE and posix api´s (fileno, strdup, fdopen, and others).

See reference https://stackoverflow.com/questions/18784112/googletest-1-6-with-cygwin-1-7-compile-error-fileno-was-not-declared-in-this

fmt/format.h
vitaut:
```
#else
 -# define FMT_HAS_STRING_VIEW 0
 +# if (FMT_HAS_INCLUDE(<experimental/string_view>) && __cplusplus > 201402L)
 +#  include <experimental/string_view>
 +#  define FMT_HAS_EXPERIMENTAL_STRING_VIEW 1
 +# else
 +#  define FMT_HAS_EXPERIMENTAL_STRING_VIEW 0
 +# endif
```

test/CMakeLists.txt
vitaut: Could you add a comment explaining this?

virgiliofornazin:
Take a look at this last commit, if it fits your requests.

vitaut:
Merged, thanks!

PR #609, opened 2017-11-12
fix warning in header: signed/unsigned comparison

jpcima:
The warning is produced by template functions in format.h with MSVC 2017 toolchain.


vitaut:
Merged, thanks!

PR #610, opened 2017-11-13
CMakeLists: Use GNUInstallDirs to set install location

mikecrowe:
CMake's GNUInstallDirs knows where particular Linux architectures and
distributions want to have their libraries installed. In particular,
Debian-derived "multi-arch" distributions keep their libraries in triplet
sudirectories under /lib. Other "bi-arch" distributions keep 64-bit
libraries in /lib64.

Including GNUInstallDirs and using CMAKE_INSTALL_LIBDIR and
CMAKE_INSTALL_INCLUDEDIR means that fmt's libraries and header files are
installed in the correct locations.

Tested with OpenEmbedded and on Debian GNU/Linux 9 (the special naming only
applies when installing in /usr.)

I would hope that GNUInstallDirs does something sensible on Windows, but I haven't tested it.

### DIFF
--- a/fmt/CMakeLists.txt
+++ b/fmt/CMakeLists.txt
@@ -49,8 +49,9 @@ endif ()
 
 # Install targets.
 if (FMT_INSTALL)
+  include(GNUInstallDirs)
   include(CMakePackageConfigHelpers)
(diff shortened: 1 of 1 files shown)


vitaut:
Merged, thanks!

PR #616, opened 2017-11-25
Avoid conflict with the macro CHAR_WIDTH

aroig:
It looks like CHAR_WIDTH is a macro in glibc. See [this](https://sourceware.org/ml/libc-alpha/2016-09/msg00225.html).


vitaut:
Merged, thanks!

Issue #619, opened 2017-12-03
_format support for string containing \0


WPMGPRoSToTeMa:
Code:
```C++
std::cout << u8"te\0st"_format() << std::endl;
std::cout << u8"te\0st"_format().size() << std::endl;
```
Output:
```
te
2
```
Expected:
```
te st
5
```

Looks like code in `master` branch doesn't support format strings containing `'\0'`. Apparently `std` branch supports them, but `_format` literal still not.

vitaut:
Good catch!

I think it can be easily fixed by changing `const Char *` to `basic_string_view` in https://github.com/fmtlib/fmt/blob/84bd2f196228e585a232f2465ff84eaa40b8398a/include/fmt/format.h#L4096 and passing the size when constructing `udl_formatter`. Could you submit a PR?

WPMGPRoSToTeMa:
@vitaut, ok. Looks like `_a` should be fixed too.

PR #620, opened 2017-12-04
Added support for format string containing '\0' in _format udl (#619)

WPMGPRoSToTeMa:


WPMGPRoSToTeMa:
@vitaut what about adding tests for this?

include/fmt/format.h
```diff
@@ -4105,7 +4105,7 @@ struct udl_formatter {
 
 template <typename Char>
 struct udl_arg {
-  const Char *str;
+  basic_string_view<Char> str;
```
vitaut: Please leave `udl_arg` as before. We want to discourage the use of `\0` in argument names.

vitaut:
Looks great, thanks for the fix!

> @vitaut what about adding tests for this?

Please do add a comment to `LiteralsTest` in `format-test.cc` and also address a small inline comment.

include/fmt/format.h
WPMGPRoSToTeMa: @vitaut ok. Is this for performance reasons as you mentioned on CppCon 2017?

WPMGPRoSToTeMa:
> Please do add a comment to `LiteralsTest` in `format-test.cc`

Do you mean: `// TODO: add test for format string containing '\0'` or something else?
> and also address a small inline comment.

What do you mean by this?

vitaut:
> Do you mean // TODO: add test for format string containing '\0' or something else?

Sorry, I meant a test, not a comment. Please add a test to `LiteralsTest`.

> What do you mean by this?

You've already addressed this (not using basic_string in `_a`).

WPMGPRoSToTeMa:
@vitaut there is also no test for `'\0'` in format string for `fmt::format`.

vitaut:
Merged, thanks!

WPMGPRoSToTeMa:
@vitaut, sorry, but my tests have an issue:
instead of `format("{}c\0d{}", "ab", 1)`
there should be `format(string_view("{}c\0d{}", 7), "ab", 1)`
and for wide strings `format(fmt::basic_string_view<wchar_t>(L"{}c\0d{}", 7), L"ab", 1)`

include/fmt/format.h
vitaut: I am not very concerned about performance here, but using a null character in argument names (as opposed to the format string itself) seems like a bad idea because these names should normally be readable.

vitaut:
> sorry, but my tests have an issue

Could you submit a PR that fixes the issue? Also it would be better to compare against literal text, not to another format string call:

```c++
   EXPECT_EQ("abc\0d1", "{}c\0d{}"_format("ab", 1));
```

> Also, can you answer here #620 (comment)?

Done.

PR #622, opened 2017-12-06
Fixed tests for format strings containing \0

WPMGPRoSToTeMa:

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -34,7 +34,6 @@
 #include <string>
 #include <type_traits>
 #include <vector>
-#include <utility>
 
 #ifdef __has_feature
 # define FMT_HAS_FEATURE(x) __has_feature(x)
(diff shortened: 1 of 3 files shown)


WPMGPRoSToTeMa:
Looks like [my assumption](https://github.com/fmtlib/fmt/issues/619#issue-278814947) was wrong, code in `std` branch still expects that format string doesn't contain `'\0'` and ends with `'\0'`:
https://github.com/fmtlib/fmt/blob/5328907f7f9952a443b9d716f44873d14fc16047/include/fmt/format.h#L1853

vitaut:
> Looks like my assumption was wrong, code in std branch still expects that format string doesn't contain '\0'. So, should I revert my _format udl changes or ...?

Ah, you are right, it's not possible to use '\0' in the format string. I suggest reverting the changes then. It seems like a rare corner case and if anyone really wants an embedded '\0' they can pass it in a format argument.

PR #631, opened 2017-12-27
v3 v4 transition helper

timblechmann:
i'd like to propose the following (ugly) helper: we're using v3 in several projects, but the transition from v3 to v4 was not straight-forward as printf functionality was moved out of format.h into a separate header.
this hack helps us to update to v4 without immediately changing all of our product code, but allows us to incrementally adapt the includes.

### DIFF
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -25,6 +25,11 @@
  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+// transition helper
+#ifdef FMT_FORMAT_PROVIDE_PRINTF
+#include "printf.h"
+#endif
+
 #ifndef FMT_FORMAT_H_
 #define FMT_FORMAT_H_
 


vitaut:
Thanks for the PR, but is there any reason you can't change the includes from `fmt/format.h` to `fmt/printf.h` in the product code, e.g. by autoreplace?

timblechmann:
auto-replace won't work, as sometimes format and sometimes printf syntax is used. but the main practical issue is that the codebase spans several repositories. the base repo pulls in fmtlib, so the transition process is (a) move the base repo to v4, enable the deprecation helper (b) move all repos to the v4 headers (c) once that is done, remove the deprecation helper again.

in a single-repo project it would be easier, but i'm having a codebase with 30+ git repos, so a deprecation strategy is crucial. i've applied this patch to my inhouse fmtlib, but maybe i'm not the only one with this problem :)

vitaut:
Makes sense, merged.

PR #633, opened 2017-12-31
core: Report useful error message if function pointer is passed

mikecrowe:
The value constructor generates a useful error message if the client code
passes a pointer to something that isn't void. Unfortunately, the compiler
doesn't get that far if a function pointer is passed because the
value(T *p) forwarding constructor ends up trying to call itself which
results in an error like:

 .../fmt/core.h:485:34: error: constructor delegates to itself
    value(T *p) : value(as_const(p)) {}

So, let's add an extra constructor that is a better match for function
pointers so that a similar useful error message can be emitted. We need to
force the compiler to decide whether to generate the message only when
something would actually cause the constructor to be called by making the
static_assert expression use the template parameter.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -468,6 +468,18 @@ class value {
     set<POINTER>(pointer, p);
   }
 
+  // Match function pointers explicitly. If we do not then the
+  // forwarding T * constructor below will end up trying to call
+  // itself when a function pointer is passed, which leads to a
+  // confusing error message.
+  template <typename T, typename ...Args>
+  value(T (*p)(Args...)) {
+    // This static_assert must use something that is dependent on T,
+    // otherwise it will always fire, so let's use the same check as
+    // above - it will never match.
+    static_assert(std::is_same<T, void>::value, "formatting of function pointers is disallowed");
+  }
+
   template <typename T>
   value(T *p) : value(as_const(p)) {}
 


include/fmt/core.h
```diff
@@ -468,6 +468,18 @@ class value {
...
+  // Match function pointers explicitly. If we do not then the
+  // forwarding T * constructor below will end up trying to call
+  // itself when a function pointer is passed, which leads to a
+  // confusing error message.
+  template <typename T, typename ...Args>
+  value(T (*p)(Args...)) {
+    // This static_assert must use something that is dependent on T,
+    // otherwise it will always fire, so let's use the same check as
+    // above - it will never match.
+    static_assert(std::is_same<T, void>::value, "formatting of function pointers is disallowed");
```
foonathan: Could you use `FMT_STATIC_ASSERT()` here instead?

mikecrowe: Thanks for the reply.

I copied the constructor above which uses static_assert. There is no mention of FMT_STATIC_ASSERT on the std branch. I can only assume that the std branch is targeting modern C++ versions that support static_assert.

I think that using std::is_pointer along with std::is_function and then std::is_member would probably produce clearer code. I'll wait for @vitaut to provide his opinion before having a go at that though.

foonathan: Oh, right. You're on the std branch. In that case you're good.

vitaut:
Thanks for the PR, @mikecrowe, but could you give an example illustrating the problem with the current error reporting?

I tried the following:
```c++
void f() {}

int main() {
  fmt::print("{}", &f);
}
```

but it already gives a reasonable error message:

```
In file included from test.cc:1:
In file included from include/fmt/format.h:39:
include/fmt/core.h:466:5: error: static_assert failed "formatting of non-void pointers is disallowed"
    static_assert(std::is_same<T, void>::value,
    ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
...
```

mikecrowe:
With GCC 6.3.0 on Debian 9 if I apply [printf-function-pointer.patch.txt](https://github.com/fmtlib/fmt/files/1596363/printf-function-pointer.patch.txt)
 to the current state of the std branch (i.e. without this PR) then I get:

```
In file included from /home/mac/git/fmt/test/printf-test.cc:32:0:
/home/mac/git/fmt/include/fmt/core.h: In instantiation of ‘fmt::internal::value<Context>::value(T*) [with T = int(int); Context = fmt::printf_context<char>]’:
/home/mac/git/fmt/include/fmt/core.h:656:10:   required from ‘typename std::enable_if<IS_PACKED, fmt::internal::value<Context> >::type fmt::internal::make_arg(const T&) [with bool IS_PACKED = true; Context = fmt::printf_context<char>; T = int (*)(int); typename std::enable_if<IS_PACKED, fmt::internal::value<Context> >::type = fmt::internal::value<fmt::printf_context<char> >]’
/home/mac/git/fmt/include/fmt/core.h:805:51:   required from ‘fmt::arg_store<Context, Args>::arg_store(const Args& ...) [with Context = fmt::printf_context<char>; Args = {int (*)(int)}]’
/home/mac/git/fmt/include/fmt/core.h:812:10:   required from ‘fmt::arg_store<Context, Args ...> fmt::make_args(const Args& ...) [with Context = fmt::printf_context<char>; Args = {int (*)(int)}]’
/home/mac/git/fmt/include/fmt/printf.h:542:62:   required from ‘std::__cxx11::string fmt::sprintf(fmt::string_view, const Args& ...) [with Args = {int (*)(int)}; std::__cxx11::string = std::__cxx11::basic_string<char>; fmt::string_view = fmt::basic_string_view<char>]’
/home/mac/git/fmt/test/printf-test.cc:453:3:   required from here
/home/mac/git/fmt/include/fmt/core.h:472:34: error: constructor delegates to itself
   value(T *p) : value(as_const(p)) {}
                                  ^
```

with GCC 7.2.0 in Debian-10-to-be I get the same error (I also get loads of other unrelated errors too.)

With this PR applied I get:

```
In file included from /home/mac/git/fmt/test/printf-test.cc:32:0:
/home/mac/git/fmt/include/fmt/core.h: In instantiation of ‘fmt::internal::value<Context>::value(T (*)(Args ...)) [with T = int; Args = {int}; Context = fmt::printf_context<char>]’:
/home/mac/git/fmt/include/fmt/core.h:668:10:   required from ‘typename std::enable_if<IS_PACKED, fmt::internal::value<Context> >::type fmt::internal::make_arg(const T&) [with bool IS_PACKED = true; Context = fmt::printf_context<char>; T = int (*)(int); typename std::enable_if<IS_PACKED, fmt::internal::value<Context> >::type = fmt::internal::value<fmt::printf_context<char> >]’
/home/mac/git/fmt/include/fmt/core.h:817:51:   required from ‘fmt::arg_store<Context, Args>::arg_store(const Args& ...) [with Context = fmt::printf_context<char>; Args = {int (*)(int)}]’
/home/mac/git/fmt/include/fmt/core.h:824:10:   required from ‘fmt::arg_store<Context, Args ...> fmt::make_args(const Args& ...) [with Context = fmt::printf_context<char>; Args = {int (*)(int)}]’
/home/mac/git/fmt/include/fmt/printf.h:542:62:   required from ‘std::__cxx11::string fmt::sprintf(fmt::string_view, const Args& ...) [with Args = {int (*)(int)}; std::__cxx11::string = std::__cxx11::basic_string<char>; fmt::string_view = fmt::basic_string_view<char>]’
/home/mac/git/fmt/test/printf-test.cc:453:3:   required from here
/home/mac/git/fmt/include/fmt/core.h:480:5: error: static assertion failed: formatting of function pointers is disallowed
     static_assert(std::is_same<T, void>::value, "formatting of function pointers is disallowed");
     ^~~~~~~~~~~~~
```

Presumably this means that you are testing with clang. I'm not sure which compiler is correct.

vitaut:
I was indeed using clang and see the "constructor delegates to itself" error on gcc now.

Unfortunately the `static_assert` in the new overload won't fire if the return type is void. Instead of adding an overload, I think it's better to address the root cause and not delegate to another constructor in `value(T *)`:

```c++

  template <typename T>
  void set_pointer(T *p) {
    static_assert(std::is_same<typename std::remove_const<T>::type, void>::value,
                  "formatting of non-void pointers is disallowed");
    set<POINTER>(pointer, p);
  }

  template <typename T>
  value(T *p) { set_pointer(p); }

  template <typename T>
  value(const T *p) { set_pointer(p); }
```

vitaut:
Fixed in https://github.com/fmtlib/fmt/commit/cea3c20747fc824dd8234df314549bf4c053168d. Thanks catching this issue, @mikecrowe!

PR #635, opened 2018-01-07
WIP: Automatically convert string/wstring arguments to match format string.

pulsa:
This commit allows mixing narrow and wide strings for issue #572:
```
std::string username = get_username_utf8();
std::wstring filename16 = fmt::sprintf(L"C:\\%s.log", username);
fmt::printf("Writing file %s\n", filename16);
```
You can use `"%ls"` if you want.  The `'l'` is ignored.  I tried doing the same for `"%c"` but got stuck and gave up for now.
Narrow/wide mismatch is no longer a compile-time error, but may throw at runtime if the source string can't be converted to the target encoding.  I'm not sure if this is OK.  The `FMT_THROW_WCONV_ERROR` macro can prevent the exception.
Because UTF-8 is not supported by MSVC I bypassed `wcrtomb` and used [the standalone version of nowide/utf.hpp](https://github.com/artyom-beilis/nowide/tree/master/standalone) (which also includes config.hpp).
```
#define FMT_WCONV_USE_NOWIDE
#include <fmt/format.h>
```
This meets my needs.  I don't know if it's a good idea.  Posting for discussion.
I think std::codecvt is the official way to do this but I just don't like it.  Is it reasonable to use nowide this way?
Also note that to write UTF-8 to the console on Windows you need this:
```
setlocale(LC_CTYPE, "C");
SetConsoleOutputCP(CP_UTF8);
```

### DIFF
--- a/fmt/format.h
+++ b/fmt/format.h
@@ -40,11 +40,17 @@
 #include <cstdio>
 #include <cstring>
 #include <limits>
+#include <limits.h>  // for MB_LEN_MAX
 #include <memory>
 #include <stdexcept>
 #include <string>
 #include <vector>
 #include <utility>  // for std::pair
+
+#ifdef FMT_WCONV_USE_NOWIDE
+# include <nowide/utf.hpp>
+#endif
+
 #undef FMT_INCLUDE
 
 // The fmt library version in the form major * 10000 + minor * 100 + patch.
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR!

One general comment is that I think we should only allow mixing multibyte and wide string in `printf` for compatibility with the standard, but not in `format`. For the latter it would be better to explicitly request conversion, e.g.

```c++
fmt::sprintf(L"%ls", "foo"); // OK
fmt::format(L"{}", "foo"); // error
fmt::format(L"{}", utf16("foo")); // explicit conversion
```

No need to implement `utf16` right now, but please make sure that the second case above gives a compile-time error as before.

Also it would be good to avoid external dependencies and use `UTF16ToUTF8` on MSVC.

pulsa:
I'll see what I can do.  I think I'd have to split the `MakeValue` class into printf and format variants.

pulsa:
Thanks for the tip.  I see I have some changes to catch up on.

vitaut:
Closing for now, but feel free to update & re-open or submit a new PR.

PR #640, opened 2018-01-20
Fix unreachable code warning when signbit returns bool

peterbell10:
Fixes the following warning:
```
warning: will never be executed [-Wunreachable-code]
                    sizeof(signbit(x)) == sizeof(int))) {
                    ^~~~~~~~~~~~~~~~~~
```


vitaut:
Nice fix, thanks!

PR #641, opened 2018-01-24
Added a fmt.pro to support build using qmake

cowo78:
For those who prefer qmake as build system or (my own situation) where it makes sense to bundle fmtlib as 3rd party dependency in a QT project.
The file builds ONLY the library as a STATIC library with DYNAMIC libc runtime.
Tested on Linux/qmake 3.1


vitaut:
Thanks for the PR. I'm a bit reluctant to add yet another build config, because it will increase maintenance burden. Is there any reason you cannot use CMake which has native support in Qt Creator?

cowo78:
The project I'm talking about is under CI management so we don't always use creator and switching everything to CMake is not an option (at least right now).
Agreed that having multiple build tools is not ideal, but maybe my scenario is not that uncommon.
Maybe move the file off to a contrib/ location or something so that it does not clutter the root while allowing for a starting point for those who are in my situation.

vitaut:
> Maybe move the file off to a contrib/ location or something so that it does not clutter the root while allowing for a starting point for those who are in my situation.

I think it's a good idea. How about moving `fmt.pro` to the `support` directory which already has build management scripts and configs?

cowo78:
PR #641 updated

vitaut:
Merged, thanks!

PR #649, opened 2018-02-13
Gradle

luncliff:
### Summary
 - Gradle support for Android projects (Especially for Android Studio IDE)
 - Config for Android `.so` file generation
 - Automatic file copy after build

**There is no change for existing `CMakeLists.txt`.**

The change includes dummy `AndroidManifest.xml` file in `support/` directory to generate Android library. Descriptions are commented in `build.gradle` file.

### Build
For command line build, `gradle assemble` will trigger build phase. 
```console
$ cd fmt;
$ export ANDROID_HOME=/path/to/Android/sdk;

$ gradle assemble;  # Trigger gradle build

> Configure project : 
Gradle CMake Plugin: 
[-std=c++14]
[-DANDROID_STL=c++_shared, -DBUILD_SHARED_LIBS=true, -DFMT_TEST=false, -DFMT_DOC=false]
The CompileOptions.bootClasspath property has been deprecated and is scheduled to be removed in Gradle 5.0. Please use the CompileOptions.bootstrapClasspath property instead.

> Task :externalNativeBuildDebug 
Build fmt arm64-v8a
...
...
...
```


vitaut:
Thanks for the PR. I'm a bit reluctant to add yet another build config, because it will increase maintenance burden, but will be OK adding it to the `support` directory.

luncliff:
@vitaut Confirmed. 
In my experience, Android Studio requires to place `build.gradle` at the root directory. But I will try to find a way to meet your opinion.

P.S. Since Travis CI's Android build doesn't contain NDK in its image, checking this build config in Travis can take more than 5 min. I will provide the detailed log when things become clear.

vitaut:
@luncliff, any luck moving the build config to a subdirectory?

luncliff:
@vitaut  Sorry for my laziness. I've been doing multiple works last month. 

### Change
 - `build.gradle` file in `support/`
 - Additional Travis CI build config

`build.gradle` file references `CMakeLists.txt` using a relative path. 

For build config, you can find a column in the build matrix. Since Travis-CI image doesn't contain NDK bundle, I had to install it manually with `sdkmanager`. I used Gradle 4.3.1 with download to save build time, [but it still takes more than 5 minute in general.](https://travis-ci.org/fmtlib/fmt/jobs/352364489)

```yaml
matrix:
  # ...
  include:
    - language: android
      android:
      components:
        - tools
        - platform-tools
        - android-21
        - sys-img-armeabi-v7a-android-21
      env:
        - ANDROID=true
      before_install:
        - git submodule update --init --recursive
        - sudo apt-get install wget unzip tree
      install:
        # Accept SDK Licenses + Install NDK
        - yes | sdkmanager --update > /dev/null 2>&1
        - sdkmanager ndk-bundle > /dev/null 2>&1
        # Download Gradle 4.3.1 
        - wget https://services.gradle.org/distributions/gradle-4.3.1-bin.zip
        - mkdir -p gradle
        - unzip -q -d ./gradle gradle-4.3.1-bin.zip
        - export GRADLE=${TRAVIS_BUILD_DIR}/gradle/gradle-4.3.1/bin/gradle
      before_script:
        - bash $GRADLE --version
        - cd ./support
      script:
        - bash $GRADLE clean assemble
      after_success:
        - cd ${TRAVIS_BUILD_DIR}
        - tree ./libs
```

At the end of the build, you can see the file tree created with Gradle task.

vitaut:
No worries, thanks a lot for updating the PR. 5+ mins is not a big problem I guess since the CI build is asynchronous.

PR #653, opened 2018-02-14
compilation fix & warnings

alabuzhev:


include/fmt/core.h
```diff
@@ -801,10 +801,10 @@ class context_base {
 // Extracts a reference to the container from back_insert_iterator.
 template <typename Container>
 inline Container &get_container(std::back_insert_iterator<Container> it) {
-  typedef std::back_insert_iterator<Container> iterator;
-  struct accessor: iterator {
-    accessor(iterator it) : iterator(it) {}
-    using iterator::container;
+  typedef std::back_insert_iterator<Container> bi_iterator;
```
alabuzhev: For some reason VS cannot compile this:
> core.h(807): error C2327: 'std::back_insert_iterator<fmt::internal::basic_buffer<char>>::container': is not a type name, static, or enumerator

renaming `iterator` to `bi_iterator` helped.

vitaut:
Merged, thanks!

PR #655, opened 2018-02-16
posix.cc: Fix compilation with -fno-exceptions

chenxiaolong:


vitaut:
Merged, thanks!

chenxiaolong:
Wow, that was quick. Thank you!

Issue #657, opened 2018-02-16
Not compatible with VS2017 string_view


alabuzhev:
In VS2017 `basic_string_view::iterator` is not a raw pointer but an internal class, `_String_view_iterator`.

fmt does not expect that and fails to compile:

> format.h(3119): error C2664: 'void fmt::basic_parse_context<char,fmt::internal::error_handler>::advance_to(std::_String_view_iterator<_Traits>)': cannot convert argument 1 from 'const char *' to 'std::_String_view_iterator<_Traits>

**Note**: there is a flaw in the current build system: VS2017 only provides <string_view> header if `/std:c++latest` is specified. fmt, however, does not set that flag and therefore uses it's own, "fallback" string_view implementation, which renders this issue invisible for your build system.

vitaut:
Fixed by using fallback string_view (#664). Thanks for reporting!

PR #658, opened 2018-02-16
Fix VS2017 string_view support

alabuzhev:
I've tried to address #657 here by encapsulating the iterator type.

I'm not sure if it's the best possible approach, but at least it works.

Note: cmake scripts really should be updated to add `/std:c++latest` compiler option to enable the "real" VS17 string_view - right now this path is not tested, which is presumably why this issue was not identified earlier.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -663,7 +663,8 @@ class basic_arg {
 template <typename Char, typename ErrorHandler = internal::error_handler>
 class basic_parse_context : private ErrorHandler {
  private:
-  basic_string_view<Char> format_str_;
+  typename basic_string_view<Char>::iterator begin_;
+  typename basic_string_view<Char>::iterator end_;
   int next_arg_id_;
 
  public:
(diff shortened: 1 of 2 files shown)


include/fmt/core.h
```diff
@@ -672,20 +673,21 @@ class basic_parse_context : private ErrorHandler {
...
+    return begin_;
   }
 
   // Returns an iterator past the end of the format string range being parsed.
-  FMT_CONSTEXPR iterator end() const FMT_NOEXCEPT { return format_str_.end(); }
+  FMT_CONSTEXPR iterator end() const FMT_NOEXCEPT { return end_; }
 
   // Advances the begin iterator to ``it``.
   FMT_CONSTEXPR void advance_to(iterator it) {
-    format_str_.remove_prefix(it - begin());
```
alabuzhev: VS17 asserts that iterators must point to the same string_views:
```C++
constexpr difference_type operator-(const _String_view_iterator& _Right) const _NOEXCEPT
{	// return difference of iterators
#if _ITERATOR_DEBUG_LEVEL >= 1
	_IDL_VERIFY(_Mydata == _Right._Mydata && _Mysize == _Right._Mysize,
		"cannot subtract incompatible string_view iterators");
```
After remove_prefix both `format_str_`'s data and size will be different and this check won't pass.
I've replaced format_str_ with a pair of string_view iterators to resolve this.

vitaut:
Fixed by using fallback string_view (#664), but thanks anyway.

Issue #388, opened 2016-09-26
Cygwin Compatibility


ixokai:
Hi. I ran into two compatibility issues with using fmt in a cygwin (gcc 5.4.0 64-bit) environment.

1) Apparently, when using c++11, gcc goes into strict mode and doesn't include any functions that aren't part of the standard. In particular, fdopen was an issue. Some googling told me you can include POSIX functions by #defining the following in posixc.c before any headers are imported:

#define _POSIX_C_SOURCE 200809L

I'm not quite sure if this is actually cygwin specific or not, as this seems to be gcc -- but I'm not in a position to test on linux yet.

2) Cygwin doesn't provide strtod_l, so I disabled FMT_LOCALE and then fmt compiled and worked.

vitaut:
I think you can workaround the first issue by adding a check for cygwin here:

```c++
#ifdef __MINGW32__
// Workaround MinGW bug https://sourceforge.net/p/mingw/bugs/2024/.
# undef __STRICT_ANSI__
#endif
```

in https://github.com/fmtlib/fmt/blob/master/fmt/posix.h#L13.

Could you submit the PR for this and the locale fix?

vitaut:
Should be fixed in #388, but I don't have a Windows machine to test. @ixokai, does it work for you?

vitaut:
Closing for now but feel free to reopen if you still get errors.

speth:
I can confirm that this works, using G++ 5.4.0 on Cygwin. Any chance of getting a 3.0.x bugfix release or branch with this at some point (esp. since `master` contains backwards-incompatible changes like moving things to `printf.h`).

vitaut:
I plan to make a new release but only after completing the first draft of the standard proposal. Hopefully it won't take long but I don't have any specific dates as I work on this in my spare time (which was a bit limited lately because of switching to a new job). If you or anyone else (with reasonable github history =)) are willing to spend some time on preparing a new release, I'll be happy to give write permissions to the repo.

vitaut:
Hey, @foonathan. There were a few requests for a new minor release of `fmt`, so if you are still interested in helping out, I can give you write permissions. Thanks!

vitaut:
[...]

Issue #345, opened 2016-06-14
strtod_l() not found for Android arm64-v8a configuration


gomons:
When I build the library using android-ndk-r10e for arm64-v8a configuration and I get error:
`fmt/posix.h:387:48: error: 'strtod_l' was not declared in this scope`

vitaut:
Thanks for the bug report. Does https://github.com/fmtlib/fmt/commit/c7739536e8e1ec844527620a2669393a6e3df130 fix the problem?

gomons:
I downloaded version 3.0.0 and make changes like in c773953 and the error disappeared. Thank you a lot!

PR #660, opened 2018-02-18
Fix build on OpenBSD

hubslave:
OpenBSD lacks `strtod_l`, so the `Locale` class won't compile. Looking around, this problem has apparently occurred in other environments as well (#388, #345), so I fixed it the same way they were fixed: by appending ` && !defined(__OpenBSD__)` to the preprocessor condition gating the definition of the `FMT_LOCALE` macro.


vitaut:
Merged, thanks!

PR #676, opened 2018-03-12
Fixed UTF8/16 converters to support empty string input

vgalka-sl:
Previously an exception was thrown since Win32 WideCharToMultiByte() API returns error on zero-length input.

### DIFF
--- a/include/fmt/format.cc
+++ b/include/fmt/format.cc
@@ -280,9 +280,16 @@ const uint64_t internal::basic_data<T>::POWERS_OF_10_64[] = {
 
 FMT_FUNC internal::utf8_to_utf16::utf8_to_utf16(string_view s) {
   static const char ERROR_MSG[] = "cannot convert string from UTF-8 to UTF-16";
-  if (s.size() > INT_MAX)
+  if (s.size() > INT_MAX || s.data() == nullptr)
     FMT_THROW(windows_error(ERROR_INVALID_PARAMETER, ERROR_MSG));
   int s_size = static_cast<int>(s.size());
+  if (s_size == 0) // MultiByteToWideChar does not support zero length, handle separately
+  {
+      buffer_.resize(1);
+      buffer_[0] = 0;
+      return;
+  }
+
   int length = MultiByteToWideChar(
       CP_UTF8, MB_ERR_INVALID_CHARS, s.data(), s_size, FMT_NULL, 0);
   if (length == 0)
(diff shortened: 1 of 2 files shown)


vitaut:
`UTF16ToUTF8` has been renamed to `utf16_to_utf8` in the `master` to follow standards conventions. Could you update the PR?

vgalka-sl:
Sure, thanks! My apologies, I ported it from v4.1.0 based branch.
Pushed, lets see if all is OK now...

include/fmt/format.cc
```diff
@@ -280,9 +280,16 @@ const uint64_t internal::basic_data<T>::POWERS_OF_10_64[] = {
 
 FMT_FUNC internal::utf8_to_utf16::utf8_to_utf16(string_view s) {
   static const char ERROR_MSG[] = "cannot convert string from UTF-8 to UTF-16";
-  if (s.size() > INT_MAX)
+  if (s.size() > INT_MAX || s.data() == nullptr)
```
vitaut: Why add a nullptr check? Won't `MultiByteToWideChar` return an error in this case already?

include/fmt/format.cc
```diff
@@ -280,9 +280,16 @@ const uint64_t internal::basic_data<T>::POWERS_OF_10_64[] = {
...
     FMT_THROW(windows_error(ERROR_INVALID_PARAMETER, ERROR_MSG));
   int s_size = static_cast<int>(s.size());
+  if (s_size == 0) // MultiByteToWideChar does not support zero length, handle separately
+  {
+      buffer_.resize(1);
+      buffer_[0] = 0;
+      return;
+  }
```
vitaut: Please follow the Google style guide (2-space indent and open brace on the same line:

```c++
if (s_size == 0) { // MultiByteToWideChar does not support zero length, handle separately
  buffer_.resize(1);
  ...
```

Same below.

include/fmt/format.cc
vgalka-sl: True, but without this check the function will now succeed for invalid input of `nullptr` and `0` length (because the size check precedes the call to `MultiByteToWideChar`). Clearly this is not the best behavior. Moreover, it fails some tests (UTF16ToUTF8Error/UTF8ToUTF16Error/UTF16ToUTF8Convert).

include/fmt/format.cc
vgalka-sl: Sure. Will fix.
Shall I submit a new (v2) pull-request, to reduce the fix commits clutter?

foonathan: No, they can be squashed together before merging.

include/fmt/format.cc
vitaut: Why is `nullptr` and `0` invalid? If size is zero than the data can be anything including nullptr.

vgalka-sl: Well, this is a conceptual discussion, and probably a matter of definition and taste... Personally, I think that `nullptr` is a special case reserved for "invalid data" and this fact should not be silently discarded. I don't see significant advantages to the other approach, but if you prefer it, the code can be easily changed by reverting this check.
However, please note that the above mentioned tests assume that (nullptr, 0) is invalid input. So, if we decide that it becomes valid, we need to redefine the tests logic.

vitaut: Empty string view has the null data pointer (https://godbolt.org/g/JtWcst), so I suggest removing the `nullptr` check here.

vitaut:
Merged with minor changes in https://github.com/fmtlib/fmt/commit/acb469ae2e18bf35a56f534e53a680890948bfbd, thanks!

include/fmt/format.cc
vgalka-sl: Ok, good point :-)
Thanks!

vitaut:
@vgalka-sl, could you please review the updated [CONTIBUTING](https://github.com/fmtlib/fmt/blob/ebec00138bd10e949ff34955bcb096387cd32acc/CONTRIBUTING.md) document, particularly the part about licensing, and let me know if you agree with it being applied to your contributions to {fmt}? The library is likely to be relicensed (#1073) so I'm collecting approval from all earlier contributors. Thanks!

Issue #678, opened 2018-03-13
Too many warnings on MSVC 2017


rcdailey:
I'm getting the following warnings:

```
Severity	Code	Description	Project	File	Line
Warning	C4244	'argument': conversion from 'wchar_t' to 'char', possible loss of data	MSRServiceInterface	E:\code\frontend\source\Core\ThirdParty\fmt\include\fmt\format.h	1464
Warning	C4244	'argument': conversion from 'wchar_t' to 'char', possible loss of data	MSRServiceInterface	E:\code\frontend\source\Core\ThirdParty\fmt\include\fmt\format.h	1455
Warning	C4244	'argument': conversion from 'wchar_t' to 'char', possible loss of data	MSRServiceInterface	E:\code\frontend\source\Core\ThirdParty\fmt\include\fmt\format.h	2468
Warning	C4244	'argument': conversion from 'wchar_t' to 'char', possible loss of data	MSRServiceInterface	E:\code\frontend\source\Core\ThirdParty\fmt\include\fmt\format.h	2701
Warning	C4244	'=': conversion from 'wchar_t' to 'char', possible loss of data	MSRServiceInterface	E:\code\frontend\source\Core\ThirdParty\fmt\include\fmt\format.h	2418
Warning	C4244	'=': conversion from 'wchar_t' to 'char', possible loss of data	MSRServiceInterface	E:\code\frontend\source\Core\ThirdParty\fmt\include\fmt\format.h	2397
Warning	C4244	'return': conversion from 'int64_t' to 'unsigned int', possible loss of data	MSRServiceInterface	e:\code\frontend\source\core\thirdparty\fmt\include\fmt\core.h	1067
Warning	C4244	'=': conversion from 'const wchar_t' to 'char', possible loss of data	MSRServiceInterface	E:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\VC\Tools\MSVC\14.12.25827\include\xutility	2933
```

I have warnings treated as errors in my code base, and I can't even use this library because of the warnings. I'm running latest on `master` at the moment. Am I missing something here?

foonathan:
I've fixed the other warnings but the last one is not from fmt.

PR #679, opened 2018-03-13
Fix warnings under MSVC

foonathan:
See #678.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -1452,7 +1452,7 @@ class arg_formatter_base {
 
   void operator()(const char_type *value) {
     internal::handle_cstring_type_spec(
-          specs_.type_, cstring_spec_handler(*this, value));
+          static_cast<char>(specs_.type_), cstring_spec_handler(*this, value));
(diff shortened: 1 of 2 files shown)


include/fmt/format.h
```diff
@@ -1452,7 +1452,7 @@ class arg_formatter_base {
 
   void operator()(const char_type *value) {
     internal::handle_cstring_type_spec(
-          specs_.type_, cstring_spec_handler(*this, value));
+          static_cast<char>(specs_.type_), cstring_spec_handler(*this, value));
```
vitaut: I think it would be better to make character type a template parameter of `handle_cstring_type_spec` instead, because otherwise an invalid type specifier can be transformed into a valid one by a narrowing cast which is undesirable.

Same in similar places below.

foonathan: Oh, you just did that, I was grepping through the code and couldn't find those casts anymore.. :D

vitaut: No worries =)

PR #681, opened 2018-03-15
Fix warnings on gcc and clang

eliaskosunen:
First commit fixes a `-Wpedantic` warning when compiling gcc and clang.
Second commit defines `FMT_UNION` as `union` rather than `struct` on clang.


vitaut:
Merged, thanks!

PR #693, opened 2018-03-26
 Removing multiple VS warning C4459

bhaskar-priya:
removed multiple Visual Studio 2017 warning (C4459)


vitaut:
I suggest keeping the names of local variables and parameters and removing the global `fill` and `width` variables https://github.com/fmtlib/fmt/blob/4fea018b2d3de314120a79e638499319f183429b/include/fmt/format.h#L3253 and corresponding tests, because the latter are part of the obsolete write API.

vitaut:
Fixed the warnings by removing the obsolete API in https://github.com/fmtlib/fmt/commit/050f3f1f0e57a13753dfb2b2eac0004ced74e6ae, but thanks anyway!

PR #694, opened 2018-03-26
Add make_printf_args

### LINKED ISSUE #687: Missing fmt::make_printf_args
I had to add `Added fmt::make_printf_args` to `fmt/printf.h` so I could pass arguments to `fmt::vsprintf`, as in 

```c++
fmt::vsprintf(format, fmt::make_printf_args(std::forward<Args>(args)...));
```

This is my implementation of it:

```c++
template <typename ...Args>
inline arg_store<printf_context<internal::buffer>::type, Args...> make_printf_args(const Args & ... args) {
  return arg_store<printf_context<internal::buffer>::type, Args...>(args...);
}
```

Kronuz:
Fixes #687

### DIFF
--- a/include/fmt/printf.h
+++ b/include/fmt/printf.h
@@ -540,6 +540,10 @@ struct printf_context {
     std::back_insert_iterator<Buffer>, typename Buffer::value_type> type;
 };
 
+template <typename ...Args>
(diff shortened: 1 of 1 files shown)


vitaut:
Merged, thanks!

PR #696, opened 2018-03-28
Fix linker errors using fmt as shared library in MSVC

sebkoenig:
When using fmt as a shared library with FMT_SHARED in MSVC, I got linker errors for the library consuming fmt.
Two definitions could not be found:
- `fmt::internal::error_handler::on_error(const char *message)`
- `char fmt::internal::thousands_sep(locale_provider *lp)`

Since those definitions are hidden in `format-inl.h` I added `FMT_API` macros to make the functions part of the DLL interface.


vitaut:
Merged, thanks!

PR #706, opened 2018-04-11
Added support for custom types with 'toString' function

PotatosFish:
Customizing output for user-defined types doesn't work as documented, and relying on ostream is not the best option, so this enables java-like toString function to format classes.

I feel like this is more understandable and easier to implement for the users.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -629,7 +629,6 @@ inline typename std::enable_if<
 
 template <typename C, typename T, typename Char = typename C::char_type>
 inline typename std::enable_if<
-    !std::is_same<decltype(sizeof(&T::toString)), size_t>::value &&
     !convert_to_int<T, Char>::value &&
     !std::is_convertible<T, basic_string_view<Char>>::value &&
     !std::is_convertible<T, std::basic_string<Char>>::value,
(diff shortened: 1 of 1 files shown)


PotatosFish:
basically you add toString() public function to your class that returns any type of string that casts to basic_string_view, and it should work.

vitaut:
Thanks for the PR, but I am a bit reluctant to add yet another extension mechanism. However, you should be able to use it in your codebase by providing an appropriate definition of `formatter` that falls back to `toString` similarly as it is done for ostream `operator<<` support.

> Customizing output for user-defined types doesn't work as documented

Could you elaborate on that? If it doesn't work then it's a bug.

PotatosFish:
the code

```cpp
#define FMT_HEADER_ONLY
#include <fmt/format.h>
#undef FMT_HEADER_ONLY

struct MyStruct { double a, b; };

void format_arg(fmt::BasicFormatter<char> &f,
  const char *&format_str, const MyStruct &s) {
  f.writer().write("[MyStruct: a={:.1f}, b={:.2f}]", s.a, s.b);
}

int main()
{
	MyStruct m = { 1, 2 };
	std::string s = fmt::format("m={}", n);
}
```

yields the error "no member named 'BasicFormatter' in namespace 'fmt'"

and a grep for BasicFormatter in the include directory shows that there is no implementation of it anywhere

vitaut:
The extension API has changed in the development (`master`) branch to match the standards proposal [P0645 Text Formatting](http://fmtlib.net/Text%20Formatting.html). The documentation for this branch is available in http://fmtlib.net/dev/api.html. Sorry for confusion.

PR #709, opened 2018-04-11
Stop newer CMake's from warning about OLD policy

henryiii:
Fmt sets the `FMT_VERSION` variables after project(), so the NEW option here is safe. Currently including Fmt causes newer versions of CMake to throw warnings. (This problem was fixed in Fmt master)


vitaut:
Merged, thanks!

PR #712, opened 2018-04-18
allow time formatting with wchar_t contexts

DanielaE:
change total specialization of 'struct formatter<std::tm>' into partial specialization 'template <typename Char> struct formatter<std::tm, Char>', backed by matching 'strftime'-like overloads


vitaut:
Merged, thanks!

PR #717, opened 2018-04-22
Fix a -Wundef warning on clang

### LINKED ISSUE #715: `FMT_USE_TRAILING_RETURN` not defined on clang, causes warnings
Please see this CE snippet: https://godbolt.org/g/CVWCax

As a solution, I propose either:
 * adding a macro `FMT_HAS_CXX11`, that checks for `__cplusplus` and/or compiler versions. 
 * defining `FMT_HAS_GXX_CXX11` as a non-zero value on clang as well

I'm not proposing:
 * defining `FMT_USE_TRAILING_RETURN` to zero instead of not defining it at all,

as this would lead to trailing returns to not be used on clang.

Once we reach a solution, I can submit a PR.

eliaskosunen:
Resolves #715

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -131,7 +131,7 @@
 # endif
 #endif
 
-#if FMT_HAS_GXX_CXX11 || FMT_MSC_VER >= 1600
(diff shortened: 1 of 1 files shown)


vitaut:
Merged, thanks!

PR #719, opened 2018-04-25
Fix warnings when with master project set to hidden

henryiii:
Setting the `OLD` version of [this policy](https://cmake.org/cmake/help/latest/policy/CMP0063.html) (or even if it is unset and CMake minimum_version is set below 3.3, which it is) does the wrong thing; it causes GCC to spew a mass of warnings if you try to use fmt from a project that has set hidden symbols. Setting this to `NEW` fixes the issue.

Patch to 4.x because both policy statements are missing from master.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,7 @@ if (POLICY CMP0048) # Version variables
 endif ()
 
 if (POLICY CMP0063) # Visibility
-  cmake_policy(SET CMP0063 OLD)
+  cmake_policy(SET CMP0063 NEW)
 endif (POLICY CMP0063)
 
 # Determine if fmt is built as a subproject (using add_subdirectory)


henryiii:
PS: 

CMake's own documentation says the following:
> While setting policies individually is supported, we encourage projects to set policies based on CMake versions

In CMake 3.12 (not yet released), the following syntax is supported:

```cmake
cmake_minimum_required(VERSION 3.1...3.12)
```

Which indicates you support 3.1+ but have tested it with 3.12 as well. This syntax, due to a trick, does not confuse older CMakes, though it does not set the policies.

A much better long term solution then would be the following:

```cmake
cmake_minimum_required(VERSION 2.8.12...3.12)

 # Tested and working with CMake 3.11
 if(${CMAKE_VERSION} VERSION_LESS 3.12)
     cmake_policy(VERSION ${CMAKE_VERSION})
 endif()
```

(The if block would not need to be changed, and does not need an else).

This should be done after CMake 3.12 comes out, to make sure no policies come out that break `fmt`.

vitaut:
Merged, thanks!

PR #725, opened 2018-04-30
Fixed arg_formatter_base::write_pointer to not mutate the format specs.

mwinterb:
This fixes cases where arg_formatters are reused, like with arg_join.


vitaut:
Merged, thanks!

PR #726, opened 2018-05-01
Fix usage of potentially undefined macro `_LIBCPP_VERSION`

eliaskosunen:
Remove checking the value of a potentially undefined macro `_LIBCPP_VERSION` and replace it with `defined(_LIBCPP_VERSION)`. Previously, this caused warnings on both gcc and clang if `-Wundef` was enabled.


vitaut:
Merged, thanks!

PR #730, opened 2018-05-07
Fix an another -Wundef

eliaskosunen:
Here's yet another -Wundef PR. Today we have FMT_CLANG_VERSION being used without being defined if FMT_GCC_VERSION is less than 600.

Even though the usage of an undefined macro has the well-defined semantics of evaluating to zero, I think doing so is bad practice and ignoring those warnings can easily lead to unintended behavior. I think that a project as widely used as fmt should compile cleanly with the most aggressive warning levels. To accomplish this, I really think CI should be more strict with warnings and use more compilers than just g++-6. I can work on a PR on that as well if you'd like.


eliaskosunen:
For example, currently when compiling with clang you'll receive a boatload of warnings, including -Wsign-conversion, because these are only disabled for g++.

vitaut:
> I really think CI should be more strict with warnings and use more compilers than just g++-6. I can work on a PR on that as well if you'd like.

I think it would be useful to add an old gcc compiler (4.4.x) to the CI config. PRs for this and stricter warnings are welcome.

vitaut:
Merged, thanks!

PR #732, opened 2018-05-08
Fix #433 for FreeBSD 12

### LINKED ISSUE #433: Fix test compilation for FreeBSD
A small patch (made against 3.0.1 tag) that fixes test for FreeBSD.
    
    
    --- test/posix-mock-test.cc.orig        2016-12-01 18:09:30 UTC
    +++ test/posix-mock-test.cc
    @@ -502,7 +502,7 @@ LocaleType newlocale(int category_mask,
       return LocaleMock::instance->newlocale(category_mask, locale, base);
     }
    
    -#ifdef __APPLE__
    +#if defined(__APPLE__) || defined(__FreeBSD__)
     typedef int FreeLocaleResult;
     #else
     typedef void FreeLocaleResult;

dankm:
The fix for #433 is correct for FreeBSD versions less than 12.

FreeBSD 12 changed the type of freelocale to the type defined by POSIX. Check the FreeBSD version when building for FreeBSD.

### DIFF
--- a/test/posix-mock-test.cc
+++ b/test/posix-mock-test.cc
@@ -476,7 +476,7 @@ LocaleType newlocale(int category_mask, const char *locale, LocaleType base) {
   return LocaleMock::instance->newlocale(category_mask, locale, base);
 }
 
-#if defined(__APPLE__) || defined(__FreeBSD__)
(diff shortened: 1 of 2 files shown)


vitaut:
Merged, thanks!

PR #733, opened 2018-05-09
Mark the whole class FormatError as FMT_API

jcelerier:
Else on windows across DLLs the vtable is not visible which causes linking error


vitaut:
Merged, thanks!

PR #735, opened 2018-05-10
Add support for ranges, containers and types tuple interface...

Remotion:


include/fmt/ranges.h
```diff
@@ -0,0 +1,224 @@
...
+	const char* p_curr = str;
+	while (*p_curr) {
+		*out++ = *p_curr++;
+	}
+}
+
+template<typename OutputIterator>
+void copy(const char ch, OutputIterator out) {
+	*out++ = ch;
+}
```
vitaut: I suggest putting all of the above in the `internal` namespace.

include/fmt/ranges.h
```diff
@@ -0,0 +1,224 @@
...
+template<typename OutputIterator>
+void copy(const char *str, OutputIterator out) {
+	const char* p_curr = str;
+	while (*p_curr) {
+		*out++ = *p_curr++;
+	}
+}
+
+template<typename OutputIterator>
+void copy(const char ch, OutputIterator out) {
```
vitaut: `const` is not needed here.

include/fmt/ranges.h
```diff
@@ -0,0 +1,224 @@
...
+	*out++ = ch;
+}
+
+} // namespace fmt
+
+
+namespace fmt {
+namespace meta {
```
vitaut: I suggest using `internal` namespace instead of `meta` here for consistency with the rest of the library.

include/fmt/ranges.h
```diff
@@ -0,0 +1,224 @@
...
+
+/// Return true value if T has std::string interface, like std::string_view.
+template<typename T>
+class is_like_std_string {
+	template<typename U> static auto check(U* p) -> decltype(
+		p->find('a')
```
vitaut: Why do we need to require `find` for string-like?

include/fmt/ranges.h
```diff
@@ -0,0 +1,224 @@
...
+		, p->length()
+		, p->data()
+		, int());
+	template<typename  > static void check(...);
+public:
+	static const bool value = !std::is_void< decltype(check<T>(nullptr)) >::value;
+};
+
+template<typename T>
+constexpr bool is_like_std_string_v = is_like_std_string<T>::value;
```
vitaut: Here and elsewhere: constexpr -> FMT_CONSTEXPR

include/fmt/ranges.h
```diff
@@ -0,0 +1,224 @@
...
+	FMT_CONSTEXPR auto parse(ParseContext &ctx) -> decltype(ctx.begin()) { return ctx.begin(); }
+};
+
+template <typename Char, typename Enable = void>
+struct formatting_range : formatting_base<Char>
+{
+	Char prefix = '{';
+	Char delimiter = ',';
+	Char postfix = '}';
+	bool add_spaces = true;
```
vitaut: Containers are usually formatted without embedded spaces, in particular in Python format which this library is modeled after:

```c++
>>> '{}'.format([1, 2])
'[1, 2]'
```
so I suggest dropping this flag (or at least setting it to false).

include/fmt/ranges.h
```diff
@@ -0,0 +1,224 @@
...
+template<typename TupleT, typename Char>
+struct formatter< TupleT, Char
+	, std::enable_if_t<fmt::meta::is_tuple_like_v<TupleT>> >
+{
+	fmt::formatting_tuple<Char> formating;
+
+	template <typename ParseContext>
+	FMT_CONSTEXPR auto parse(ParseContext &ctx) -> decltype(ctx.begin()) {
+		return formating.parse(ctx);
+	}
```
vitaut: This won't be needed if you inherit from formatting_tuple.

include/fmt/ranges.h
```diff
@@ -0,0 +1,224 @@
...
+
+	template <typename FormatContext = format_context>
+	auto format(const TupleT &values, FormatContext &ctx) -> decltype(ctx.out()) {
+		auto out = ctx.out();
+		std::ptrdiff_t i = 0;
```
vitaut: Why not size_t?

include/fmt/ranges.h
```diff
@@ -0,0 +1,224 @@
...
+
+		return ctx.out();
+	}
+};
+
+} // namespace fmt
+
+
+
+namespace fmt {
```
vitaut: Can be merged with the above namespace

include/fmt/ranges.h
```diff
@@ -0,0 +1,224 @@
...
+
+
+template<typename RangeT, typename Char>
+struct formatter <RangeT, Char, std::enable_if_t<fmt::meta::is_range_v<RangeT>> >
+{
+	static constexpr std::ptrdiff_t range_length_limit = FMT_RANGE_OUTPUT_LENGTH_LIMIT; // output only up to N items from the range.
```
vitaut: Why not size_t?

include/fmt/ranges.h
Remotion: Done

include/fmt/ranges.h
Remotion: Done

include/fmt/ranges.h
Remotion: Done

include/fmt/ranges.h
Remotion: Done

include/fmt/ranges.h
Remotion: Done, set add_spaces to false per default.

include/fmt/ranges.h
Remotion: Done, replaced by size_t.
I needed signed type in older implementation but now it can be size_t.

include/fmt/ranges.h
Remotion: Done

include/fmt/ranges.h
Remotion: Done, ditto.

include/fmt/ranges.h
```diff
@@ -0,0 +1,226 @@
...
+  Char delimiter = ',';
+  Char postfix = '}';
+  static FMT_CONSTEXPR_DECL const bool add_spaces = false;
+};
+
+template <typename Char, typename Enable = void>
+struct formatting_tuple : formatting_base<Char> {
+  Char prefix = '[';
+  Char delimiter = ',';
+  Char postfix = ']';
```
vitaut: Tuples are normally delimited by parentheses `()` rather than square brackets `[]`.

include/fmt/ranges.h
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  template <typename FormatContext = format_context>
+  auto format(const TupleT &values, FormatContext &ctx) -> decltype(ctx.out()) {
+    auto out = ctx.out();
+    std::size_t i = 0;
+    internal::copy(formatting.prefix, out);
+    internal::for_each(values, [&](const auto &v) {
+      if (i++ > 0) {
+        internal::copy(formatting.delimiter, out);
```
vitaut: I suggest adding a space after delimiter as in `(1, 2)`.

include/fmt/ranges.h
```diff
@@ -0,0 +1,226 @@
...
+    return ctx.out();
+  }
+};
+
+template <typename RangeT, typename Char>
+struct formatter< RangeT, Char,
+    typename std::enable_if<fmt::internal::is_range<RangeT>::value>::type> {
+
+  static FMT_CONSTEXPR_DECL const std::size_t range_length_limit =
+      FMT_RANGE_OUTPUT_LENGTH_LIMIT; // output only up to N items from the range.
```
vitaut: Why do we need a limit? User can always pass a subrange to `fmt::join` if they don't want to format everything.

test/ranges-test.cc
```diff
@@ -0,0 +1,90 @@
...
+}
+
+TEST(RangesTest, FormatTuple) {
+  std::tuple<int64_t, float, std::string> tu1{42, 3.14159265358979f,
+                                              "this is tuple"};
+  EXPECT_EQ("[42,3.14159,this is tuple]", fmt::format("{}", tu1));
+}
+
+#if (__cplusplus > 201402L) || \
+    (defined(_MSVC_LANG) && _MSVC_LANG > 201402L && _MSC_VER >= 1910)
```
vitaut: Is this to check if `constexpr` is supported?

test/ranges-test.cc
```diff
@@ -0,0 +1,90 @@
...
+// All rights reserved.
+//
+// For the license information refer to format.h.
+//
+// Copyright (c) 2018 - present, Remotion (Igor Schulz)
+// All Rights Reserved
+// {fmt} support for ranges, containers and types tuple interface.
+
+#ifdef WIN32
+#define _CRT_SECURE_NO_WARNINGS
```
vitaut: Why is this needed?

test/ranges-test.cc
Remotion: This is to check if  'if constexpr' is supported.

test/ranges-test.cc
Remotion: This can be removed.

include/fmt/ranges.h
Remotion: Because it could be and infinite range, or nested range where user can not use fmt::join.
`std::tuple<std::vector<std::vector<int>>>`

include/fmt/ranges.h
Remotion: Using `()` in last commit now.

include/fmt/ranges.h
Remotion: This can be configured now by `bool add_delimiter_spaces`

vitaut:
Merged with minor formatting tweaks in https://github.com/fmtlib/fmt/commit/e3f7f3a2e9544ec2db3d6997c78b77f498948ff8. Thanks!

PR #736, opened 2018-05-11
Add more compilers to CI and increase FMT_PEDANTIC warning levels

eliaskosunen:
In reference to #730 

This PR contains:
1. Add the following compilers/environments to Travis (all on Linux with Debug configuration):
  - clang 6.0 with C++14
  - clang 4.0 with C++11
  - g++ 4.8 with C++11
  - g++ 4.4 with C++11
2. Add Visual Studio 2013 and 2015 to Appveyor, while removing Release builds to decrease build time
3. Include a _lot_ more warnings in FMT_PEDANTIC, see CMakeLists.txt for more information
4. Add FMT_WERROR, which is now used on Travis but not on Appveyor
5. Fix every warning on g++ and clang, either by explicitly ignoring them through compiler pragmas or actually fixing them
6. Restructure appveyor.yml and remove appveyor-build.py

Rationale behind some of the changes:
1. Clang is a major compiler on Linux, and is not equivalent to Apple clang. Dismissing it entirely would be silly. g++ 4.8, because it's the default compiler on Trusty. Also, there are now also tests with C++11, which I think is beneficial.
2. There are now 6 build jobs on Appveyor. That number would increase to 12 should release builds be added back. I think it's more important to test multiple platforms rather than multiple build configurations that really only change the compiler optimization level. Because the free Travis version only allows 4 build jobs to be run in parallel, doing the same thing there would decrease the total number of jobs from 10 to 8 and thus decrease build time significantly, but I'm not sure if we want to do that
3. \-
4. Not used on Appveyor because there are plenty of warnings yet to be fixed
5. \-
6. This way we'll see the actual build messages in the dedicated Appveyor interface

(Potentially) blocking issues before merging:
- [x] The recent ranges-PR #735 doesn't support C++11 and causes build errors
- [ ] Build errors on g++ 4.8 and 4.4
- [ ] Build errors on Visual Studio 2013
- [x] Test failures on all Visual Studio versions over at Appveyor
- [x] Warnings on Visual Studio
- [x] I'm not sure which g++ and clang versions introduced which warnings, so adding/suppressing these could lead to errors on compiler versions that aren't tested


eliaskosunen:
Testing with C++98 could also be useful.

.travis.yml
```diff
@@ -1,27 +1,85 @@
...
-  matrix:
-    - BUILD=Doc
-    - BUILD=Debug STANDARD=14
-    - BUILD=Release STANDARD=14
 
 matrix:
   exclude:
-    - os: osx
-      env: BUILD=Doc
+    - env: TRAVIS_EMPTY_JOB_WORKAROUND=true
```
vitaut: What is this for?

.travis.yml
```diff
@@ -1,27 +1,85 @@
...
   include:
+      # Documentation
+    - env: CXX_COMPILER=g++-6 BUILD=Doc
+      sudo: required
+      compiler: gcc
```
vitaut: `CXX_COMPILER` and `compiler` are probably unneeded for building docs

.travis.yml
```diff
@@ -51,17 +109,22 @@ matrix:
...
-# Install gcc-6 for extended constexpr support.
-addons:
-  apt:
-    sources:
-    - ubuntu-toolchain-r-test
-    packages:
-    - g++-6
+install:
+    - DEPS_DIR="${TRAVIS_BUILD_DIR}/deps"
+    - mkdir -p ${DEPS_DIR} && cd ${DEPS_DIR}
```
vitaut: Where is `DEPS_DIR` used?

include/fmt/core.h
```diff
@@ -30,10 +30,10 @@
 # define FMT_HAS_INCLUDE(x) 0
 #endif
 
-#if defined(__GNUC__) && !defined(__clang__)
-# define FMT_GCC_VERSION (__GNUC__ * 100 + __GNUC_MINOR__)
```
vitaut: Please bring `FMT_GCC_VERSION` back as it is used in this header.

include/fmt/core.h
```diff
@@ -545,6 +545,11 @@ struct typed_value : value<Context> {
 template <typename Context, typename T>
 FMT_CONSTEXPR basic_format_arg<Context> make_arg(const T &value);
 
+#if FMT_GCC_VERSION >= 406
+# pragma GCC diagnostic push
+# pragma GCC diagnostic ignored "-Wuseless-cast"
+#endif
```
vitaut: I suggest either fixing the warning or not setting `-Wuseless-cast` in CMake config if the warning is too noisy. Similarly in other cases since it's better not to pollute the code with numerous ifdefs for spurious warnings.

support/appveyor.yml
```diff
@@ -1,22 +1,54 @@
-configuration:
-  - Debug
-  - Release
+configuration: Debug
```
vitaut: I think we should build both in Debug and Release modes.

support/appveyor.yml
```diff
@@ -1,22 +1,54 @@
...
+    
+ - ps: cmake -G "$($env:CMAKE_GENERATOR)$($env:CMAKE_GENERATOR_SUFFIX)" -DFMT_WERROR=OFF -DFMT_PEDANTIC=ON -DCMAKE_BUILD_TYPE=Debug ..
+
+build:
+  project: C:\projects\fmt\build\fmt.sln
+  parallel: true
+  verbosity: normal
 
-build_script:
-  - python support/appveyor-build.py
```
vitaut: Please keep the Python script, since it's easier to deal with configs in one language and we might want to reintroduce MinGW at some point.

test/gtest-extra-test.cc
```diff
@@ -195,33 +195,39 @@ TEST(ExpectSystemErrorTest, DoesNotGenerateUnreachableCodeWarning) {
 }
 
 TEST(AssertionSyntaxTest, ExceptionAssertionBehavesLikeSingleStatement) {
-  if (::testing::internal::AlwaysFalse())
+  if (::testing::internal::AlwaysFalse()) {
```
vitaut: I don't think it's necessary, just adds syntactic noise.

include/fmt/core.h
eliaskosunen: It's also used in format.h before the inclusion of core.h, but sure.

include/fmt/core.h
eliaskosunen: Not enabling the warning during testing is only ignoring the problem. People may and will have that warning enabled when including fmt, and they'll have to resort to similar pragma-trickery in order to silence the compiler about issues they have no control over. It's better to explicitly disable a single warning for a small piece of code.

For that specific case, though, fixing the warning should be possible with ease. I don't think the same can be said for the other instances.

.travis.yml
eliaskosunen: When creating a Travis build matrix, that is sometimes necessary. I just forgot it there.

.travis.yml
eliaskosunen: True.

.travis.yml
eliaskosunen: Nothing, removed.

support/appveyor.yml
eliaskosunen: That'll increase the build time from ~30 mins to an hour (6 -> 12 jobs, 5 min/job), but sure. I think we should look into some ways to decrease this time.

support/appveyor.yml
eliaskosunen: I agree that the Python script was much more pleasant to work with. By using the Appveyor's built-in build functionality, we can have build messages and test results reported in a more aesthetically pleasing (tm) way, [as you can see here](https://ci.appveyor.com/project/vitaut/fmt/build/build%201357/job/qugyqem428ehobor). I'll look into incorporating the Python script back, prioritizing bringing it back over the PowerShell trickery.

eliaskosunen:
How serious are we about C++98 support or older compilers? Reading the docs I got the impression that C++98 would be dropped for 5.0, is this true? What comes to older compilers, g++ 4.8 and 4.4 and Visual Studio 2013 builds are currently failing.

mwinterb:
Rvalue references, variadic templates, decltype, trailing return types, and deleted functions all seem unconditionally required in the master branch. Unrestricted unions are sort-of required (there's a fallback to struct for GCC 4.4 only).
Feature-wise it seems like VS2015 (unrestricted unions, 2013 for variadic templates), GCC 4.4 (trailing return types), and Clang 3.0 (deleted functions) are probably the minimum baselines for the "big three".
I doubt supporting a compiler without variadic template support is something that's going to come back in v5.

eliaskosunen:
It looks like that's indeed the case. [The documentation](http://fmtlib.net/dev/#portability) should definitely be updated prior to release, as it claims support for VS 2010 and C++98 with minor restrictions.

vitaut:
> How serious are we about C++98 support or older compilers? Reading the docs I got the impression that C++98 would be dropped for 5.0, is this true?

C++98 is only supported on the `4.x` branch. The `master` branch and the forthcoming 5.0 release requires some C++11 features such as variadic templates and type traits.

> Rvalue references, variadic templates, decltype, trailing return types, and deleted functions all seem unconditionally required in the master branch. Unrestricted unions are sort-of required (there's a fallback to struct for GCC 4.4 only). Feature-wise it seems like VS2015 (unrestricted unions, 2013 for variadic templates), GCC 4.4 (trailing return types), and Clang 3.0 (deleted functions) are probably the minimum baselines for the "big three". I doubt supporting a compiler without variadic template support is something that's going to come back in v5.

You are completely right. I plan to update the docs before the next release.

eliaskosunen:
As of now, ranges.h requires C++14 for its use of generic lambdas. This is breaking the builds using C++11. There should probably be a way to exclude range tests during build configuration or to fix ranges.h to only use C++11 features.

vitaut:
> As of now, ranges.h requires C++14 for its use of generic lambdas. 

We can temporarily exclude `ranges-test.cc` from the build until C++11 compatibility is fixed.

ukreator:
Is gcc 4.4 support really necessary for 5.0? It's 9 years old already

eliaskosunen:
To me it seems like this would be ready for merging. Are you still concerned of the disabled warnings?

vitaut:
> Is gcc 4.4 support really necessary for 5.0? It's 9 years old already

I think gcc 4.4 achieves a good balance between being old enough to be available even on legacy systems, but new enough to provide most important C++11 features that we need.

> Are you still concerned of the disabled warnings?

Yes. I'm happy to merge the CI improvements, but IMHO explicitly disabling warnings in the code adds too much noise for little value since warnings not included in `-Wall` often give too many false positives. Could you remove additional diagnostic `#pragma`s from the code and corresponding switches that trigger them from CMake? I'm open to discussing enabling specific warnings, but not wholesale and better in separate PRs.

vitaut:
Compilation on gcc 4.4 should be fixed now.

include/fmt/ranges.h
```diff
@@ -126,16 +126,46 @@ struct is_tuple_like {
     is_tuple_like_<T>::value && !is_range_<T>::value;
 };
 
-template <size_t... Is, class Tuple, class F>
-void for_each(std::index_sequence<Is...>, Tuple &&tup, F &&f) noexcept {
+// Check for integer_sequence
+#if defined(__cpp_lib_integer_sequence) || FMT_MSC_VER >= 1910
```
mwinterb: Why 1910 for MSVC? cppreference lists that it should be in VS2015, which is 1900. https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros

appveyor.yml
```diff
@@ -0,0 +1 @@
+support/appveyor.yml
```
vitaut: I don't think this is necessary. `appveyor.yml` is intentionally placed in `support` to reduce clutter in the top-level directory.

.travis.yml
```diff
@@ -1,29 +1,94 @@
...
 matrix:
-  exclude:
-    - os: osx
-      env: BUILD=Doc
   include:
+      # Documentation
+    - env: BUILD=Doc
+      sudo: required
+      # g++ 6 on Linux with C++14
+    - env: CXX_COMPILER=g++-6 BUILD=Debug STANDARD=14
```
vitaut: Why not use `CXX` instead of `CXX_COMPILER` and simplify `before_script`?

include/fmt/ranges.h
```diff
@@ -128,24 +128,46 @@ struct is_tuple_like {
...
+using make_index_sequence = std::make_index_sequence<N>;
 #else
-# define FMT_USE_INTEGER_SEQUENCE 0
+template <typename T, T... N>
+struct integer_sequence {
+    typedef T value_type;
+
+    static FMT_CONSTEXPR std::size_t size() {
+        return sizeof...(N);
+    }
```
vitaut: Please use 2-space indent for consistency (https://github.com/fmtlib/fmt/blob/master/CONTRIBUTING.rst).

support/appveyor-build.py
```diff
@@ -21,13 +23,25 @@
...
+        generator = 'Visual Studio 12 2013'
+    elif image == 'Visual Studio 2015':
+        generator = 'Visual Studio 14 2015'
+    elif image == 'Visual Studio 2017':
+        generator = 'Visual Studio 15 2017'
     if platform == 'x64':
         generator += ' Win64'
     cmake_command.append('-G' + generator)
+    # build_command = ['msbuild', r'C:\projects\fmt\build\fmt.sln', '/m:4',
+    #        r'/logger:"C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"']
```
vitaut: Remove?

support/appveyor-build.py
```diff
@@ -21,13 +23,25 @@
...
     build_command = ['cmake', '--build', '.', '--config', config, '--', '/m:4']
-    test_command = ['ctest', '-C', config]
+    test_command = ['ctest', '-C', config, '-T', 'Test']
+    with open('DartConfiguration.tcl', 'w+') as file:
```
vitaut: What is it for?

support/appveyor-test-upload.ps1
```diff
@@ -0,0 +1,11 @@
...
+  [Parameter(Mandatory=$true)][string]$configuration,
+  [Parameter(Mandatory=$true)][string]$jobid
+)
+
+$XSLInputElement = New-Object System.Xml.Xsl.XslCompiledTransform
+$XslInputElement.Load("https://raw.githubusercontent.com/rpavlik/jenkins-ctest-plugin/master/ctest-to-junit.xsl")
+$file = $(ls Testing\*\Test.xml) | Select -first 1
+$XSLInputElement.Transform((Resolve-Path $file), (Join-Path (Resolve-Path .) "ctest-to-junit-results.xml"))
+$wc = New-Object 'System.Net.WebClient'
+$wc.UploadFile("https://ci.appveyor.com/api/testresults/junit/$jobid", (Resolve-Path .\ctest-to-junit-results.xml))
```
vitaut: Is this file a debugging artefact? Does it need to be in the PR?

eliaskosunen:
Sorry for taking so long with this. I was busy with my finals, but now that my summer vacation has begun, I was able to take another look at this.

Builds are now succeeding, except for VS 2013. I feel like this would now finally be ready for merging.

.travis.yml
eliaskosunen: Travis exports `CXX` as `g++` in between of `matrix` and `before_script`, so the value set in `matrix` would be lost.

support/appveyor-build.py
eliaskosunen: Remove.

vitaut:
Merged with minor tweaks, thanks!

PR #739, opened 2018-05-14
MSVC improvements and data truncation cleanup.

mwinterb:
MSVC is timid about evaluating constexpr functions unless it has to, so the "TYPES" variables end up in read-write memory even though the optimizer removes the initializer. Making TYPES constexpr causes MSVC to try harder to initialize these variables at compile time, which also ends up completely removing the (named) variable from the final compiled binary.
Fixed a data truncation warning being reported in ostream-test.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -73,7 +73,8 @@
 # endif
 #endif
 
-#if FMT_HAS_FEATURE(cxx_explicit_conversions)
+#if FMT_HAS_FEATURE(cxx_explicit_conversions) || \
+    FMT_MSC_VER >= 1800
 # define FMT_EXPLICIT explicit
 #else
 # define FMT_EXPLICIT
@@ -986,8 +987,17 @@ class format_arg_store {
 
   friend class basic_format_args<Context>;
 
+  static FMT_CONSTEXPR uint64_t get_types() {
+    return IS_PACKED ? internal::get_types<Context, Args...>()
+                : -static_cast<int64_t>(NUM_ARGS);
+  }
+
  public:
+#if FMT_USE_CONSTEXPR
+  static constexpr uint64_t TYPES = get_types();
+#else
   static const uint64_t TYPES;
+#endif
 
 #if FMT_GCC_VERSION && FMT_GCC_VERSION <= 405
   // Workaround an array initialization bug in gcc 4.5 and earlier.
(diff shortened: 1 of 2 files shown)


include/fmt/core.h
```diff
@@ -73,7 +73,8 @@
 # endif
 #endif
 
-#if FMT_HAS_FEATURE(cxx_explicit_conversions)
+#if FMT_HAS_FEATURE(cxx_explicit_conversions) || \
+    FMT_MSC_VER >= 1800
```
vitaut: MSVC 18.0 is the minimum supported version in `master`, so the check is not needed.

include/fmt/core.h
```diff
@@ -987,7 +988,13 @@ class format_arg_store {
   friend class basic_format_args<Context>;
 
  public:
+#if FMT_USE_CONSTEXPR
+  static constexpr uint64_t TYPES =
+      IS_PACKED ? internal::get_types<Context, Args...>()
+                : -static_cast<int64_t>(NUM_ARGS);
```
vitaut: Please factor out the initializer into a separate function and reuse here and below to reduce copy-paste.

vitaut:
> Since part of the reason for the timidity about not evaluating internal::get_types at compile time is because they've been buggy, do you think there should be explicit tests validating the correctness of TYPES, or do you believe the existing tests should be good enough?

What do you mean by get_types being buggy? They are not `constexpr` only for compatibility with gcc 4.4, but normally all the computation is optimized away. In any case, I think existing tests should provide enough coverage.

mwinterb:
Sorry, I don't mean get_types is buggy. I meant something along the lines of "are the existing tests enough of a sanity check to be 'certain' that get_types doesn't come across any bugs in MSVC's fragile constexpr implementation?" I didn't want to make things worse for MSVC users for something that may not have any tangible benefits. As you noted, the computation is optimized away, but it leaves behind a mutable global. New mutable globals trigger some internal reviews at my company. It's easy enough to filter them out in this instance, but not needing to filter them out is even easier.

include/fmt/core.h
mwinterb: Since MSVC doesn't support __has_feature, surely there needs to be some check? FMT_MSC_VER > 0 seems like a silly test.

vitaut: My bad, not sure what I was thinking about =).

vitaut:
> Sorry, I don't mean get_types is buggy. I meant something along the lines of "are the existing tests enough of a sanity check to be 'certain' that get_types doesn't come across any bugs in MSVC's fragile constexpr implementation?"

Got it. I think existing tests should be enough.

> As you noted, the computation is optimized away, but it leaves behind a mutable global. New mutable globals trigger some internal reviews at my company. It's easy enough to filter them out in this instance, but not needing to filter them out is even easier.

Sure. I'm OK with the change, but please address the remaining inline comment above.

mwinterb:
Addressed. I wanted to wait for your guidance on the explicit conversion thing since I figured I was missing something.

vitaut:
Merged, thanks!

PR #759, opened 2018-06-05
Make is_range public API, fix #751

### LINKED ISSUE #751: Can't specialize formatter for templated ranged class
If I include `fmt/ranges.h` I cannot specialize `fmt::formatter` for my type that is ranged and templated.
Say I have this class:

```c++
template<typename T>
class Foo {
public:
    ...
    const_iterator begin() const { ... }
    const_iterator cbegin() const { ... }
    ...
};
```

When trying to provide a specialization of `fmt::formatter` for this class:

```c++
template<typename T>
struct formatter<Foo<T>> {
    ...
};
```

I get an ambiguous template instantiation error because of the specialization in `fmt/ranges.h`:

```
include/fmt/core.h:529:56: error: ambiguous template instantiation for ‘struct fmt::v5::formatter<Foo<int>, char, void>’
     typename Context::template formatter_type<T>::type f;
                                                        ^
In file included from ...:
include/fmt/ranges.h:199:8: note: candidates are: ‘template<class RangeT, class Char> struct fmt::v5::formatter<RangeT, Char, typename std::enable_if<fmt::v5::internal::is_range<T>::value>::type> [with RangeT = Foo<int>; Char = char]’
 struct formatter< RangeT, Char,
        ^~~~~~~~~~~~~~~~~~~~~~~~
     typename std::enable_if<fmt::internal::is_range<RangeT>::value>::type> {
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from ...:
foo.cpp: note:                 ‘template<class T> struct fmt::v5::formatter<Foo<T> > [with T = int]’
     struct formatter<Foo<T>> {
            ^~~~~~~~~~~~~~~~~
```

As a workaround I can force `fmt::internal::is_range` to be false:

```c++
namespace fmt::internal {

template<typename T>
struct is_range<Foo<T>> : std::false_type {};

} // namespace fmt::internal
```

But obviously I would want to avoid defining anything in an internal namespace.

drrlvn:


Remotion:
It would be consistent to move is_tuple_like<> to fmt namespace too.

drrlvn:
@Remotion what about `is_like_std_string`, should I move that to `fmt` as well?

drrlvn:
Moved `is_tuple_like` as well.

Remotion:
Then may be it was not great idea to do this in this PR as well ?
https://github.com/fmtlib/fmt/pull/763

vitaut:
Merged, thanks!

PR #763, opened 2018-06-06
quotes  for string support.

Remotion:
std::string and types like it are not printed in single quotes `'str'`.
fmt::is_range and fmt::is_tuple_like are now in fmt namespace. 

```cpp

#include <tuple>
#include <vector>
#include <string_view>

#define FMT_HEADER_ONLY
#include <fmt/format.h>
#include <fmt/ranges.h>
#include <fmt/printf.h>

int main() {
    fmt::print("{}\n", tup);

    std::vector vec = {'1', '2'};
    fmt::print("{}\n", vec);

    std::vector vec2 = {"2", "3"};
    fmt::print("{}\n", vec2);
}
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,7 +112,7 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
 endif ()
 
 if (MSVC)
-  set(PEDANTIC_COMPILE_FLAGS /W4)
+  set(PEDANTIC_COMPILE_FLAGS /W3)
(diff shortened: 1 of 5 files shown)


vitaut:
Thanks for the PR! I suggest using double quotes (`"..."`) for consistency with C++ strings. Also could you rebase getting rid of multiple "Merge branch ..." commits?

Remotion:
I made a new clear PR with all the changes.

vitaut:
> I made a new clear PR with all the changes.

Thanks!

PR #766, opened 2018-06-07
quotes for string support.

Remotion:
std::string and types like it are now printed in double quotes "str".
chars are now printed in single quotes 'A'.

```cpp
#include <tuple>
#include <vector>
#include <string_view>

#define FMT_HEADER_ONLY
#include <fmt/ranges.h>

int main() {
    fmt::print("{}\n", tup); // (1, '1', "10", "20", "30")

    std::vector vec = {'1', '2'};
    fmt::print("{}\n", vec); // {'1', '2'}

    std::vector vec2 = {"2", "3"};
    fmt::print("{}\n", vec2); // {"2", "3"}
}
```

### DIFF
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -165,13 +165,15 @@ void for_each(Tuple &&tup, F &&f) {
 
 template<typename Arg>
 FMT_CONSTEXPR const char* format_str_quoted(bool add_space, const Arg&, 
-    typename std::enable_if<!is_like_std_string<typename  std::decay<Arg>::type>::value>::type* = nullptr) {
+  typename std::enable_if<
+    !is_like_std_string<typename std::decay<Arg>::type>::value>::type* = nullptr) {
   return add_space ? " {}" : "{}";
 }
 
 template<typename Arg>
 FMT_CONSTEXPR const char* format_str_quoted(bool add_space, const Arg&, 
-    typename std::enable_if<is_like_std_string<typename  std::decay<Arg>::type>::value>::type* = nullptr) {
+  typename std::enable_if<
+    is_like_std_string<typename std::decay<Arg>::type>::value>::type* = nullptr) {
   return add_space ? " \"{}\"" : "\"{}\"";
 }
 
@@ -212,7 +214,10 @@ private:
         }
         internal::copy(formatting.delimiter, out);
       }
-      format_to(out, internal::format_str_quoted((formatting.add_delimiter_spaces && i > 0), v), v);
+      format_to(out,
+                internal::format_str_quoted(
+                    (formatting.add_delimiter_spaces && i > 0), v),
+                v);
       ++i;
     }
 
@@ -275,7 +280,10 @@ struct formatter<RangeT, Char,
         }
         internal::copy(formatting.delimiter, out);
       }
-      format_to(out, internal::format_str_quoted((formatting.add_delimiter_spaces && i > 0), *it), *it);
+      format_to(out,
+                internal::format_str_quoted(
+                    (formatting.add_delimiter_spaces && i > 0), *it),
+                *it);
       if (++i > formatting.range_length_limit) {
         format_to(out, " ... <other elements>");
         break;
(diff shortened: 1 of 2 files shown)


include/fmt/ranges.h
```diff
@@ -162,6 +162,33 @@ void for_each(Tuple &&tup, F &&f) {
   const auto indexes = get_indexes(tup);
   for_each(indexes, std::forward<Tuple>(tup), std::forward<F>(f));
 }
+
+template<typename Arg>
+FMT_CONSTEXPR const char* format_str_quoted(bool add_space, const Arg&, 
+    typename std::enable_if<!is_like_std_string<typename  std::decay<Arg>::type>::value>::type* = nullptr) {
```
vitaut: nit: could you wrap long (>80 char length) lines per the style guide (https://google.github.io/styleguide/cppguide.html#Line_Length) here and below?

Sarcasm:
It would be nice if the quoting worked the same as [std::quoted](http://en.cppreference.com/w/cpp/io/manip/quoted).
That is, embedded quotes are escaped too.

Also, I'm would really love it if there was something similar to python format's value conversion flag:
- https://pyformat.info/#conversion_flags

Notably, an equivalent to the `!r` flag.

That is:

```cpp
std::string s = "abc \"def\" hij";
fmt::print("s is: {!r}\n", s);
```

Would output:

```
s is: "abc \"def\" hij"
```


The reason I'm asking for this, is when formatting error messages.
If I get an invalid string value, which might be a path, an executable name, or anything else.
I'd like to make sure my error messages fits on one line and is somewhat unambiguous:

No quote at all:

    error: Value abc def is not valid.
    error: Value abc "def is not valid.
    error: Value abc
    def is not valid.
    error: Invalid value: abc def

Vs "dumb quotes" (fmt::format("\"{}\"", value)):

    error: Value "abc def" is not valid.
    error: Value "abc "def" is not valid.
    error: Value "abc
    def" is not valid.
    error: Invalid value: "abc def  "

Vs string representation:

    error: Value "abc def" is not valid.
    error: Value "abc \"def" is not valid.
    error: Value "abc\ndef" is not valid.
    error: Invalid value: "abc def  "

Note: `std::quoted` does not handle newline specifically.

Sorry if this comment is out of scope for this PR.

vitaut:
> It would be nice if the quoting worked the same as std::quoted.

That makes sense, but I suggest doing it in a separate PR.

SuperWig:
Would it be a good idea to have an is_char type trait so people have the option to add their own chars?

Remotion:
Adding  fmt::is_char<char>  and use it instate of function specialization for char ?

How about  char, wchar_t, char16_t, char32_t ?

SuperWig:
Yeah was something like
```
template <typename T>
struct is_char : std::integral_constant<bool,
	std::is_same<char, T>::value ||
	std::is_same<char16_t, T>::value ||
	std::is_same<char32_t, T>::value> {};
---------------
template<typename Arg>
FMT_CONSTEXPR const char* format_str_quoted(bool add_space, const Arg&, 
    typename std::enable_if<!is_like_std_string<typename  std::decay<Arg>::type>::value &&
	!is_char<typename std::decay<Arg>::type>::value>::type* = nullptr) {
  return add_space ? " {}" : "{}";
}
------------------------
template <typename Arg>
FMT_CONSTEXPR const char* format_str_quoted(bool add_space, const Arg&,
	typename std::enable_if<is_char<typename std::decay<Arg>::type>::value>::type* = nullptr) {
    return add_space ? " '{}'" : "'{}'";
}
```
Though wouldn't you need to separate wchar_t since it uses different a different format? And apparently char16_t and char32_t aren't supported by fmt. I get this error
```
fmt\core.h(541) error C2079: 'f' uses undefined struct 'fmt::v5::formatter<T,char,void>'
```

vitaut:
@Remotion, merged, thanks! Could you by any chance submit a section about the new ranges functionality including quotation to the docs?

PR #767, opened 2018-06-07
Added support for rgb color output.

Remotion:
```cpp
#include <fmt/colors.h>

int main() {
    fmt::print("{}\n", tup);
    fmt::print(fmt::rgb(10,100,200),"{}\n", tup);
    fmt::print(fmt::rgb(200,100,200),"{}\n", tup);
    fmt::print(fmt::rgb(20,200,20),"{}\n", tup);
    fmt::print(fmt::rgb(20,0,250),"{}\n", tup);
}
```


include/fmt/colors.h
```diff
@@ -0,0 +1,238 @@
...
+}
+
+// rgb foreground color and background color
+template <typename... Args>
+inline void print(rgb fd, rgb bg, string_view format_str, const Args & ... args) {
+  vprint_rgb(fd, bg, format_str, make_format_args(args...));
+}
+
+
+enum hex : uint32_t { // use enum class ?
```
eliaskosunen: This would probably be better as an enum class. Also, maybe rename it to something more descriptive, like `color`? I'm also not sure of the CamelCasing of the variants, and if it's consistent with the rest of the library.

Remotion: > This would probably be better as an enum class.
Probably but I really do not want this to be too long.

```cpp
fmt::print(fmt::Coral, "message {}\n", msg);
fmt::print(fmt::colors::Coral, "message {}\n", msg);
```

Also color names are from this list 
https://www.w3schools.com/colors/colors_names.asp

Also Visual Studio and VS Code can directly show colors in the editor using this names.

https://marketplace.visualstudio.com/items?itemName=anseki.vscode-color

https://marketplace.visualstudio.com/items?itemName=lihui.vs-color-picker

Unfortunately I could not find a way to support hex colors like #FFFFFF  in C++.

eliaskosunen: What I meant was more like using `snake_case` instead of `CamelCase` since the library and C++ code in general tends to prefer the former over the latter. But you're right about those extensions, they seem to require the color names to be `lowercase` or `CamelCase`.

vitaut: I agree with @eliaskosunen's points that it should be a enum class `color` with snake case enum constants for consistency with the rest of the code:

```c++
enum class colors : uint32_t {
  alice_blue = ...
};
```

vitaut:
Merged with minor tweaks, thanks!

Issue #770, opened 2018-06-08
GCC warnings if [-Wshadow] is enabled. 


Remotion:
```
In file included from \test\warnings_test.cpp:3:0:
fmt/format.h: In function 'Char* fmt::v5::internal::format_decimal(Char*, UInt, unsigned int, ThousandsSep)':
fmt/format.h:1081:55: warning: declaration of 'buffer' shadows a global declaration [-Wshadow]
                             ThousandsSep thousands_sep) {
                                                       ^
In file included from fmt/format.h:67:0,
                 from \test\warnings_test.cpp:3:
fmt/core.h:390:28: note: shadowed declaration is here
 typedef basic_buffer<char> buffer;
                            ^~~~~~
In file included from \test\warnings_test.cpp:3:0:
fmt/format.h: In function 'Iterator fmt::v5::internal::format_decimal(Iterator, UInt, unsigned int, ThousandsSep)':
fmt/format.h:1111:59: warning: declaration of 'buffer' shadows a global declaration [-Wshadow]
   char_type buffer[std::numeric_limits<UInt>::digits10 + 2];
                                                           ^
In file included from fmt/format.h:67:0,
                 from \test\warnings_test.cpp:3:
fmt/core.h:390:28: note: shadowed declaration is here
 typedef basic_buffer<char> buffer;
                            ^~~~~~
In file included from \test\warnings_test.cpp:3:0:
fmt/format.h: In function 'Char* fmt::v5::internal::format_uint(Char*, UInt, unsigned int, bool)':
fmt/format.h:1123:44: warning: declaration of 'buffer' shadows a global declaration [-Wshadow]
                          bool upper = false) {
                                            ^
In file included from fmt/format.h:67:0,
                 from \test\warnings_test.cpp:3:
fmt/core.h:390:28: note: shadowed declaration is here
 typedef basic_buffer<char> buffer;
                            ^~~~~~
In file included from \test\warnings_test.cpp:3:0:
fmt/format.h: In function 'It fmt::v5::internal::format_uint(It, UInt, unsigned int, bool)':
fmt/format.h:1139:64: warning: declaration of 'buffer' shadows a global declaration [-Wshadow]
   char buffer[std::numeric_limits<UInt>::digits / BASE_BITS + 2];
                                                                ^
In file included from fmt/format.h:67:0,
                 from \test\warnings_test.cpp:3:
fmt/core.h:390:28: note: shadowed declaration is here
 typedef basic_buffer<char> buffer;
[...]

PR #771, opened 2018-06-08
 Fixed GCC version test. 

Remotion:
Please see here for more info https://github.com/fmtlib/fmt/issues/770

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -51,7 +51,7 @@
 # define FMT_ICC_VERSION 0
 #endif
 
-#if (defined(__GNUC__) && !defined(__clang__) && __GNUC__ >= 406) || \
+#if (defined(__GNUC__) && !defined(__clang__) && (__GNUC__ * 100 + __GNUC_MINOR__) >= 406) || \
     FMT_CLANG_VERSION
 # pragma GCC diagnostic push
 
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -165,13 +165,15 @@ void for_each(Tuple &&tup, F &&f) {
 
 template<typename Arg>
 FMT_CONSTEXPR const char* format_str_quoted(bool add_space, const Arg&, 
-    typename std::enable_if<!is_like_std_string<typename  std::decay<Arg>::type>::value>::type* = nullptr) {
+  typename std::enable_if<
+    !is_like_std_string<typename std::decay<Arg>::type>::value>::type* = nullptr) {
   return add_space ? " {}" : "{}";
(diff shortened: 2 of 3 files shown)


include/fmt/format.h
```diff
@@ -51,7 +51,7 @@
 # define FMT_ICC_VERSION 0
 #endif
 
-#if (defined(__GNUC__) && !defined(__clang__) && __GNUC__ >= 406) || \
+#if (defined(__GNUC__) && !defined(__clang__) && (__GNUC__ * 100 + __GNUC_MINOR__) >= 406) || \
```
vitaut: I think you could use `FMT_GCC_VERSION` instead of `__GNUC__ * 100 + __GNUC_MINOR__` (possibly after reordering defines).

eliaskosunen: `FMT_GCC_VERSION` is defined in core.h, which is included after this. This also needs to be before the inclusion of core.h to disable warnings from there. On your first review in #736 you said that `FMT_GCC_VERSION` needs to be in core.h.

So that leaves two options that I can think of:
1) Leave it as is
2) Duplicate these warning-disabling maneuvers in both core.h and format.h and move the inclusion of core.h before the disabling of the warnings in format.h, so both can use `FMT_GCC_VERSION`, like this:

```cpp
// core.h
#define FMT_GCC_VERSION /* ... */

#if FMT_GCC_VERSION
// disable
#endif

// ...

#if FMT_GCC_VERSION
// re-enable
#endif
```

```cpp
// format.h

#include "core.h"

#if FMT_GCC_VERSION
// disable
#endif

// ...

#if FMT_GCC_VERSION
// re-enable
#endif
```

eliaskosunen: I said that very incoherently. What I was trying to say was that the warnings need to be disabled before the inclusion of core.h, but `FMT_GCC_VERSION` is defined in core.h. So we can either not use `FMT_GCC_VERSION` in disabling the warnings or disable them in both headers. The latter would probably be preferable, because now the warnings aren't disabled at all if you only include core.h.

vitaut: I think we can live without disabling these warnings in `fmt/core.h` for now, so I suggest just moving `fmt/core.h` to the front and using `FMT_GCC_VERSION` here.

eliaskosunen: The original point of this PR was to fix the version check so there wouldn't be any -Wshadow warnings.

vitaut: Yes and this will fix the reported warnings in `fmt/format.h` (but not in `fmt/core.h`).

PR #774, opened 2018-06-08
Renamed enum color to colors.

Remotion:
Added enum colors conversion to rgb struct.
Added colors_test.cpp.

Now it looks a bit strange. 
We have `enum color`  from format.h and `enum class colors` from colors.h.

### DIFF
--- a/include/fmt/colors.h
+++ b/include/fmt/colors.h
@@ -16,6 +16,8 @@
 
 FMT_BEGIN_NAMESPACE
 
+enum class colors : uint32_t;
+
 // rgb is a struct for red, green and blue colors.
 // We use rgb as name because some editors will show it as color direct in the
 // editor.
@@ -25,6 +27,8 @@ struct rgb {
     : r(r_), g(g_), b(b_) {}
   FMT_CONSTEXPR_DECL rgb(uint32_t hex) 
       : r((hex >> 16) & 0xFF), g((hex >> 8) & 0xFF), b((hex) & 0xFF) {}
+  FMT_CONSTEXPR_DECL rgb(colors hex) 
+      : r((uint32_t(hex) >> 16) & 0xFF), g((uint32_t(hex) >> 8) & 0xFF), b(uint32_t(hex) & 0xFF) {}
   uint8_t r;
   uint8_t g;
   uint8_t b;
@@ -40,7 +44,7 @@ FMT_CONSTEXPR inline void to_esc(uint8_t c, char out[], int offset) {
 
 } // namespace internal
 
-FMT_FUNC void vprint_rgb(rgb fd, string_view format, format_args args) {
+inline void vprint_rgb(rgb fd, string_view format, format_args args) {
   char escape_fd[] = "\x1b[38;2;000;000;000m";
   static FMT_CONSTEXPR_DECL const char RESET_COLOR[] = "\x1b[0m";
   internal::to_esc(fd.r, escape_fd, 7);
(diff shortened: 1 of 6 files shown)


eliaskosunen:
`(v)print_colored` and `enum color` should probably be entirely replaced by this. They aren't even documented, so I don't think it would be a breaking change. This would also allow renaming the enum to the more logical singular-form `color`.

Remotion:
Agree name `color` is better but then there are 2 different enums with same name. 
How can this be resolved then ? Rename or remove enum color from format.h  ?

vitaut:
I suggest removing the old `color` and `print_colored`. Those are undocumented and experimental anyway and the new functionality seems strictly more general.

Remotion:
What about  `fmt::printf(rgb, "");` ?  I could add this too but then colors.h will need to include printf.h.

vitaut:
I don't think we need `fmt::printf(rgb, "")`. The whole point of `printf` is to simplify transition from legacy `printf` and the latter doesn't have color support.

Remotion:
What do we miss here ?

include/fmt/colors.h
```diff
@@ -70,18 +74,34 @@ FMT_FUNC void vprint_rgb(rgb fd, rgb bg, string_view format, format_args args) {
...
 
-// rgb foreground color
+/**
+  Formats a string and prints it to stdout using ANSI escape sequences to
+  specify foreground color 'fd'.
+  Example:
+    fmt::print(fmt::color::red, "Elapsed time: {0:.2f} seconds", 1.23);
+ */
 template <typename... Args>
 inline void print(rgb fd, string_view format_str, const Args & ... args) {
```
vitaut: Why have both `print` and `print_rgb`?

eliaskosunen:
You could try fixing the g++ 4.8 build. https://github.com/Remotion/fmt-fork/blob/74107c9bd8c1df2b74d3cb4c0d98f0dcc60813d7/include/fmt/colors.h#L39-L43

`FMT_CONSTEXPR` expands to `constexpr` when relaxed C++14 constexpr is available and `inline` otherwise. Because `constexpr` functions are implicitly `inline`, you should remove the extra `inline`.

include/fmt/colors.h
Remotion: print_rgb is now removed.

vitaut:
Merged in https://github.com/fmtlib/fmt/commit/ce500635c7a0fbf899a2d158043d6fa8f04304b3 with minor tweaks to prevent breaking the API (we can remove the old color API during the next major release). Thanks!

eliaskosunen:
Is it really necessary to wait for the next major release, though? The previous color API was not documented anywhere, so the feature wasn't really stabilized and it wouldn't really be a breaking change to remove it.

vitaut:
It is still part of the public API, so I'd rather remove it in stages. Maybe not in the major version but at least we need to provide a good migration plan.

vitaut:
I added a comment that the old color code is deprecated (https://github.com/fmtlib/fmt/blob/master/include/fmt/core.h#L1185) and we can make `FMT_EXTENDED_COLORS` the default in one of the minor versions.

vitaut:
@Remotion, could you please review the updated [CONTIBUTING](https://github.com/fmtlib/fmt/blob/ebec00138bd10e949ff34955bcb096387cd32acc/CONTRIBUTING.md) document, particularly the part about licensing, and let me know if you agree with it being applied both to your contributions to {fmt}? The library is likely to be relicensed (#1073) so I'm collecting approval from all earlier contributors.

PR #775, opened 2018-06-09
Fixed GCC version test.

Remotion:
Here is clear PR for this https://github.com/fmtlib/fmt/pull/771.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -51,7 +51,7 @@
 # define FMT_ICC_VERSION 0
 #endif
 
-#if (defined(__GNUC__) && !defined(__clang__) && __GNUC__ >= 406) || \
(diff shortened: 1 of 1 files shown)


include/fmt/format.h
```diff
@@ -51,7 +51,9 @@
 # define FMT_ICC_VERSION 0
 #endif
 
-#if (defined(__GNUC__) && !defined(__clang__) && __GNUC__ >= 406) || \
+#include "core.h"
+
+#if (defined(__GNUC__) && !defined(__clang__) && FMT_GCC_VERSION >= 406) || \
```
eliaskosunen: You only need to check `FMT_GCC_VERSION` since it has a non-zero value iff `defined(__GNUC__) && !defined(__clang__)` is `true`

Remotion:
Is this Ok now ?

vitaut:
Merged in https://github.com/fmtlib/fmt/commit/47268ecd808b86eb35b617d3db3078daf4641fcc, thanks!

PR #790, opened 2018-06-22
Fix sign conversion warnings

foonathan:
Just a couple static casts to fix -Wsign-conversion warnings.

There are a couple of weird things though, please take a look.
For example, why is precision an `int` and not an `unsigned`?


vitaut:
Thanks for the PR, just one small comment inline.

> why is precision an `int` and not an `unsigned`?

-1 is used to denote "no precision" since 0 is a valid precision.

PR #792, opened 2018-06-28
fix redefinition warning for RESET_COLOR

Dark-Passenger:


include/fmt/format-inl.h
```diff
@@ -86,8 +86,8 @@ inline int fmt_snprintf(char *buffer, size_t size, const char *format, ...) {
 # define FMT_SWPRINTF swprintf
 #endif // defined(_WIN32) && defined(__MINGW32__) && !defined(__NO_ISOCEXT)
 
-const char RESET_COLOR[] = "\x1b[0m";
-const wchar_t WRESET_COLOR[] = L"\x1b[0m"; 
+static FMT_CONSTEXPR_DECL const char RESET_COLOR[] = "\x1b[0m";
+static FMT_CONSTEXPR_DECL const wchar_t WRESET_COLOR[] = L"\x1b[0m";
```
vitaut: Why static?

vitaut:
Merged with some tweaks in https://github.com/fmtlib/fmt/commit/ec218a3ad16feb9eb69a3d467b054cfbb7f47f4d. Thanks!

include/fmt/format-inl.h
Dark-Passenger: that would be copy-pasta mistake on my part.

Dark-Passenger:
awesome sorry about the late reply and some pretty cool tweaks if I may say so myself.

PR #797, opened 2018-07-02
Add hint to readme to include ostream.h for user-defined types.

lawrencem99:


vitaut:
Thanks for the PR, but support for user-defined types doesn't require `ostream.h`. As the name suggest the latter provides support for `std::ostream` including `operator<<`. See [Formatting user-defined types](http://fmtlib.net/dev/api.html#formatting-user-defined-types) and [std::ostream support](http://fmtlib.net/dev/api.html#std-ostream-support) in the docs.

Issue #802, opened 2018-07-09
Two warnings fixed in 4.x present in 5.1


henryiii:
There are two warnings that were fixed in the 4.x series that have popped back up again:

In the CMake configuration, when mixing with something that sets the default visibility, I see:

```
  Policy CMP0063 is not set: Honor visibility properties for all target
  types.  Run "cmake --help-policy CMP0063" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.

  Target "fmt" of type "STATIC_LIBRARY" has the following visibility
  properties set for CXX:

    CXX_VISIBILITY_PRESET

  For compatibility CMake is not honoring them for this target.
This warning is for project developers.  Use -Wno-dev to suppress it.
```

This was fixed in #719.

And, when building with NVCC, this warning pops up thousands of times:

```
include/fmt/format.h(1479): warning: statement is unreachable
```

Fixed in #708.

vitaut:
Thanks for porting the fixes to version 5.

PR #803, opened 2018-07-09
Better support for newer CMake's

henryiii:
This fixes the first half of #802 , and is what was recommended in #719: basically, if CMake version is less than 3.12, use the latest features. If cmake is equal or more to 3.12, just take the features from 3.11. (For each new CMake version, once it comes out and is tested, the one number that needs changing is the one in `cmake_minimum_required`, since this syntax is supported in 3.12+).

I've also moved the CMake version printout so that it does not print if this is not the main project (cleaner add as subdirectory).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,12 +1,15 @@
-message(STATUS "CMake version: ${CMAKE_VERSION}")
+cmake_minimum_required(VERSION 3.1.0...3.11)
 
-cmake_minimum_required(VERSION 3.1.0)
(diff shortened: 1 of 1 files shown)


vitaut:
Merged in https://github.com/fmtlib/fmt/commit/0eb01b832c095c9a9e87d58adb5ed73b987c437e, thanks!

vitaut:
Unfortunately I had to revert this change because it was causing problem in MSVC (#809).

PR #804, opened 2018-07-09
Fixing return unreachable warning on NVCC

henryiii:
Fixes the other half of #802.

The return is unreachable since all paths either return or break, which gets them out past the return.

PS: This bug is a showstopper for me for upgrading to fmt 5.1.0, since I get a massive number of copies of warnings in the build process.


vitaut:
Merged, thanks!

Issue #801, opened 2018-07-05
FMT_NOEXCEPT should not depend on FMT_EXCEPTIONS


qis:
As the title suggests, `FMT_NOEXCEPT` should not depend on `FMT_EXCEPTIONS`.

This will allow `std::is_nothrow_*` traits to detect the actual behavior properly and things like `std::move_if_noexcept` to work the same way in both, normal and `-fno-exceptions` builds.

```diff
diff --git i/include/fmt/core.h w/include/fmt/core.h
index 1303f55..96a98f9 100644
--- i/include/fmt/core.h
+++ w/include/fmt/core.h
@@ -131,11 +131,7 @@
 #endif

 #ifndef FMT_NOEXCEPT
-# if FMT_EXCEPTIONS
 #  define FMT_NOEXCEPT FMT_DETECTED_NOEXCEPT
-# else
-#  define FMT_NOEXCEPT
-# endif
 #endif

 // This is needed because GCC still uses throw() in its headers when exceptions
```

EricWF:
`FMT_DETECTED_NOEXCEPT` is sometimes defined to `throw()`. I don't think you want to add `throw()` when exceptions are disabled but `noexcept` is not available.

qis:
@EricWF So the default `FMT_NOEXCEPT` value should also depend on the `noexcept` support?

```diff
diff --git i/include/fmt/core.h w/include/fmt/core.h
index 1303f55..35941c9 100644
--- i/include/fmt/core.h
+++ w/include/fmt/core.h
@@ -126,16 +126,17 @@
     (FMT_GCC_VERSION >= 408 && FMT_HAS_GXX_CXX11) || \
     FMT_MSC_VER >= 1900
 # define FMT_DETECTED_NOEXCEPT noexcept
+# define FMT_HAS_NOEXCEPT 1
 #else
 # define FMT_DETECTED_NOEXCEPT throw()
+# define FMT_HAS_NOEXCEPT 0
 #endif

 #ifndef FMT_NOEXCEPT
-# if FMT_EXCEPTIONS
+# if FMT_HAS_NOEXCEPT
 #  define FMT_NOEXCEPT FMT_DETECTED_NOEXCEPT
 # else
 #  define FMT_NOEXCEPT
-# endif
 #endif
```

qis:
@EricWF So the default `FMT_NOEXCEPT` value should also depend on `noexcept` support?

```diff
[...]

PR #810, opened 2018-07-13
Keep noexcept specifier when exceptions are disabled.

qis:
As discussed in <https://github.com/fmtlib/fmt/issues/801>, this allows builds with disabled exceptions to still benefit from the C++11 `noexcept` specifier.


vitaut:
Merged, thanks!

Issue #809, opened 2018-07-13
Error "Version string portion was too short or too long" for msvc with native cmake


aletschhorn:
When adding fmtlib as a subfolder to any of my projects in MSVC 2017 (Windows 10) with native cmake support (cmake: 3.11.18040201-MSVC_2) I get an error :
```
1> Version string portion was too short or too long.
   at System.Version.VersionResult.SetFailure(ParseFailureKind failure, String argument)
   at System.Version.TryParseVersion(String version, VersionResult& result)
   at System.Version.Parse(String input)
   at System.Version..ctor(String version)
   at libcmake.CMakeProjectBase`3.<ExtractInfoGeneric>d__94`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at libcmake.CMakeLists.<ExtractInfo>d__114.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at libcmake.CMakeLists.<ParseWithoutCacheAsync>d__113.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.VisualStudio.CMake.Project.CMakeListsUtils.<ParseAsync>d__20.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task task)
   at Microsoft.VisualStudio.CMake.Project.CMakeListsCache.<ParseAsync>d__28.MoveNext()
```
I think it is somehow related to the version requirements in the CMakeLists.txt:
```
[...]

PR #811, opened 2018-07-16
Reduce warnings, support #809

henryiii:
Slightly uglier than #803 , but should support MSVC's built-in CMake.


henryiii:
@aletschhorn, please see if this works.

vitaut:
Merged, thanks!

PR #815, opened 2018-07-17
Add unicode support for print and vprint_rgb

phoenixmcallister:
Unicode support was missing for print and vprint_rgb.
Edit: updated because changed char to wchar_t and std::fputs to std::fputws
Edit again: its fucked up because i need to account for wchar_t being two bytes

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -535,7 +535,35 @@ FMT_FUNC void vprint_rgb(rgb fd, string_view format, format_args args) {
   std::fputs(internal::data::RESET_COLOR, stdout);
 }
 
+FMT_FUNC void vprint_rgb(rgb fd, wstring_view format, wformat_args args) {
+  char escape_fd[] = "\x1b[38;2;000;000;000m";
+  internal::to_esc(fd.r, escape_fd, 7);
+  internal::to_esc(fd.g, escape_fd, 11);
+  internal::to_esc(fd.b, escape_fd, 15);
+
+  std::fputs(escape_fd, stdout);
+  vprint(format, args);
+  std::fputs(internal::data::RESET_COLOR, stdout);
+}
+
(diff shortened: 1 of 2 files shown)


include/fmt/format-inl.h
```diff
@@ -522,6 +522,12 @@ FMT_CONSTEXPR void to_esc(uint8_t c, char out[], int offset) {
   out[offset + 1] = static_cast<char>('0' + c / 10 % 10);
   out[offset + 2] = static_cast<char>('0' + c % 10);
 }
+
+FMT_CONSTEXPR void to_esc(uint8_t c, wchar_t out[], int offset) {
```
vitaut: This can be merged into `to_esc` overload above by making the latter a template.

include/fmt/format-inl.h
```diff
@@ -535,6 +541,17 @@ FMT_FUNC void vprint_rgb(rgb fd, string_view format, format_args args) {
   std::fputs(internal::data::RESET_COLOR, stdout);
 }
 
+FMT_FUNC void vprint_rgb(rgb fd, wstring_view format, wformat_args args) {
```
vitaut: Similarly, to prevent copy-paste move the implementation to a new function template, `internal::vprint_rgb`, and reuse here and in `string_view` overload.

include/fmt/format-inl.h
```diff
@@ -551,6 +568,23 @@ FMT_FUNC void vprint_rgb(rgb fd, rgb bg, string_view format, format_args args) {
   vprint(format, args);
   std::fputs(internal::data::RESET_COLOR, stdout);
 }
+
+FMT_FUNC void vprint_rgb(rgb fd, rgb bg, wstring_view format, wformat_args args) {
```
vitaut: Same here.

vitaut:
Closing because this cannot go in as is unfortunately, but feel free to re-open addressing the requested changes.

PR #827, opened 2018-08-07
Re-enable compile-time format-string checking

medithe:
Since the commit 691a7a9 compile-time string checking is (accidently?) not possible anymore.
With this change, it should be possible again.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -120,7 +120,7 @@ FMT_END_NAMESPACE
 # endif
 #endif
 
-#if FMT_USE_USER_DEFINED_LITERALS && !defined(FMT_ICC_VERSION) && \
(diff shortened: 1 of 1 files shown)


vitaut:
Merged, thanks!

vitaut:
Compile-time format checking is in 5.1.0, only the UDLs are not which is not that important considering that they are non-portable anyway.

medithe:
But as an example in the spdlog library, the constexpr checking doesn't work without this fix.
Therefore, it may be a good idea, to release a 5.1.1.

vitaut:
> But as an example in the spdlog library, the constexpr checking doesn't work without this fix.

Could you elaborate why it doesn't work with spdlog? The main way of using `constexpr` checks is by wrapping the format string in the `fmt` macro which is unaffected by this issue.

Issue #146, opened 2015-04-04
MSVC warnings with \W4 about `C4127: conditional expression is constant`


Naios:
aab64b55a4c5db598c123e1a2770b9eb6507d7d8

Example:
```
dep\cppformat\format.h(2715): warning C4127: conditional expression is constant
src\server\shared\Utilities\StringFormat.h(30) : see reference to function template instantiation 'std::string fmt::sprintf<uint32>(fmt::StringRef,const uint32 &)' being compiled
src\server\shared\Logging\Log.h(67) : see reference to function template instantiation 'std::string Trinity::StringFormat<uint32>(const char *,const uint32 &)' being compiled
src\server\shared\database\Field.h(330) : see reference to function template instantiation 'void Log::outMessage<uint32>(const std::string &,const LogLevel,const char *,const uint32 &)' being compiled
```

Or

```
6>C:\Source\TrinityCore335\dep\cppformat\format.h(872): warning C4127: conditional expression is constant (C:\Source\TrinityCore335\src\server\worldserver\Main.cpp)
6>          C:\Source\TrinityCore335\dep\cppformat\format.h(871) : while compiling class template member function 'fmt::internal::MakeValue<char>::MakeValue(unsigned long)'
6>          C:\Source\TrinityCore335\dep\cppformat\format.h(2715) : see reference to function template instantiation 'fmt::internal::MakeValue<char>::MakeValue(unsigned long)' being compiled
6>          C:\Source\TrinityCore335\dep\cppformat\format.h(2715) : see reference to class template instantiation 'fmt::internal::MakeValue<char>' being compiled
6>          C:\Source\TrinityCore335\src\server\shared\Utilities\StringFormat.h(30) : see reference to function template instantiation 'std::string fmt::sprintf<uint32>(fmt::StringRef,const uint32 &)' being compiled
```

The actual sprintf format call is `"SQL::SizeForType(): invalid field type %u", uint32(field->type)`.

We didn't find out what causes this warning, but we have a lot of it in our TrinityCore/TrinityCore@589296da9044be13f921724e8c2b1312bebde4c8 build when compiling with \W4: http://pastebin.com/HNhKtzBZ after upgrading from revision bf8636c9596fbfddded3d0f5879abc7579c9f4dd.

vitaut:
Should be fixed in https://github.com/cppformat/cppformat/commit/fd53bb6fb88a23e38ec4fe331bfe95d7372d49c9. Thanks for reporting!

[...]

PR #838, opened 2018-08-16
Fix C4127 warning in basic_writer<Range>::write_double

eliasdaler:
Hello. 

The situation is similar to #146. 
Got a lot of C4127 warnings (constant expression) on /W4 in VS 2017. This fixes it via the same approach that was previously used.


vitaut:
Merged, thanks!

zoller27osu:
In C++17, these kinds of things are best fixed by using http://wg21.link/p0292 .

PR #839, opened 2018-08-17
Reduce executable size overhead for embedded targets using newlib nano

teajay-fr:
This pull request makes libfmt usable for small embedded targets.

Detecting newlib is necessary as it doesn't provide the function strtod_l.

The second aspect is the dependency on std::locale which bring loads of code overhead (100kb in my tests) just to get hold of the  separation character for thousands. To work around this, a static separator can be configured through the define FMT_STATIC_THOUSANDS_SEPARATOR.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -18,7 +18,9 @@
 #include <cmath>
 #include <cstdarg>
 #include <cstddef>  // for std::ptrdiff_t
+#if !defined(FMT_STATIC_THOUSANDS_SEPARATOR)
 #include <locale>
+#endif
 
 #if defined(_WIN32) && defined(__MINGW32__)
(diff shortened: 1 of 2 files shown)


vitaut:
Merged in https://github.com/fmtlib/fmt/commit/abde38b4fba8c9578d503ea561e1bcc153669d0d, thanks!

vitaut:
@teajay-fr, could you please review the updated [CONTIBUTING](https://github.com/fmtlib/fmt/blob/ebec00138bd10e949ff34955bcb096387cd32acc/CONTRIBUTING.md) document, particularly the part about licensing, and let me know if you agree with it being applied to your contributions to {fmt}? The library is likely to be relicensed (#1073) so I'm collecting approval from all earlier contributors. Thanks!

PR #844, opened 2018-08-27
Remove conversion compiler warnings

medithe:
When compiling with g++8, I get the following two errors:
```
include/fmt/format-inl.h:400:29: error: conversion from ‘int’ to ‘char’ may change value [-Werror=conversion]
       buffer[size++] = zero + static_cast<char>(digit);
                        ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
include/fmt/format-inl.h:416:28: error: conversion from ‘int’ to ‘char’ may change value [-Werror=conversion]
       buffer[size++] = '0' + digit;
                        ~~~~^~~~~~~
```

With this change, the errors are gone.


vitaut:
Merged, thanks!

PR #845, opened 2018-08-28
Fix warnings in pedantic mode in Visual Studio 2017

chronoxor:
format-inl.h(532): warning C4267: '+=': conversion from 'size_t' to 'int', possible loss of data


vitaut:
Thanks for the PR. #844 should fix this and other warnings once merged.

PR #852, opened 2018-09-07
Get rid of implicit-fallthrough warn. in GCC 7 and 8

medithe:
This change enables FMT_FALLLTHROUGH also for GCC-Compilers in the versions 7 and 8.


vitaut:
Merged, thanks!

PR #854, opened 2018-09-10
Get rid of conversion warning in gcc-4.8

medithe:
Get rid of the following warning:
conversion to ‘int’ from ‘long unsigned int’ may alter its value [-Werror=conversion]

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -371,7 +371,7 @@ class fp {
   explicit fp(Double d) {
     // Assume double is in the format [sign][exponent][significand].
     typedef std::numeric_limits<Double> limits;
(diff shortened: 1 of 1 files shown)


include/fmt/format-inl.h
```diff
@@ -371,7 +371,7 @@ class fp {
   explicit fp(Double d) {
     // Assume double is in the format [sign][exponent][significand].
     typedef std::numeric_limits<Double> limits;
-    const int double_size = sizeof(Double) * char_size;
+    const size_t double_size = sizeof(Double) * char_size;
```
vitaut: This might cause narrowing warnings below. I suggest static casting to int instead.

vitaut:
Merged, thanks!

PR #863, opened 2018-09-16
[Docs] Changed to use scoped enum

SuperWig:
Using unscoped enum formats to ints


vitaut:
LGTM, thanks for the PR.

PR #867, opened 2018-09-17
Provide more overloads for the wide string flavour

DanielaE:

### DIFF
--- a/include/fmt/color.h
+++ b/include/fmt/color.h
@@ -209,8 +209,85 @@ struct rgb {
   uint8_t b;
 };
 
-void vprint_rgb(rgb fd, string_view format, format_args args);
-void vprint_rgb(rgb fd, rgb bg, string_view format, format_args args);
+namespace internal {
+
+template <typename Char>
+struct ansi_color_escape {
+  FMT_CONSTEXPR ansi_color_escape(rgb color, const char * esc) FMT_NOEXCEPT {
+    for (int i = 0; i < 7; i++) {
+      buffer[i] = static_cast<Char>(esc[i]);
+    }
+    to_esc(color.r, buffer +  7, ';');
+    to_esc(color.g, buffer + 11, ';');
+    to_esc(color.b, buffer + 15, 'm');
+    buffer[19] = static_cast<Char>(0);
+  }
+  FMT_CONSTEXPR operator const Char *() const FMT_NOEXCEPT { return buffer; }
+
+private:
+  Char buffer[7 + 3 * 4 + 1];
+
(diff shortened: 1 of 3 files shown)


DanielaE:
Great - I totally love to see this happening! I've already changed my code with respect to the `format(rgb color, ...)` family, now on to the `vprint_rgb(rgb color, ...)` family.

DanielaE:
@vitaut Victor, do you have an idea why gcc 4.8 is failing to compile the templated `printf(rgb color, ...)` function, while all other compilers have no problems with it?

vitaut:
gcc 4.8 is finicky. You need to explicitly construct `basic_format_args` to make it happy (see https://github.com/fmtlib/fmt/blob/master/include/fmt/core.h#L1447 for example).

include/fmt/color.h
```diff
@@ -209,18 +209,103 @@ struct rgb {
   uint8_t b;
 };
 
-void vprint_rgb(rgb fd, string_view format, format_args args);
-void vprint_rgb(rgb fd, rgb bg, string_view format, format_args args);
+namespace internal {
+FMT_CONSTEXPR_DECL const char FOREGROUND_COLOR[] = "\x1b[38;2;000;000;000m";
+FMT_CONSTEXPR_DECL const char BACKGROUND_COLOR[] = "\x1b[48;2;000;000;000m";
```
vitaut: These should go to `internal::data`. Also you don't need the `000;000;000m` part because you construct it in `ansi_color_escape`.

include/fmt/color.h
```diff
@@ -209,18 +209,103 @@ struct rgb {
...
+
+template <typename Char>
+struct ansi_color_escape {
+  FMT_CONSTEXPR ansi_color_escape(rgb color, const char esc[]) FMT_NOEXCEPT {
```
vitaut: nit: here and elsewhere `const char []` -> `const char *`

include/fmt/color.h
```diff
@@ -209,18 +209,103 @@ struct rgb {
...
+    to_esc(color.g, buffer, 11, ';');
+    to_esc(color.b, buffer, 15, 'm');
+    buffer[19] = static_cast<Char>(0);
+  }
+  FMT_CONSTEXPR operator const Char *() const FMT_NOEXCEPT { return buffer; }
+
+private:
+  Char buffer[sizeof(FOREGROUND_COLOR)];
+
+  FMT_CONSTEXPR void to_esc(uint8_t c, Char out[], int offset,
```
vitaut: I suggest getting rid of the `offset` parameter and passing buffer pointer adjusted by offset instead.

vitaut:
Thanks!

iz

Issue #865, opened 2018-09-17
compiling error if included both "fmt/ranges.h" and "fmt/ostream.h"


hhggit:
The following code would compile failed with error `ambiguous template instantiation for ‘struct fmt::v5::formatter<fmt::v5::basic_string_view<char>, char, void>’`  if uncommented the line `//#include <fmt/ranges.h>`
``` c++
#define FMT_HEADER_ONLY
#include <fmt/format.h>
#include <fmt/ostream.h>
//#include <fmt/ranges.h>

struct foo {
  friend std::ostream& operator<<(std::ostream& os, const foo&) {
    return os << "foo";
  }
};
int main() {
  fmt::print("{}", foo{});
}
```

vitaut:
Fixed in https://github.com/fmtlib/fmt/commit/70a6a4bb010db3b278c850569d894d95f01400f3. Thanks for reporting and fixing the issue!

PR #868, opened 2018-09-18
prevent ""fmt/range.h"" from specializing fmt::basic_string_view (#865)

hhggit:


vitaut:
Merged in https://github.com/fmtlib/fmt/commit/70a6a4bb010db3b278c850569d894d95f01400f3, thanks!

Issue #742, opened 2018-05-18
compilation fails with C++14 and Intel C++ 2019 beta


kc9jud:
If compiling with the Intel C++ 2019 beta and C++14 enabled, compilation fails:

```
In file included from libraries/relative/relative_me.cpp(14):
./libraries/fmt/format.h(3595): error: a literal operator template must have a template parameter list equivalent to "<char ...>"
  FMT_CONSTEXPR internal::udl_formatter<Char, CHARS...> operator""_format() {
                                                        ^
```

Host info:
```
>> lsb_release -a
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04 LTS
Release:	18.04
Codename:	bionic
```

Compiler info:
```
>> icpc --version
icpc (ICC) 19.0.0.046 20180317
Copyright (C) 1985-2018 Intel Corporation.  All rights reserved.
```
```
>> g++ --version
g++ (Ubuntu 7.3.0-16ubuntu3) 7.3.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

kc9jud:
Related to #311 and #312?

vitaut:
It looks like the Intel C++ compiler doesn't support UDL templates. Does replacing

```c++
#if FMT_USE_USER_DEFINED_LITERALS && \
    ((FMT_GCC_VERSION >= 600 && __cplusplus >= 201402L) || \
    (defined(FMT_CLANG_VERSION) && FMT_CLANG_VERSION >= 304))
```

with 

```c++
#if FMT_USE_USER_DEFINED_LITERALS && !defined(FMT_ICC_VERSION) && \
    ((FMT_GCC_VERSION >= 600 && __cplusplus >= 201402L) || \
    (defined(FMT_CLANG_VERSION) && FMT_CLANG_VERSION >= 304))
```

in https://github.com/fmtlib/fmt/blob/master/include/fmt/format.h#L118 fixes the issue?

vitaut:
Committed a tentative fix (https://github.com/fmtlib/fmt/commit/fba352a92ad22cdbb6de53856ccaa90eabb2621e). However, I wasn't able to verify it since I don't have and Intel compiler at hand, so feel free to reopen if the problem persists.

kc9jud:
fba352a appears to have fixed it. Thanks!

PR #875, opened 2018-09-21
Fix building under NVIDIA's CUDA compiler

CandyGumdrop:
The issue #742 seems to also affect other compilers using the EDG C++ Front End, such as the NVIDIA CUDA compiler.

`nvcc` from CUDA 9.2 gives the following error for me when including fmt 5.2.0 in a CUDA source file built with `-std=c++14`:

    /usr/include/fmt/format.h(3677): error: a literal operator template must have a template parameter list equivalent to "<char ...>"

This patch fixes the issue by also disabling `FMT_UDL_TEMPLATE` when built under nvcc.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -51,6 +51,12 @@
 # define FMT_ICC_VERSION 0
 #endif
 
+#ifdef __NVCC__
+# define FMT_CUDA_VERSION (__CUDACC_VER_MAJOR__ * 100 + __CUDACC_VER_MINOR__)
+#else
+# define FMT_CUDA_VERSION 0
+#endif
+
 #include "core.h"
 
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks!

vitaut:
@CandyGumdrop, could you please review the updated [CONTIBUTING](https://github.com/fmtlib/fmt/blob/ebec00138bd10e949ff34955bcb096387cd32acc/CONTRIBUTING.md) document, particularly the part about licensing, and let me know if you agree with it being applied to your contributions to {fmt}? The library is likely to be relicensed (#1073) so I'm collecting approval from all earlier contributors. Thanks!

CandyGumdrop:
@vitaut I agree with this change in licence.

PR #882, opened 2018-09-26
Fix compilation with intel compilers (ICC/ICPC) v14.0

Luthaf:
The `auto` change was a warning

```
warning #3373: nonstandard use of "auto" to both deduce the type from an initializer and to announce a trailing return type
    auto begin = format_str.data(), end = begin + format_str.size();
```

The `explicit` change was an error

```
error: "explicit" is not allowed
    explicit operator String() const { return String("foo", 3u); }
```

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -89,7 +89,8 @@
 # endif
 #endif
 
-#if FMT_HAS_FEATURE(cxx_explicit_conversions) || FMT_MSC_VER >= 1800
+#if FMT_HAS_FEATURE(cxx_explicit_conversions) || \
+  FMT_GCC_VERSION >= 405 || FMT_MSC_VER >= 1800
 # define FMT_EXPLICIT explicit
 #else
 # define FMT_EXPLICIT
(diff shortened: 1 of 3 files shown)


test/format-test.cc
```diff
@@ -1577,7 +1577,7 @@ struct explicitly_convertible_to_string_like {
       typename String,
       typename = typename std::enable_if<
         std::is_constructible<String, const char*, std::size_t>::value>::type>
-  explicit operator String() const { return String("foo", 3u); }
+  operator String() const { return String("foo", 3u); }
```
vitaut: `explicit` is required here. Please use `FMT_EXPLICIT` instead.

Luthaf: Oh, yeah, this makes sense. Corrected!

include/fmt/core.h
```diff
@@ -89,7 +89,8 @@
 # endif
 #endif
 
-#if FMT_HAS_FEATURE(cxx_explicit_conversions) || FMT_MSC_VER >= 1800
+#if FMT_HAS_FEATURE(cxx_explicit_conversions) || \
+  FMT_GCC_VERSION >= 405 || FMT_MSC_VER >= 1800
```
Luthaf: I used https://en.cppreference.com/w/cpp/compiler_support to check for the GCC version

vitaut:
Merged in https://github.com/fmtlib/fmt/commit/e864acfdbd2a81c6efcc80d93c771e31a0433c62, thanks!

PR #885, opened 2018-09-27
Parameterize core functions on the type of the format string.

DanielaE:
Take #1 of n

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -1290,8 +1290,9 @@ template <typename S>
 struct is_format_string : std::is_empty<format_string_traits<S>> {};
 
 template <typename S>
-struct is_compile_string :
-    std::integral_constant<bool, std::is_base_of<compile_string, S>::value> {};
+struct is_compile_string : std::is_base_of<compile_string, S> {};
+
+#define FMT_CHAR(Str) typename internal::format_string_traits<Str>::char_type
 
 template <typename... Args, typename S>
 inline typename std::enable_if<!is_compile_string<S>::value>::type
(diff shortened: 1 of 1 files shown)


include/fmt/core.h
```diff
@@ -1341,26 +1341,20 @@ typename std::enable_if<
...
-inline typename std::enable_if<
-  is_contiguous<Container>::value, std::back_insert_iterator<Container>>::type
-    format_to(std::back_insert_iterator<Container> out,
-              string_view format_str, const Args &... args) {
-  format_arg_store<format_context, Args...> as{args...};
-  return vformat_to(out, format_str, as);
-}
-
-template <typename Container, typename... Args>
+#define FMT_CHAR(Str) typename internal::format_string_traits<Str>::char_type
```
vitaut: Please add a blank line to separate the macro from the function definition.

include/fmt/core.h
```diff
@@ -1402,21 +1396,21 @@ FMT_API void vprint(std::FILE *f, wstring_view format_str, wformat_args args);
...
- */
-template <typename... Args>
-inline void print(std::FILE *f, string_view format_str, const Args &... args) {
-  format_arg_store<format_context, Args...> as(args...);
-  vprint(f, format_str, as);
-}
-/**
-  Prints formatted data to the file *f* which should be in wide-oriented mode
+
+  **Hint**::
```
vitaut: I don't think `::` is correct here. It's probably better to just have a normal sentence: 

```
Prints formatted data to the file *f*. For wide format strings, *f* should be in wide-oriented mode ...
```

DanielaE: Well, if you say so 😄 I have no clue about rst and just copied the syntax from three lines above. Thanks for the correction and the improved wording.

include/fmt/core.h
vitaut: The macro is there to support gcc versions older than 4.7, e.g. gcc 4.6.4 released in 2013. I don't like it either and eventually it will be replace with the type alias you wrote or similar.

It looks like `FMT_CHAR` is defined twice now.

DanielaE: Oops, too much git history editing without coffee. Fixed now.

vitaut:
Merged in https://github.com/fmtlib/fmt/commit/267fdc7a1bf60d694a6adf9d69cf8f86ed8713ed. Thanks!

PR #887, opened 2018-10-01
Parameterize core functions on the type of the format string.

DanielaE:
Take #2 of n

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -1340,23 +1340,14 @@ template <typename Char>
 struct is_contiguous<internal::basic_buffer<Char> >: std::true_type {};
 
 /** Formats a string and writes the output to ``out``. */
-template <typename Container>
+template <typename Container, typename S>
 typename std::enable_if<
   is_contiguous<Container>::value, std::back_insert_iterator<Container>>::type
     vformat_to(std::back_insert_iterator<Container> out,
-               string_view format_str, format_args args) {
+               const S &format_str,
+               basic_format_args<typename buffer_context<FMT_CHAR(S)>::type> args) {
   internal::container_buffer<Container> buf(internal::get_container(out));
-  vformat_to(buf, format_str, args);
-  return out;
-}
-
-template <typename Container>
-typename std::enable_if<
-  is_contiguous<Container>::value, std::back_insert_iterator<Container>>::type
-    vformat_to(std::back_insert_iterator<Container> out,
-               wstring_view format_str, wformat_args args) {
-  internal::container_buffer<Container> buf(internal::get_container(out));
-  vformat_to(buf, format_str, args);
+  vformat_to(buf, internal::to_string_view(format_str), args);
   return out;
 }
 
@@ -1366,8 +1357,8 @@ inline typename std::enable_if<
   std::back_insert_iterator<Container>>::type
     format_to(std::back_insert_iterator<Container> out, const S &format_str,
               const Args &... args) {
-  return vformat_to(out, internal::to_string_view(format_str),
-                    internal::checked_args<S, Args...>(format_str, args...));
+  internal::checked_args<S, Args...> ca(format_str, args...);
+  return vformat_to(out, internal::to_string_view(format_str), *ca);
 }
 
 template <typename S>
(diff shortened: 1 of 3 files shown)


include/fmt/core.h
```diff
@@ -1334,21 +1334,12 @@ template <typename Char>
...
-  return out;
-}
-
-template <typename Container>
+template <typename Container, typename Char>
 typename std::enable_if<
   is_contiguous<Container>::value, std::back_insert_iterator<Container>>::type
     vformat_to(std::back_insert_iterator<Container> out,
-               wstring_view format_str, wformat_args args) {
+               basic_string_view<Char> format_str,
```
vitaut: Will this work on older gcc if `format_str` is a string literal? Could you add a test?

include/fmt/format.h
```diff
@@ -3563,19 +3557,31 @@ inline format_to_n_result<OutputIt> vformat_to_n(
...
-inline format_to_n_result<OutputIt> format_to_n(
-    OutputIt out, std::size_t n, wstring_view format_str,
-    const Args &... args) {
+
+template <typename OutputIt, typename String, typename... Args>
+inline typename std::enable_if<
+  internal::is_format_string<String>::value &&
+  !std::is_same<FMT_CHAR(String), char>::value,
+  format_to_n_result<OutputIt>>::type format_to_n(
+    OutputIt out, std::size_t n, const String &format_str, const Args &... args) {
```
vitaut: Why do we need two overloads?

DanielaE: Well, `format_to_n` in `master` has different implementations in the function body depending on the actual char type. Therefore I mapped one overload (resembling the old `string_view` one) to the `char` case, and the other (resembling the old `wstring_view` one) to all other `char` types. The `enable_if` switches between the two. At this stage, I haven't dug into the different iterator types used in the function bodies to figure out if they can be merged - opening the road to remove this `enable_if` switch.

include/fmt/core.h
DanielaE: Is this particular overload of `vformat_to` with non-variadic parameters meant to be a public interface? AFAICS it is called only from other interface functions which do have a variadic interface and explicitly convert the given `format_string` to a `basic_string_view`. I am asking because I have a hard time to come up with a test to exercise this `vformat_to` overload.

DanielaE: Nevermind, I've widened the interface and added some tests.

include/fmt/format.h
DanielaE: Whoa, the widened but constrained interface triggers a spectacular failure on all gccs and clangs! I suspect a weird interaction with their stl implementation. MSVC is just fine ...

include/fmt/core.h
DanielaE: Whoa, the widened but contrained interface triggers a spectacular compile failure on all versions of gcc and clang. I suspect a weird interaction with their stl implementation. MSVC is happily chugging along, though ...

DanielaE: For the sake of documentation: [Travis log](https://travis-ci.org/fmtlib/fmt/builds/436191568)
Removing the constraint now ...

DanielaE: No matter what I do, gcc 4.8 is choking. I'd rather stick with my original code!

include/fmt/format.h
vitaut: I think the difference between the two is incidental (I just forgot to update the `wchar_t` overload after changing the `char` one). Both should forward to `vformat_to_n` and can be merged.

include/fmt/core.h
```diff
@@ -1366,8 +1357,10 @@ inline typename std::enable_if<
   std::back_insert_iterator<Container>>::type
     format_to(std::back_insert_iterator<Container> out, const S &format_str,
               const Args &... args) {
+  typedef typename buffer_context<FMT_CHAR(S)>::type context;
+  internal::checked_args<S, Args...> ca(format_str, args...);
   return vformat_to(out, internal::to_string_view(format_str),
-                    internal::checked_args<S, Args...>(format_str, args...));
+                    basic_format_args<context>(ca));
```
vitaut: You can just do `*ca` instead of `basic_format_args<context>(ca)`.

DanielaE: Cool!

DanielaE:
THis should be fine now 🤞

vitaut:
Awesome, thanks a lot!

PR #895, opened 2018-10-07
Fix non-matching char types.

DanielaE:
The non-`char` contexts were probably never excercised in this regard.


vitaut:
Merged, thanks!

PR #896, opened 2018-10-07
tiptoeing into class template argument deduction.

DanielaE:

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -139,6 +139,13 @@
 # endif
 #endif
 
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR but I don't think `fmt::basic_string_view` should provide argument deduction guides. The latter are a C++17 feature where `std::string_view` is available and should be used instead of `fmt::string_view`. `fmt::string_view` is just a workaround for older platforms with a minimal usable API.

PR #897, opened 2018-10-07
Parameterize printf functions on the type of the format string.

DanielaE:

### DIFF
--- a/include/fmt/printf.h
+++ b/include/fmt/printf.h
@@ -577,17 +577,15 @@ struct printf_context {
     std::back_insert_iterator<Buffer>, typename Buffer::value_type> type;
 };
 
-template <typename ...Args>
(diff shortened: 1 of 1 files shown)


include/fmt/printf.h
```diff
@@ -577,17 +577,12 @@ struct printf_context {
...
 };
 
-template <typename ...Args>
-inline format_arg_store<printf_context<internal::buffer>::type, Args...>
-    make_printf_args(const Args & ... args) {
-  return format_arg_store<printf_context<internal::buffer>::type, Args...>(
-      args...);
-}
-typedef basic_format_args<printf_context<internal::buffer>::type> printf_args;
-typedef basic_format_args<printf_context<internal::wbuffer>::type> wprintf_args;
```
vitaut: Removing `printf_args` and `wprintf_args` might break users' code. I suggest keeping them.

PR #901, opened 2018-10-08
minor documentation corrections

kookjr:
In the Format API section, it says fmt:print writes to a file, but it
writes to a stream as referenced from fopen(3), "stream open
functions". Also in the Safety section a typo; tim should be time.


vitaut:
Thanks!

PR #913, opened 2018-10-24
fix: 'format_to_n' compiles 'std::back_inserter' arguments

DanielaE:
std::back_insert_iterators model the OutputIterator concept but differ considerably in their traits and behavior. In particular the former made compilation to fail when format_to_n is given a back_inserter as first argument. The emulation of an OutputIterator is not perfect due to the behavioural differences of back_insert_iterators (e.g. assignment always implies increment) but good enough to be used within fmt's machinery.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -738,35 +738,49 @@ class counting_iterator {
   T &operator*() const { return blackhole_; }
 };
 
-// An output iterator that truncates the output and counts the number of objects
-// written to it.
 template <typename OutputIt>
-class truncating_iterator {
- private:
-  typedef std::iterator_traits<OutputIt> traits;
-
+class truncating_iterator_base {
+ protected:
   OutputIt out_;
   std::size_t limit_;
   std::size_t count_;
-  mutable typename traits::value_type blackhole_;
+
+  truncating_iterator_base(OutputIt out, std::size_t limit)
+    : out_(out), limit_(limit), count_(0) {}
 
  public:
   typedef std::output_iterator_tag iterator_category;
-  typedef typename traits::value_type value_type;
-  typedef typename traits::difference_type difference_type;
-  typedef typename traits::pointer pointer;
-  typedef typename traits::reference reference;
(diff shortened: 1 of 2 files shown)


include/fmt/format.h
```diff
@@ -779,6 +783,40 @@ class truncating_iterator {
...
+      out_ = val;
+    return *this;
+  }
+
+  truncating_iterator& operator++() { return *this; }
+
+  truncating_iterator& operator++(int) { return *this; }
+
+  truncating_iterator& operator*() { return *this; }
+};
```
vitaut: Is it possible to move common code between the two specializations into a superclass to reduce copy-paste?

DanielaE:
Done ...

include/fmt/format.h
```diff
@@ -776,7 +790,34 @@ class truncating_iterator {
...
+
+template <typename OutputIt>
+class truncating_iterator<OutputIt, std::true_type>:
+  public truncating_iterator_base<OutputIt> {
+ public:
+  typedef typename OutputIt::container_type::value_type value_type;
+  typedef void difference_type;
+  typedef void pointer;
+  typedef void reference;
```
vitaut: Why are these `void` and are they different from their counterparts in `iterator_traits`?

DanielaE: `void` actually **is** their type in [`iterator_traits`](https://en.cppreference.com/w/cpp/iterator/back_insert_iterator). We are talking about a `back_insert_operator` here. I've basically just redefined the `value_type` from `void` to the `value_type`of the underlying container.

vitaut: Then they can be moved to `truncating_iterator_base` I think.

DanielaE: I agree, even if it loses information about the underlying iterator type. Anyway, I've refactored these three `void` typedefs into the common base class, the rest must stay as is: these are the essential differences between "regular output iterators" and "back insert interators".

vitaut:
Merged, thanks!

PR #916, opened 2018-10-24
add fmt.pc for pkgconfig

colemickens:
This makes `fmt` easier to use from `meson` or other tools that use `pkgconfig`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ CMakeFiles
 FMT.build
 Makefile
 run-msbuild.bat
+fmt.pc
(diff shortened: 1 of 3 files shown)


fmt.pc
```diff
@@ -0,0 +1,11 @@
+prefix=@out@
```
Mic92: I suppose upstream probably also wants to see some sort of build system integration into cmake using `configure_file`:

https://dailycommit.blogspot.com/2016/10/how-to-generate-pkg-config-file-with.html

vitaut:
Merged, thanks!

Mic92:
Wow, this was fast!

PR #919, opened 2018-10-25
additional test for print with background color

DanielaE:
checks for clashes with other function overloads in fmt with templated parameters on the first or second position in the parameter list


DanielaE:
This will probably fail ea4010d7040bf0b8d98509710da0408d6d769f40

vitaut:
Merged, thanks!

PR #921, opened 2018-10-25
check for property 'mutable iterator'

DanielaE:
constrain the `OutputIt` function parameters to be models of `mutable iterators`.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -3395,6 +3395,51 @@ inline typename buffer_context<Char>::type::iterator format_to(
                               basic_format_args<context>(as));
 }
 
+namespace internal {
+
+// Detect the iterator category of *any* given type in a SFINAE-friendly way.
+// Unfortunately, older implementations of std::iterator_traits are not safe
+// for use in a SFINAE-context.
+
+// the gist of C++17's void_t magic
+template<typename... Ts>
+struct void_ { typedef void type; };
+
+template <typename T, typename Enable = void>
+struct it_category : std::false_type {};
+
+template <typename T>
+struct it_category<T*> { typedef std::random_access_iterator_tag type; };
+
+template <typename T>
+struct it_category<T, typename void_<typename T::iterator_category>::type> {
+  typedef typename T::iterator_category type;
+};
+
+// Detect if *any* given type models the OutputIterator concept
+template <typename It>
+class is_output_iterator {
+  // Check for mutability because all iterator categories derived from
+  // std::input_iterator_tag *may* also meet the requirements of an
+  // OutputIterator, thereby falling into the category of 'mutable iterators'
+  // [iterator.requirements.general] clause 4.
+  // The compiler reveals this property only at the point of *actually
+  // dereferencing* the iterator!
+  template <typename U>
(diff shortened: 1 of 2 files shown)


DanielaE:
please disregard so far, VS2017 passes the tests, but none of the compilers on Travis :disappointed: May be their library implementation is not SFINAE friendly when it comes to `std::iterator_traits` ?

DanielaE:
The hostility towards SFINAE was indeed the killer! Now MSVC, gcc, and clang are happy campers again 😄

include/fmt/format.h
```diff
@@ -3395,6 +3395,43 @@ inline typename buffer_context<Char>::type::iterator format_to(
                               basic_format_args<context>(as));
 }
 
+namespace internal {
+
+template<typename... Ts>
+struct void_ { typedef void type; };
+
+template <typename T, typename Enable = void_<>::type>
```
vitaut: Why `void_<>::type` and not just `void`? A comment might be good if this is required.

include/fmt/format.h
```diff
@@ -3395,6 +3395,43 @@ inline typename buffer_context<Char>::type::iterator format_to(
...
+struct it_category { typedef int type; };
```
vitaut: Do I understand correctly that `int` is arbitrary here and anything that is not `std::output_iterator_tag` would do? Why not use `void`? =)

include/fmt/format.h
```diff
@@ -3395,6 +3395,43 @@ inline typename buffer_context<Char>::type::iterator format_to(
...
+
+template <typename T>
+struct it_category<T*> { typedef std::output_iterator_tag type; };
+
+template <typename T>
+struct it_category<const T*> { typedef int type; };
+
+template <typename T>
+struct it_category<T, typename void_<typename T::iterator_category>::type> {
```
vitaut: So this is basically a SFINAE detection of presence of `T::iterator_category`? A comment would be nice as well.

include/fmt/format.h
```diff
@@ -3395,6 +3395,43 @@ inline typename buffer_context<Char>::type::iterator format_to(
...
+  typedef typename T::iterator_category type;
+};
+
+template <typename It>
+class is_output_iterator {
+  //check for mutability
+  template <typename U>
+  static decltype(*(internal::declval<U>())) test(std::input_iterator_tag);
```
vitaut: Why not just return `const char&`?

include/fmt/format.h
DanielaE: Well spotted, this is no longer necessary !

include/fmt/format.h
DanielaE: Well, `void` cannot be used as an argument type to a function call. I've settled on `std:.false_type`now.

include/fmt/format.h
DanielaE: Yeah, I've added a comment to make this clear.

include/fmt/format.h
vitaut: Makes sense. I was confused by the use of `input_iterator_tag` in the write context, but since iterator tags form a class hierarchy it makes sense now.

vitaut:
Merged, thanks!

Issue #922, opened 2018-10-25
Compile time format limited to char


XZiar:
I don't know why FMT_STRING is limited to use `char` only since fmt itself support both `char` and `wchar_t`. Is it designed to be that?

There several pieces of code hard coded the char_type:

https://github.com/fmtlib/fmt/blob/3e75ad9822980e41bc591938f26548f24eb88907/include/fmt/format.h#L1257

https://github.com/fmtlib/fmt/blob/3e75ad9822980e41bc591938f26548f24eb88907/include/fmt/format.h#L2273-L2274

https://github.com/fmtlib/fmt/blob/3e75ad9822980e41bc591938f26548f24eb88907/include/fmt/format.h#L3684-L3685

It is easy to solve this problem by adding typedef of char_type in the `struct S`.

vitaut:
> Is it designed to be that?

No, just noone generalized it yet. It is indeed easy to solve and PRs are welcome.

PR #924, opened 2018-10-27
Add support to compile time non-char format

XZiar:
PR for issue #922

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -472,10 +472,10 @@ struct compile_string {};
 template <typename S>
 struct is_compile_string : std::is_base_of<compile_string, S> {};
 
-template <typename S, typename Enable = std::enable_if<is_compile_string<S>::value>>
-inline auto to_string_view(const S &s) -> basic_string_view<typename S::Char> { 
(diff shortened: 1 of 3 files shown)


include/fmt/core.h
```diff
@@ -472,9 +472,11 @@ struct compile_string {};
 template <typename S>
 struct is_compile_string : std::is_base_of<compile_string, S> {};
 
-template <typename S>
-inline typename std::enable_if<is_compile_string<S>::value, string_view>::type
-  to_string_view(const S &s) { return {s.data(), s.size() - 1}; }
+template <typename S, typename Enable = typename std::enable_if<is_compile_string<S>::value>::type>
+inline auto to_string_view(const S &s) -> basic_string_view<typename S::char_type> {
```
vitaut: Is there any reason to move `enable_if` from return type to template parameter?

XZiar: I am not sure if it is possible to use `S::char_type` inside `enable_if` since `S` may not satisfy `is_compile_string`, so I choose my familiar way to implement this.

vitaut:
Merged, thanks!

PR #926, opened 2018-11-02
Update of Android build (with Gradle)

luncliff:
I'm glad to make another PR for this library. Appreciate for your maintenance works.   
More changes can be required when Gradle 5.0 is released. Please mention me if the build fails. 

### Summary

Updated Gradle and Travis YAML for Android build. [Build Log](https://travis-ci.org/fmtlib/fmt/builds/449942981) 

> Seems like the build log is crazy. :(

 - Applied latest Gradle (ver. 4.10.2) and plugin (ver 3.1.3)
 - Latest NDK and `c++17` for cxx flag

For more detailed explanation, please reference this commit: https://github.com/fmtlib/fmt/pull/926/commits/65134fe3390d732895aa98201e5b850a7686e087

### History
 - #649

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,7 +83,7 @@ matrix:
             - g++-4.4
           sources:
             - ubuntu-toolchain-r-test
-     # Android
+      # Android
     - language: android
       addons:
         apt:
@@ -115,7 +115,7 @@ matrix:
       before_script:
         - pushd ./support
       script:
-        - bash $GRADLE clean assemble
+        - bash ../$GRADLE clean assemble
       after_success:
         - popd;
         - tree ./libs
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,39 +83,42 @@ matrix:
             - g++-4.4
           sources:
             - ubuntu-toolchain-r-test
-      # Android
-#    - language: android
-#      android:
-#      addons:
-#        apt:
-#          update: true
-#      components:
-#        - tools
-#        - platform-tools
-#        - android-21
-#        - sys-img-armeabi-v7a-android-21
-#      env:
-#        - ANDROID=true
-#      before_install:
-#        - git submodule update --init --recursive
-#        - sudo apt-get install wget unzip tree
-#      install:
-#        # Accept SDK Licenses + Install NDK
-#        - yes | sdkmanager --update > /dev/null 2>&1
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -84,38 +84,41 @@ matrix:
...
-#        - unzip -q -d ./gradle gradle-4.3.1-bin.zip
-#        - export GRADLE=${TRAVIS_BUILD_DIR}/gradle/gradle-4.3.1/bin/gradle
-#      before_script:
-#        - bash $GRADLE --version
-#        - cd ./support
-#      script:
-#        - bash $GRADLE clean assemble
-#      after_success:
-#        - cd ${TRAVIS_BUILD_DIR}
-#        - tree ./libs
```
vitaut: Please change in place instead of commenting out the previous config.

support/AndroidManifest.xml
```diff
@@ -1 +1 @@
-<manifest package="fmt" />
\ No newline at end of file
+<manifest package="org.fmtlib" />
```
vitaut: This should probably be `net.fmtlib` because that's the fmtlib's domain (reversed).

support/build.gradle
```diff
@@ -4,12 +4,17 @@ buildscript {
         google()
         jcenter()
     }
-
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.0.1'
+        // Requires Gradle 4.4+
```
vitaut: Why does the comment say `4.4+` and the classpath `3.1.3`?

.travis.yml
luncliff: Sorry, I can't get it. I uncommented them on the right side. Or did you mean to maintain scripts(`install`, `script` etc) in place?

support/build.gradle
luncliff: We have 2 version there. One is the version of Gradle, and the other one is the version of Gradle plugin **for Android**.
I designated the plugin version to `3.1.3` and Gradle `4.4+` can support the plugin. 

I thought commenting those detail might be confused and placed a URL, but seems like I was wrong. I will comment them either.

.travis.yml
vitaut: My bad, I totally misread the diff =). Please ignore my comment.

Issue #932, opened 2018-11-17
Regression in Clang (r346892) with C++2a 


christinaa:
The problem is here:

```
enum char8_t : unsigned char {};
```

LLVM/Clang revision 346892 makes `char8_t` a fundamental type in C++2a mode. Would it be possible to introduce a workaround in fmtlib to account for that? I could do a PR myself but I'm not sure what the best way of approaching it is.

Thank you.

christinaa:
Quick and dirty fix:

```
#if !defined(__cpp_char8_t)
// A UTF-8 code unit type.
enum char8_t : unsigned char {};
#endif
```

PR #933, opened 2018-11-17
Fix incompatibilities with c++2a mode in clang

christinaa:
This closes https://github.com/fmtlib/fmt/issues/932.


christinaa:
Minor correction, it was a fundamental type as-is but was gated behind an option. Since Clang 346892 it is now enabled by default. 

This makes use of a rather simple feature test:
```
Builder.defineMacro("__cpp_char8_t", "201811L");
```

vitaut:
Good catch, thanks!

PR #934, opened 2018-11-18
add make_printf_args and make_wprintf_args functions

tnovotny:
Add utitlity functions to allow the easy creation of `fmt::printf_args` and `fmt::wprintf_args`.

### DIFF
--- a/include/fmt/printf.h
+++ b/include/fmt/printf.h
@@ -581,6 +581,33 @@ struct printf_context {
 typedef basic_format_args<printf_context<internal::buffer>::type> printf_args;
 typedef basic_format_args<printf_context<internal::wbuffer>::type> wprintf_args;
 
+typedef printf_context<internal::buffer>::type printf_char_context;
+typedef printf_context<internal::wbuffer>::type printf_wchar_context;
+
+
+/**
+  \rst
+  Constructs an `~fmt::format_arg_store` object that contains references to
+  arguments and can be implicitly converted to `~fmt::printf_args`. `Context`
+  can be omitted in which case it defaults to `~fmt::printf_char_context`.
+  \endrst
+ */
+template<typename Context=printf_char_context, typename... Args>
+inline format_arg_store<Context, Args...>
+  make_printf_args(const Args &... args) { return {args...}; }
+
+/**
+  \rst
+  Constructs an `~fmt::format_arg_store` object that contains references to
+  arguments and can be implicitly converted to `~fmt::wprintf_args`. `Context`
+  can be omitted in which case it defaults to `~fmt::printf_wchar_context`.
+  \endrst
+ */
+template<typename Context = printf_wchar_context, typename... Args>
+inline format_arg_store<Context, Args...>
+  make_wprintf_args(const Args &... args) { return {args...}; }
+
+
 template <typename S, typename Char = FMT_CHAR(S)>
 inline std::basic_string<Char>
 vsprintf(const S &format,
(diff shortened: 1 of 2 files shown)


tnovotny:
Hmm, not sure why the g++ 4.8/4.4 are failing. Any ideas?

include/fmt/printf.h
```diff
@@ -578,8 +578,32 @@ struct printf_context {
     std::back_insert_iterator<Buffer>, typename Buffer::value_type> type;
 };
 
-typedef basic_format_args<printf_context<internal::buffer>::type> printf_args;
-typedef basic_format_args<printf_context<internal::wbuffer>::type> wprintf_args;
+typedef printf_context<internal::buffer>::type printf_char_context;
+typedef printf_context<internal::wbuffer>::type printf_wchar_context;
```
vitaut: `printf_char_context` and `printf_wchar_context` should be called `printf_context` and `wprintf_context` for consistency with `format_context` and `wformat_context`. The old `printf_context` template should be renamed to `basic_printf_context`.

test/printf-test.cc
```diff
@@ -518,3 +518,19 @@ void check_format_string_regression(fmt::string_view s, const Args&... args) {
 TEST(PrintfTest, CheckFormatStringRegression) {
   check_format_string_regression("%c%s", 'x', "");
 }
+
+
```
vitaut: nit: no need for an extra empty line

tnovotny:
I don't know how to make the problematic tests pass for the gcc versions and I want a tests that mirrors the intended usage, so I #defined it out for gcc 4, for lack of a better idea.

tnovotny:
In the additional test, the type returned by fmt seems to be different for msvc 2013. It returns the string literals as const array where in all of the other cases it isn't const. That kinda seems like a bug in fmtlib.

test/printf-test.cc
```diff
@@ -518,3 +517,43 @@ void check_format_string_regression(fmt::string_view s, const Args&... args) {
 TEST(PrintfTest, CheckFormatStringRegression) {
   check_format_string_regression("%c%s", 'x', "");
 }
+
+TEST( PrintfTest, VSPrintfMakeArgsExample ) {
```
vitaut: nit: spaces after `(` and before `)` are not needed

vitaut:
There are a few compile errors on MSVC2013: https://ci.appveyor.com/project/vitaut/fmt/builds/20446501/job/tu84x312yjrwp6h7

include/fmt/printf.h
tnovotny: I ended up using `basic_printf_context_t` because it caused a name conflict with the inner typedef. That naming style is also with the format context

    template <typename OutputIt, typename Char = char>
    //using format_context_t = basic_format_context<OutputIt, Char>;
    struct format_context_t { typedef basic_format_context<OutputIt, Char> type; };

I hope that is ok with you.

test/printf-test.cc
tnovotny: removed

test/printf-test.cc
tnovotny: I removed all the superfluous spaces clang formatting introduced. I think I got them all.

vitaut:
Merged, thanks!

vitaut:
@tnovotny, could you please review the updated [CONTIBUTING](https://github.com/fmtlib/fmt/blob/ebec00138bd10e949ff34955bcb096387cd32acc/CONTRIBUTING.md) document, particularly the part about licensing, and let me know if you agree with it being applied to your contributions to {fmt}? The library is likely to be relicensed (#1073) so I'm collecting approval from all earlier contributors. Thanks!

Issue #931, opened 2018-11-15
warnings in msvc2015 around localtime and gmtime


andry81:
> fmt/time.h(51): warning C4996: 'localtime': This function or variable may be unsafe. Consider using localtime_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.
> fmt/time.h(86): warning C4996: 'gmtime': This function or variable may be unsafe. Consider using gmtime_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.

These warnings goes spammed almost everythere in sources.

vitaut:
Fixed in https://github.com/fmtlib/fmt/commit/fecb2d6f0d6c97986c72c7b28f6ff52b3af6348c.

PR #943, opened 2018-11-25
Eliminate msvc compiler warnings (#931)

DanielaE:
The sheer presence of 'std::gmtime' or 'std::localtime' gives rise to C4996 warnings in Microsofts compilers. Alas, the 'fallback(internal::null<>)'functions containing these library calls are never ADL-picked in the respective 'handle(internal::null<>)' selectors. Therefore hiding the fallbacks from msvc is perfectly fine.


vitaut:
Merged, thanks!

PR #946, opened 2018-11-25
Fix conversion warnings with clang7

0x8000-0000:
Add 'int' overloads for certain methods to limit the spread of static casts through the code.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -562,15 +562,32 @@ struct prettify_handler {
     size += n;
   }
 
+  template <typename F>
+  void insert(int pos, int n, F f) {
+    FMT_ASSERT(pos >= 0, "negative character offset");
+    FMT_ASSERT(n >= 0, "negative count of characters");
+    insert(static_cast<size_t>(pos), static_cast<size_t>(n), f);
+  }
+
   void insert(size_t pos, char c) {
     std::memmove(data + pos + 1, data + pos, size - pos);
     data[pos] = c;
     ++size;
   }
 
+  void insert(int pos, char c) {
+    FMT_ASSERT(pos >= 0, "negative character offset");
+    insert(static_cast<size_t>(pos), c);
+  }
+
   void append(size_t n, char c) {
     std::uninitialized_fill_n(data + size, n, c);
     size += n;
+  }
+
+  void append(int n, char c) {
+    FMT_ASSERT(n >= 0, "negative count of characters");
+    append(static_cast<size_t>(n), c);
   }
 
   void append(char c) { data[size++] = c; }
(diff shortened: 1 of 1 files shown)


0x8000-0000:
Casts are a code smell, and I think {fmt} has way too many of those. We should not need to keep overriding the compiler. Instead of adding 5 or 6 at the call sites, at least this centralizes the casting.

The C++ committee considers the "unsigned / size_t" interface to the standard containers a mistake, and suggests using ints (with suitable asserts that they are positive where they need to be) in order to simplify all the code that deals with signed arithmetic. It's clearly too late to fix the standard C++ library, but not too late to fix {fmt}.

0x8000-0000:
We should also remove the -Wno-xyz command line option from the CMakeLists and make sure the code is clean. There are code bases with strict warning checking, and they could not take advantage of {fmt}.

vitaut:
There are perfectly valid uses of casts, particularly when mixed signed and unsigned operations are necessarily involved.

In this particular case, fixing the implementation of the function is better than lifting the warning suppression workarounds to its interface.

vitaut:
BTW could you post the warnings that you are trying to fix?

0x8000-0000:
> BTW could you post the warnings that you are trying to fix?

/home/florin/work/geometry/external/lib/fmt/include/fmt/format-inl.h:953:13: note: in instantiation of function template specialization 'fmt::v5::internal::vformat_to<wchar_t>' requested here
  internal::vformat_to(buffer, format_str, args);
            ^
/home/florin/work/geometry/external/lib/fmt/include/fmt/format-inl.h:639:22: error: implicit conversion changes signedness: 'int' to 'size_t' (aka 'unsigned long') [-Werror,-Wsign-conversion]
      handler.append(num_zeros, '0');
              ~~~~~~ ^~~~~~~~~
/home/florin/work/geometry/external/lib/fmt/include/fmt/format-inl.h:643:20: error: implicit conversion changes signedness: 'int' to 'size_t' (aka 'unsigned long') [-Werror,-Wsign-conversion]
    handler.insert(full_exp, '.');
            ~~~~~~ ^~~~~~~~
/home/florin/work/geometry/external/lib/fmt/include/fmt/format-inl.h:654:25: error: implicit conversion changes signedness: 'int' to 'size_t' (aka 'unsigned long') [-Werror,-Wsign-conversion]
    handler.insert(0, 2 - full_exp, fill{to_unsigned(-full_exp)});
            ~~~~~~    ~~^~~~~~~~~~

vitaut:
Closing in favor of https://github.com/fmtlib/fmt/commit/b8b06e3e10345a57431cb986e6e99f1adaad57e0 which changes a few types to signed without introducing extra complexity into the handler (type sizes should probably be unified more but that can be addressed separately).

Issue #618, opened 2017-12-03
Proposal: prepared statements 


gabime:
While #613 would be great, but will probably only supported for c++17.

Another approach would be to have kind of prepared templates (or prepared statements - a bit like in db, but fmt need not to store or manage the handle anywhere - the handle itself would contain all the needed info).

The user would pass a format string he expects to use a lot and get a handle to a "compiled" version of the format string:

```c++
// fmt returns a handle object that can be executed like this:
auto handle = fmt::prepare("My name is {}. I am {} years old."); 


// and then the user can make heavy use of it
auto result = handle->format(arg1, arg2);
// or maybe
auto result = fmt::format(handle, arg1, arg2);
```

The handle would contain an efficient representation of the format -  list of actual formatters to use, their flags, and their offsets in the format string.
I am not sure if the result would be much faster for simple format strings,  but complex format strings, it might give big performance  boost.

vitaut:
Thanks for the suggestion. It sounds very similar to what Bengt Gustafsson's proposed in https://groups.google.com/a/isocpp.org/forum/#!msg/std-proposals/4wOU-1_3D0A/xiNQSmO1CAAJ:

> I think it may make sense to allow pre-creating a formatting object for an entire format string, so that the parsing of that string occurs only once.

and, in fact, I've already done most of the work to make this possible, namely separated parsing and formatting as part of compile-time format string parsing support. It still needs to be packaged in an API similar to the one you proposed with the main difference that `prepare` function or its equivalent should take argument types in addition to format string:

```c++
auto formatter = fmt::prepare<std::string, int>("My name is {}. I am {} years old.");
```

vitaut:
BTW #613 can work with C++14.

gabime:
Do you think adding this feature, or will you do #613 first ?

vitaut:
Not sure, #613 will give much better runtime performance but, as you noticed, it will require a modern compiler. This one is a small optimization, mostly for the case when you need to process the same format string multiple times. My main focus right now is addressing standards proposal feedback (#518), but I'll get back to this idea afterwards.

stryku:
Hello, does this proposal need to wait for some other dependency than #581, or I can fool around a little and try to implement this? Or you want to postpone it?

vitaut:
No this doesn't have to wait on anything.

>  I can fool around a little and try to implement this?

Please do. Contributions are always welcome!

stryku:
[...]

Issue #613, opened 2017-11-24
Generate optimal formatters at compile time


vitaut:
As pointed out by Louis Dionne in #546:

> Not only could you perform the safety checks at compile-time, but you could also generate optimal code by parsing everything at compile-time.

This can make the format API as efficient as the write API and eliminate the need in the latter.

fiblit:
Perhaps I'm misunderstanding this issue, but is this even possible right now?

From my brief attempts at playing around, it would seem you need constexpr function parameters to do this. You would somehow need to split a string in a function at compile-time, yet I know no valid way to achieve this:

```
constexpr void split(constexpr const char* s) {
    // I'm not super familiar with fmt's internals/API
    // count_replacements would be constexpr and count the number of {}'s, or similar.
    std::array<const char*, count_replacements(s) + 1> text;
    std::array<fmt::?, count_replacements(s)> formatting;
    //...
}
```

The problem is this *also* needs to be able to run at run-time, which makes no sense for the template of std::array (or C arrays).

Perhaps this is possible with odd manipulation of template character packs? I haven't figured out a proof-of-concept for that, though.

vitaut:
> is this even possible right now?

It should be possible.

> it would seem you need constexpr function parameters to do this

This is correct and {fmt} emulates them with some `constexpr` lambda and macro magic: https://github.com/fmtlib/fmt/blob/846c644e848ca1a11c679b168cfc690bc3c89c61/include/fmt/format.h#L3559

You can overload on `constexpr`ness of a format string with the help of the `is_compile_string` trait.

vitaut:
A prototype of generating optimal formatters at compile time for a subset of std::format/{fmt}'s syntax by Hana: https://github.com/hanickadot/compile-time-regular-expressions/commit/13afb2c6f1a3a3ec5467a54bba0a84f2c0ec85b4

vitaut:
In C++20 we could do this:

```c++
#include <string>
#include <array>

template <typename Char, size_t N>
class basic_fixed_string {
 private:
  Char data[N] = {};

 public:
  constexpr basic_fixed_string(const Char (&s)[N]) {
    for (size_t i = 0; i < N; ++i)
      data[i] = s[i];
  }
};

template <typename Char, size_t N>
basic_fixed_string(const Char (&s)[N]) ->
  basic_fixed_string<Char, N>;

template <basic_fixed_string, typename... Args>
std::string format(const Args&... args) {
  return {};
}

[...]

PR #949, opened 2018-11-27
[WIP] Implementation of fmt::prepare. Implements #618 (and probably #613)

stryku:
# Foreword
Hello (: First of, this thing is huge. Please, take your time while reviewing it. I'm open for a discussion and aware that this overview doesn't explain everything, so don't hesitate to ask questions, I'll try to explain why I did this in such way, that in other way etc.

# Abstract
This pull request's purpose is to implement a `fmt::prepare<>()` function and all functionality around it.

# Problems
1. Format should be prepared once, at creation. Later on, it should just format user arguments.
2. If format string can be handled at compile time, format should be prepared at compile time.
3. User should be able to declare prepared format as a e.g. class member.
4. User should be able to do with prepared format whatever he wants (copy, move etc.).
5. Because of above, prepared format can't work with raw string views, thus unlike current fmt features, it is statefull.
6. Prepared format is statefull, because of that it needs to be allocator-aware.

# TL;DR solutions
1. Indeed, format is prepared at creation - in ctor.
2. If `FMT_USE_CONSTEXPR` is enabled and `fmt::prepare` detects that passed format `is_compile_string`, then format will be prepared at compile time.
3. User can declare format as a member, using `fmt::prepared_format<>` typedef/alias. 
4. Prepared format is copyable and movable. 
5. Prepared format keeps a copy of passed format string as a `std::basic_string<>`.
6. Prepared format has two members that the user (who wants to provide his own allocators) needs to know about: a copy of the format string and something that I call prepared parts. Prepared parts are parts of the format, that format was able to prepare at a creation time (e.g. in `"My name is {0}."` format, there are three parts: text `"My name is "`, argument id `0`, text `"."`). User is able to provide his own types for the format and prepared parts container via a `fmt::basic_prepared_format<>` (I'll talk about it later).

# API
### format_part<> class
```cpp
template <typename Char>
class format_part;
```
A class that represents a format prepared part. It's not internal, to give user possibility to pass his own parts container type (see below).

### parts_container<> typedef
```cpp
template <typename Char, typename Container = std::vector<format_part<Char>>
class parts_container;
```
A typedef of a class that is used as a parts container by prepared format. User can provide his own underlying container type (in, of course, the second parameter).

### basic_prepared_format<> typedef
```cpp
template <typename Format, typename PartsContainer, typename... Args>
using basic_prepared_format = /* ... */;
```
A typedef of a class that represents a fully customizable prepared format. User can provide his own Format and PartsContainer types.

### prepared_format<> typedef
```cpp
template <typename... Args>
using prepared_format = basic_prepared_format<std::string, parts_container<char>, Args...>;
```
A simple alias for basic_prepared_format with default Format and PartsContainer types.

### wprepared_format<> typedef
```cpp
template <typename... Args>
using wprepared_format = basic_prepared_format<std::wstring, parts_container<wchar_t>, Args...>;
```
Same as above, but for wide strings.

### prepare<>()
```cpp
template <typename... Args, typename Format>
prepare(const Format& format_str)
```
A function that returns a prepared format object. `Args` are the argument types that prepared format accepts in format functions.

Prepared formats created by all of the above ways, provide the same functionality (same as plain fmt::format*() functions):
-format()
-format_to()
-format_to_n()
-formatted_size()

# Tests
There was a need to a little refactor them. I have had to change tests strictly related to formatting, to be able to test multiple ways of formatting. Those tests are moved to a new file: `common-format-test.cc`, and they test the old `fmt::format()` and two versions of `fmt::prepare()`: one with runtime format string passed and one with compiletime format string.

In `common-format-test`  I've used gtest's TYPED_TESTs. There are three wrappers: 
-FormatFunctionWrapper: simply call `fmt::format*()`.
-RuntimePreparedFormatWrapper: explicitly convert passed format string to a runtime format (which is std::basic_string, with proper character type). Then, create a prepared format using the shiny new fmt::prepare(). The last step is to format string using prepared format's `format()` method.
-CompiletimePreparedFormatWrapper: this one is enabled only if FMT_USE_CONSTEXPR is enabled. It prepares format (using compiletime format string) and then, formats stuff using prepared format's `format()` method.

As you can see, in almost every test, format string is wrapped by the ~`TEST_FMT_STRING`~`FMT`. That's because I couldn't think of a better, generic way to pass compiletime format string to wrappers. I wanted to do this, to tests `fmt::prepare()` with compiletime format string.

[DONE]~(Now I realized, that I probably should create one more wrapper: `RuntimeFormatFunctionWrapper` (and rename `FormatFunctionWrapper` to `CompiletimeFormatFunctionWrapper`) that would explicitly convert a compiletime string to a runtime one and test `fmt::format()` against it. What do you think?)~

# Documentation
Didn't write it, simply because a lot can be changed during the review. I'll provide it as a last step of this pull request.

# Hacks (or not)
## `basic_format_context::prepared_specs*` member and `dynamic_formatter`.
`basic_format_context` has a new `prepared_specs*` member. It is set, only in `internal::prepared_format::vformat_to()` method. Then, it's used by `dynamic_formatter` to check whether it needs to parse specification, or it is already parsed and ready to use.
Actually, this workaround is needed for only one of the tests, the one with variant:
```cpp
TYPED_TEST(RuntimeFormattersTest, DynamicFormatter) {
    auto num = variant(42);
    auto str = variant("foo");
    /* ... */
    EXPECT_EQ(" 42 foo ",
              TypeParam::format(TEST_FMT_STRING("{:{}} {:{}}"), num, 3, str, 4));
}
```
It fails with old approach because:
-when using `fmt::format`, automatic argument id is tracked by a parse context. Later on, dynamic formatter gets that id from the parsed context.
-when using `fmt::prepare`, all argument ids are tracked by the prepared format. Specification is parsed and prepared, long before actual formatting. Parse context doesn't track the id anymore, so when dynamic formatter was asking for it, parse context returned a corrupted id. Now, this prepared specification is handled by prepared format and set in `basic_format_context::prepared_specs*`. If dynamic_formatter detects that it can use an already prepared specs, it doesn't bother the parse context.

An alternative would be to create a new context e.g. `prepared_format_context`, that would have this member. Then in `dynamic_formatter`, we would overload `format` method to handle the new context. I decided to not do that (at least in scope of this pull request) because it would require a refactor, to pass format context template parameter in a couple (a lot?) of places. This pull request is already big, even without this change.

## Unrestricted unions
These are the unions with e.g. non-POD types. I needed them in `format_part` and changed `arg_ref`.
The simplest solution when they are not supported was to simply use a structure instead. I agree that it is not the best solution but unrestricted unions are supported since GCC 4.6, Clang 3.3 and Visual Studio 15 (MSVS >= 1900), so we support a couple of really old compilers that doesn't support that.
An alternative would be to use PODs in all unions and implement separate class-wrappers, that would implement needed methods handling those PODs. This would introduce a fine amount of new code (e.g. I'd need to introduce a new POD type for parsed_format_specs + a wrapper with methods for it, and more..). That's why I went this way.

# Other
## `basic_prepared_format<>`
This typedef gives user a possibility to provide his own types for held stuff. Thanks to that, we're fully allocator-aware. User can provide:
-A format string type.
-A parts container type.

Parts container needs to fulfill the PartsContainer concept, which is:
```cpp
class {
public:
    typedef /*held format part type */ format_part_type;
    void add(format_part_type part); // Adds part to the end of already held parts.
    void substitute_last(format_part_type part); // Substitutes part at the end of held parts with the provided one. fmt lib cares of not calling it on a empty container.
    format_part_type last(); // Returns part from end of held parts.
    /*format_part_type iterator*/ begin(); // Returns begin of the held parts.
    /*format_part_type iterator*/ end(); // Returns end of the held parts.
};
```

Concept is checked at a compile-time by the `parts_container_concept_check` struct.

## parts_container_concept_check (prepare.h)
There is a lot of boilerplate. It could be done with a macro, but I decided that macros for such stuff are yuck.
[DONE] ~(I just realized that I probably should add checks for default, move and copy ctors there. What do you think?)~

## prepare.h
I decided to create a new file for features related to `fmt::prepare()`. Thanks to that, user can avoid compilation of heavy templates when he doesn't need the `fmt::prepare()`.


vitaut:
Wow, that's an impressive PR, thank you for working on it!

I'll take a look in details but first a few quick questions regarding the design:

> Prepared format is statefull, because of that it needs to be allocator-aware.

Could you elaborate how statefulness relates to allocator-awareness? Is it because you use a container (string) to store some of the state?

> Prepared format keeps a copy of passed format string as a std::basic_string<>.

How does it work with `constexpr`?

> Now it returns formatter_parse_result, which is basically a bool (whether it was parsed successfully) and the iterator (to where the formatter has stopped parsing).

Why and how does it work with existing parsers that report errors via exceptions?

stryku:
>Could you elaborate how statefulness relates to allocator-awareness? Is it because you use a container (string) to store some of the state?

Exactly. Prepared format has two members that can allocate memory - a format and a parts container. We need to provide a way to specify allocators that are used to allocate memory for these members.


>How does it work with constexpr?

Sorry, I forgot to mention the constexpr case. When a format's type is the fmt's compile time string, the stored member is an instance of the struct that wraps the string literal (the struct, returned from the FMT_STRING's lambda), otherwise it's a `std::basic_string<>`.
I just realized, that there is probably an issue in my implementation. For runtime format strings, I always try to convert them to a `std::basic_string<>`. When user would pass his own format string type, this would mostly like fail during a compilation. I will take a look and write it properly.

>Why and how does it work with existing parsers that report errors via exceptions?

Well, while writing an answer to this, I realized that I can take advantage of a parse_context to check whether a specification was successfully parsed by a custom formatter. Thanks to that, the change in the public API won't be needed. I'll do the changes and push a commit probably tomorrow.

stryku:
I'll squash commits just before the merge. It'd be more convenient for me to keep them separated while working on this pr.

include/fmt/format-inl.h
```diff
@@ -353,21 +353,21 @@ class fp {
...
+                                                    << double_significand_size;
 
  public:
   significand_type f;
   int e;
 
-  static FMT_CONSTEXPR_DECL const int significand_size =
-    sizeof(significand_type) * char_size;
+  static FMT_CONSTEXPR_DECL int significand_size =
+      sizeof(significand_type) * char_size;
```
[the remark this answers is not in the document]
stryku: By 'constexpr-related' you mean `FMT_CONSTEXPR_DECL`-related changes or all constexpr-related code from this pr (constexpr fmt::prepare() etc.)?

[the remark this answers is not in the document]
vitaut: `FMT_CONSTEXPR_DECL`-related and any similar changes like making `check_arg_id` `constexpr` that aren't related to `fmt::prepare`.

include/fmt/core.h
```diff
@@ -619,8 +624,9 @@ class value {
     // `printf_formatter<T>` for `printf`.
     typename Context::template formatter_type<T>::type f;
     auto &&parse_ctx = ctx.parse_context();
-    parse_ctx.advance_to(f.parse(parse_ctx));
-    ctx.advance_to(f.format(*static_cast<const T*>(arg), ctx));
+    const auto stopped_at = f.parse(parse_ctx);
+    parse_ctx.advance_to(stopped_at);
+    ctx.advance_to(f.format(*static_cast<const T *>(arg), ctx));
```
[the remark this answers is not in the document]
stryku: Forgot to revert these chages. Will do.

include/fmt/format-inl.h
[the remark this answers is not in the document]
stryku: Sure, I'll revert `FMT_CONSTEXPR_DECL` changes, but making `check_arg_id` `constexpr` is crucial for compile-time preparing.

include/fmt/core.h
```diff
@@ -67,7 +67,7 @@
 # define FMT_CONSTEXPR_DECL constexpr
 #else
 # define FMT_CONSTEXPR inline
-# define FMT_CONSTEXPR_DECL
+#define FMT_CONSTEXPR_DECL const
```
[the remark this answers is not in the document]
stryku: I'm sorry for whitespaces changes. I ran clang-format on my changes, with a google style. Apparently cf does not indent preprocessor directives. Btw, how do you format code while working with {fmt}?

About the change. All uses of the `FMT_CONSTEXPR_DECL` in the current code are followed by a `const`. If code compiles with a relaxed constexpr enabled, that `const` is redundant, if it compiles without the relaxed constexpr, const is useful. That's why I put it directly in the define.
On the other hand I can see that removing `const` from here and using is directly in the code is more explicit. What do you think?

include/fmt/format.h
```diff
@@ -1609,7 +1617,9 @@ class custom_formatter: public function<bool> {
   }
 
   template <typename T>
-  bool operator()(T) const { return false; }
+  bool operator()(T) const {
+    return false;
+  }
```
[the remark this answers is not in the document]
stryku: Sorry, clang-format. I'll try to revert such changes.

stryku:
>A user format string should be convertible to `basic_string_view` via `to_string_view` so you can use that to construct a `basic_string`.

Didn't know that user format string needs to be a variation of a `std::basic_string`. Thanks for info (:
Anyway, I still think that we need to handle user's type of a format string, to handle cases when the string is not a simple `std::basic_string<char/wchar_t>` but e.g. a string with user provided allocator.
Actually I've already made changes to handle that. If passed format string's type is `const Char*` or `const Char[]`, it is converted to a `std::basic_string<Char>`. In any other case, format is just `std::move`d all the way to the `prepare_format`'s member. That handles above cases.

>Great! So the "Current API breaking changes" section is no longer relevant?

Sure, done.

>That seems OK, but I it would be nice to have a shorter name to make tests more readable. Maybe just fmt(...)? 

>Yes, please. 

Will do.

include/fmt/format.h
```diff
@@ -1822,83 +1844,208 @@ class specs_handler: public specs_setter<typename Context::char_type> {
...
 };
 
 // Format specifiers with width and precision resolved at formatting rather
 // than parsing time to allow re-using the same parsed specifiers with
 // differents sets of arguments (precompilation of format strings).
 template <typename Char>
 struct dynamic_format_specs : basic_format_specs<Char> {
-  arg_ref<Char> width_ref;
-  arg_ref<Char> precision_ref;
+  typedef arg_ref<Char, string_value<Char>> arg_reference;
```
vitaut: I suggest renaming to `arg_ref_type` for consistency with similar type aliases elsewhere.

include/fmt/format.h
```diff
@@ -1822,83 +1844,208 @@ class specs_handler: public specs_setter<typename Context::char_type> {
...
+class dynamic_arg_ref_creator {
+  typedef typename ParseContext::char_type char_type;
+
+ public:
+  typedef arg_ref<char_type, string_value<char_type>> arg_ref_type;
+
+  FMT_CONSTEXPR dynamic_arg_ref_creator(ParseContext &ctx) : context_(ctx) {}
+
+  FMT_CONSTEXPR arg_ref_type create(unsigned arg_id) {
     context_.check_arg_id(arg_id);
```
vitaut: I think this belongs to `dynamic_specs_handler`. Same with the calls to `check_arg_id` and `next_arg_id` below. Then this class won't need to store a reference to parse context.

include/fmt/format.h
```diff
@@ -1822,83 +1844,208 @@ class specs_handler: public specs_setter<typename Context::char_type> {
...
+ public:
+  typedef arg_ref<char_type, string_view_metadata> arg_ref_type;
+
+  FMT_CONSTEXPR
+  prepared_arg_ref_creator(ParseContext &ctx,
+                           basic_string_view<char_type> format_str)
+      : context_(ctx), format_(format_str) {}
+
+  FMT_CONSTEXPR arg_ref_type create(basic_string_view<char_type> id) {
+    context_.check_arg_id(id);
```
vitaut: Same as above.

include/fmt/format.h
stryku: You're right, I'll move `check_arg_id` back to the `dynamic_specs_handler`, but the context reference is still needed to get the id when `create(auto_id)` is called.

include/fmt/core.h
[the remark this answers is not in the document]
stryku: >Manually like a savage =).

Wait, what? I've just created an issue for this (:

>I agree...

Cool, I'll introduce this changes in a separate pr, anyway.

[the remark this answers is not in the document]
stryku: `#define FMT_CONSTEXPR_DECL const` change has been reverted in the latest commit.

include/fmt/format.h
```diff
@@ -1704,68 +1714,80 @@ class specs_setter {
   basic_format_specs<Char> &specs_;
 };
 
+template <typename ErrorHandler>
+class speck_checker {
```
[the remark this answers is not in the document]
stryku: Done.

include/fmt/format.h
```diff
@@ -1822,83 +1844,208 @@ class specs_handler: public specs_setter<typename Context::char_type> {
 };
 
 // An argument reference.
-template <typename Char>
+template <typename Char, typename Name>
```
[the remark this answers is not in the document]
stryku: Done.

include/fmt/format.h
```diff
@@ -3306,7 +3525,7 @@ struct formatter<arg_join<It, Char>, Char>:
     formatter<typename std::iterator_traits<It>::value_type, Char> {
   template <typename FormatContext>
   auto format(const arg_join<It, Char> &value, FormatContext &ctx)
-      -> decltype(ctx.out()) {
+      -> decltype(ctx.begin()) {
```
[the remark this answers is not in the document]
stryku: I hope I've change all deprecated `begin` calls.

include/fmt/format.h
stryku: Done.

include/fmt/format.h
stryku: Moved checking argument id back to the `dynamic_specs_handler`.

include/fmt/core.h
```diff
@@ -620,7 +625,7 @@ class value {
     typename Context::template formatter_type<T>::type f;
     auto &&parse_ctx = ctx.parse_context();
     parse_ctx.advance_to(f.parse(parse_ctx));
-    ctx.advance_to(f.format(*static_cast<const T*>(arg), ctx));
+    ctx.advance_to(f.format(*static_cast<const T *>(arg), ctx));
```
vitaut: Unnecessary whitespace change.

include/fmt/core.h
```diff
@@ -902,20 +909,37 @@ class basic_parse_context : private ErrorHandler {
   FMT_CONSTEXPR unsigned next_arg_id();
 
   FMT_CONSTEXPR bool check_arg_id(unsigned) {
-    if (next_arg_id_ > 0) {
+    if (is_indexing_arguments_automatically()) {
```
vitaut: Changes to indexing seem unrelated, please submit as a separate PR to make this one manageable.

include/fmt/format-inl.h
```diff
@@ -360,7 +360,7 @@ class fp {
   static FMT_CONSTEXPR_DECL const int double_significand_size =
     std::numeric_limits<double>::digits - 1;
   static FMT_CONSTEXPR_DECL const uint64_t implicit_bit =
-    1ull << double_significand_size;
+      1ull << double_significand_size;
```
vitaut: Unnecessary.

include/fmt/format.h
```diff
@@ -81,6 +81,16 @@
 # define FMT_SECURE_SCL 0
 #endif
 
+// Check whether we can use unrestricted unions (e.g. unions with non-PODs
+// members). If no, use struct.
+#ifndef FMT_UNRESTRICTED_UNION
+#if (FMT_MSC_VER >= 1900 || FMT_GCC_VERSION >= 406 || FMT_CLANG_VERSION >= 303)
```
vitaut: Please indent nested code for readability.

include/fmt/core.h
```diff
@@ -902,20 +909,37 @@ class basic_parse_context : private ErrorHandler {
...
       on_error("cannot switch from automatic to manual argument indexing");
       return false;
     }
     next_arg_id_ = -1;
     return true;
   }
-  void check_arg_id(basic_string_view<Char>) {}
+
+  FMT_CONSTEXPR bool check_arg_id(internal::auto_id) {
```
vitaut: I don't think it is a good idea to add this implementation detail to the public API. `next_arg_id` is there for automatic indexing.

include/fmt/format.h
```diff
@@ -1822,85 +1843,201 @@ class specs_handler: public specs_setter<typename Context::char_type> {
...
 };
 
 // Format spec handler that saves references to arguments representing dynamic
 // width and precision to be resolved at formatting time.
-template <typename ParseContext>
-class dynamic_specs_handler :
-    public specs_setter<typename ParseContext::char_type> {
+template <typename Specs, typename ParseContext, typename ArgRefCreator>
+class dynamic_specs_handler
+    : public specs_setter<typename ParseContext::char_type> {
```
vitaut: I suggest privately inheriting from `ArgRefCreator` to make use of empty base calls optimization (this will also require making `dynamic_arg_ref_creator` stateless - see comment below).

include/fmt/format.h
```diff
@@ -1822,85 +1843,201 @@ class specs_handler: public specs_setter<typename Context::char_type> {
...
+
+  FMT_CONSTEXPR arg_ref_type create(unsigned id) { return arg_ref_type(id); }
+
+  FMT_CONSTEXPR arg_ref_type create(basic_string_view<char_type> id) {
+    return arg_ref_type(string_value<char_type>{id.data(), id.size()});
   }
 
-  FMT_CONSTEXPR arg_ref_type make_arg_ref(auto_id) {
+  FMT_CONSTEXPR arg_ref_type create(auto_id) {
     return arg_ref_type(context_.next_arg_id());
```
vitaut: I think `context_.next_arg_id()` should be called in `dynamic_specs_handler` and the ID passed to `create`. Then `dynamic_arg_ref_creator` won't depend on parse context.

include/fmt/format.h
```diff
@@ -1822,85 +1843,201 @@ class specs_handler: public specs_setter<typename Context::char_type> {
...
   }
 
-  dynamic_format_specs<char_type> &specs_;
+ private:
   ParseContext &context_;
 };
 
+struct string_view_metadata {
```
vitaut: Why do we need this class? Can it be replaced with `string_view`?

include/fmt/format.h
```diff
@@ -1822,85 +1843,201 @@ class specs_handler: public specs_setter<typename Context::char_type> {
...
+
+  FMT_CONSTEXPR arg_ref_type create(basic_string_view<char_type> id) {
+    const auto id_metadata = string_view_metadata(format_, id);
+    return arg_ref_type(id_metadata);
+  }
+
+  FMT_CONSTEXPR arg_ref_type create(unsigned id) { return arg_ref_type(id); }
+
+  FMT_CONSTEXPR arg_ref_type create(auto_id) {
+    return arg_ref_type(context_.next_arg_id());
```
vitaut: As above, `next_arg_id` should be moved up the stack and parse context dependency removed.

include/fmt/format.h
```diff
@@ -2137,8 +2274,12 @@ FMT_CONSTEXPR void parse_format_string(
...
       for (;;) {
         const Char *p = FMT_NULL;
-        if (!find<IS_CONSTEXPR>(begin, end, '}', p))
-          return handler_.on_text(begin, end);
+        if (!find<IS_CONSTEXPR>(begin, end, '}', p)) {
+          if (begin != end) {
+            handler_.on_text(begin, end);
+          }
+          return;
+        }
```
vitaut: This looks unrelated.

include/fmt/format.h
```diff
@@ -2220,8 +2361,8 @@ class format_string_checker {
   FMT_CONSTEXPR const Char *on_format_specs(iterator it) {
     auto p = pointer_from(it);
     context_.advance_to(p);
-    return to_unsigned(arg_id_) < NUM_ARGS ?
-          parse_funcs_[arg_id_](context_) : p;
+    return to_unsigned(arg_id_) < NUM_ARGS ? parse_funcs_[arg_id_](context_)
+                                           : p;
```
vitaut: Please revert all formatting-only changes to simplify review.

include/fmt/format.h
stryku: Parts preparation uses the `parse_format_string()` function to count how many parts the format has. Without the `if (begin != end)`, the `handler_.on_text()` is called, even if `begin == end` and that causes creation of an additional, empty format part. With the change, prepared format has one part less.

include/fmt/format.h
stryku: Prepared format needs to be copyable and movable. User can store it as a class member, in containers etc. 
Prepared format stores a copy of the format string, from which it has been created. Beside the format, it stores format parts, and format part can store a 'view' to a part of the primary format string, e.g. to store a name of an argument, or so.
Now consider such situation with usage of `string_view`:
1. User creates a prepared format, e.g. `"{arg}"`. Let's call it `PreparedFormat_1`.
1.1 Prepared format stores a copy of the format string as a `std::basic_string<>("{arg"})`. Let's call it `FormatString_1`.
1.2 Prepared format creates format parts. In fact one part with a 'view' that is a `string_view`, which represents a view of string `"arg"` that is a part of the above `FormatString_1`.
2. User copies `PreparedFormat_1` to a `PreparedFormat_2`.
2.1 `PreparedFormat_2` creates a copy of the `FormatString_1`. Let's call it `FormatString_2`.
2.2 `PreparedFormat_2` creates a copy of the `PreparedFormat_1`'s format parts.
3. User destroys `PreparedFormat_1`.
3.1 `FormatString_1` and `PreparedFormat_1`'s format parts are destroyed.
4. User ends up with a `PreparedFormat_2`, but `PreparedFormat_2`'s format parts are corrupted. The part that represents a name of the argument "arg" is a `string_view` that refers to the destroyed `FormatString_1`, not the `FormatString_2`.

That's why the `string_view_metadata` was introduced. It's representation of a string view that is not bind to any actual string. Later on, when the view is needed, and we have a primary string, we can create a `string_view`, that refers to that primary string. It's done in the `prepared_format` class, where it actually needs the `string_view`s.

An alternative would be to store views as `string_view`s, and recreate them (in fact, not only them but all the prepared parts) on copy/move of a prepared format. I decided to not go this way because IMO it's more error-prone and complicated than the `string_view_metadata` way.

include/fmt/core.h
stryku: The `dynamic_specs_handler` has two template methods to check arguments. I wanted to avoid adding an explicit `on_dynamic_width(auto_id)` and `on_dynamic_precision(auto_id)` overloads. But yes, you're right. Keeping the public API clean is more important (in fact, I didn't realize that it'll be exposed to the public API). I'll change it.

include/fmt/core.h
stryku: It was introduced to simplify the `bool check_arg_id(internal::auto_id)` implementation. I'll revert it (see below).

include/fmt/format.h
```diff
@@ -3162,48 +3309,66 @@ class dynamic_formatter {
...
-    internal::specs_checker<null_handler>
-        checker(null_handler(), internal::get_type<FormatContext, T>::value);
-    checker.on_align(specs_.align());
-    if (specs_.flags == 0);  // Do nothing.
-    else if (specs_.has(SIGN_FLAG))
-      specs_.has(PLUS_FLAG) ? checker.on_plus() : checker.on_space();
-    else if (specs_.has(MINUS_FLAG))
+    basic_format_specs<Char> specs;
+    if (ctx.has_prepared_specs()) {
+      specs = ctx.get_prepared_specs();
```
vitaut: It seems strange to have two different specs here, `specs_` and the ones returned by `ctx.get_prepared_specs()`. Can we avoid that?

Ideally format_context shouldn't be aware of specs at all. Those should be managed by each individual formatter.

include/fmt/format.h
stryku: >Why do we need unrestricted union

TBH I don't know about which union are you asking. Could you please elaborate?

vitaut: `arg_ref::value`

stryku: So, do you suggest to simply use `struct` there? Why? `union` can be used with almost zero cost. Yes, we need to type more code, but couple of bytes are saved with every `arg_ref`. `arg_ref`s are used in prepared format parts, so we're also saving couple user's memory bytes. Let me know if that answers your concerns. I'm still not sure if I got it right.

include/fmt/format.h
```diff
@@ -1822,85 +1843,201 @@ class specs_handler: public specs_setter<typename Context::char_type> {
...
+  FMT_CONSTEXPR arg_ref_getter(const arg_ref_type &ref,
+                               basic_string_view<Char> format_str)
+      : arg_ref_getter_base<arg_ref_type>(ref), format_(format_str) {}
+  FMT_CONSTEXPR basic_string_view<Char> name() const {
+    return this->ref_.val.name.to_view(format_);
+  }
+
+ private:
+  basic_string_view<Char> format_;
+};
```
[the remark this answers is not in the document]
stryku: Good idea. Now, in `prepared_format`, `arg_ref<Char, string_view_metadata>` is converted to a regular `arg_ref<Char, string_value<Char>>` (to be specific, in `prepared_format::metadata_to_string_value_arg_ref()`). Thanks to that, getters are not needed anymore.

stryku:
I hope that I removed all whitespaces-only changes, sorry again for that. Also, sorry for taking so long to reply/make changes - I had tough week with a cold.

I know that I didn't reply for one comment, I won't be able to do it today because this topic needs a deeper discussion and I want to prepare my point of view. I'll try to do it tomorrow. Now I wanted to address as much as I could today.

include/fmt/format.h
vitaut: > So, do you suggest to simply use struct there?

No, I'm suggesting to use a normal (not unrestricted) union and don't fallback to `struct`. This is a minor point though and can be addressed later.

include/fmt/format.h
vitaut: > Because of that dynamic_specs_handler is not able to inherit from its ArgRefCreator.

I don't think this is the case. Base class gives you a subobject in the same way as a member does, it's just optimized away if empty, so you should be able to do:

```c++
...
class dynamic_specs_handler
    : private NameRefCreator, public specs_setter<typename ParseContext::char_type> {
 public:
  FMT_CONSTEXPR dynamic_specs_handler(
      Specs &specs, ParseContext &ctx,
      NameRefCreator name_ref_creator = NameRefCreator())
      : NameRefCreator(name_ref_creator),
        specs_setter<char_type>(specs),
        specs_(specs),
        context_(ctx) {}
  ...
};
```

include/fmt/core.h
```diff
@@ -771,6 +776,8 @@ enum : unsigned long long { is_unpacked_bit = 1ull << 63 };
 
 template <typename Context>
 class arg_map;
+
+struct auto_id {};
```
vitaut: This can be moved back now.

include/fmt/format.h
```diff
@@ -608,7 +618,7 @@ template <typename Char>
 class null_terminating_iterator;
 
 template <typename Char>
-FMT_CONSTEXPR_DECL const Char *pointer_from(null_terminating_iterator<Char> it);
+FMT_CONSTEXPR const Char *pointer_from(null_terminating_iterator<Char> it);
```
vitaut: Why is this changed?

include/fmt/format.h
```diff
@@ -1194,6 +1204,7 @@ struct core_format_specs {
 
   FMT_CONSTEXPR core_format_specs() : precision(-1), flags(0), type(0) {}
   FMT_CONSTEXPR bool has(unsigned f) const { return (flags & f) != 0; }
+  FMT_CONSTEXPR bool has_precision() const { return precision != -1; }
```
vitaut: 👍

include/fmt/format.h
stryku: Ahh, I didn't think that it can be implemented like that. Good idea, will do!

include/fmt/format.h
stryku: That's a minor improvement. I just saw that it's a function, so `FMT_CONSTEXPR` should be used. IMO this doesn't pollute the review too much, so I allowed myself to change it.

include/fmt/format.h
stryku: I see. I think that's not possible (I mean it is, but it'd be more problematic).
1. Old compilers need restricted union - union without any ctor etc.
2. Modern compilers need constexpr union - union with at least one constexpr ctor (in this case, we need those three ctors).

There is a way that I can think of - create two separate unions and ifdef them for old and modern compiler. I decided that it's more problematic and using the `FMT_UNRESTRICTED_UNION` is actually a cleaner way. What do you think?

include/fmt/format.h
vitaut: I think `FMT_CONSTEXPR_DECL` is more appropriate here because `FMT_CONSTEXPR` expands to `inline` and this is not a definition (and, more abstractly, this is a declaration).

stryku: +1, will revert.

include/fmt/format.h
vitaut: > What do you think?

Wouldn't the second option suffer from the same problem?

In general I think the only way to communicate format specifiers should be a formatter object which keeps very clear separation of formatting state from the rest. At least that's the current design and I would very much try to keep it this way unless there are very strong reasons why it's impossible.

In this particular case, I think we should only use `dynamic_format_specs` for formatting and convert relocatable `prepared_format_specs` into them just before calling the `format` function.

stryku: Probably it would but a new member would not be added to the `base_format_context`.

I see your point and I would not want to change the desing too.

Sure, specs can be converted before formatting but the problem remains. There is no way to pass prepared specs into a custom formatter. The formatter used to parse and format custom value is created deep in the `value::format_custom_arg()`. We'd need to create the formatter with/without prepared specs before and somehow pass it to the `value::format_custom_arg()`.
Another problem is that the type of the value is needed to determine the type of its formatter.

stryku: Or maybe instead of passing whole formatter, we could pass only specs. That would solve the problem with figuring out the formatter type.

include/fmt/format.h
vitaut: Fair enough. Let's keep `FMT_UNRESTRICTED_UNION` here.

include/fmt/format.h
vitaut: Can we always use relocatable `prepared_format_specs` instead of `dynamic_format_specs` and avoid this problem in the first place? This would also simplify the code because we won't need two separate classes to represent specs for one narrow use case (variant-like types). Or am I still missing something?

include/fmt/format.h
```diff
@@ -3669,16 +3843,16 @@ operator"" _a(const wchar_t *s, std::size_t) { return {s}; }
...
+#define FMT_STRING(s)                                                     \
+  [] {                                                                    \
+    struct str : fmt::compile_string {                                    \
+      typedef typename std::remove_cv<std::remove_pointer<                \
+          typename std::decay<decltype(s)>::type>::type>::type char_type; \
+      FMT_CONSTEXPR operator fmt::basic_string_view<char_type>() const {  \
+        return {s, sizeof(s) / sizeof(char_type) - 1};                    \
+      }                                                                   \
+    };                                                                    \
+    return str{};                                                         \
```
vitaut: What is this change for?

stryku: I moved the `char_type` typedef into the `str` struct, because it's useful for `prepare()` while handling compile-time format strings. Then, while formatting the pull request's code, clang-format detected that the define has changes, so it formatted whole define. That's why there is so many changes. The actual code change is renaming `ct` => `char_type` and moving it into the `str`.

I know that because of reformatting it mostly looks like whitespace change but can we keep it like this? It'll be pain to track manual and clang-format formatting while working on this pr and with this define. The problem will be resolved with #988

include/fmt/format.h
```diff
@@ -3222,6 +3351,51 @@ typename basic_format_context<Range, Char>::format_arg
   return arg;
 }
 
+template <typename ArgFormatter, typename Char, typename Context,
+          typename Range>
+class format_part_writer {
```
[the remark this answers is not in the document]
stryku: Sure, let's give it a try.

include/fmt/format.h
vitaut: > I know that because of reformatting it mostly looks like whitespace change but can we keep it like this?

Sure.

stryku:
I squashed the commits and made changes as agreed:
* `fmt::prepare()` doesn't support custom formatters (user types, arg joins etc). The prepared specs members was removed from the context.
* `dynamic_format_specs` uses `string_view_metadata` instead of `string_value`. Thanks to that, named argument reference is relocatable.
* removed `format_part_writer`.

I'm aware that android build fails (not sure why gradle is not able to find Ninja). I'll handle this asap.

include/fmt/core.h
```diff
@@ -877,16 +881,18 @@ FMT_CONSTEXPR typename internal::result_of<Visitor(int)>::type
 template <typename Char, typename ErrorHandler = internal::error_handler>
 class basic_parse_context : private ErrorHandler {
  private:
-  basic_string_view<Char> format_str_;
-  int next_arg_id_;
+   basic_string_view<Char> primary_format_str_;
```
vitaut: Why is `primary_format_str_` needed? Shouldn't `format_str_` be enough to resolve references?

Also the indent is a bit off.

include/fmt/format.h
```diff
@@ -1585,68 +1596,79 @@ class specs_setter {
   basic_format_specs<Char> &specs_;
 };
 
+template <typename ErrorHandler> class numeric_specs_checker {
```
vitaut: This seems no longer needed and can be integrated back into `specs_checker` (just revert `specs_checker` changes).

include/fmt/core.h
stryku: AFAIU, while parsing/formatting, `basic_parse_context::advance_to()` can be called, which calls `format_str_.remove_prefix()`, which 'invalidates' `format_str_` and `arg_ref::name` (which is `string_view_metadata`) can not be created because it needs primary format string to calculate offset from the beginning of the format. Does that make sense, or I just messed up something?

include/fmt/format.h
stryku: Hmm, I think it's still useful because it's used in two places: `dynamic_specs_handler` and `prepared_format` while formatting, to check already prepared specs.

vitaut: Ah, I missed the second use.

vitaut:
I merged the main part of this PR with some tweaks. The part that is not merged is test refactoring as I'm not sure that it's a good idea to type parameterize the tests. Will explain in more details a bit later. Thanks for working on this!

include/fmt/core.h
vitaut: Right now it is not a problem because we agreed to address custom formatting separately.

vitaut:
To elaborate on my comment about the tests, I think it would be nice to avoid type parametrization because

1. It introduces another layer of calls which makes tests different from the usual way the API is used. This may potentially obscure bugs in the API and in general makes tests more complicated.

2. We do not need to test the formatting logic multiple times through different APIs. This is why most of the tests are done through `fmt::format`. With prepare we only need to test the parts that behave differently, not all the possible format specifiers.

include/fmt/core.h
stryku: So `basic_parse_context::advance_to()` is called only for custom formatters?

stryku:
I am glad that the main part has been merged! Thanks for the time that you put into review.

Good points about the tests. I'll take a closer look on them, figure out what really needs to be tested and provide a pull request or create an issue to start a discussion.

include/fmt/core.h
vitaut: More or less. The whole context machinery is there to support extension mechanism (custom formatters).

vitaut:
@stryku, could you please review the updated [CONTIBUTING](https://github.com/fmtlib/fmt/blob/ebec00138bd10e949ff34955bcb096387cd32acc/CONTRIBUTING.md) document, particularly the part about licensing, and let me know if you agree with it being applied to your contributions to {fmt}? The library is likely to be relicensed (#1073) so I'm collecting approval from all earlier contributors. Thanks!

PR #950, opened 2018-11-28
Avoid warning for deprectated 'strerror' when compiling with clang-cl.

iprtel:
Clang-cl apparently mimics the behaviour of MSVC very closely. And gives an unnecessary warning.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -45,12 +45,17 @@
 #endif
 
 #ifdef _MSC_VER
-# pragma warning(push)
-# pragma warning(disable: 4127)  // conditional expression is constant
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR. I think a somewhat cleaner solution is to conditionally compiler the fallback function that calls`strerror` (similarly to https://github.com/fmtlib/fmt/commit/fecb2d6f0d6c97986c72c7b28f6ff52b3af6348c).

vitaut:
Closing in favor of a somewhat simpler fix: https://github.com/fmtlib/fmt/commit/9dcf127fa5ba833c273cb18c8dd2d01cf4289cb0.

PR #953, opened 2018-11-30
Use full paths for fmt.pc.in

hotwatermorning:
CMAKE_INSTALL_INCLUDEDIR used in the fmt.pc.in is converted to just "include", not a full path, on OSX. (not tested on other envrionments).

Thus `pkg-config fmt --cflags` results `-Iinclude`.
The same applies to CMAKE_INSTALL_LIBDIR. 

I suppose that they should be CMAKE_INSTALL_FULL_x variables.

Releated information:
* https://gitlab.freedesktop.org/poppler/poppler/commit/c45efa2d7e0db8a4ace0e8c6955b3fa48dc6d070
* https://bugzilla.redhat.com/show_bug.cgi?id=795542


vitaut:
Merged, thanks.

PR #954, opened 2018-12-01
Make return type of basic_format_args::max_size() consistent.

DanielaE:
This squashes nasty narrowing conversion warnings.


vitaut:
Merged, thanks!

PR #956, opened 2018-12-02
Fix sign-conversion warnings reported by Clang7

0x8000-0000:
This patch fixes the last Clang7 build warnings.

Of course, please change ssize_t to ptrdiff_t if you think they the latter is more standard-conformant.


vitaut:
Thanks for the PR. Could you post the warnings that you are trying to fix?

Issue #955, opened 2018-12-02
Implicit conversion warning and x86 only signed/unsigned comparison warnings


AGraber:
Sometimes, this warning appears on MSVC 17:
`\lib\fmt\include\fmt\core.h(1284): warning C4244: 'return': conversion from 'unsigned __int64' to 'unsigned int', possible loss of data`

https://github.com/fmtlib/fmt/blob/e7e2ab107035a54216c6927b14aa8d140c5c761a/include/fmt/core.h#L1284-L1285

When compiling on x86, two more warnings appear:
`\lib\fmt\include\fmt\format-inl.h(625): warning C4018: '<': signed/unsigned mismatch`
https://github.com/fmtlib/fmt/blob/e7e2ab107035a54216c6927b14aa8d140c5c761a/include/fmt/format-inl.h#L625

And:
`\lib\fmt\include\fmt\format-inl.h(649): warning C4018: '>': signed/unsigned mismatch`
https://github.com/fmtlib/fmt/blob/e7e2ab107035a54216c6927b14aa8d140c5c761a/include/fmt/format-inl.h#L649

vitaut:
Warnings in `core.h` and `format-inl.h` should be fixed now: https://ci.appveyor.com/project/vitaut/fmt/builds/20795873/job/qolsauqk7psv2dtt

PR #957, opened 2018-12-02
change signature of 'grisu2_prettify' (#955)

DanielaE:
grisu2_prettify(..,ptrdiff_t,..,..) -> grisu2_prettify(..,size_t,..,..) because the second parameter is called only with arguments of types 'unsigned int' or 'size_t'. This squashes compiler warnings when compiled for 32 bit targets where 'size_t' may have the same size as 'int' and different implicit conversion/promotion rules apply.

While at it, remove an obsolete warning suppression.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -48,9 +48,6 @@
 # pragma warning(push)
 # pragma warning(disable: 4127)  // conditional expression is constant
 # pragma warning(disable: 4702)  // unreachable code
-// Disable deprecation warning for strerror. The latter is not called but
-// MSVC fails to detect it.
-# pragma warning(disable: 4996)
 #endif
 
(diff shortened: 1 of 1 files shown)


vitaut:
The idea was to propagate signedness to the call sites to reduce the number of mixed signed/unsigned computations since Grisu mostly operates on signed types.

DanielaE:
But you still want to keep `gen_digits_params::num_digits` unsigned and guarantee the precondition that the `size` argument is never negative, right? Otherwise there is no semantically correct comparison between those possible on 32 bit targets.

vitaut:
> But you still want to keep gen_digits_params::num_digits unsigned and guarantee the precondition that the size argument is never negative, right?

I was planning to change it to signed. It should be in the range `[0, INT_MAX]` which neither signed nor unsigned guarantee.

PR #959, opened 2018-12-03
Do not override user provided compile flag

Zefz:
Currently core.h overrides my -DFMT_EXCEPTIONS=0 flag supplied from my CMake project which includes fmtlib as a third-party dependency.

This minor edit allows a parent CMake file to disable exceptions by explicitly setting the FMT_EXCEPTIONS flag to 0.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -112,11 +112,13 @@
 #endif
 
 // Check if exceptions are disabled.
+#ifndef FMT_EXCEPTIONS
 #if (defined(__GNUC__) && !defined(__EXCEPTIONS)) || \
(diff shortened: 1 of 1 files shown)


vitaut:
Merged, thanks!

vitaut:
@Zefz, could you please review the updated [CONTIBUTING](https://github.com/fmtlib/fmt/blob/ebec00138bd10e949ff34955bcb096387cd32acc/CONTRIBUTING.md) document, particularly the part about licensing, and let me know if you agree with it being applied to your contributions to {fmt}? The library is likely to be relicensed (#1073) so I'm collecting approval from all earlier contributors. Thanks!

PR #961, opened 2018-12-04
Add experimental emphasis support

Rakete1111:
This patch adds support for text emphasis: bold and underline.

```c++
#define FMT_HEADER_ONLY
#include "include/fmt/color.h"
#include "include/fmt/format.h"

int main() {
  fmt::print(fmt::emphasis::bold, "bold\n");
  fmt::print(fmt::emphasis::underline, "underline\n");

  fmt::print(fmt::emphasis::bold, fmt::color::red, "bold and red\n");
  fmt::print(fmt::emphasis::underline, fmt::color::red, fmt::color::yellow,
             "underline, red and yellow\n");
}
```

### DIFF
--- a/include/fmt/color.h
+++ b/include/fmt/color.h
@@ -293,15 +293,15 @@ public:
     return static_cast<uint8_t>(ems);
   }
   FMT_CONSTEXPR_DECL rgb get_foreground() const FMT_NOEXCEPT {
-    assert(set_foreground_color);
+    assert(has_foreground() && "no foreground specified for this style");
     return foreground_color;
   }
   FMT_CONSTEXPR_DECL rgb get_background() const FMT_NOEXCEPT {
-    assert(set_background_color);
+    assert(has_background() && "no background specified for this style");
     return background_color;
   }
   FMT_CONSTEXPR emphasis get_emphasis() const FMT_NOEXCEPT {
-    assert(ems);
+    assert(has_emphasis() && "no emphasis specified for this style");
     return ems;
   }
 
--- a/include/fmt/color.h
+++ b/include/fmt/color.h
@@ -192,6 +192,11 @@ enum class color : uint32_t {
   yellow_green            = 0x9ACD32  // rgb(154,205,50)
 };  // enum class color
 
+enum class emphasis : uint8_t {
+  bold = 1,
+  underline = 4,
+};  // enum class emphasis
+
 // rgb is a struct for red, green and blue colors.
 // We use rgb as name because some editors will show it as color direct in the
 // editor.
@@ -222,6 +227,13 @@ struct ansi_color_escape {
     to_esc(color.b, buffer + 15, 'm');
     buffer[19] = static_cast<Char>(0);
   }
+  FMT_CONSTEXPR ansi_color_escape(emphasis em) FMT_NOEXCEPT {
+    buffer[0] = '\x1b';
+    buffer[1] = '[';
+    buffer[2] = '0' + (uint8_t)em;
+    buffer[3] = 'm';
+    buffer[4] = '\0';
+  }
   FMT_CONSTEXPR operator const Char *() const FMT_NOEXCEPT { return buffer; }
 
 private:
@@ -245,6 +257,11 @@ template <typename Char>
 FMT_CONSTEXPR ansi_color_escape<Char>
 make_background_color(rgb color) FMT_NOEXCEPT {
   return ansi_color_escape<Char>(color, internal::data::BACKGROUND_COLOR);
+}
+
+template <typename Char>
+FMT_CONSTEXPR ansi_color_escape<Char> make_emphasis(emphasis em) FMT_NOEXCEPT {
+  return ansi_color_escape<Char>(em);
 }
 
 template <typename Char>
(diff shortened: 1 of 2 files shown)


Rakete1111:
I was thinking about that, but then I decided against it, because I don't know how to differentiate between foreground and background colors. `fmt::fg(fmt::color::red)`? But that's not nice to look at IMO. Do you have any idea?

foonathan:
Right, good catch.

I'm actually not too opposed to something like `fg(fmt::color::red)`.

vitaut:
I agree with @foonathan that we should have a single formatting style struct to keep the number of overloads manageable. Using `fg(color)` and `bg(color)` to differentiate seems fine. I like it better than the current positional API with arbitrary order.

vitaut:
And thanks for the PR! This is a cool feature.

Rakete1111:
So, I introduced a new class `text_format`, which stores the text formatting:

```
fmt::print(fmt::emphasis::bold | fg(fmt::color::red), "Hello in bold and red\n");

fmt::text_format hackerman = bg(fmt::color::black) | fg(fmt::color::green);
fmt::print(hackerman | fmt::emphasis::bold, "imma hack u\n");
```

include/fmt/color.h
```diff
@@ -192,6 +191,13 @@ enum class color : uint32_t {
   yellow_green            = 0x9ACD32  // rgb(154,205,50)
 };  // enum class color
 
+enum class emphasis : uint32_t {
+  bold = 1 << 1,
+  italic = 1 << 3,
```
vitaut: Why gaps in the bit pattern?

include/fmt/color.h
```diff
@@ -209,6 +215,126 @@ struct rgb {
   uint8_t b;
 };
 
+// Experimental text formatting support.
+class text_format {
```
vitaut: `format` is a bit of an overloaded term in {fmt}. Can we choose some other name for this struct like `style`?

include/fmt/color.h
```diff
@@ -209,6 +215,126 @@ struct rgb {
...
+    if (!set_background_color) {
+      set_background_color = rhs.set_background_color;
+      background_color = rhs.background_color;
+    } else if (rhs.set_background_color) {
+      background_color.r |= rhs.background_color.r;
+      background_color.g |= rhs.background_color.g;
+      background_color.b |= rhs.background_color.b;
+    }
+
+    ems = (emphasis)((uint32_t)ems | (uint32_t)rhs.ems);
```
vitaut: Please use `static_cast`, possibly wrapped in some helper function, instead of C-style casts as it may cause warnings on higher warning levels.

include/fmt/color.h
```diff
@@ -270,54 +424,34 @@ inline void reset_color<wchar_t>(FILE *stream) FMT_NOEXCEPT {
...
-                basic_format_args<typename buffer_context<Char>::type> args) {
-  internal::fputs<Char>(internal::make_foreground_color<Char>(fd), stdout);
-  vprint(format, args);
-  internal::reset_color<Char>(stdout);
-}
-
-template <
-  typename S, typename Char = typename internal::char_t<S>::type>
-void vprint_rgb(rgb fd, rgb bg, const S &format,
+void vprint_text_format(const text_format &tf, const S &format,
```
vitaut: I think this is unambiguous and can be just `vprint`.

include/fmt/color.h
```diff
@@ -270,54 +424,34 @@ inline void reset_color<wchar_t>(FILE *stream) FMT_NOEXCEPT {
...
                 basic_format_args<typename buffer_context<Char>::type> args) {
-  internal::fputs<Char>(internal::make_foreground_color<Char>(fd), stdout);
-  internal::fputs<Char>(internal::make_background_color<Char>(bg), stdout);
+  internal::fputs<Char>(internal::make_emphasis<Char>(tf.get_emphasis()), stdout);
```
vitaut: Shouldn't we only do this when emphasis is set?

test/format-impl-test.cc
```diff
@@ -25,9 +25,9 @@
 #undef max
 
 #if FMT_HAS_CPP_ATTRIBUTE(noreturn)
-# define FMT_NORETURN [[noreturn]]
+#define FMT_NORETURN [[noreturn]]
```
vitaut: Seems unnecessary

include/fmt/color.h
```diff
@@ -270,54 +424,34 @@ inline void reset_color<wchar_t>(FILE *stream) FMT_NOEXCEPT {
...
-  vprint_rgb(fd, format_str, basic_format_args<context_t>(as));
-}
-
-/**
-  Formats a string and prints it to stdout using ANSI escape sequences to
-  specify foreground color 'fd' and background color 'bg'.
+  specify text formatting.
   Example:
-    fmt::print(fmt::color::red, fmt::color::black,
+    fmt::print(fmt::emphasis::bold | fmt::color::red,
```
vitaut: Do we need to wrap `fmt::color::red` in `fg`?

include/fmt/color.h
Rakete1111: That gets handled elsewhere (`ansi_color_escape`). I can also handle it here if you'd like.

include/fmt/color.h
Rakete1111: The numbers were originally for the escape code, but since many different emphasis can be ORed together, I used this, but I didn't change the actual code. I guess I can remove the bits haha :)

test/format-impl-test.cc
Rakete1111: Oops, forgot to turn off clang-format.

include/fmt/color.h
Rakete1111: Yes of course. :)

include/fmt/color.h
Rakete1111: I don't think creating a new function is worth it. I'm using those casts in two cases, and they're not the same.

include/fmt/color.h
vitaut: Yeah, let's remove the gaps.

include/fmt/color.h
vitaut: I'd suggest omitting the call to `fputs` altogether if emphasis is not set.

vitaut:
Merged, thanks!

Rakete1111:
@vitaut Can you retroactively add a default constructor for `text_style` that I completely forgot? Thanks! :)

vitaut:
> Can you retroactively add a default constructor for text_style that I completely forgot? 

Sure, added.

PR #971, opened 2018-12-08
Add a simple test for strict compiler warnings

### LINKED ISSUE #970: Add a simple "hello fmt" project which tests the compiler flags
There seem to be an ebb and flow of regressions when building with stricter compilation flags. I suggest adding a simple "hello fmt" project that is built using the following flags for GCC and Clang:

-Wall -Wextra -Werror -pedantic -Wconversion -Wsign-conversion -Wunused -Wshadow -Wdouble-promotion -Wcast-align -Wnull-dereference 

With this small project the automated builders could catch the build breaks.

0x8000-0000:
Fixes #970

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -200,4 +200,8 @@ if (FMT_PEDANTIC AND NOT WIN32)
     "-DCMAKE_CXX_STANDARD=${CMAKE_CXX_STANDARD}"
     "-DPEDANTIC_COMPILE_FLAGS=${PEDANTIC_COMPILE_FLAGS}"
     "-DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}")
+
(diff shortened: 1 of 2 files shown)


DanielaE:
If the Travis builds fail because of these compiler flags, then should the AppVeyor builds fail, too. The MSVC compiler flags giving the same level of static analysis are `/W4 /WX`.

DanielaE:
Well, I'm a total noob when it comes to CMake. I don't use it in any shape or form. I am pretty sure there exists some kind of CMake syntax to raise the warning level to /W4 and treat warnings as errors or apply the respective commandline flags conditionally for MSVC.
AFAICS your additional test isn't even excercised on AppVeyor.

DanielaE:
Nice!

vitaut:
Closing per comment in #970.

PR #973, opened 2018-12-09
Disable printing the reset escape code when no style modifiers where applied.

Rakete1111:
As the title says. This avoids printing "\x1b[0m" when no ASCII escape code was applied.


vitaut:
Merged, thanks!

PR #979, opened 2018-12-14
DOCS: VS .sln filename changed from FORMAT.sln to FMT.sln

HecticSerenity:
after executing `cmake <path-to-fmt>, the output .sln file is FMT.sln not FORMAT.sln.


vitaut:
Merged, thanks.

PR #983, opened 2018-12-16
Fix conversion and other pedantic warnings

0x8000-0000:
Supersedes #981 and #982 . Rebased onto current master.

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -377,7 +377,7 @@ struct formatter<std::chrono::duration<Rep, Period>, Char> {
       -> decltype(ctx.begin()) {
     auto begin = ctx.begin(), end = ctx.end();
     end = parse_chrono_format(begin, end, internal::chrono_format_checker());
-    format_str = basic_string_view<Char>(&*begin, end - begin);
+    format_str = basic_string_view<Char>(&*begin, static_cast<size_t>(end - begin));
     return end;
   }
 
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -250,7 +250,7 @@ struct chrono_formatter {
   void write(int value, int width) {
     typedef typename int_traits<int>::main_type main_type;
     main_type n = to_unsigned(value);
-    int num_digits = static_cast<int>(internal::count_digits(n));
+    int num_digits = internal::count_digits(n);
     if (width > num_digits)
       out = std::fill_n(out, width - num_digits, '0');
     out = format_decimal<char_type>(out, n, num_digits);
--- a/include/fmt/color.h
+++ b/include/fmt/color.h
@@ -246,7 +246,7 @@ struct color_type {
   }
   FMT_CONSTEXPR color_type(rgb rgb_color) FMT_NOEXCEPT
     : is_rgb(true), value{} {
-    value.rgb_color = (rgb_color.r << 16) + (rgb_color.g << 8) + rgb_color.b;
+    value.rgb_color = (static_cast<uint32_t>(rgb_color.r) << 16) | (static_cast<uint32_t>(rgb_color.g) << 8) | rgb_color.b;
   }
   FMT_CONSTEXPR color_type(terminal_color term_color) FMT_NOEXCEPT
     : is_rgb(), value{} {
@@ -395,22 +395,22 @@ struct ansi_color_escape {
     // sequence.
     if (!text_color.is_rgb) {
       bool is_background = esc == internal::data::BACKGROUND_COLOR;
-      uint8_t value = text_color.value.term_color;
+      uint32_t value = text_color.value.term_color;
       // Background ASCII codes are the same as the foreground ones but with
       // 10 more.
       if (is_background)
-        value += 10;
+        value += 10u;
 
       std::size_t index = 0;
       buffer[index++] = static_cast<Char>('\x1b');
       buffer[index++] = static_cast<Char>('[');
 
-      if (value >= 100) {
+      if (value >= 100u) {
         buffer[index++] = static_cast<Char>('1');
-        value %= 100;
+        value %= 100u;
       }
-      buffer[index++] = static_cast<Char>('0' + value / 10);
-      buffer[index++] = static_cast<Char>('0' + value % 10);
+      buffer[index++] = static_cast<Char>('0' + value / 10u);
+      buffer[index++] = static_cast<Char>('0' + value % 10u);
 
       buffer[index++] = static_cast<Char>('m');
       buffer[index++] = static_cast<Char>('\0');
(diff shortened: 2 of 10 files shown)


0x8000-0000:
With 8ea3c13, if we enable sign-conversion warnings as part of the FMT_PEDANTIC flag set then we don't get any more warnings in the headers.

There are however several conversion warnings in the test code - but it is heavily templated so I'm not sure how to fix them.

include/fmt/chrono.h
```diff
@@ -377,7 +377,7 @@ struct formatter<std::chrono::duration<Rep, Period>, Char> {
       -> decltype(ctx.begin()) {
     auto begin = ctx.begin(), end = ctx.end();
     end = parse_chrono_format(begin, end, internal::chrono_format_checker());
-    format_str = basic_string_view<Char>(&*begin, end - begin);
+    format_str = basic_string_view<Char>(&*begin, static_cast<size_t>(end - begin));
```
vitaut: I suggest using `internal::to_unsigned` which also checks if conversion is valid instead of the cast.

include/fmt/color.h
```diff
@@ -246,7 +246,7 @@ struct color_type {
   }
   FMT_CONSTEXPR color_type(rgb rgb_color) FMT_NOEXCEPT
     : is_rgb(true), value{} {
-    value.rgb_color = (rgb_color.r << 16) + (rgb_color.g << 8) + rgb_color.b;
+    value.rgb_color = (static_cast<uint32_t>(rgb_color.r) << 16) | (static_cast<uint32_t>(rgb_color.g) << 8) | rgb_color.b;
```
vitaut: Please break the long line (https://google.github.io/styleguide/cppguide.html#Line_Length).

include/fmt/core.h
```diff
@@ -223,7 +223,7 @@ class basic_buffer {
 
  protected:
   // Don't initialize ptr_ since it is not accessed to save a few cycles.
-  basic_buffer(std::size_t sz) FMT_NOEXCEPT: size_(sz), capacity_(sz) {}
+  basic_buffer(std::size_t sz) FMT_NOEXCEPT: ptr_(FMT_NULL), size_(sz), capacity_(sz) {}
```
vitaut: `ptr_` is not initialized intentionally here. This have measurable performance impact on some benchmarks.

include/fmt/core.h
```diff
@@ -1328,7 +1328,9 @@ struct named_arg_base {
   mutable char data[
     sizeof(basic_format_arg<typename buffer_context<Char>::type>)];
 
-  named_arg_base(basic_string_view<Char> nm) : name(nm) {}
+  named_arg_base(basic_string_view<Char> nm) : name(nm) {
+     std::memset(data, 0, sizeof(data));
```
vitaut: We should avoid unnecessary work here as well.

include/fmt/format-inl.h
```diff
@@ -172,7 +172,7 @@ void format_error_code(internal::buffer &out, int error_code,
     abs_value = 0 - abs_value;
     ++error_code_size;
   }
-  error_code_size += internal::count_digits(abs_value);
+  error_code_size += static_cast<size_t>(internal::count_digits(abs_value));
```
vitaut: cast -> `to_unsigned`

include/fmt/format.h
```diff
@@ -996,25 +994,25 @@ inline Iterator format_decimal(
...
-inline Char *format_uint(Char *buffer, UInt value, unsigned num_digits,
+inline Char *format_uint(Char *buffer, UInt value, int num_digits,
                          bool upper = false) {
   buffer += num_digits;
   Char *end = buffer;
   do {
     const char *digits = upper ? "0123456789ABCDEF" : "0123456789abcdef";
     unsigned digit = (value & ((1 << BASE_BITS) - 1));
-    *--buffer = static_cast<Char>(BASE_BITS < 4 ? '0' + digit : digits[digit]);
+    *--buffer = static_cast<Char>(BASE_BITS < 4 ? static_cast<char>('0' + digit) : digits[digit]);
```
vitaut: Line width limit.

include/fmt/format.h
```diff
@@ -1963,7 +1961,9 @@ FMT_CONSTEXPR bool find(Ptr first, Ptr last, T value, Ptr &out) {
 template <>
 inline bool find<false, char>(
     const char *first, const char *last, char value, const char *&out) {
-  out = static_cast<const char*>(std::memchr(first, value, last - first));
+  const std::ptrdiff_t length = last - first;
+  FMT_ASSERT(length >= 0, "invalid input range");
+  out = static_cast<const char*>(std::memchr(first, value, static_cast<size_t>(length)));
```
vitaut: `std::memchr(first, value, to_unsigned(last - first))` will do the check.

include/fmt/format.h
```diff
@@ -2043,10 +2043,11 @@ FMT_CONSTEXPR const typename ParseContext::char_type *
 
 template <typename Char, typename ErrorHandler, typename... Args>
 class format_string_checker {
+  enum { INVALID_ARG_ID = 0xffffffffu };
```
vitaut: `0xffffffffu` -> `std::numeric_limits<unsigned>::max()` to improve portability.

include/fmt/format.h
```diff
@@ -2289,19 +2290,19 @@ class basic_writer {
   //   <left-padding><prefix><numeric-padding><digits><right-padding>
   // where <digits> are written by f(it).
   template <typename Spec, typename F>
-  void write_int(unsigned num_digits, string_view prefix,
+  void write_int(int num_digits, string_view prefix,
                  const Spec &spec, F f) {
-    std::size_t size = prefix.size() + num_digits;
+    std::size_t size = prefix.size() + static_cast<size_t>(num_digits);
```
vitaut: `to_unsigned`

include/fmt/format.h
```diff
@@ -2906,9 +2907,9 @@ inline void format_decimal(char *&buffer, T value) {
     *buffer++ = internal::data::DIGITS[index + 1];
     return;
   }
-  unsigned num_digits = internal::count_digits(abs_value);
+  int num_digits = internal::count_digits(abs_value);
   internal::format_decimal<char>(
-        internal::make_checked(buffer, num_digits), abs_value, num_digits);
+        internal::make_checked(buffer, static_cast<size_t>(num_digits)), abs_value, num_digits);
```
vitaut: `to_unsigned`

include/fmt/time.h
```diff
@@ -122,7 +122,7 @@ struct formatter<std::tm, Char> {
     auto end = it;
     while (end != ctx.end() && *end != '}')
       ++end;
-    tm_format.reserve(end - it + 1);
+    tm_format.reserve(static_cast<size_t>(end - it + 1));
```
vitaut: `to_unsigned`

include/fmt/chrono.h
0x8000-0000: ok

include/fmt/color.h
0x8000-0000: ok

include/fmt/format-inl.h
0x8000-0000: ok

include/fmt/core.h
0x8000-0000: OK - as long as the header does not produce warnings when included in user-level code.

0x8000-0000:
Superseded buy #989

Issue #958, opened 2018-12-03
Introduce clang-format


stryku:
It would be really helpful. I think there should be a `.clang-format` file created and a cf version mentioned in the `CONTRIBUTING.rst`

luncliff:
Sounds reasonable.  I've been using `.clang-format` something like below. 

Following links will be helpful. 
 - Designing `.clang-format` : https://clangformat.com/ 
 - Support in Visual Studio : https://blogs.msdn.microsoft.com/vcblog/2018/03/13/clangformat-support-in-visual-studio-2017-15-7-preview-1/
 - Support in Visual Studio Code: https://code.visualstudio.com/docs/languages/cpp#_code-formatting

```
BasedOnStyle: Google
Language: Cpp
Standard: Cpp11

UseTab: Never
IndentWidth: 4
ColumnLimit: 80
PointerAlignment: Left

AlignTrailingComments: true
AllowShortFunctionsOnASingleLine: false
AllowShortIfStatementsOnASingleLine: false
AllowShortLoopsOnASingleLine: false
AlwaysBreakTemplateDeclarations: true 

BreakBeforeBinaryOperators: true
BreakBeforeBraces: Allman
BreakBeforeTernaryOperators: true

ConstructorInitializerAllOnOneLineOrOnePerLine: true
Cpp11BracedListStyle: true

IndentCaseLabels: false
IndentFunctionDeclarationAfterType: true
NamespaceIndentation: None
```

foonathan:
I've run https://github.com/johnmcfarlane/unformat and it created the following .clang-format file. It only did some minor changes to format.h, I recommend you try it out @vitaut.
[clang-format.txt](https://github.com/fmtlib/fmt/files/2652619/clang-format.txt)

(Had to rename the file for github)

DanielaE:
Jonathan's `.clang-format` can be simplified to just
```
BasedOnStyle: Google
AlwaysBreakTemplateDeclarations: true
```
Enjoy!

vitaut:
Done in https://github.com/fmtlib/fmt/commit/58b6f8db48792a6f8b43a0b6caef1de5fac96f03.

PR #988, opened 2018-12-19
Clang Format

luncliff:
## Note
 - Created/Applied `.clang-format` based on Google Style
 - The format to header / src / test files

Each clang format file dominates its sub-directories. So the file had to be placed at the root.  
Consider the style fits your intent and let me know.

You can discard the PR if it feels unnecessary. :)

### History
 - #958

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,12 @@
+BasedOnStyle: Google
+Language: Cpp
+Standard: Cpp11
+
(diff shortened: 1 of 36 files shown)


vitaut:
Thanks, but it had to wait until later because this will introduce a lot of conflicts and there are some important changes in flight.

PR #989, opened 2018-12-19
Fix conversion warnings

0x8000-0000:
Supersedes #983. Rebased onto current master and resolved all review comments.


include/fmt/format.h
```diff
@@ -1460,7 +1460,7 @@ FMT_CONSTEXPR unsigned parse_nonnegative_int(
   }
   unsigned value = 0;
   // Convert to unsigned to prevent a warning.
-  unsigned max_int = (std::numeric_limits<int>::max)();
+  unsigned max_int = internal::to_unsigned(std::numeric_limits<int>::max());
```
0x8000-0000: This fails to build on Windows. Somehow I get a feeling we're bitten by some max() macro coming from system headers or from the infamous <windows.h>

Ugh... so that's why it looked like that...

https://stackoverflow.com/questions/11544073/how-do-i-deal-with-the-max-macro-in-windows-h-colliding-with-max-in-std

vitaut:
@0x8000-0000, could you please review the updated [CONTIBUTING](https://github.com/fmtlib/fmt/blob/ebec00138bd10e949ff34955bcb096387cd32acc/CONTRIBUTING.md) document, particularly the part about licensing, and let me know if you agree with it being applied both to your contributions to {fmt}? The library is likely to be relicensed (#1073) so I'm collecting approval from all earlier contributors. Thanks!

0x8000-0000:
Victor,


I do agree with re-licensing my contributions under any license you find 
best for the project. Thank you for your efforts so far, and for future 
efforts in integrating this into C++20.


Thank you,

florin

Issue #993, opened 2018-12-30
Allow fmt::format and friends to use colors


samvv:
Would it be possible to add a color specifier to `fmt::format`, `fmt::vformat` just like it is possible `fmt::print` and `fmt::vprint`? I'm having an issue where I only want part of a string to be colored, so I'd like to wrap it in `fmt::format` and pass the colored result to a `fmt::print`, but currently that is not possible. I thought about changing it myself but it required more changes than expected and I don't fully know how the code works.

Great library, by the way!

Naios:
Here we go #994, I'm looking forward for your feedback.

PR #994, opened 2018-12-30
Add support for using text_style in format and vformat directly (#993)

Naios:
This PR extends the `format` and `vformat` function to also work with a provided text_style:

```
std::string message = fmt::format(fmt::emphasis::bold | fg(fmt::color::red),
                                  "The answer is {}", 42);
std::cout << message << std:.endl;
```


ricco19:
Looks like this causes a compilation failure with clang: [goldbolt](https://godbolt.org/z/4pGc2x)

Issue #980, opened 2018-12-15
Get rid of null_terminating_iterator


vitaut:
Rewrite the `printf` parsing logic without `null_terminating_iterator` and remove the latter (similar to https://github.com/fmtlib/fmt/commit/8668639ae200631c77cecddd9569fb8ea1371b9c).

PR #995, opened 2018-12-31
Get rid of 'null_terminating_iterator' in printf.h (#980)

DanielaE:
The iterator class itself and its helpers 'to_pointer' are removed after reformulating the formatting loop in terms of a 'basic_string_view<Char>::iterator'.

### DIFF
--- a/include/fmt/printf.h
+++ b/include/fmt/printf.h
@@ -16,130 +16,6 @@
 FMT_BEGIN_NAMESPACE
 namespace internal {
 
-// An iterator that produces a null terminator on *end. This simplifies parsing
-// and allows comparing the performance of processing a null-terminated string
-// vs string_view.
-template <typename Char>
-class null_terminating_iterator {
- public:
-  typedef std::ptrdiff_t difference_type;
-  typedef Char value_type;
-  typedef const Char* pointer;
-  typedef const Char& reference;
-  typedef std::random_access_iterator_tag iterator_category;
-
-  null_terminating_iterator() : ptr_(0), end_(0) {}
-
-  FMT_CONSTEXPR null_terminating_iterator(const Char *ptr, const Char *end)
-    : ptr_(ptr), end_(end) {}
-
-  template <typename Range>
-  FMT_CONSTEXPR explicit null_terminating_iterator(const Range &r)
-    : ptr_(r.begin()), end_(r.end()) {}
-
-  FMT_CONSTEXPR null_terminating_iterator &operator=(const Char *ptr) {
-    assert(ptr <= end_);
-    ptr_ = ptr;
-    return *this;
-  }
-
-  FMT_CONSTEXPR Char operator*() const {
-    return ptr_ != end_ ? *ptr_ : Char();
-  }
-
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks a lot for the PR! Looks good, but I think `parse_header` which is called from `format` still assumes null termination, so it might be unsafe to merge as is.

DanielaE:
I may be blind, but I can't find any check for a null termination there: neither in `parse_header` nor in `parse-nonnegative_int`. The only somewhat zero-related comparison that I can see is with `'0'` rather than `0`. Can you point me to a line number or so, please?

vitaut:
There are no explicit checks for null terminator but `parse_header` doesn't do any range checks relying on null termination when dereferencing the iterator passed to it.

DanielaE:
Well, call me legally blind 😄 But I'm in good company: Howard had a similar facepalm moment when I was pointing him out an equally stupid mistake in "date" recently. As a grown woman I can stand such situations.

Ok, another take - it's becoming nasty ...

include/fmt/printf.h
```diff
@@ -579,57 +449,60 @@ unsigned basic_printf_context<OutputIt, Char, AF>::parse_header(
...
 void basic_printf_context<OutputIt, Char, AF>::format() {
   auto &buffer = internal::get_container(this->out());
-  auto start = iterator(this->parse_context());
+  const auto range = this->parse_context();
+  const Char * end = range.end();
+  const Char * start = range.begin();
   auto it = start;
-  using internal::pointer_from;
-  while (*it) {
+  while (it < end) {
```
vitaut: I don't think all iterators support relational operators, so `!=` might be a better choice. Same in other places. This will make supporting output iterators easier.

include/fmt/printf.h
```diff
@@ -539,33 +412,30 @@ void basic_printf_context<OutputIt, Char, AF>::parse_flags(
...
-        --it;
         return;
     }
   }
 }
 
 template <typename OutputIt, typename Char, typename AF>
 typename basic_printf_context<OutputIt, Char, AF>::format_arg
-  basic_printf_context<OutputIt, Char, AF>::get_arg(
-    iterator it, unsigned arg_index) {
```
vitaut: LOL, I have no idea why it was taking an iterator.

include/fmt/printf.h
DanielaE: There is no iterator involved here, only pointers. And for those all two-way relational operators are defined as long as they point into the same array (they do).  It is all about requiring as few comparisons as possible because they potentially stall cpu pipelines (I know, a weak argument).
I can't see a connection to the output iterators. These are mostly orthogonal to the input range as may prepared PR shows.
Well, I will dig even more into the data flow analysis and pepper some more comparisions into the code where required.

DanielaE:
An updated PR is in flight.

include/fmt/printf.h
vitaut: You are right of course. I was thinking about output iterators for some reason which makes no sense in this context. Sorry for the noise.

vitaut:
Merged, thanks!

PR #998, opened 2019-01-02
Replaced usage of gtest's internal scoped_ptr with unique_ptr.

mwinterb:
scoped_ptr was removed in with gtest google/googletest@e857f9cdd998136b9aad634272301f5b2d0476ea.


vitaut:
Merged, thanks!

Issue #917, opened 2018-10-24
Printf support size limit


fancl20:
Does fmt::printf or fmt::sprintf support size limit like std::snprintf or fmt::format_to_n ? c_str Arguments contains dirty data may hang the program without size limit.

vitaut:
No, but it should be relatively straightforward to implement one in the same way as `format_to_n`. PRs are welcome!

DanielaE:
@vitaut : the requested feature is implemented, may be you like it. Merging it requires a merge of #995 before.

vitaut:
@DanielaE, awesome, looking forward for the PR!

PR #999, opened 2019-01-04
Implement 'snprintf(size_t n, const S &format, const Args & ... args)…

DanielaE:
…' (#917)

Mostly equivalent to 'sprintf(const S &format, const Args & ... args)' but generates at most 'n' characters.

### DIFF
--- a/include/fmt/printf.h
+++ b/include/fmt/printf.h
@@ -192,6 +192,13 @@ void printf(basic_buffer<Char>& buf, basic_string_view<Char> format,
             basic_format_args<Context> args) {
   Context(std::back_inserter(buf), format, args).format();
 }
+
+template <typename OutputIt, typename Char, typename Context>
+internal::truncating_iterator<OutputIt> printf(
+    internal::truncating_iterator<OutputIt> it, basic_string_view<Char> format,
+    basic_format_args<Context> args) {
+  return Context(it, format, args).format();
+}
 }  // namespace internal
 
 using internal::printf;  // For printing into memory_buffer.
@@ -217,7 +224,8 @@ class printf_arg_formatter
   typedef typename Range::value_type char_type;
   typedef decltype(internal::declval<Range>().begin()) iterator;
   typedef internal::arg_formatter_base<Range> base;
-  typedef basic_printf_context<iterator, char_type> context_type;
+  typedef basic_printf_context<iterator, char_type, printf_arg_formatter>
+      context_type;
 
   context_type& context_;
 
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks a lot for the PR. It looks great but I think `fmt::snprintf` should work more like `format_to` and the standard `snprintf` taking an output iterator and a size, because the usual use case for this is to write to a preallocated buffer.

(And sorry for a slow response, I've been busy traveling and preparing for the Kona meeting.)

DanielaE:
Well, I was following suit `fmt::sprintf` which also doesn't take an output iterator. After all, this is C++ rather than C and `fmt::sprintf` does take this into account. I have no problem with returning the untruncated length to, but isn't the other proposed change inconsistent?

vitaut:
> isn't the other proposed change inconsistent?

You mean it's inconsistent with `fmt::sprintf`? To some extent yes, because the latter diverged from the standard API. In the case of `snprintf` there is no reason to substantially diverge from the standard API because it is not as bad. More importantly, we need to support the expected use case (writing to an external buffer). If there is demand we could add a more standard-compliant `sprintf` in the future but I haven't heard from anyone interested in it yet.

DanielaE:
Well, after all you are the mc of your lib 😄 
My vacation is over now, so I'll probably won't be in a hurry to change the PR accordingly.

vitaut:
> Well, after all you are the mc of your lib

I try to listen to user input too sometimes =)

> My vacation is over now, so I'll probably won't be in a hurry to change the PR accordingly.

I very much appreciate that you found time during your vacation to work on this and of course there is no hurry.

DanielaE:
The PR is updated such that it hopefully matches your expectations better now.

include/fmt/printf.h
```diff
@@ -465,21 +470,21 @@ unsigned basic_printf_context<OutputIt, Char, AF>::parse_header(
...
+      continue;
+    }
     if (it != end && *it == c) {
-      buffer.append(start, it);
-      start = ++it;
+      ++it;
+      *(out++) = c;
       continue;
     }
-    buffer.append(start, it - 1);
```
vitaut: I think it's better to keep the old logic of writing ranges of characters rather than character by character. Just replace `buffer.append` with `std::copy`. This will be easier to optimize so that reallocations are minimized.

vitaut:
Thanks! Could you also clang format this?

DanielaE:
Done!

vitaut:
Merged, thanks!

Sciencentistguy:
This appears to have been removed as "broken" in https://github.com/fmtlib/fmt/commit/018d8b57f6ff76fc5bb5abaa243ff3f805bf297f. In what way was it broken? I'm interested in using this feature, if possible

vitaut:
I don't remember the details and would recommend using `format_to_n` instead the legacy `printf` API.

Sciencentistguy:
Yes, and that is the longer-term goal. But the codebase I had in mind is rather old and very large, and contains a lot of printf-syntax formatting calls. I'd like to be able to take advantage of the type-safe formatting without rewriting all callsites to use the fmt syntax, at least straight away.

PR #1001, opened 2019-01-08
Support any string_type as result type of 'fmt::format_as<string_type…

DanielaE:
…>(...)'.

In particular, support 'std::basic_string' with non-defaulted allocators like f.e. 'std::pmr::polymorphic_allocator'.

This requires two overloads 'fmt::format_as<string_type>(...)' for stateless allocators and 'fmt::format_as<string_type>(const allocator &, ...)' for stateful allocators. A user-supplied conversion operator 'to_string(const fmt::basic_memory_buffer &, const string_type *)' found by ADL does the actual conversion to the requested 'string_type'.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -1372,6 +1372,12 @@ std::basic_string<Char> vformat(
   basic_string_view<Char> format_str,
   basic_format_args<typename buffer_context<Char>::type> args);
 
+template <typename Char, typename OutString>
+OutString vformat_as(
+  const OutString *tag,
+  basic_string_view<Char> format_str,
+  basic_format_args<typename buffer_context<Char>::type> args);
+
 template <typename Char>
 typename buffer_context<Char>::type::iterator vformat_to(
   internal::basic_buffer<Char> &buf, basic_string_view<Char> format_str,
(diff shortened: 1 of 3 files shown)


DanielaE:
@vitaut Victor, what do you think about this?

This is basically the last major step on my quest for universal string support for {fmt}.

DanielaE:
Wow, the Android build is failing for some reason in the build configuration?

vitaut:
> what do you think about this?

I think that the functionality is useful, but I'm not 100% convinced that we should provide this exact API. First, it is already possible to format to any string type via `format_to`:

```c++
  my_string s;
  fmt::format_to(std::back_inserter(s), "{}", 42);
```

It is just not as convenient as using `std::format`. It might be OK if the number of call sites is small. If the number is large though, I'm not sure if `format_as` is convenient enough either because you'll need to spell out the string type every time. I think we should enable (and encourage) users to easily create formatting functions for their string types, e.g. 

```c++
template <typename... Args>
my_string my_format(fmt::string_view format, const Args&... args) {
  my_string s;
  fmt::format_to(std::back_inserter(s), format, args...);
  return s;
}
```
instead of trying to provide a generic solution.

Second, I was thinking of using `format_as` as a simpler alternative extension point for the common case when objects of one type can be formatted as objects as another, e.g.

```c++
struct my_int {
  int value;
};
int format_as(my_int i) { return i.value; }

auto s = format("{}", my_int{42});
```

The second point is less important though as names are subject to bikeshedding.

vitaut:
> Wow, the Android build is failing for some reason in the build configuration?

Looks like it cannot find ninja for some reason.

DanielaE:
> I think that the functionality is useful, but I'm not 100% convinced that we should provide this exact API. First, it is already possible to format to any string type via `format_to`.  It is just not as convenient as using fmt::format. It might be OK if the number of call sites is small. If the number is large though, I'm not sure if format_as is convenient enough either because you'll need to spell out the string type every time. I think we should enable (and encourage) users to easily create formatting functions for their string types, instead of trying to provide a generic solution.

Using `format_to` was my first thought as well until colleagues of mine came along telling me that this is not viable for their use cases: they deal with a huge codebase that is built using MFC/ATL based strings in the user-facing code layers, i.e. where most of the formatting takes place. None of these string types support iterators for manipulation! In fact, every non-C string type in the Windows API (in other words: every non-legacy API) uses such string types. Even worse, the current standard string type `HSTRING` is immutable after construction. These interfaces will haunt me in an upcoming mixed-language project. For more on that matter I'd like to refer to [Victor Ciura's Meeting C++ 2018 talk](https://meetingcpp.com/mcpp/slides/2018/These%20Aren't%20the%20COM%20Objects%20You're%20Looking%20For%20-%20Victor%20Ciura%20-%20Meeting%20C++%202018.pdf)

I am pretty sure that problems like these do affect people beyond the Windows world, too: there are so many string types out there which don't interface well with STL in general and STL iterators in particular.

OTOH, **every** string type that I have encountered so far supports construction from contiguous character arrays of known length, possibly augmented with some kind of allocator (not necessarily STL allocators). This led me to the design of the proposed interface where building the full formatted string is done _internally_ in a `fmt::basic_memory_buffer` (with its optimized memory allocation), and constructing a possibly immutable result string object (optionally using its custom allocator) from the final formatting result _externally_. This strategy is the most versatile with the least amount of temporary, intermediate, throw-away memory allocations.

On the subject of convenience regarding the string_type template parameter: we think that this better than the alternatives we could think of so far. A defaulted template parameter doesn't play well with variadic template arguments afaik.

vitaut:
> None of these string types support iterators for manipulation!

I don't think it is a problem because, as you've already found out, it is possible to use an intermediate memory buffer to construct them.

> On the subject of convenience regarding the string_type template parameter: we think that this better than the alternatives we could think of so far. 

Why is it better than a non-template function like

```c++
template <typename... Args>
my_string my_format(fmt::string_view format, const Args&... args) {
  memory_buffer buf;
  fmt::format_to(std::back_inserter(buf), format, args...);
  return {buf.data(), buf.size()};
}
```
?

While it seems appealing to support all of the versatile string types out-of-the-box, I think the unfortunate side effect of this will be that the user code will be plagued with the ugly and verbose templated calls. This is why I think we should not provide such general facility and encourage users (via documentation and examples) to implement non-template formatting functions for specific string types they are interested in. The ability to easily and efficiently implement such functions is one of the main features of {fmt}. For some widely used string types like `QString` we can eventually add support in the library (via optional headers).

DanielaE:
Yeah, you are proably right. I've been mullover this idea the past couple of days and think I like your idea better. So I gladly revoke my PR :smile:

PR #1006, opened 2019-01-13
Fix compile errors due to name-hiding and an unused function argument.

DanielaE:
These problems are pointed out by clang-tidy and fail compilation on msvc at /W4 /WX.

### DIFF
--- a/test/prepare-test.cc
+++ b/test/prepare-test.cc
@@ -627,7 +627,7 @@ template <typename T> struct user_allocator {
     return new value_type[cnt];
   }
 
-  void deallocate(pointer p, size_type cnt) { delete[] p; }
+  void deallocate(pointer p, size_type) { delete[] p; }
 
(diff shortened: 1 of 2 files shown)


test/prepare-test.cc
```diff
@@ -625,7 +625,7 @@ template <typename T> struct user_allocator {
     return new value_type[cnt];
   }
 
-  void deallocate(pointer p, size_type cnt) { delete[] p; }
+  void deallocate(pointer p, size_type /*cnt*/) { delete[] p; }
```
vitaut: nit: I'd just remove the parameter instead of commenting out.

DanielaE: Great, wasn't sure about your policy regarding this. This is my preference as well.

DanielaE:
Done!

vitaut:
Merged, thanks!

Issue #1004, opened 2019-01-11
std::chrono duration formatting is lacking floating point specifiers


DanielaE:
Consider this test code:
```
typedef std::chrono::duration<float, std::milli> fms;
EXPECT_EQ("1.234ms", fmt::format("{}", fms(1.234)));
typedef std::chrono::duration<double, std::milli> dms;
EXPECT_EQ("1.234ms", fmt::format("{}", dms(1.234)));
EXPECT_EQ("1.2ms", fmt::format("{:.2}", dms(1.234)));
```
The first and second test passes, i.e. floating point representation types in `std::chrono::duration` work in general. The last fails by returning `.2` rather than formatting the given time quantity of 1.2345 ms to the given precision. There is seemingly no application of floating point specifiers possible.

DanielaE:
I've implemented some stuff to handle the precision specification, but this is probably a dead end.
https://github.com/DanielaE/fmt/commit/1cec2928b7b9d072e4a88d3ff4469cfedb41f4c6

DanielaE:
Right, in this case (static precision, static formatting string), the workaround of shifting the precision part of the formatting process onto the user certainly works. In the general case it doesn't and maintenance in non-trivial codebases with the prospect of engineering changes might become a nightmare. After all, {fmt}'s added value and most noble goal is relieving users from the burden of formatting.
I have no problem if chrono's output stream operator doesn't handle this (at least, this is my conclusion after reading it's specification). It only means that {fmt} should take this into consideration and possibly not rely on chrono's facilities in _all_ formatting scenarios.

vitaut:
Supporting precision for duration types with floating-point representation makes a lot of sense to me. 

> this is probably a dead end 

@DanielaE why do you think so?

DanielaE:
Well, I am not sure about the direction of the implementation of this feature. I am glad that you support the idea. My experimental stuff does a consistent parsing so far, but it's totally unclear to me on how to proceed from there. I've learned from Howards comment that backing the formatting process on chrono's output streaming operator is not the way to go in this case. Currently I envision an implementation which checks the formatting string for `chrono` formatting specifiers and defers to `chrono`-based formatting in this case with exactly it's features and limitations. If there are none, {fmt}'s facilities are superior and implementing `chrono::duration` formatting is trivial as I've shown in my Meeting C++ 2018 presentation. Taking such a route makes my experiment obsolete: therefore the 'dead end'.

vitaut:
[...]

PR #1012, opened 2019-01-19
Implement precision for floating-point durations.

DanielaE:
The formatting syntax follows p1361r0, augmented by a precision field as proposed in #1004.

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -362,8 +362,10 @@ template <typename Rep, typename Period, typename Char>
 struct formatter<std::chrono::duration<Rep, Period>, Char> {
  private:
   align_spec spec;
+  int precision;
   typedef internal::arg_ref<Char> arg_ref_type;
   arg_ref_type width_ref;
+  arg_ref_type precision_ref;
   mutable basic_string_view<Char> format_str;
   typedef std::chrono::duration<Rep, Period> duration;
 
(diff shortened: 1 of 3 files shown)


include/fmt/chrono.h
```diff
@@ -362,8 +362,10 @@ template <typename Rep, typename Period, typename Char>
 struct formatter<std::chrono::duration<Rep, Period>, Char> {
  private:
   align_spec spec;
+  core_format_specs pspec;
```
vitaut: We only need `int precision` instead of the whole `core_format_specs` here.

include/fmt/chrono.h
```diff
@@ -362,8 +362,10 @@ template <typename Rep, typename Period, typename Char>
...
   typedef internal::arg_ref<Char> arg_ref_type;
   arg_ref_type width_ref;
+  internal::arg_ref<Char> precision_ref;
```
vitaut: `internal::arg_ref<Char>` -> `arg_ref_type`

include/fmt/chrono.h
```diff
@@ -391,12 +393,62 @@ struct formatter<std::chrono::duration<Rep, Period>, Char> {
...
+      specs.type = 'f';
+    }
+    typedef typename format_context_t<OutputIt, Char>::type context;
+    context ctx(out, {}, {}, fctx.locale());
+    arg_formatter<output_range<OutputIt, Char>> af(ctx, &specs);
+    internal::custom_formatter<Char, context> cf(ctx);
+    const auto arg = internal::make_arg<context>(val);
+    if (!visit_format_arg(cf, arg))
+      visit_format_arg(af, arg);
+    return ctx.out();
```
vitaut: I think this can be significantly simplified by using `format_to(out, "{:.{}f}", val, precision)` or something like that.

include/fmt/chrono.h
```diff
@@ -408,6 +460,9 @@ struct formatter<std::chrono::duration<Rep, Period>, Char> {
     begin = internal::parse_align(begin, end, handler);
     if (begin == end) return begin;
     begin = internal::parse_width(begin, end, handler);
+    if (begin == end) return begin;
+    begin = parse_precision(begin, end, handler, std::is_floating_point<Rep>{});
```
vitaut: Do we actually need tag dispatched `parse_precision`? Wouldn't `if (std::is_floating_point<Rep>{})` work and be much simpler?

test/chrono-test.cc
```diff
@@ -186,3 +186,32 @@ TEST(ChronoTest, Locale) {
...
+  EXPECT_EQ("1.23ms", fmt::format("{:.{}}", dms(1.234), 2));
+}
+
+TEST(ChronoTest, FormatFullSpecs) {
+  EXPECT_EQ("1.2ms ", fmt::format("{:6.1}", dms(1.234)));
+  EXPECT_EQ("  1.23ms", fmt::format("{:>8.{}}", dms(1.234), 2));
+  EXPECT_EQ(" 1.2ms ", fmt::format("{:^{}.{}}", dms(1.234), 7, 1));
+  EXPECT_EQ(" 1.23ms ", fmt::format("{0:^{2}.{1}}", dms(1.234), 2, 8));
+  EXPECT_EQ("=1.234ms=", fmt::format("{:=^{}.{}}", dms(1.234), 9, 3));
+  EXPECT_EQ("*1.2340ms*", fmt::format("{:*^10.4}", dms(1.234), 10, 4));
```
vitaut: 👍

include/fmt/chrono.h
DanielaE: Sure, but with my performance-hat on I tried to avoid reparsing.

include/fmt/chrono.h
DanielaE: We don't need it. The tag dispatch is just a clear hint to the compiler to not even try instantiating dead code.

include/fmt/chrono.h
DanielaE: That was my original implementation. I switched over to `core_format_specs` because of the `has-precision` predicate defined there.

DanielaE:
BTW, what about other `wchar_t` and other character types - any plans on that? All these string literals may become annoying then.

vitaut:
Merged, thanks!

> BTW, what about other wchar_t and other character types - any plans on that?

I don't have any plans regarding this at the moment. The whole exercise was mainly to support the chrono integration proposal.

PR #1019, opened 2019-01-25
Implement 'chrono' formatting specifiers '%Q' and '%q' (#1004)

DanielaE:
Howard Hinnant's 'date' library recently gained these two new formatting specifiers. This implementation in {fmt} includes support for 'std::chrono::duration' specializations with floating-point representation types and user-definable precision.

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -19,6 +19,32 @@
 FMT_BEGIN_NAMESPACE
 
 namespace internal {
+template <typename Period> FMT_CONSTEXPR const char* get_units() {
+  return FMT_NULL;
+}
+template <> FMT_CONSTEXPR const char* get_units<std::atto>() { return "as"; }
+template <> FMT_CONSTEXPR const char* get_units<std::femto>() { return "fs"; }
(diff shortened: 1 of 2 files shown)


include/fmt/chrono.h
```diff
@@ -18,6 +18,33 @@
 
 FMT_BEGIN_NAMESPACE
 
+template <typename Period> FMT_CONSTEXPR const char* get_units() {
```
vitaut: This and specializations should probably be internal.

include/fmt/chrono.h
```diff
@@ -118,6 +145,12 @@ FMT_CONSTEXPR const Char* parse_chrono_format(const Char* begin,
     case 'p':
       handler.on_am_pm();
       break;
+    case 'Q':
+      handler.on_value();
```
vitaut: `on_value` is a bit too generic. What kind of value is allowed here, time?

include/fmt/chrono.h
DanielaE: Right, done

include/fmt/chrono.h
DanielaE: Renamed to clarify the intent.

PR #1021, opened 2019-01-29
format-inl.h: fixing compare double with 0

medithe:
If compiling with -Wfloat-equal, I get the following error:
    comparing floating point with == or != is unsafe [-Werror=float-equal]


vitaut:
Thanks for the PR, but unfortunately comparing against epsilon is incorrect here. The bogus warning should be silenced in https://github.com/fmtlib/fmt/commit/7ca8fc3b170f14c4ca9acee4d4af51957ec7576e. Interestingly the new code also generates slightly better assembly.

PR #1022, opened 2019-01-29
Mark deprecated functionality with attributes and remove usage of such features

eliaskosunen:
This PR marks features commented as deprecated with attributes; `[[deprecated]]` if compiled with C++14 and `__attribute__((deprecated))` or `__declspec(deprecated)` if not.

I was able to replace most of the cases where deprecated functionality was used in the library, but one case lingers: https://github.com/eliaskosunen/fmt/blob/599632ee226d74d6c34d89290ebabd7320c90afe/include/fmt/format.h#L3061

`arg_map::init` takes `const basic_format_args<Context>&` as its sole argument, which can only be extracted from `context_base` with the deprecated member function `args()`.

### DIFF
--- a/include/fmt/color.h
+++ b/include/fmt/color.h
@@ -15,21 +15,33 @@ FMT_BEGIN_NAMESPACE
 #ifdef FMT_DEPRECATED_COLORS
 
 // color and (v)print_colored are deprecated.
-enum color { black, red, green, yellow, blue, magenta, cyan, white };
-FMT_API void vprint_colored(color c, string_view format, format_args args);
-FMT_API void vprint_colored(color c, wstring_view format, wformat_args args);
+enum FMT_DEPRECATED color {
+  black,
+  red,
+  green,
+  yellow,
+  blue,
+  magenta,
+  cyan,
+  white
+};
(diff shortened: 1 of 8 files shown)


eliaskosunen:
Travis builds are erroring because of `-Werror` and the usage of deprecated `context_base::args()`.

vitaut:
This is great, thanks!

> arg_map::init takes const basic_format_args<Context>& as its sole argument, which can only be extracted from context_base with the deprecated member function args().

I suggest making args protected and adding a using declaration in `basic_format_context` for backward compatibility. Not sure if using declaration can be deprecated.

eliaskosunen:
`[[deprecated]]` cannot be used with _using-declaration_s, to that option is off the table. I circumvented the issue by simply commenting the `FMT_DEPRECATED` out and adding a comment explaining it. I couldn't come up with any alternative ways of fixing the issue apart from some messy ones requiring way too much work. I think this should be fixed at some later time.

There was also some usage of the deprecated `format_decimal` overload in `format-test.cc`. For that, I simply disabled the warning with pragmas. I don't know that part of the library all that well, so I'm not really one to fix it.

vitaut:
Merged, thanks!

PR #1029, opened 2019-02-03
(Unwillingly) protect against overloaded comma operators in decltype

morinmorin:
It seems that the doc doesn't explicitly ban overloading comma operators. So this PR adds guards against overloaded comma operators by sprinkling void casts.

Here is a (nonsense) test case:
```
#include <iostream>

#include "fmt/format.h"
#include "fmt/ostream.h"

template <typename OStream>
struct ostream_with_comma_op
{
    std::ostream& os;

    ostream_with_comma_op(std::ostream& _os) : os(_os) {}

    template <typename T>
    ostream_with_comma_op& operator<<(T const& t)
    {
        os << t;
        return *this;
    }
};

template <typename OStream, typename T>
void operator,(ostream_with_comma_op<OStream> const&, T&&);

struct ouch {};

ostream_with_comma_op<std::ostream> operator<<(std::ostream& os, ouch)
{
    os << "ouch";
    return ostream_with_comma_op<std::ostream>(os);
}

int main ()
{
    std::cout << fmt::format("{}", ouch{}) << '\n';
    
    return 0;
}
```

Another options would be: adding a note to the doc that pathological types are not supported.

### DIFF
--- a/include/fmt/ostream.h
+++ b/include/fmt/ostream.h
@@ -56,8 +56,8 @@ template <typename Char> struct test_stream : std::basic_ostream<Char> {
 template <typename T, typename Char> class is_streamable {
  private:
   template <typename U>
-  static decltype(internal::declval<test_stream<Char>&>()
-                      << internal::declval<U>(),
+  static decltype((void)(internal::declval<test_stream<Char>&>()
+                      << internal::declval<U>()),
                   std::true_type())
   test(int);
 
--- a/include/fmt/prepare.h
+++ b/include/fmt/prepare.h
@@ -538,14 +538,15 @@ struct parts_container_concept_check : std::true_type {
 
   template <typename T> static std::false_type has_add_check(check_second);
   template <typename T>
-  static decltype(declval<T>().add(declval<typename T::format_part_type>()),
+  static decltype((void)declval<T>().add(
+                      declval<typename T::format_part_type>()),
                   std::true_type()) has_add_check(check_first);
   typedef decltype(has_add_check<PartsContainer>(check_first())) has_add;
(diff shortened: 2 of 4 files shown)


morinmorin:
TL;DR - this is only theoretical.

Formatted output is one of headaches in teaching C++. So I sincerely hope that this great library, fmt, will be adopted in C++20! As for teachability, I'm quite happy that fmt also provides a way to support custom types via `operator <<`, since overloading is easier for novice C++ programmers than template specializations. I was curious about how `format` uses `operator <<`, so I checked the code (and learned `streambuf`). Then I found this
```
decltype(
  internal::declval<test_stream<Char>&>() << internal::declval<U>(), std::true_type()
)
```
I don't think this code breaks in real-world use cases at all, but...
- I thought that it might make sense to have robust codes, since this library is an implementation of the future stdlib component ;)
- innocent type, say `Ostreamable`, would break by inserting _three_ lines of codes
```
struct type_with_comma_op {};
template <typename T> void operator,(type_with_comma_op, T&&);
template <typename T> type_with_comma_op operator<<(T&&, Ostreamable);
```
- `… << … , …` is a real-world code (an Eigen's initialization syntax).

I'm generally in favor of adding constraints that the type should not have overloaded comma operators; but in this case, the (formal) constraint that we need is not so simple because `Ostreamable` is not the type to forbid overloaded comma operators. So I updated the code rather than the doc.

vitaut:
Fair enough. Could you by any chance add a test at least for the ostream case?

morinmorin:
Thanks, done (but only for the ostream case..., sorry).

vitaut:
Merged, thanks!

PR #1031, opened 2019-02-06
Fix library deprecation warning

DanielaE:
Since C++17, using template specialization `std::allocator<void>` in any shape or form (even just mentioning it) is deprecated! A simple workaround is replacing `void` by another (preferably empty) type, e.g. `fmt::monostate`.

Found by Clang 9 in Visual Studio.


vitaut:
Merged, thanks!

PR #1051, opened 2019-02-21
Clarify lifetimes of named_arg parameters

mikelui:
Document that fmt::arg takes a non-owning
reference, even if that reference is to
a temporary. As such, users should make sure
the lifetime of the reference lasts as long
as the named argument.

Disable doxygen for deleted overload.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -1385,10 +1385,10 @@ typename buffer_context<Char>::type::iterator vformat_to(
   \rst
   Returns a named argument to be used in a formatting function.
 
-  Usage is analogous to `std::reference_wrapper
-  <https://en.cppreference.com/w/cpp/utility/functional/reference_wrapper>`_,
-  except that rvalues references are not disabled.
-  The user should take care to only pass in temporaries when
+  The named argument holds a reference and does not extend the lifetime
+  of its arguments.
+  Consequently, a dangling reference can accidentally be created.
+  The user should take care to only pass this function temporaries when
   the named argument is itself a temporary, as per the following example.
 
   **Example**::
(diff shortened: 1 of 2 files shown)


include/fmt/core.h
```diff
@@ -1384,6 +1385,12 @@ typename buffer_context<Char>::type::iterator vformat_to(
   \rst
   Returns a named argument to be used in a formatting function.
 
+  Usage is analogous to `std::reference_wrapper
+  <https://en.cppreference.com/w/cpp/utility/functional/reference_wrapper>`_,
+  except that rvalues references are not disabled.
```
epruesse: should be `rvalue references`

mikelui: Actually I think it should just be rvalues, since rvalue references are lvalues.

vitaut: Is the rvalue part useful here? I'd drop it altogether.

include/fmt/core.h
```diff
@@ -1394,9 +1401,11 @@ inline internal::named_arg<T, FMT_CHAR(S)> arg(const S& name, const T& arg) {
   return {name, arg};
 }
 
+/// \cond
```
vitaut: This is not needed. The ambiguity is caused by signature mismatch, not presence of extra overload.

include/fmt/core.h
mikelui: I'll change it to just dangling references. Not all users would know what an rvalue is (or reference_wrapper for that matter)

include/fmt/core.h
mikelui: Ah okay. I'll change it in doc/api.rst

vitaut:
Merged, thanks!

vitaut:
@mikelui, could you please review the updated [CONTIBUTING](https://github.com/fmtlib/fmt/blob/ebec00138bd10e949ff34955bcb096387cd32acc/CONTRIBUTING.md) document, particularly the part about licensing, and let me know if you agree with it being applied to your contributions to {fmt}? The library is likely to be relicensed (#1073) so I'm collecting approval from all earlier contributors. Thanks!

PR #1052, opened 2019-02-22
Documentation/api.rst: Use same template argument order as in the pro…

t-b:
…totype

The formatter prototype is

// A formatter for objects of type T.
template <typename T, typename Char = char, typename Enable = void>
struct formatter {
  explicit formatter(internal::dummy_formatter_arg);
  };

so we should use this order in the example as well.

In addition we use Char instead of plain char to not confuse the reader
as this name is totally arbitrary.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -165,8 +165,8 @@ You can also write a formatter for a hierarchy of classes::
     virtual std::string name() const { return "B"; }
   };
 
-  template <typename T>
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR, but I don't think adding an extra template parameter is necessary in this example. The default character type is `char` and the example is there to illustrate formatting of hierarchy of classes, not supporting various character types.

t-b:
@vitaut okay, fine as well. thanks for the speedy feedback.

PR #1054, opened 2019-02-23
Fixes for some pedantic warnings

ricco19:
Some small changes required for me to compile with stricter compiler flags

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -62,6 +62,10 @@
 #if FMT_GCC_VERSION >= 406 || FMT_CLANG_VERSION
 #  pragma GCC diagnostic push
 
+// Disable warning about not handling all enums in switch statement even with
+// a default case
+#  pragma GCC diagnostic ignored "-Wswitch-enum"
+
 // Disable the warning about declaration shadowing because it affects too
 // many valid cases.
 #  pragma GCC diagnostic ignored "-Wshadow"
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -1798,7 +1798,8 @@ class specs_handler : public specs_setter<typename Context::char_type> {
   typedef typename Context::format_arg format_arg;
 
   FMT_CONSTEXPR format_arg get_arg(auto_id) {
-    return internal::get_arg(context_, parse_context_.next_arg_id());
+    return internal::get_arg(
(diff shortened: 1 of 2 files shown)


include/fmt/format.h
```diff
@@ -62,6 +62,10 @@
 #if FMT_GCC_VERSION >= 406 || FMT_CLANG_VERSION
 #  pragma GCC diagnostic push
 
+// Disable warning about not handling all enums in switch statement even with
+// a default case
+#  pragma GCC diagnostic ignored "-Wswitch-enum"
```
vitaut: Could you post specific warnings this is causing? Might be better to fix those instead of suppressing.

include/fmt/format.h
```diff
@@ -1794,7 +1798,8 @@ class specs_handler : public specs_setter<typename Context::char_type> {
   typedef typename Context::format_arg format_arg;
 
   FMT_CONSTEXPR format_arg get_arg(auto_id) {
-    return internal::get_arg(context_, parse_context_.next_arg_id());
+    return internal::get_arg(
+        context_, static_cast<unsigned>(parse_context_.next_arg_id()));
```
vitaut: Why is it needed? `next_arg_id()` already returns `unsigned`.

vitaut:
Thanks for the PR! Just two minor inline comments.

include/fmt/format.h
ricco19: This was to suppress a warning I was getting with [clang-trunk](https://i.imgur.com/rd2U6Np.png) (-Wshorten-64-to-32)

include/fmt/format.h
ricco19: screenshot_2019-02-23_13-29-00

I thought having a default case is cleaner and just disabling this warning is preferable.

include/fmt/format.h
vitaut: I couldn't repro this: https://godbolt.org/z/tfq3_w. Could you provide an example that reproduces this warning?

ricco19: Can't seem to reproduce it on godbolt, but the warning went away when I included `fmt/format.h` instead of `format` - I was getting this warning with just a call to `fmt::print`.

vitaut:
Merged, thanks!

PR #1060, opened 2019-03-01
Add support for '%' type to output floating point values as a percentage.

gawain-bolton:
This helps with compatibility with Python's format strings.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -2646,10 +2646,12 @@ template <typename Range> class basic_writer {
 
   struct inf_or_nan_writer {
     char sign;
+    bool as_percentage;
     const char* str;
 
     size_t size() const {
-      return static_cast<std::size_t>(INF_SIZE + (sign ? 1 : 0));
+      return static_cast<std::size_t>(INF_SIZE + (sign ? 1 : 0) +
+                                      (as_percentage ? 1 : 0));
     }
     size_t width() const { return size(); }
 
@@ -2657,6 +2659,7 @@ template <typename Range> class basic_writer {
       if (sign) *it++ = static_cast<char_type>(sign);
       it = internal::copy_str<char_type>(
           str, str + static_cast<std::size_t>(INF_SIZE), it);
+      if (as_percentage) *it++ = static_cast<char_type>('%');
     }
   };
 
@@ -2832,8 +2835,8 @@ struct float_spec_handler {
   }
 
   void on_percent() {
-      fixed = true;
-      as_percentage = true;
+    fixed = true;
+    as_percentage = true;
   }
 
   void on_hex() {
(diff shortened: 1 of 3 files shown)


vitaut:
Thanks for the PR, but please fix the CI build, apply clang-format to your changes and add a unit test.

gawain-bolton:
> Thanks for the PR, but please fix the CI build, apply clang-format to your changes and add a unit test.

I have applied clang-format and added unit tests.
The CI builds also now pass.

include/fmt/format.h
```diff
@@ -2854,26 +2867,39 @@ void basic_writer<Range>::write_double(T value, const format_specs& spec) {
...
-  if (internal::fputil::isinfinity(value))
-    return write_inf_or_nan(handler.upper ? "INF" : "inf");
+  if (internal::fputil::isnotanumber(value)) {
+    write_inf_or_nan(handler.upper ? "NAN" : "nan");
+    return;
+  }
+  if (internal::fputil::isinfinity(value)) {
+    write_inf_or_nan(handler.upper ? "INF" : "inf");
+    return;
+  }
```
vitaut: This change looks unrelated. Please revert.

include/fmt/format.h
```diff
@@ -2854,26 +2867,39 @@ void basic_writer<Range>::write_double(T value, const format_specs& spec) {
...
+
   bool use_grisu = fmt::internal::use_grisu<T>() &&
                    (!spec.type || handler.fixed) &&
                    internal::grisu2_format(static_cast<double>(value), buffer,
                                            precision, handler.fixed, exp);
   if (!use_grisu) internal::sprintf_format(value, buffer, spec);
+
+  if (handler.as_percentage) {
+    buffer.push_back('%');
+    --exp;  // Adjust decimal place position.
```
vitaut: Shouldn't this be -= 2?

include/fmt/format.h
gawain-bolton: You're right, I've reverted this changes.

include/fmt/format.h
gawain-bolton: The decimal place position is shifted by one place because only one character, the percent sign, is added.

vitaut:
Merged in https://github.com/fmtlib/fmt/commit/79b79f329ea126c14d150d9b75668ed137b33c42, thanks!

vitaut:
@gawain-bolton, could you please review the updated [CONTIBUTING](https://github.com/fmtlib/fmt/blob/ebec00138bd10e949ff34955bcb096387cd32acc/CONTRIBUTING.md) document, particularly the part about licensing, and let me know if you agree with it being applied to your contributions to {fmt}? The library is likely to be relicensed (#1073) so I'm collecting approval from all earlier contributors. Thanks!

PR #1061, opened 2019-03-02
format: Fix compilation with -fno-exceptions

alexpana:
Including `<format>` with `-fno-exceptions` results in compile errors:

    [...]/format:184:5: error: cannot use 'throw' with exceptions disabled


vitaut:
Thanks for the PR, but `<format>` is there for testing the paper wording, not for general use.

alexpana:
What is "paper wording"?

vitaut:
ISO C++ standard wording for https://wg21.link/p0645.

PR #1068, opened 2019-03-06
Declare the size of RESET_COLOR.

mwinterb:
This is so that the format with a text_tyle will compile even if
header-only mode isn't enabled. Addresses #1063.


mwinterb:
The existing tests in format-impl-test were enough to detect if RESET_COLOR was ever over-allocated, so I didn't add an explicit test covering that.

DanielaE:
Great! This passes the test-suite on VS2019.RC as well:
fmt head deec2d8

vitaut:
Merged, thanks!

Issue #1073, opened 2019-03-08
Relicense {fmt} to make it more friendly for standard library vendors


vitaut:
According to Billy O'Neal, the standard library vendors (including libstdc++) all can't look at BSD-licensed code. To fix this we need to drop all requirements on binary redistribution, e.g. introduce a clause similar to "LLVM Exceptions to the Apache 2.0 License":

> As an exception, if, as a result of your compiling your source code, portions
of this Software are embedded into an Object form of such source code, you
may redistribute such embedded portions in such Object form without complying
with the conditions of Sections 4(a), 4(b) and 4(d) of the License.

> In addition, if you combine or link compiled forms of this Software with
software that is licensed under the GPLv2 ("Combined Software") and if a
court of competent jurisdiction determines that the patent provision (Section
3), the indemnity provision (Section 9) or other Section of the License
conflicts with the conditions of the GPLv2, you may retroactively and
prospectively choose to deem waived or otherwise exclude such Section(s) of
the License, but only in their entirety and only with respect to the Combined
Software.

See also http://llvm.org/foundation/relicensing/

Also might be a good time to introduce CLA: https://github.com/cla-assistant/cla-assistant

 List of current library contributors from blame: 

* [x] Victor Zverovich (vitaut)
* [x] Ivan Shynkarenka (chronoxor): https://github.com/fmtlib/fmt/pull/1075#issuecomment-471654313
* [x] Jason Shuo Zang (jasonszang): https://github.com/fmtlib/fmt/pull/1078#issuecomment-472484229
* [x] Daniela Engert (DanielaE): https://github.com/fmtlib/fmt/issues/1073#issuecomment-473688894
* [x] Elias Kosunen (eliaskosunen): https://github.com/fmtlib/fmt/issues/1073#issuecomment-473705496
* [x] Jonathan Müller (foonathan): https://github.com/fmtlib/fmt/pull/1083#issue-261871368
* [x] Brian Ricciardelli (ricco19, Brian): https://github.com/fmtlib/fmt/pull/1086#issue-262083016
* [x] Michael Winterberg (mwinterb): https://github.com/fmtlib/fmt/pull/1089#issue-263072033
* [x] Mario Werner (niosHD): https://github.com/fmtlib/fmt/issues/1073#issuecomment-475152902
* [x] Lars Gullik Bjønnes: https://github.com/fmtlib/fmt/issues/1073#issuecomment-476341407
[...]

PR #1075, opened 2019-03-11
Visual Studio 2017: warning C4456: declaration of 'result' hides prev…

chronoxor:
…ious local declaration


vitaut:
Hi @chronoxor, thanks for the PR!

Could you please review https://github.com/fmtlib/fmt/blob/ebec00138bd10e949ff34955bcb096387cd32acc/CONTRIBUTING.md, particularly the part about licensing, and let me know if you agree with it being applied both to this PR and your earlier contributions?

chronoxor:
Yes I agree. See no problems with the license.

vitaut:
Great, thanks for confirmation. Also note that the future versions may be relicensed (probably as BSD + an optional clause to drop all requirements on binary redistribution): #1073

chronoxor:
Does the PR require some other updates due to new licensing?

vitaut:
> Does the PR require some other updates due to new licensing?

Not really. Merged and thanks again for the PR!

PR #1079, opened 2019-03-15
Bcc32 compatibility

starkmapper:
I did some work to get fmt working on the "classic" Embarcadero compiler.

### DIFF
--- a/fmt/format.cc
+++ b/fmt/format.cc
@@ -83,7 +83,10 @@ static inline fmt::internal::Null<> strerror_s(char *, std::size_t, ...) {
 
 namespace fmt {
 
-FMT_FUNC internal::RuntimeError::~RuntimeError() FMT_DTOR_NOEXCEPT {}
+namespace internal {
+FMT_FUNC RuntimeError::~RuntimeError() FMT_DTOR_NOEXCEPT {}
+} // namespace internal
+
 FMT_FUNC FormatError::~FormatError() FMT_DTOR_NOEXCEPT {}
 FMT_FUNC SystemError::~SystemError() FMT_DTOR_NOEXCEPT {}
 
@@ -231,8 +234,9 @@ FMT_FUNC void SystemError::init(
   base = std::runtime_error(w.str());
 }
 
+namespace internal {
 template <typename T>
-int internal::CharTraits<char>::format_float(
+int CharTraits<char>::format_float(
     char *buffer, std::size_t size, const char *format,
     unsigned width, int precision, T value) {
   if (width == 0) {
@@ -246,7 +250,7 @@ int internal::CharTraits<char>::format_float(
 }
 
 template <typename T>
-int internal::CharTraits<wchar_t>::format_float(
+int CharTraits<wchar_t>::format_float(
     wchar_t *buffer, std::size_t size, const wchar_t *format,
     unsigned width, int precision, T value) {
   if (width == 0) {
@@ -260,7 +264,7 @@ int internal::CharTraits<wchar_t>::format_float(
 }
 
 template <typename T>
-const char internal::BasicData<T>::DIGITS[] =
+const char BasicData<T>::DIGITS[] =
     "0001020304050607080910111213141516171819"
     "2021222324252627282930313233343536373839"
     "4041424344454647484950515253545556575859"
@@ -279,12 +283,12 @@ const char internal::BasicData<T>::DIGITS[] =
   factor * 1000000000
 
 template <typename T>
-const uint32_t internal::BasicData<T>::POWERS_OF_10_32[] = {
+const uint32_t BasicData<T>::POWERS_OF_10_32[] = {
   0, FMT_POWERS_OF_10(1)
 };
 
 template <typename T>
-const uint64_t internal::BasicData<T>::POWERS_OF_10_64[] = {
+const uint64_t BasicData<T>::POWERS_OF_10_64[] = {
   0,
   FMT_POWERS_OF_10(1),
   FMT_POWERS_OF_10(ULongLong(1000000000)),
@@ -293,7 +297,7 @@ const uint64_t internal::BasicData<T>::POWERS_OF_10_64[] = {
   ULongLong(1000000000) * ULongLong(1000000000) * 10
 };
 
-FMT_FUNC void internal::report_unknown_type(char code, const char *type) {
+FMT_FUNC void report_unknown_type(char code, const char *type) {
   (void)type;
   if (std::isprint(static_cast<unsigned char>(code))) {
     FMT_THROW(FormatError(
@@ -306,7 +310,7 @@ FMT_FUNC void internal::report_unknown_type(char code, const char *type) {
 
 #if FMT_USE_WINDOWS_H
 
-FMT_FUNC internal::UTF8ToUTF16::UTF8ToUTF16(StringRef s) {
+FMT_FUNC UTF8ToUTF16::UTF8ToUTF16(StringRef s) {
   static const char ERROR_MSG[] = "cannot convert string from UTF-8 to UTF-16";
   if (s.size() > INT_MAX)
     FMT_THROW(WindowsError(ERROR_INVALID_PARAMETER, ERROR_MSG));
@@ -323,14 +327,14 @@ FMT_FUNC internal::UTF8ToUTF16::UTF8ToUTF16(StringRef s) {
   buffer_[length] = 0;
 }
 
-FMT_FUNC internal::UTF16ToUTF8::UTF16ToUTF8(WStringRef s) {
+FMT_FUNC UTF16ToUTF8::UTF16ToUTF8(WStringRef s) {
   if (int error_code = convert(s)) {
     FMT_THROW(WindowsError(error_code,
         "cannot convert string from UTF-16 to UTF-8"));
   }
 }
 
-FMT_FUNC int internal::UTF16ToUTF8::convert(WStringRef s) {
+FMT_FUNC int UTF16ToUTF8::convert(WStringRef s) {
   if (s.size() > INT_MAX)
     return ERROR_INVALID_PARAMETER;
   int s_size = static_cast<int>(s.size());
@@ -346,6 +350,7 @@ FMT_FUNC int internal::UTF16ToUTF8::convert(WStringRef s) {
   buffer_[length] = 0;
   return 0;
 }
+} // namespace internal
 
 FMT_FUNC void WindowsError::init(
     int err_code, CStringRef format_str, ArgList args) {
@@ -356,7 +361,8 @@ FMT_FUNC void WindowsError::init(
   base = std::runtime_error(w.str());
 }
 
-FMT_FUNC void internal::format_windows_error(
+namespace internal {
+FMT_FUNC void format_windows_error(
     Writer &out, int error_code, StringRef message) FMT_NOEXCEPT {
   FMT_TRY {
     MemoryBuffer<wchar_t, INLINE_BUFFER_SIZE> buffer;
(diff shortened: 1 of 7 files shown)


vitaut:
Hi @starkmapper, thanks for the PR!

Since this is your first contribution to {fmt}, please review [CONTRIBUTING.md](https://github.com/fmtlib/fmt/blob/ebec00138bd10e949ff34955bcb096387cd32acc/CONTRIBUTING.md), particularly the part about licensing, and comment here whether you agree with it being applied to your contributions including this PR.

starkmapper:
I have read contributing.md and agree that my contribution will be licensed under the fmt license.

fmt/format.cc
```diff
@@ -83,10 +83,12 @@ static inline fmt::internal::Null<> strerror_s(char *, std::size_t, ...) {
...
 namespace fmt {
 
-FMT_FUNC internal::RuntimeError::~RuntimeError() FMT_DTOR_NOEXCEPT {}
+namespace internal {
+FMT_FUNC RuntimeError::~RuntimeError() FMT_DTOR_NOEXCEPT {}
+} // namespace internal
+
 FMT_FUNC FormatError::~FormatError() FMT_DTOR_NOEXCEPT {}
 FMT_FUNC SystemError::~SystemError() FMT_DTOR_NOEXCEPT {}
-
```
vitaut: Please avoid unnecessary whitespace changes.

fmt/format.h
```diff
@@ -1331,11 +1331,11 @@ template <typename T, T> struct LConvCheck {
 // Returns the thousands separator for the current locale.
 // We check if ``lconv`` contains ``thousands_sep`` because on Android
 // ``lconv`` is stubbed as an empty struct.
-template <typename LConv>
+/*template <typename LConv>
 inline StringRef thousands_sep(
     LConv *lc, LConvCheck<char *LConv::*, &LConv::thousands_sep> = 0) {
   return lc->thousands_sep;
-}
+}*/
```
vitaut: This should be conditionally compiled instead of commented out.

fmt/format.h
```diff
@@ -2138,8 +2138,10 @@ class ArgFormatterBase : public ArgVisitor<Impl, void> {
 
   // workaround MSVC two-phase lookup issue
   typedef internal::Arg Arg;
+  typedef Char          Char_Type;
+  typedef BasicWriter<Char> Writer_Type;
```
vitaut: `Char_Type` -> `CharType`, `Writer_Type` -> `WriterType` for consistency with other names.

fmt/format.h
```diff
@@ -2432,6 +2433,19 @@ struct ArgArray<N, false/*IsPacked*/> {
   template <typename Formatter, typename T>
   static Arg make(const T &value) { return MakeArg<Formatter>(value); }
 };
+#else
+template <std::size_t N>
+struct ArgArray
+{
```
vitaut: nit: please follow the Google Style: https://google.github.io/styleguide/cppguide.html, in particular 

```c++
struct ArgArray {
```

fmt/format.cc
starkmapper: I did a sweep through all changes to resolve all whitespace issues. Should be OK now, please let me know if I missed anything.

fmt/format.h
starkmapper: That was what commit eec3fcdeabc52f85fcb58a567fae0a6d5d700123 did :-)

fmt/format.h
starkmapper: Should be fixed in 8915b2754397b5c4090267599d03cd31f1e323f0

fmt/format.h
starkmapper: I did my best to adhere to the style guide. I'm not sure if single line functions are allowed, but if so, this should be ok now :-).

starkmapper:
I didn't want to rebase+squash just yet, but let me know if that would help you.

starkmapper:
I'm also a little worried about the CI failing, is there anything I can do about that?

vitaut:
> I didn't want to rebase+squash just yet, but let me know if that would help you.

That would help.

> I'm also a little worried about the CI failing, is there anything I can do about that?

Please check the error at https://travis-ci.org/fmtlib/fmt/jobs/507890085. Looks like they are mostly namespace-related.

starkmapper:
Squashed and fixed non-windows builds (waiting for travis to confirm :-) )

starkmapper:
> I'm not sure what's going on with the appveyor build. Did I break that or is something else going on?
Tried the appveyor script locally and that seems to work.
image

fmt/format.h
```diff
@@ -2469,74 +2498,6 @@ inline uint64_t make_type(FMT_GEN15(FMT_ARG_TYPE_DEFAULT)) {
...
-  FMT_CTOR(ctor, func, arg0_type, arg1_type, 3) \
-  FMT_CTOR(ctor, func, arg0_type, arg1_type, 4) \
-  FMT_CTOR(ctor, func, arg0_type, arg1_type, 5) \
-  FMT_CTOR(ctor, func, arg0_type, arg1_type, 6) \
-  FMT_CTOR(ctor, func, arg0_type, arg1_type, 7) \
-  FMT_CTOR(ctor, func, arg0_type, arg1_type, 8) \
-  FMT_CTOR(ctor, func, arg0_type, arg1_type, 9) \
-  FMT_CTOR(ctor, func, arg0_type, arg1_type, 10)
-#endif
-
```
vitaut: Why move this block?

fmt/format.h
```diff
@@ -2593,11 +2687,8 @@ class SystemError : public internal::RuntimeError {
        throw fmt::SystemError(errno, "cannot open file '{}'", filename);
    \endrst
   */
-  SystemError(int error_code, CStringRef message) {
-    init(error_code, message, ArgList());
-  }
```
vitaut: Please keep this ctor.

fmt/format.h
```diff
@@ -3446,9 +3537,6 @@ class WindowsError : public SystemError {
      }
    \endrst
   */
-  WindowsError(int error_code, CStringRef message) {
-    init(error_code, message, ArgList());
-  }
```
vitaut: And this

fmt/format.h
starkmapper: It is generated by the FMT_VARIADIC_ macro.

fmt/format.h
starkmapper: This one is also generated by the FMT_VARIADIC_ macro.

fmt/format.h
vitaut: It's here for documentation purposes (note the apidoc above). If it is causing problems for bcc, I suggest conditionally compiling instead of removing.

fmt/format.h
starkmapper: The only thing left of this block is the `FMT_MAKE_REF` macro. I moved those to have a single `#if FMT_USE_VARIADIC_TEMPLATES` block.

fmt/format.h
starkmapper: It's not a bcc issue, it's because all `FMT_VARIADIC_*` macros now use the `FMT_VARIADIC_` macro, this normalizing variadic template handing, and less macro magic to maintain. The `FMT_VARIADIC_` macro generates functions with 0..15 templated arguments to mimic variadic templates, where the 0 argument one has the exact same signature as this manually defined one.
I propose to make it consistent with functions `printf`, `format` etc and add an overload with an extra `ArgList` parameter:
```cpp
SystemError(int error_code, CStringRef message, ArgList args) {
    init(error_code, message, args);
  }
```
This would be the best of both worlds: singe `FMT_VARIADIC_` macro and documentation (and more consistency as a bonus :-) ).

starkmapper:
I'll make the proposed changes. Would you like me to squash again, or just push updates? Also, should I check the documentation using doxygen, or apidoc like you mentioned?

starkmapper:
Adding the `..` didn't fix the appveyor build unfortunately.

starkmapper:
> 
> 
> Adding the `..` didn't fix the appveyor build unfortunately. (https://ci.appveyor.com/project/vitaut/fmt/builds/23233738)

But adding `.` did, seems that adding the source path to the commandline became required in v3.13.

fmt/format.h
starkmapper: This should be resolved with the newly added overload.

fmt/format.h
vitaut: If this is not bcc issue, could you remove the macro changes from the current PR?

vitaut:
> But adding . did, seems that adding the source path to the commandline became required in v3.13.

Thanks for fixing this.

> Would you like me to squash again, or just push updates?

Yes, please squash and remove the macro changes as discussed above.

> Also, should I check the documentation using doxygen, or apidoc like you mentioned?

I don't think it's necessary.

fmt/format.h
starkmapper: The need for the macro changes *is* because of a bcc issue. If you feel very strongly about not changing the constructor signature in the documentation I can change the `FMT_VARIADIC_CTOR` macro not to generate the conflicting overload.

### Background

The original macro did c-style array initialization:
```cpp
ArgArray<2>::Type array = { MakeValue<BasicFormatter<char>>(v0), 
                            MakeValue<BasicFormatter<char>>(v1)};
```
Where `ArgArray<2>::Type` is an array of `Value`.
Somehow bcc tried to assign the results of `MakeValue<BasicFormatter<char>>(vN)` to the first member of `Value` (`int int_value` in the anonymous union) causing a "Cannot convert `MakeValue<BasicFormatter<char>>(vN) to int" error.

When I started fixing this issue I found the `FMT_VARIADIC_` function and started using that. This had already solved this issue by generating the individual assignment statements for the array:
```cpp
ArgArray<2>::Type arr;
arr[0] = MakeValue<BasicFormatter<char>>(v0), arr[1] = MakeValue<BasicFormatter<char>>(v1);
```

note: I left out namespaces for clarity

starkmapper:
I changed the macros so the `WindowsError` and `SystemError` constructors can remain unchanged. Please let me know if there is anything else.

fmt/format.h
vitaut: I see. The macro logic is fairly complicated and version 4.x is in maintenance mode, so I'd like the changes in this PR to be minimal, just enough for bcc32 compatibility. Right now it's hard to see what actually changed because the macros have been moved around. Could you return them to their original places and make changes there?

starkmapper: I started doing that, but found that the solution would depend on the (newer and more versatile) macros defined later on in the file. That is why I moved the large macros block up. To verify that I only moved them I used kdiff3 with manual alignment. I'll post those diffs so you can get a better idea of the changes I made.

starkmapper: Here you can see what I changed  to get this working for bcc with the existing macros. Not only achieving bcc compatibility, but also removing the second set of macro's, resulting in less "magic" to maintain:
Macro-diff

And the diff for the `FMT_MAKE_REF` macro's:
Macro-diff - old macros

The alignment of the diff is shown below. I marked the manual alignment markers with red. 
Macro-diff alignment

I hope this makes it clear that I only made minor changes, apart from moving the macros.

vitaut:
Merged everything except the macro changes in https://github.com/fmtlib/fmt/commit/d2744bc848f63b222ed9bab3d93a9a3472f63eca. The macros need a bit more work because they are in their present location for a reason: to make the `system_header` pragma apply to them and not to the other parts of the code. Also it's not entirely clear to me how `FMT_VARIADIC_VAR_ARG` fixes anything.

starkmapper:
I'll try to find a less invasive way to make the macros work.

Issue #1085, opened 2019-03-18
explicit specialization of 'is_string<std::FILE *>' after instantiation


ricco19:
Example [here](https://godbolt.org/z/jKIGv3). I believe per commit c21c6b8

This error will happen when you include `format.h`, format with a `std::FILE*` handle, and later include `color.h` (ie from another translation unit). 

Specialization happens here: https://github.com/fmtlib/fmt/blob/02a6f16b524970112fd93be3f4f7e0aad01ecef5/include/fmt/color.h#L521-L526

Can these be forward declared in `core.h` before instantiation?

PR #1086, opened 2019-03-18
Forward declare is_string<FILE*> specializations

ricco19:
Not sure if this is the ideal or correct solution, but this is how I got around issue #1085 

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -618,6 +618,10 @@ struct is_string
                               decltype(to_string_view(declval<S>()))>::value> {
 };
 
+// Forward declare FILE* specialization defined in color.h
+template <> struct is_string<std::FILE*>;
+template <> struct is_string<const std::FILE*>;
+
 template <typename S> struct char_t {
   typedef decltype(to_string_view(declval<S>())) result;
   typedef typename result::char_type type;


vitaut:
Thanks for the PR. I think the root cause is that we allow anything as a character type, so a better solution would be to only provide overloads of `to_string_view` for `const char*` and `const wchar_t*` in https://github.com/fmtlib/fmt/blob/b742f622ab2fe2c2a7c7cff4939374b42c64c04b/include/fmt/core.h#L500 and remove those weird specializations in https://github.com/fmtlib/fmt/blob/b742f622ab2fe2c2a7c7cff4939374b42c64c04b/include/fmt/color.h#L525.

ricco19:
> only provide overloads of `to_string_view` for `const char*` and `const wchar_t*` 

Correct me if I'm wrong, but wouldn't you at least need `char16_t` and `char32_t` as well just to pass unit tests? (and `nullptr_t` would become ambiguous). Also, wouldn't this be a breaking change for anyone doing something with weird char types?

vitaut:
Hmm, you are right. It's not enough to handle `char` and `wchar_t`, we need some kind of `is_char` trait. Anyway, I'll merge your workaround for now to unbreak `master`. Thanks again!

PR #1087, opened 2019-03-18
Allow passing UDL format strings to the format function

marc-groundctl:


vitaut:
Thanks for the PR, but could you elaborate why you propose this? The `_format` UDL is there to allow usage like this:

```c++
auto s = "{}"_format(42);
```
as a workaround for missing compile-time arguments in C++. I'm not sure if we should reuse it for anything else.

vitaut:
As commented, I don't think we should support this.

PR #1089, opened 2019-03-20
Added missing typename to FMT_STRING.

mwinterb:
This is so that FMT_STRING can be used in a template.


mwinterb:
https://gcc.godbolt.org/z/4UqZOF Just because I was surprised it was an issue. Clang's error message is much nicer than GCC's.

vitaut:
Merged, thanks!

PR #1091, opened 2019-03-25
install pkg-config file into libdir

olafhering:
Most pkgconfig files contain a Libs: variable, which is either /usr/lib
or /usr/lib64. If a 32bit and a 64bit variant of fmt libraries is
installed, the last one wins. As a result compiling for the other
bitsize will fail.

Instead of sharedir use libdir as install target.

Fixes commit 9d0c9c4bb145a286f725cd38c90331eee7addc7f
Fixes commit 287342dab10a5a96479b3e21cd11ce671ed8125c

Signed-off-by: Olaf Hering <olaf@aepfle.de>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -206,7 +206,7 @@ if (FMT_INSTALL)
   set(FMT_INC_DIR ${CMAKE_INSTALL_INCLUDEDIR}/fmt CACHE STRING
       "Installation directory for include files, relative to ${CMAKE_INSTALL_PREFIX}.")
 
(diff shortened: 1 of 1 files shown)


olafhering:
I've reviewed CONTRIBUTING.md and I agree with the licensing. Thank you!

vitaut:
Thanks for the PR!

Is it a standard practice to install `.pc` files in the library directory?

olafhering:
Only if the .pc files references $libdir. For plain data .pc files the /usr/share/pkgconfig should be ok.

PR #1092, opened 2019-03-25
Extend basic_writer with write() method for doubles with optional format_specs argument

chronoxor:

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -2703,7 +2703,9 @@ template <typename Range> class basic_writer {
     write_int(value, s);
   }
 
-  void write(double value) { write_double(value, format_specs()); }
+  void write(double value, const format_specs& spec = format_specs()) {
+    write_double(value, spec);
+  }
 
   /**
     \rst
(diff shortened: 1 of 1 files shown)


chronoxor:
This is important for me, because I want to write doubles using basic_writer with format_specs in a same way I can write ints.

include/fmt/format.h
```diff
@@ -2703,15 +2703,19 @@ template <typename Range> class basic_writer {
     write_int(value, s);
   }
 
-  void write(double value) { write_double(value, format_specs()); }
+  void write(double value, const format_specs& spec = format_specs()) {
```
vitaut: Can `format_specs()` be replaced with `{}` or will gcc 4.4 complain?

chronoxor: Fixed. This compiles with gcc 4.4

vitaut: I meant dropping the type name altogether and just using braces as in ` = {}`, not to repeat the argument type again.

chronoxor: Sorry for misunderstanding. Unfortunately `= {}` not works for gcc 4.4 so lets currently keep the solution with `= format_specs()`

chronoxor:
Does the PR require any other fixes from me?

vitaut:
> Does the PR require any other fixes from me?

Nope, all good. Merged and thanks for the PR!

chronoxor:
Thanks a lot!

Issue #335, opened 2016-05-26
Request:  Custom argument formatting for PrintfFormatter


spacemoose:
As discussed in issue 331

I would find it very helpful to be able to specify custom argument formatting for the printf interface.

Please let me know if I can help in any way.

vitaut:
Thanks for the suggestion. Adding custom argument formatter support to `PrintfFormatter` should be very similar to `BasicFormatter` (#235). Basically the argument formatter class will have to be passed as an additional template argument. If you could contribute this feature using `BasicFormatter` as an example that would be great. Otherwise I can look into it myself but this might take a while.

spacemoose:
Okay, so I hacked out a proof of concept extension that seems to have the desired behavior.  I was a little out of my comfort zone here, which means it was rewarding work, but I'd appreciate a review and
suggestions for improvement.  I didn't have any great ideas regarding naming so please let me know your preferences there.  You can see the work at my fork of the project here: https://github.com/spacemoose/fmt

I'd like some feedback from you regarding what you'd like to see before pulling.  Here are some of my thoughts:

0. Am I on the right track? Obviously I should test more.
1. Are you using a published formatting style?  Otherwise I'll do my best.
2. I extracted a file printf_formatter.h for the modified code.  I  put the stuff which was previously in the anonymous namespace into the internal namespace.  Reasonable?
3. I'd be inclined to try to split the two interfaces into two header  files if you're amenable to that.  In my opinion format.h could be split up a little.
4.  Like with 235, I guess I should pull the ArgFormatterBase out internal
5. If I do 4, maybe we could make the naming of ArgFormatterBase more consistent with the naming of BasicFormatter?

vitaut:
> [1] Am I on the right track? Obviously I should test more.

Yes, although I haven't tested your changes they do look right.

> [2] Are you using a published formatting style?  Otherwise I'll do my best.

I'm following [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html) with two exceptions: (1) exceptions are allowed, (2) function names are lowercase.

> [3] I extracted a file printf_formatter.h for the modified code.  I  put the stuff which was previously in the anonymous namespace into the internal namespace.  Reasonable?

Makes sense. I'd only rename `printf_formatter.h` to just `printf.h`.

> [4] I'd be inclined to try to split the two interfaces into two header  files if you're amenable to that.  In my opinion format.h could be split up a little.

[...]

PR #1093, opened 2019-03-27
Restructure printf_arg_formatter to make it customizable

ToolsDevler:

I work for the Profidata AG which is using the libfmt version 4.1.0. A former colleague of mine (spacemoose) requested the expandability of the PrintfArgFormatter (now printf_arg_formatter) in the following issue: https://github.com/fmtlib/fmt/issues/335.

I have the task to update the libfmt dependency to 5.3.0 and ensure that we still have this feature available. Since I have a little bit of time pressure to finish this I had to balance the time of learning the code structure and the fixing itself. So I hope my approach in this change is ok.

Brief explanation what I did:
In order to be able to override the behaviour of the printf_arg_formatter I had to decouple the printf_arg_formatter and the basic_printf_context by creating a new class printf_arg_formatter_proxy. This new class keeps the context (so we can remove it from printf_arg_formatter) and decides either to call the basic_format_arg<Context>::handle with its context instance or a argument formatter which is passed by a template parameter. To simplify the setup I overloaded vprintf which accepts an argument formatter as template argument. I wrote a test (custom-printf-formatter-test.cc) for it.

Looking forward hearing from you.

### DIFF
--- a/include/fmt/printf.h
+++ b/include/fmt/printf.h
@@ -197,9 +197,7 @@ using internal::printf;  // For printing into memory_buffer.
 
 template <typename Range> class printf_arg_formatter;
 
-template <typename OutputIt, typename Char,
-          typename ArgFormatter =
-              printf_arg_formatter<back_insert_range<internal::buffer<Char>>>>
+template <typename OutputIt, typename Char>
 class basic_printf_context;
 
 /**
@@ -212,12 +210,12 @@ class printf_arg_formatter
     : public internal::function<
           typename internal::arg_formatter_base<Range>::iterator>,
       public internal::arg_formatter_base<Range> {
+ public:
+  typedef decltype(internal::declval<Range>().begin()) iterator;
  private:
   typedef typename Range::value_type char_type;
-  typedef decltype(internal::declval<Range>().begin()) iterator;
   typedef internal::arg_formatter_base<Range> base;
-  typedef basic_printf_context<iterator, char_type, printf_arg_formatter>
-      context_type;
+  typedef basic_printf_context<iterator, char_type> context_type;
 
   context_type& context_;
 
@@ -328,7 +326,7 @@ template <typename T> struct printf_formatter {
 };
 
 /** This template formats data and writes the output to a writer. */
-template <typename OutputIt, typename Char, typename ArgFormatter>
+template <typename OutputIt, typename Char>
 class basic_printf_context {
  public:
   /** The character type for the output. */
@@ -379,11 +377,12 @@ class basic_printf_context {
   }
 
   /** Formats stored arguments and writes the output to the range. */
+  template<typename ArgFormatter = printf_arg_formatter<back_insert_range<internal::buffer<Char>>>>
   OutputIt format();
 };
 
-template <typename OutputIt, typename Char, typename AF>
-void basic_printf_context<OutputIt, Char, AF>::parse_flags(format_specs& spec,
+template <typename OutputIt, typename Char>
+void basic_printf_context<OutputIt, Char>::parse_flags(format_specs& spec,
                                                            const Char*& it,
                                                            const Char* end) {
   for (; it != end; ++it) {
@@ -409,9 +408,9 @@ void basic_printf_context<OutputIt, Char, AF>::parse_flags(format_specs& spec,
   }
 }
 
-template <typename OutputIt, typename Char, typename AF>
-typename basic_printf_context<OutputIt, Char, AF>::format_arg
-basic_printf_context<OutputIt, Char, AF>::get_arg(unsigned arg_index) {
+template <typename OutputIt, typename Char>
+typename basic_printf_context<OutputIt, Char>::format_arg
+basic_printf_context<OutputIt, Char>::get_arg(unsigned arg_index) {
   if (arg_index == std::numeric_limits<unsigned>::max())
     arg_index = parse_ctx_.next_arg_id();
   else
@@ -419,8 +418,8 @@ basic_printf_context<OutputIt, Char, AF>::get_arg(unsigned arg_index) {
   return internal::get_arg(*this, arg_index);
 }
 
-template <typename OutputIt, typename Char, typename AF>
-unsigned basic_printf_context<OutputIt, Char, AF>::parse_header(
+template <typename OutputIt, typename Char>
+unsigned basic_printf_context<OutputIt, Char>::parse_header(
     const Char*& it, const Char* end, format_specs& spec) {
   unsigned arg_index = std::numeric_limits<unsigned>::max();
   char_type c = *it;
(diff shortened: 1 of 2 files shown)


include/fmt/printf.h
```diff
@@ -327,6 +319,37 @@ template <typename T> struct printf_formatter {
   }
 };
 
+// Redirects argument formatting visitor to argument formatters or format arg
+// handler.
+template<typename OutputIt, typename Context, typename FormatSpecs,
+         typename ArgFormatter>
+class printf_arg_formatter_proxy : public internal::function<
```
vitaut: A proxy shouldn't be needed. `ArgFormatter` should be able to deal with both `T` and `handle` case.

include/fmt/printf.h
```diff
@@ -212,14 +212,12 @@ class printf_arg_formatter
     : public internal::function<
           typename internal::arg_formatter_base<Range>::iterator>,
       public internal::arg_formatter_base<Range> {
- private:
+ public:
```
vitaut: I don't think these should be public because the children won't be able to use these types directly anyway.

include/fmt/printf.h
```diff
@@ -306,12 +304,6 @@ class printf_arg_formatter
     write_null_pointer(char_type());
     return this->out();
   }
-
-  /** Formats an argument of a custom (user-defined) type. */
-  iterator operator()(typename basic_format_arg<context_type>::handle handle) {
```
vitaut: This should remain here. Please see `arg_formatter` for an example.

test/custom-printf-formatter-test.cc
```diff
@@ -0,0 +1,57 @@
...
+  return custom_vformat(format_str, va);
+}
+
+TEST(CustomFormatterTest, Format) {
+  EXPECT_EQ("0.00", custom_format("%.2f", -.00001));
+  EXPECT_EQ("0.00", custom_format("%.2f", .00001));
+  EXPECT_EQ("1.00", custom_format("%.2f", 1.00001));
+  EXPECT_EQ("-1.00", custom_format("%.2f", -1.00001));
+}
+#endif
```
vitaut: Please move this to `printf-test`.

include/fmt/printf.h
ToolsDevler: Agreed. But I have to keep the iterator public (like in the arg_formatter) to get my approach working.

include/fmt/printf.h
ToolsDevler: I agree that this is not really a pretty and clean way. But the problem I tried to solve is, that the printf_arg_formatter did construct a basic_printf_context containing a reference to printf_arg_formatter again. When I try to extend the printf_arg_formatter the basic_printf_context would still be constructed with printf_arg_formatter as template parameter and ignore my child class (and its overloaded **operator()**).

I can also not pass my formatter child class via a template parameter to printf_arg_formatter since I would disable the possibility to specify a default parameter for the forward declaration of printf_arg_formatter (on line 198).

I'm open to implement a complete different approach.

include/fmt/printf.h
ToolsDevler: See comment at printf_arg_formatter_proxy

include/fmt/printf.h
vitaut: `basic_printf_context` should not be parameterized on `ArgFormatter`, it's just an artefact of some earlier design. Please remove it and then you'll be able to address other comments on the PR.

include/fmt/printf.h
vitaut: Right, making `iterator` public makes sense.

include/fmt/printf.h
ToolsDevler: When I understand you correctly you mean the ArgFormatter template parameter from basic_printf_context.

That one is still used for the visit_format_arg call in basic_printf_context::format at the end. This was the problem I stumbled over before I decided to make the proxy class because I couldn't come up with another alternative. But I will stick my head into it maybe I've overlooked something or I find a cleaner way.

Would it be better to pass the `ArgFormatter` as template parameter for the format `function` directly?

vitaut: > When I understand you correctly you mean the ArgFormatter template parameter from basic_printf_context.

Yes.

> Would it be better to pass the ArgFormatter as template parameter for the format function directly?

Sure.

ToolsDevler:
Sorry for the delay. I'm still struggling with a problem (looks like a bug) of the msvc 18.x compiler.

include/fmt/printf.h
```diff
@@ -197,11 +197,10 @@ using internal::printf;  // For printing into memory_buffer.
 
 template <typename Range> class printf_arg_formatter;
 
-template <typename OutputIt, typename Char,
-          typename ArgFormatter =
-              printf_arg_formatter<back_insert_range<internal::buffer<Char>>>>
+template <typename OutputIt, typename Char>
 class basic_printf_context;
 
+
```
vitaut: Please remove unnecessary whitespace changes.

include/fmt/printf.h
```diff
@@ -246,7 +245,7 @@ class printf_arg_formatter
 
   template <typename T, FMT_ENABLE_IF(std::is_integral<T>::value)>
   iterator operator()(T value) {
-    // MSVC2013 fails to compile separate overloads for bool and char_type so
+    // MSVC2013 fails to compile separate overloads for bool  and char_type so
```
vitaut: ditto

test/printf-test.cc
```diff
@@ -554,3 +554,51 @@ TEST(PrintfTest, VSPrintfMakeWArgsExample) {
...
+
+  custom_printf_arg_formatter(formatter_t::iterator iter, formatter_t::format_specs& spec, context_t& ctx)
+      : formatter_t(iter, spec, ctx) {}
+
+  using formatter_t::operator();
+
+#if FMT_MSC_VER > 0 && FMT_MSC_VER <= 1804
+      template <typename T, FMT_ENABLE_IF(std::is_floating_point<T>::value)>
+      iterator operator()(T value) {
+#else
```
vitaut: Can we use the above SFINAE-based definition for all compilers and remove conditional compilation (adding a comment clarifying that this is a workaround for MSVC).

ToolsDevler: Unfortunately this doesn't work. Other compilers (clang and gcc as far as I recall) complain then.

include/fmt/printf.h
ToolsDevler: Done

include/fmt/printf.h
ToolsDevler: Done

test/printf-test.cc
vitaut: OK, let's keep it as is.

Issue #1098, opened 2019-03-28
ignored fwrite results trigger a gcc-8.2 -Wunused-result


BillyDonahue:
We don't use cmake, so I'm doing a naive compile of a libfmt subtree in my own toolchain.
(using fmt @ e979c78)

I could easily add the -Wno-unused-result to my build, but maybe you'd want to tackle it upstream. This is gcc-8.2 on an Ubuntu 18.04 workstation, btw.
These could perhaps be mitigated with some `(void)` casts or pragmas or something.

```

 [2019/03/28 17:32:10.092] /opt/mongodbtoolchain/v3/bin/g++ \
    -o build/cached/third_party/fmt/dist/src/format.o \
    -c -Woverloaded-virtual -Wno-maybe-uninitialized -std=c++17 \
    -fno-omit-frame-pointer -fno-strict-aliasing -ggdb -pthread \
    -Wall -Wsign-compare -Wno-unknown-pragmas -Winvalid-pch -Werror \
    -O2 \
    -Wno-unused-local-typedefs -Wno-unused-function -Wno-deprecated-declarations -Wno-unused-const-variable -Wno-unused-but-set-variable -Wno-missing-braces \
    -fstack-protector-strong -fno-builtin-memcmp -fPIE \
    -DPCRE_STATIC -DNDEBUG -D_XOPEN_SOURCE=700 -D_GNU_SOURCE -D_FORTIFY_SOURCE=2 -DBOOST_SYSTEM_NO_DEPRECATED -DBOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS -DABSL_FORCE_ALIGNED_ACCESS \
    -Isrc/third_party/fmt/dist/include -Ibuild/cached -Isrc \
    src/third_party/fmt/dist/src/format.cc


 [2019/03/28 17:32:11.072] In file included from src/third_party/fmt/dist/src/format.cc:8:
 [2019/03/28 17:32:11.072] src/third_party/fmt/dist/include/fmt/format-inl.h: In function 'void fmt::v5::{anonymous}::report_error(fmt::v5::{anonymous}::FormatFunc, int, fmt::v5::string_view)':
 [2019/03/28 17:32:11.072] src/third_party/fmt/dist/include/fmt/format-inl.h:190:14: error: ignoring return value of 'size_t fwrite(const void*, size_t, size_t, FILE*)', declared with attribute warn_unused_result [-Werror=unused-result]
[...]

PR #1099, opened 2019-03-29
handle fwrite results #1098

BillyDonahue:

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -181,13 +181,21 @@ void format_error_code(internal::buffer& out, int error_code,
   assert(out.size() <= inline_buffer_size);
 }
 
+// try an fwrite, FMT_THROW on failure
+void fwrite_fully(const void *ptr, size_t size, size_t count, FILE *stream) {
+  size_t written = std::fwrite(ptr, size, count, stream);
+  if (written < count) {
+    FMT_THROW(system_error(errno, "cannot write to file"));
+  }
+}
+
 void report_error(FormatFunc func, int error_code,
                   string_view message) FMT_NOEXCEPT {
   memory_buffer full_message;
   func(full_message, error_code, message);
   // Use Writer::data instead of Writer::c_str to avoid potential memory
   // allocation.
-  std::fwrite(full_message.data(), full_message.size(), 1, stderr);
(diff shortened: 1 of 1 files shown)


include/fmt/format-inl.h
```diff
@@ -181,13 +181,21 @@ void format_error_code(internal::buffer& out, int error_code,
   assert(out.size() <= inline_buffer_size);
 }
 
+// try an fwrite, FMT_THROW on failure
+void fwrite_fully(const void *ptr, size_t size, size_t count, FILE *stream) {
```
vitaut: Please clang-format.

include/fmt/format-inl.h
```diff
@@ -181,13 +181,21 @@ void format_error_code(internal::buffer& out, int error_code,
...
+}
+
 void report_error(FormatFunc func, int error_code,
                   string_view message) FMT_NOEXCEPT {
   memory_buffer full_message;
   func(full_message, error_code, message);
   // Use Writer::data instead of Writer::c_str to avoid potential memory
   // allocation.
-  std::fwrite(full_message.data(), full_message.size(), 1, stderr);
+  fwrite_fully(full_message.data(), 1, full_message.size(), stderr);
```
vitaut: Why swap the order of arguments?

BillyDonahue: Oh, it's not important but it looked like a mistake to me. There are 3 fwrite calls.
The other two calls put the char size first.

    fwrite(s.data, 1, s.size, fp)
    fwrite(ws.data, sizeof(wchar_t), ws.size, fp)

... which match fwrite's intended semantics of having the record_size followed by the record_count.

include/fmt/format-inl.h
BillyDonahue: Done

include/fmt/format-inl.h
madscientist: I disagree.  The new way tells the system to write COUNT records of 1 byte each.  Conceptually I think it's more correct to say we want to write 1 record of COUNT bytes, which is what the old code did.  Although I'm quite positive that virtually every real implementation simply multiplies these numbers together, one could imagine an implementation where the new change would have the underlying system call write() invoked COUNT times with a size of 1 byte, which would be super inefficient, whereas the original code would invoke the underlying system call 1 time with COUNT bytes.

PR #1103, opened 2019-04-01
Add compile features for cmake 3.8+

Lectem:
This pull requests adds the minimal requirements of `fmtlib` to the CMake targets.
This is needed so that the users don't have to do it on all their target linking with `fmt::*`.
A cmake version check is added since the `cxx_std_11` compile feature is not available before CMake 3.8.

Note that without this feature required by the user, such an error can occur:

    In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/6.5.0/../../../../include/c++/6.5.0/type_traits:35:/usr/bin/../lib/gcc/x86_64-linux-gnu/6.5.0/../../../../include/c++/6.5.0/bits/c++0x_warning.h:32:2: error: This file requires compiler and library support for the ISO C++ 2011 standard. This support must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
    #error This file requires compiler and library support \

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,6 +66,8 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH}
 include(cxx14)
 include(CheckCXXCompilerFlag)
 
+set(FMT_REQUIRED_FEATURES cxx_auto_type cxx_variadic_templates)
+
 if (CMAKE_CXX_COMPILER_ID MATCHES "GNU")
   set(PEDANTIC_COMPILE_FLAGS -pedantic-errors -Wall -Wextra -pedantic
       -Wold-style-cast -Wundef
@@ -158,10 +160,7 @@ if (FMT_PEDANTIC)
   target_compile_options(fmt PRIVATE ${PEDANTIC_COMPILE_FLAGS})
 endif ()
 
-if (NOT ${CMAKE_VERSION} VERSION_LESS 3.8)
-  target_compile_features(fmt INTERFACE cxx_std_11)
-endif ()
-
+target_compile_features(fmt INTERFACE ${FMT_REQUIRED_FEATURES})
 
 target_include_directories(fmt PUBLIC
   $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -158,6 +158,11 @@ if (FMT_PEDANTIC)
   target_compile_options(fmt PRIVATE ${PEDANTIC_COMPILE_FLAGS})
 endif ()
 
+if (NOT ${CMAKE_VERSION} VERSION_LESS 3.8)
+  target_compile_features(fmt INTERFACE cxx_std_11)
```
vitaut: This is a bit too strict. {fmt} requires only a small subset of C++11 features, mostly variadic templates, traits and auto.

CMakeLists.txt
```diff
@@ -158,6 +158,11 @@ if (FMT_PEDANTIC)
...
+endif ()
+
+
```
vitaut: nit: extra newline

CMakeLists.txt
Lectem: Sadly there is no compile feature for `<type_traits>`, however there are `cxx_auto_type` and `cxx_variadic_templates`, which are even available since CMake 3.1.
So if we assume that any compiler with those 2 featutres has type_traits, I think we could remove the CMake version check, and just use those 2 compile features ?

CMakeLists.txt
Lectem: Will fix after confirmation of the other potential change

CMakeLists.txt
vitaut: Thinking more of it, what is the benefit of `target_compile_features`? If it only adds configuration-time checks, then I don't think it is needed because compilation will fail one way or another if some compiler features are not available. So adding this will only add more work at configuration time and potential for mismatch between the source and configuration.

foonathan: It adds the -std=c++11 flag.

vitaut: Ah, makes sense. What if the user wants to specify higher standard version though?

foonathan: I'm pretty sure it just works, that's the whole idea behind it (as opposed to manually setting the cxx standard variable or similar)

vitaut: > we could remove the CMake version check, and just use those 2 compile features ?

Sounds good then.

Lectem:
I did the modifications to require only `cxx_auto_type` and `cxx_variadic_templates`.
As a note, I think we could also remove the `cxx14.cmake` helper and use [Conditional Compilation Options](https://cmake.org/cmake/help/v3.1/manual/cmake-compile-features.7.html#id5) instead. (For a later PR?)

Most of the current macros have a mapping:

Supported by Cmake 3.1

* SUPPORTS_VARIADIC_TEMPLATES > `cxx_variadic_macros`
* SUPPORTS_INITIALIZER_LIST > `cxx_generalized_initializers`
* SUPPORTS_ENUM_BASE > `cxx_strong_enums` (a bit more aggressive, but enum base and strongly typed enums are from the same proposal)
* SUPPORTS_USER_DEFINED_LITERALS > `cxx_user_literals`

Unsupported by Cmake 3.1

* SUPPORTS_TYPE_TRAITS

vitaut:
Merged, thanks!

>  I think we could also remove the cxx14.cmake helper and use Conditional Compilation Options instead. (For a later PR?)

Good idea.

`SUPPORTS_VARIADIC_TEMPLATES` can just go away since variadic templates is a hard requirement now.

PR #1110, opened 2019-04-08
Visual Studio: Fix C4702: unreachable code in pedantic mode

chronoxor:

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -57,6 +57,11 @@
 #  define FMT_CUDA_VERSION 0
 #endif
 
+#if defined(_MSC_VER)
+#pragma warning(push)
+#pragma warning(disable: 4702) // C4702: unreachable code
+#endif
+
 #include "core.h"
 
 #if FMT_GCC_VERSION >= 406 || FMT_CLANG_VERSION
@@ -3721,6 +3726,10 @@ FMT_END_NAMESPACE
 // Restore warnings.
 #if FMT_GCC_VERSION >= 406 || FMT_CLANG_VERSION
 #  pragma GCC diagnostic pop
+#endif
+
+#if defined(_MSC_VER)
+#pragma warning(pop)
 #endif
 
 #endif  // FMT_FORMAT_H_


chronoxor:
fmt\include\fmt\core.h(563): warning C4702: unreachable code
fmt\include\fmt\format.h(1144): warning C4702: unreachable code
fmt\include\fmt\format.h(1646): warning C4702: unreachable code
fmt\include\fmt\format.h(2014): warning C4702: unreachable code
fmt\include\fmt\format.h(2037): warning C4702: unreachable code
fmt\include\fmt\format.h(2057): warning C4702: unreachable code
fmt\include\fmt\format.h(2059): warning C4702: unreachable code

vitaut:
Thanks for the PR. Is it possible to fix/workaround the warnings instead of suppressing them?

chronoxor:
We cannot just remove the code after handler.on_error() because the implementation can be different. But VS 2019 warns for the default implementation which does FMT_THROW.

chronoxor:
I suggest suppressing the Visual Studio warning for this case. I think this warning is false positive in this case.

vitaut:
Does this only occur on VS2019? I was trying to repro on VC2017 on godbolt, but couldn't: https://godbolt.org/z/qO1zZa

chronoxor:
Yes this issue appears only in new VS2019

vitaut:
Hmm, I wasn't able to repro on 2019 either: https://godbolt.org/z/TulfIT (at least not with /W4). What compiler options do you use?

chronoxor:
I use Level4 (/W4) & /WX to treat warnings as errors:

image
image

chronoxor:
Just tried to rebuild without suppressing of the warnings and get the following output:

image

vitaut:
OK, I managed to repro: https://godbolt.org/z/B2egiG

vitaut:
Closing in favor of a simpler workaround: https://github.com/fmtlib/fmt/commit/ccd70f59eeed94770ab75817a3a08915148767b4

chronoxor:
Thanks a lot!

PR #1121, opened 2019-04-18
fmt::ptr: support unique_ptr and shared_ptr.

sighingnow:
It is a bit noisy to add `.get()` when formatting unique_ptr and shared_ptr. This patch also convert `std::unique_ptr/std::shared_ptr` to `const void *` in `fmt::ptr`.


vitaut:
Thanks!

PR #1134, opened 2019-04-28
Fix DLL visibility of explicit instantiation "declaration" of internal::basic_data<void> in header format.h and the explicit instantiation "definition" in format.cc

denchat:
In format.h, the explicit instantiation *declaration* of internal::basic_data<void> should mirror DLL visibility of FMT_API

As the explicit instantiation *declaration* of `internal::basic_data<void>` in format.h, this explicit instantiation *definition* should mirror FMT_API also.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -189,10 +189,14 @@
 #    define FMT_API __declspec(dllexport)
 #  elif defined(FMT_SHARED)
 #    define FMT_API __declspec(dllimport)
+#    define FMT_EXTERN_TEMPLATE_API FMT_API
 #  endif
 #endif
 #ifndef FMT_API
 #  define FMT_API
+#endif
+#ifndef FMT_EXTERN_TEMPLATE_API
+#  define FMT_EXTERN_TEMPLATE_API
 #endif
 
 #ifndef FMT_ASSERT
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -199,7 +199,7 @@
 #  define FMT_EXTERN_TEMPLATE_API
 #endif
 
-#if !define(FMT_HEADER_ONLY)
+#ifndef FMT_HEADER_ONLY
 #  define FMT_EXTERN extern
 #else
 #  define FMT_EXTERN
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -199,6 +199,12 @@
 #  define FMT_EXTERN_TEMPLATE_API
 #endif
 
+#if !define(FMT_HEADER_ONLY)
+#  define FMT_EXTERN extern
+#else
+#  define FMT_EXTERN
+#endif
+
 #ifndef FMT_ASSERT
 #  define FMT_ASSERT(condition, message) assert((condition) && message)
 #endif
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -716,7 +716,7 @@ template <typename T = void> struct FMT_EXTERN_TEMPLATE_API basic_data {
   static const wchar_t WRESET_COLOR[5];
 };
 
-extern template struct basic_data<void>;
+FMT_EXTERN template struct basic_data<void>;
 
 // This is a struct rather than a typedef to avoid shadowing warnings in gcc.
 struct data : basic_data<> {};
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -169,6 +169,26 @@ FMT_END_NAMESPACE
 #  endif
 #endif
 
+#if FMT_USE_EXTERN_TEMPLATES
+#  if !defined(FMT_HEADER_ONLY) && defined(_WIN32)
+#    ifdef FMT_EXPORT
+//     When DLL exporting an "explicit class template instantiation declaration",
+//     we should not designate __declspec(dllexport) at both 
+//     - class template definition 
+//     - explicit class template instantiation declaration.
+//     
+//     (Instead, designate __declspec(dllexport) somewhere else
+//     and only at a single point where we have
+//     the "explicit class template instantiation definition.")
+#      define FMT_EXTERN_TEMPLATE_HEADER_FILE_API
+#    endif
+#  endif
+#endif
+#ifndef FMT_EXTERN_TEMPLATE_HEADER_FILE_API
+// In other cases, we can just follow FMT_API there.
+#  define FMT_EXTERN_TEMPLATE_HEADER_FILE_API FMT_API
+#endif
+
 #if FMT_HAS_GXX_CXX11 || FMT_HAS_FEATURE(cxx_trailing_return) || \
     FMT_MSC_VER >= 1600
 #  define FMT_USE_TRAILING_RETURN 1
@@ -712,7 +732,7 @@ template <typename T> struct int_traits {
 
 // Static data is placed in this class template to allow header-only
 // configuration.
-template <typename T = void> struct FMT_API basic_data {
+template <typename T = void> struct FMT_EXTERN_TEMPLATE_HEADER_FILE_API basic_data {
   static const uint64_t POWERS_OF_10_64[];
   static const uint32_t ZERO_OR_POWERS_OF_10_32[];
   static const uint64_t ZERO_OR_POWERS_OF_10_64[];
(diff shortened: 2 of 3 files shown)


include/fmt/format.h
```diff
@@ -727,7 +727,7 @@ template <typename T = void> struct FMT_API basic_data {
 };
 
 #if FMT_USE_EXTERN_TEMPLATES
-extern template struct basic_data<void>;
+extern template struct FMT_API basic_data<void>;
```
vitaut: I don't think this is right, e.g. https://godbolt.org/z/kGqqeS gives

```
<source>(6): warning C4910: 'basic_data<void>': '__declspec(dllexport)' and 'extern' are incompatible on an explicit instantiation
```

denchat: Thank you. You are right. It seems the instantiation declaration implicitly has got the same `__declspec(dllexport)` from the class template definition.

So we can resolve this at the class template definition alone.

It looks like travis-ci compilation failed by some reason. How can I make them recompile?

denchat: My problem is when I tried to compile this code on Windows
```
#include <cstdio>
#include <fmt/printf.h>

int main()
{
  fmt::printf("Hello, %s!\n", "world");
}
```
The linker can not find `internal::basic_data<void>::ZERO_OR_POWERS_OF_10_32`
```
lld-link: error: undefined symbol: __declspec(dllimport) public: static unsigned int const *const fmt::v5::internal::basic_data<void>::ZERO_OR_POWERS_OF_10_32
```

This is because we were trying to dllexport "declaration", not "definition" in `format.cc`

vitaut: I don't think `extern` is actually useful here, so I suggest just removing it together with the whole `FMT_USE_EXTERN_TEMPLATES` block.

denchat: `extern` is telling other translation units that they all could rely on the definition of `internal::basic_data<void>` in `format.cc`. So that all translation units using `internet::basic_dat<void>`, except `format.cc`, don't need to instantiate it by themselves. This makes compilation faster. This is the purpose of c++11 explicit class template declaration feature.

If `extern` in format.h is removed, so it becomes the definition itself. All translation units using `format.h` will be required to instantiate `internal::basic_data<vod>` every single time which that doesn't need to, only if using c++11 explicit class template declaration feature. 

Besides, what about the definition in format.cc? Now there would be already another definition in `format.h`, so the definition in `format.cc` must be removed altogether with `extern` then.

However, if you are affirmed that the explicit class template declaration feature would make lesser benefits than the complexity it'd gain here, I will make the definition stay in `format.h` as you suggested.

vitaut: You are right, it's better to keep `extern template`, but please remove the macro `FMT_USE_EXTERN_TEMPLATES`. We already require C++11 and this feature is widely available. Hopefully this will simplify the logic around `FMT_API` a bit.

denchat: I tried removing
- the whole FMT_USE_EXTERN_TEMPLATES block and the condition that embraces the declaration.
- the `extern` keyword (so the declaration becomes definition)
- the definition in `format.cc`

It succesfully compiles DLL but with a bunch of warning when compiling `posix.cc` as below
```
C:\Users\User\AppData\Roaming\fmt>ninja -j 2
[1/3] Building CXX object CMakeFiles\fmt.dir\src\posix.cc.obj
In file included from C:\Users\User\AppData\Roaming\fmt-master\src\posix.cc:13:
In file included from C:\Users\User\AppData\Roaming\fmt-master\include\fmt/posix.h:28:
C:\Users\User\AppData\Roaming\fmt-master\include\fmt/format.h(764,25): warning: instantiation of variable 'fmt::v5::internal::basic_data<void>::ZERO_OR_POWERS_OF_10_64' required here, but no definition is available [-Wundefined-var-template]
  return t - (n < data::ZERO_OR_POWERS_OF_10_64[t]) + 1;
                        ^
C:\Users\User\AppData\Roaming\fmt-master\include\fmt/format.h(739,25): note: forward declaration of template entity is here
  static const uint64_t ZERO_OR_POWERS_OF_10_64[];
                        ^
C:\Users\User\AppData\Roaming\fmt-master\include\fmt/format.h(764,25): note: add an explicit instantiation declaration to suppress this warning if 'fmt::v5::internal::basic_data<void>::ZERO_OR_POWERS_OF_10_64' is explicitly instantiated in another translation unit
  return t - (n < data::ZERO_OR_POWERS_OF_10_64[t]) + 1;
                        ^
C:\Users\User\AppData\Roaming\fmt-master\include\fmt/format.h(854,36): warning: instantiation of variable 'fmt::v5::internal::basic_data<void>::DIGITS' required here, but no definition is available [-Wundefined-var-template]
    std::memcpy(buffer_ + N, data::DIGITS + index * 2, 2);
                                   ^
C:\Users\User\AppData\Roaming\fmt-master\include\fmt/format.h(742,21): note: forward declaration of template entity is here
  static const char DIGITS[];
                    ^
C:\Users\User\AppData\Roaming\fmt-master\include\fmt/format.h(854,36): note: add an explicit instantiation declaration to suppress this warning if 'fmt::v5::internal::basic_data<void>::DIGITS' is explicitly instantiated in another translation unit
    std::memcpy(buffer_ + N, data::DIGITS + index * 2, 2);
                                   ^
C:\Users\User\AppData\Roaming\fmt-master\include\fmt/format.h(903,25): warning: instantiation of variable 'fmt::v5::internal::basic_data<void>::ZERO_OR_POWERS_OF_10_32' required here, but no definition is available [-Wundefined-var-template]
  return t - (n < data::ZERO_OR_POWERS_OF_10_32[t]) + 1;
                        ^
C:\Users\User\AppData\Roaming\fmt-master\include\fmt/format.h(738,25): note: forward declaration of template entity is here
  static const uint32_t ZERO_OR_POWERS_OF_10_32[];
                        ^
C:\Users\User\AppData\Roaming\fmt-master\include\fmt/format.h(903,25): note: add an explicit instantiation declaration to suppress this warning if 'fmt::v5::internal::basic_data<void>::ZERO_OR_POWERS_OF_10_32' is explicitly instantiated in another translation unit
  return t - (n < data::ZERO_OR_POWERS_OF_10_32[t]) + 1;
                        ^
3 warnings generated.
[3/3] Linking CXX shared library bin\fmt.dll
   Creating library fmt.lib and object fmt.exp
```
I don't quite understand the reason why `posix.cc` seems not to see the definition (that I already remove the `FMT_USE_EXTERN_TEMPLATES` condition as below.) and demand `extern` there.
```
note: add an explicit instantiation declaration to suppress this warning if 'fmt::v5::internal::basic_data<void>::ZERO_OR_POWERS_OF_10_64' is explicitly instantiated in another translation unit
```
```
//#if FMT_USE_EXTERN_TEMPLATES
template struct basic_data<void>;
//#endif
```

denchat: I've already removed the whole `FMT_USE_EXTERN_TEMPLATES` block.

Anyway, I guess
-  __declspec(DLL) api of the template definition having any extern declarations must have its own condition other than `FMT_API`. 
- the `extern` in front of extern template declaration itself must also have its own condition as well.

I'm not any good at naming. I only could come up with `FMT_EXTERN_TEMPLATE_API` and `FMT_EXTERN` respectively and make them added in `core.h`. Any suggestions are welcome.

vitaut:
Merged, thanks a lot!

include/fmt/core.h
```diff
@@ -189,11 +189,21 @@
 #    define FMT_API __declspec(dllexport)
```
NikitaFeodonit: It should be like this:
```
#if !defined(FMT_HEADER_ONLY) && defined(_WIN32)
#  ifdef FMT_EXPORT
#    define FMT_API __declspec(dllexport)
#    define FMT_EXTERN_TEMPLATE_API FMT_API
#  elif defined(FMT_SHARED)
#    define FMT_API __declspec(dllimport)
#    define FMT_EXTERN_TEMPLATE_API FMT_API
#  endif
#endif
```

With the addition
```
#    define FMT_EXTERN_TEMPLATE_API FMT_API
```
the MinGW shared build is fixed.

vitaut: @NikitaFeodonit, if you get an error in MinGW please open a new issue (or a pull request) with more details.

PR #1135, opened 2019-04-28
trigger assertion in to_unsigned

pauldreik:
The following assertion was triggered by fuzzing with chrono duration:
```
constexpr typename std::make_unsigned<_Tp>::type fmt::v5::internal::to_unsigned(Int) [with Int = long int; typename std::make_unsigned<_Tp>::type = long unsigned int]: Assertion `(value >= 0) && "negative value"' failed.
```

This pull request adds a unit test that triggers the same assertion.


pauldreik:
The unit test fails on CI, as intended.

PR #1139, opened 2019-05-03
Eliminate shadowed variable warnings from gcc-7.2

gsjaardema:
The gcc-7.2.0 compiler (and others) were giving shadowed variable warnings for this include file.  A simple renaming of a couple local variables eliminates the warnings.


vitaut:
Thanks!

Issue #1143, opened 2019-05-05
Compile error under MSVC due to unsupported #warning directive.


rcane:
The `#warning` directive in time.h is not recognized by MSVC and therefor produces a compile error.

We could use `#pragma message("fmt/time.h is deprecated...")` under MSVC instead to at least get the message output.

vitaut:
> We could use #pragma message("fmt/time.h is deprecated...") under MSVC instead to at least get the message output.

Seems reasonable. Could you submit a PR?

PR #1144, opened 2019-05-05
Fixed a compile error under MSVC (#1143).

rcane:
See #1143.


vitaut:
Thanks!

PR #1147, opened 2019-05-07
 Fix DLL visibility of `count_digits<4>(internal::uintptr_t)` in `format-inl.h`

denchat:
Previously, test-format.exe linking failed because `count_digits<4>(internal::uintptr_t)` not found in the DLL.
Also suppresses warning of MSVC's `_CRT_INSECURE_DEPRECATE` functions from `gtest.h` included in custom-formatter-test.cc

### DIFF
--- a/test/custom-formatter-test.cc
+++ b/test/custom-formatter-test.cc
@@ -5,10 +5,8 @@
 //
 // For the license information refer to format.h.
 
-#ifdef _MSC_VER
-#  ifndef _CRT_SECURE_NO_WARNINGS
-#    define _CRT_SECURE_NO_WARNINGS
-#  endif
+#ifndef _CRT_SECURE_NO_WARNINGS
+#  define _CRT_SECURE_NO_WARNINGS
 #endif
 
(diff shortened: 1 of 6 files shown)


vitaut:
Thanks for the PR. Since `count_digits` is the implementation detail, its test should be moved to `format-impl-test` and it shouldn't be exported.

test/custom-formatter-test.cc
```diff
@@ -5,6 +5,12 @@
 //
 // For the license information refer to format.h.
 
+#ifdef _MSC_VER
```
vitaut: Here and below: I'd drop `_MSC_VER_` altogether. There is no harm in always defining `_CRT_SECURE_NO_WARNINGS`.

denchat: Done.

Anyway I notice `format-imple-test` is exempted when building DLL. When building DLL it includes `format-inl.h` in which cout_digits resides, I wonder if `format-imple-test` should be re-enabled also?

vitaut:
Thanks!

Issue #1140, opened 2019-05-05
Compile Error VS2019


urusaiurusai:
- Error	C2993	 'fmt::v5::internal::value': illegal type for non-type template parameter '_Test'	F:\project\fmt\include\fmt\core.h	904	

- Error	C2039	 'type': is not a member of 'std'	F:\project\fmt\include\fmt\core.h	903	

I just `#include <fmt/format.h>` and do nothing.  everything works fine when building the source.

vitaut:
The error looks weird because there is no template `_Test` anywhere in the library. Perhaps you are seeing some macro conflict from elsewhere.

vitaut:
Also I couldn't repro this on godbolt: https://godbolt.org/z/jHsYkN

urusaiurusai:
The errors disapeared if directly append these source code `format.h`, `core.h`, `format-inl.h` and `format.cc` to a helloworld project instead of using the compiled library `fmtd.lib` and `#include <fmt/format.h>`.

There must be something wrong but i can't figure it out.  Anyway, I am happy to see it works now.

Thanks.

PR #1150, opened 2019-05-08
Fix compilation error under MSVC 19.21 (#1140)

mocabe:
`I agree that my contributions are licensed under the {fmt} license, and agree to future changes to the licensing.`

See #1140


vitaut:
Thanks!

Issue #1158, opened 2019-05-12
MSVC basic_string_view<charT>::const_iterator still won't compile


denchat:
When I include `<format>`, there are errors :
```
C:\Users\User\AppData\Roaming\fmt-master\include\format(611,26): error: no viable conversion from 'const char *' to
      'std::basic_format_parse_context<char>::const_iterator' (aka '_String_view_iterator<std::char_traits<char> >')
    parse_ctx.advance_to(p);
                         ^
C:\Users\User\AppData\Roaming\fmt-master\include/fmt/format.h(2152,15): note: in instantiation of member function
      'std::detail::format_handler<std::detail::arg_formatter<fmt::v5::back_insert_range<fmt::v5::internal::buffer<char> > >, char,
      std::basic_format_context<std::back_insert_iterator<fmt::v5::internal::buffer<char> >, char> >::on_replacement_field' requested here
      handler.on_replacement_field(p);
              ^
C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Tools\MSVC\14.20.27508\include\xstring(851,7): note: candidate constructor
      (the implicit copy constructor) not viable: no known conversion from 'const char *' to 'const std::_String_view_iterator<std::char_traits<char> > &' for 1st
      argument
class _String_view_iterator { // iterator for character buffer wrapper
      ^
C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Tools\MSVC\14.20.27508\include\xstring(851,7): note: candidate constructor
      (the implicit move constructor) not viable: no known conversion from 'const char *' to 'std::_String_view_iterator<std::char_traits<char> > &&' for 1st argument
C:\Users\User\AppData\Roaming\fmt-master\include\format(153,46): note: passing argument to parameter 'it' here
    constexpr void advance_to(const_iterator it);
                                             ^
C:\Users\User\AppData\Roaming\fmt-master\include\format(618,26): error: no viable conversion from 'const char *' to
      'std::basic_format_parse_context<char>::const_iterator' (aka '_String_view_iterator<std::char_traits<char> >')
    parse_ctx.advance_to(begin);
                         ^~~~~
C:\Users\User\AppData\Roaming\fmt-master\include/fmt/format.h(2161,21): note: in instantiation of member function
      'std::detail::format_handler<std::detail::arg_formatter<fmt::v5::back_insert_range<fmt::v5::internal::buffer<char> > >, char,
[...]

PR #1159, opened 2019-05-13
Fix advance_to() and begin() using iterator

denchat:
This seems to fixes errors when include <format> in issue #1158


vitaut:
Thanks!

Please do not use `<format>` in real code - it's for testing purposes only.

PR #1163, opened 2019-05-15
Use C++ headers instead of deprecated C headers

agmt:


vitaut:
Thanks for the PR but some of the headers are used to pull in symbols not defined by the C++ standard, so strictly speaking this may not work (and will also require namespace qualification).

PR #1167, opened 2019-05-17
Include as system to suppress warnings

agmt:
Including as system header (gcc flag `-isystem`) allows non-standard extensions in pedantic mode without warnings.


vitaut:
Thanks for the PR. The library generally tries to avoid non-standard extensions and I think its headers shouldn't be included as system.

foonathan:
I mean, this would get rid of all issues and PRs of the form "fixed this annoying warning with this particular compiler configuration"...

vitaut:
Warnings are sometimes helpful, we shouldn't be suppressing them wholesale.

PR #1177, opened 2019-05-29
Add C++17 builds to travis

pauldreik:
I think it would be nice to have the existing C++17 support being proved by a CI job.


vitaut:
Thanks!

PR #1187, opened 2019-06-01
inlines count_code_points(basic_string_view<char8_t>)

cjdb:
count_code_points(basic_string_view<char8_t>) was defined in
fmt/format-inl.h, and only declared in fmt/format.h, but not defined
in src/format.cc. This commit moves the overload to fmt/format.h and
inlines it.


vitaut:
Thanks!

PR #1206, opened 2019-06-24
Fix undefined reference & and one warning

### LINKED ISSUE #1205: Undefined reference fmt::v5::format_error::~format_error() when building dll
Hi, same setup as in #1204.

I get multiple
> \<file>: (.text+0x2c): undefined reference to `fmt::v5::format_error::~format_error()'

The destructor was added in commit 1c3197b8d84884b0e16ba76b2c211ac35d2fa679 defined in `format-inl.h`, but in <file> I only `#include <fmt/format.h>`. I think it should be added there, or the class has to be exported.

HazardyKnusperkeks:

Fixes #1205

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -417,7 +417,7 @@ typedef basic_memory_buffer<char> memory_buffer;
 typedef basic_memory_buffer<wchar_t> wmemory_buffer;
 
 /** A formatting error such as invalid format string. */
-class format_error : public std::runtime_error {
+class FMT_API format_error : public std::runtime_error {
  public:
   explicit format_error(const char* message) : std::runtime_error(message) {}
   explicit format_error(const std::string& message)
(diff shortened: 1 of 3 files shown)


include/fmt/format.h
```diff
@@ -417,7 +417,7 @@ typedef basic_memory_buffer<char> memory_buffer;
 typedef basic_memory_buffer<wchar_t> wmemory_buffer;
 
 /** A formatting error such as invalid format string. */
-class format_error : public std::runtime_error {
+class FMT_API format_error : public std::runtime_error {
```
vitaut: Only the destructor needs to be marked with `FMT_API`.

include/fmt/format.h
```diff
@@ -2198,7 +2198,7 @@ class arg_formatter : public internal::arg_formatter_base<Range> {
  An error returned by an operating system or a language runtime,
  for example a file opening error.
 */
-class system_error : public std::runtime_error {
+class FMT_API system_error : public std::runtime_error {
```
vitaut: ditto

include/fmt/format.h
HazardyKnusperkeks: That was my first pull request, but then I got undefined reference to the vtable, which is not exported in that case. If you know how to export the vtable explicitly I will change it.

vitaut: Ah, OK then.

PR #1230, opened 2019-07-15
Remove const qualification in compile-time checks

jackoalan:
A regression appeared in dd8cc8b for compile-time checking of const references.

Prior to dd8cc8b, all callers of `make_args_checked` had the equivalent of `remove_const_t` and `remove_reference_t` via parameter type deduction. Now only `remove_reference_t` is present which means custom type formatter specializations will not be selected unless they have been explicitly specialized with a const type. I am assuming that this is not the intended behavior and have added `remove_const_t` in addition to `remove_reference_t`.


vitaut:
Thanks!

PR #1231, opened 2019-07-15
Opt-in macro to enforce use of compile-time format strings

jackoalan:
This adds an opt-in macro `FMT_ENFORCE_COMPILE_STRING` to ensure universal usage of compile strings. Some projects may wish to enable this policy in order to catch issues that would otherwise manifest at runtime.


vitaut:
Thanks!

PR #1235, opened 2019-07-17
Use `FMT_THROW` in fmt/color.h.

DaanDeMeyer:
This allows fmt to be compiled with `-fno-exceptions`.


vitaut:
Thanks!

PR #1238, opened 2019-07-19
Stop setting `CMAKE_CXX_STANDARD` explicitly.

### LINKED ISSUE #1237: CMAKE_CXX_STANDARD 11 is incorrect when using windows clang gnu frontend
CMake 3.15 just released with support for Clang's GNU frontend on Windows (clang with gnu like options instead of clang-cl). When using Clang with the GNU frontend as CMake's compiler, fmt fails to compile because it specifies the C++ version as C++11 while the MSVC standard library headers are already written using C++14.

To solve this issue, we could set `CMAKE_CXX_STANDARD` to C++14 when we detect that Clang's GNU frontend is used on Windows. Is this an acceptable solution?

DaanDeMeyer:
This allows the compiler default to be used which fixes #1237 without
explicitly checking for a specific compiler variant.

This fixes the first part of #1237. We let all targets be compiled with the compiler default C++ version if it is not explicitly specified. Clang's GNU frontend on Windows defaults to C++14 and as a result can parse the MSVC headers correctly.

### DIFF
--- a/support/cmake/cxx14.cmake
+++ b/support/cmake/cxx14.cmake
@@ -3,10 +3,11 @@
 include(CheckCXXSourceCompiles)
 include(CheckCXXCompilerFlag)
 
-if (NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 11)
+if (CMAKE_CXX_STANDARD)
+  message(STATUS "CXX_STANDARD: ${CMAKE_CXX_STANDARD}")
+else()
+  message(STATUS "CXX_STANDARD: Default")
 endif()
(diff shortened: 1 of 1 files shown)


vitaut:
Unfortunately this won't work with older compilers such as gcc 4.8 that default to C++98. I making an exception only for the problematic compiler in #1237 instead.

DaanDeMeyer:
The issue is possibly being dealt with by CMake itself. I think its best to wait for the next CMake bugfix release to see if that fixes the problem first.

https://gitlab.kitware.com/cmake/cmake/issues/19496

PR #1254, opened 2019-07-27
User-defined literals improvements

sylveon:

This is a few couple improvements to literals formatting API.
- It uses forwarding references in the UDL template version instead of const references.
- Simplifies the overly verbose return type of `udl_formatter`.
- Makes use of `basic_string_view` to store the format string and argument name instead of raw character pointers.
- Makes all the literal operators `constexpr`.
- ~~Adds the possibility of defining `FMT_USING_LITERALS` to allow all code at a project level to use the `_format` and `_a` operator.~~

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -3455,13 +3455,13 @@ namespace internal {
 template <typename Char, Char... CHARS> class udl_formatter {
  public:
   template <typename... Args>
-  std::basic_string<Char> operator()(const Args&... args) const {
+  std::basic_string<Char> operator()(Args&&... args) const {
     FMT_CONSTEXPR_DECL Char s[] = {CHARS..., '\0'};
     FMT_CONSTEXPR_DECL bool invalid_format =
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks a lot for the PR. All looks good but I'd rather not have a using directive in a header even as an opt-in. If users want something like that they can create a small wrapper header.

include/fmt/format.h
```diff
@@ -3455,29 +3455,28 @@ namespace internal {
...
 };
 #  else
 template <typename Char> struct udl_formatter {
-  const Char* str;
+  basic_string_view<Char> str;
 
   template <typename... Args>
-  auto operator()(Args&&... args) const
-      -> decltype(format(str, std::forward<Args>(args)...)) {
+  std::basic_string<Char> operator()(Args&&... args) const {
```
vitaut: 👍

sylveon:
Fair enough, I've rebased the branch to drop those two commits.

PR #1263, opened 2019-08-07
Add FMT_NOEXCEPT to basic_string_view constructor

gabime:
This solves clang-tidy warnings when creating static string_view objects (https://github.com/gabime/spdlog/issues/1170), by adding `FMT_NOEXCEPT` to the `basic_string_view(const Char* s)` constructor.

It seems safe because ```std::char_traits<Char>::length(s)``` never throws according to the [standard](https://en.cppreference.com/w/cpp/string/char_traits/length).

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -262,7 +262,7 @@ template <typename Char> class basic_string_view {
     the size with ``std::char_traits<Char>::length``.
     \endrst
    */
-  basic_string_view(const Char* s)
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR, but this ctor shouldn't be noexcept for consistency with [`std::basic_string_view`](https://en.cppreference.com/w/cpp/string/basic_string_view/basic_string_view) and because it has a narrow contract (passing a null pointer is a UB) - Lakos rule.

PR #1265, opened 2019-08-13
Fix documentation for importing header only target with CMake

sdebionne:

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -81,7 +81,7 @@ You can detect and use an installed version of {fmt} as follows::
 
 Setting up your target to use a header-only version of ``fmt`` is equally easy::
 
-   target_link_libraries(<your-target> PRIVATE fmt-header-only)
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR, but `fmt-header-only` is an alias to `fmt::fmt-header-only` so the change is unnecessary: https://github.com/fmtlib/fmt/blob/9f09b8eed11017ac4aba23694de5e94c78bcc381/CMakeLists.txt#L200

sdebionne:
When using fmt with `find_package()`, that is as an imported target, I think that `fmt::fmt-header-only` has to be used since `::` tells CMake that it is an imported target. I came up with this PR after trying the following from the doc:

```
find_package(fmt)
target_link_libraries(<your-target> PRIVATE fmt-header-only)
```

This fails on my setup with a link error on the line of "unable to find `fmt-header-only.so`". But no big deal...

vitaut:
> This fails on my setup with a link error on the line of "unable to find fmt-header-only.so".

Interesting. Let's merge this then.

Issue #1273, opened 2019-08-26
fmt v6.0.0 has a compiler error under Intel Compiler v19.0


sharpe5:
Firstly, thank you for your fantastic work on this library. It really is soooo much nicer than chevron hell, especially when doing something simple like changing number of significant figures for a double.

So everything is working under Visual Studio 2017 v15.9. However, when I switch to Intel Compiler v19.0, it generates compiler errors.

Intel actually supplies a free "Community Edition" of their compiler for free for Windows and Linux, so it should be possible to test without spending money to buy it:
[Free "Community Edition": Intel System Studio](https://software.intel.com/en-us/system-studio/choose-download).

If you have any troubles installing it, let me know. There is a trick to it - download as a device manufacturer, generate a license key on their website, then open "Intel Software Manager" and add the license. It will be valid for one year, and you can renew for free after that. It is also free to install under Linux. 

Compiler errors:

```
3>[snip]\include\fmt\core.h(477): error : attribute does not apply to any entity
3>  using parse_context FMT_DEPRECATED = basic_parse_context<char>;
3>                      ^
3>
3>[snip]\fmt\include\fmt\core.h(478): error : attribute does not apply to any entity
3>  using wparse_context FMT_DEPRECATED = basic_parse_context<wchar_t>;
3>                       ^
3>
3>[snip]\fmt\include\fmt\format.h(2599): error : attribute does not apply to any entity
3>  using writer FMT_DEPRECATED = internal::writer;
3>               ^
3>
3>[snip]\fmt\include\fmt\format.h(2600): error : attribute does not apply to any entity
3>  using wwriter FMT_DEPRECATED =
3>                ^
3>
3>compilation aborted for main.cpp (code 2)
3>Done building project "main.vcxproj" -- FAILED.
9>------ Build started: Project: Project, Configuration: Debug x64 ------
9>  Project -> load.exe
========== Build: 8 succeeded, 1 failed, 46 up-to-date, 0 skipped ==========
```

sharpe5:
Found a workaround. Compiler with header only and define `FMT_DEPRECATED`:

```
#define FMT_HEADER_ONLY
#define FMT_DEPRECATED // Necessary in v6.0.0 of library to avoid an Intel C++ v19.0 compiler error.
#include <fmt/core.h>
#include <fmt/format.h>

int main()
{
  std::cout << "Hello World!\n";
  fmt::print("Hello, {}!", "world");
}
```

sharpe5:
[...]

PR #1279, opened 2019-08-27
Add out-of-the-box support for compiling with Intel Compiler v19.0

sharpe5:
PR for Issue #1273.

Out of the box, the Intel Compiler v19.0 does not understand the `[[deprecated]]` attribute and generates a compiler error: 

```
1>.\DemoFmt\github\fmt\include\fmt\core.h(478): error : attribute does not apply to any entity
1>  using wparse_context FMT_DEPRECATED = basic_parse_context<wchar_t>;
1>                       ^
```

However, it does accept the `[[gnu:deprecated]]` attribute.

I have updated the following StackOverflow question to document this:
https://stackoverflow.com/questions/295120/c-mark-as-deprecated/57680802#57680802

Tested with the following language levels:

* C++17 Support (/Qstd=c++17)
* C++14 Support (/Qstd=c++14)
* C++11 Support (/Qstd=c++11)
* C11 Support (/Qstd=c11)
* C99 Support (/Qstd=c99)

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -119,7 +119,7 @@
 #endif
 
 #ifndef FMT_DEPRECATED
-#  if defined(__INTEL_COMPILER)
+#  if defined(__INTEL_COMPILER) // Tested with Intel Compiler v19.0 at all available language levels.
 #    define FMT_DEPRECATED [[gnu::deprecated]]
 #  elif (FMT_HAS_CPP_ATTRIBUTE(deprecated) && __cplusplus >= 201402L) || \
       FMT_MSC_VER >= 1900
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -119,17 +119,17 @@
 #endif
 
 #ifndef FMT_DEPRECATED
-#  if (FMT_HAS_CPP_ATTRIBUTE(deprecated) && __cplusplus >= 201402L) || \
+#  if defined(__INTEL_COMPILER)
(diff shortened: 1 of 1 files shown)


sharpe5:
To test, install the Intel Compiler v19.0, the use the following in a header-only mode:

```
#include "pch.h"

#define FMT_HEADER_ONLY 
#include <fmt/core.h> // String formatting. See: https://github.com/fmtlib/fmt
#include <fmt/format.h> // String formatting. See: https://github.com/fmtlib/fmt
#include <iostream>

int main()
{
    std::cout << "Hello World!\n";
	fmt::print("Hello, {}!", "world");
}
```

sharpe5:
To add specific support for Intel Compiler v19.00, this works:

#if (FMT_HAS_CPP_ATTRIBUTE(deprecated)) && __INTEL_COMPILER >= 1900
#    define FMT_DEPRECATED [[gnu::deprecated]]
#endif

I'm not sure how far back `[[gnu::deprecated]]` is supported by the Intel Compiler.

sharpe5:
Closing this request as further testing shows it is not entirely correct.

The Intel Compiler generally accepts `[[deprecated]]`, but it does not accept `[[deprecated]]` on certain language elements, such as this:

```
using wparse_context [[deprecated]] = basic_parse_context<wchar_t>;
```

The use of [[gnu::deprecated]] appears to be misleading, as it also accepts this:

```
using wparse_context [[gnu::deprecatedZZZZZ]] = basic_parse_context<wchar_t>;
```

PR #1286, opened 2019-08-29
Map not int enum to correct underlying_type

agmt:

This PR correctly prints values of such enums:
`
enum E{
    I = 5000000000ULL
};
`

Test update is needed because "0" may not detect an error.

### DIFF
--- a/test/format-test.cc
+++ b/test/format-test.cc
@@ -1906,9 +1906,9 @@ TEST(FormatTest, FormatterNotSpecialized) {
 }
 
 #if FMT_HAS_FEATURE(cxx_strong_enums)
-enum TestFixedEnum : short { B };
+enum TestFixedEnum : short { B = 1 };
 
-TEST(FormatTest, FixedEnum) { EXPECT_EQ("0", fmt::format("{}", B)); }
+TEST(FormatTest, FixedEnum) { EXPECT_EQ("1", fmt::format("{}", B)); }
 #endif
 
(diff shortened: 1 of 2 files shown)


test/format-test.cc
```diff
@@ -1906,9 +1906,9 @@ TEST(FormatTest, FormatterNotSpecialized) {
 }
 
 #if FMT_HAS_FEATURE(cxx_strong_enums)
-enum TestFixedEnum : short { B };
+enum TestFixedEnum : short { B = 1 };
```
vitaut: Is it actually formatted differently if we don't map to underlying type?

vitaut:
Anyway, I merged the PR and will add a regression test separately. Thanks!

test/format-test.cc
agmt: I don't understand the problem clearly but if map() returns short (which doesn't enumerated in enum `type`) 
`
FMT_ENABLE_IF(std::is_enum<T>::value &&
                          !has_formatter<T, Context>::value &&
                          !has_fallback_formatter<T, Context>::value)>
  FMT_CONSTEXPR auto map(const T& val) -> decltype(static_cast<typename std::underlying_type<T>::type>(val)) {
    return static_cast<typename std::underlying_type<T>::type>(val);
  }
`
then after that will be called map(short) with correct argument and map(int) with garbage ( =0 almost always). It means that fmt::format(short_enum_value) will return "0" regardless of the argument.

PR #1287, opened 2019-08-29
Add support for built-in int128 when available

denizevrenci:

This is an initial implementation. I want to hear opinions before trying to extend it to other platforms and adding tests for it.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -671,12 +671,12 @@ FMT_TYPE_CONSTANT(const Char*, cstring_type);
 FMT_TYPE_CONSTANT(basic_string_view<Char>, string_type);
 FMT_TYPE_CONSTANT(const void*, pointer_type);
 
-FMT_CONSTEXPR bool is_integral(type t) {
+FMT_CONSTEXPR bool is_integral_type(type t) {
   FMT_ASSERT(t != named_arg_type, "invalid argument type");
   return t > none_type && t <= last_integer_type;
 }
 
-FMT_CONSTEXPR bool is_arithmetic(type t) {
+FMT_CONSTEXPR bool is_arithmetic_type(type t) {
   FMT_ASSERT(t != named_arg_type, "invalid argument type");
   return t > none_type && t <= last_numeric_type;
 }
@@ -905,8 +905,8 @@ template <typename Context> class basic_format_arg {
 
   internal::type type() const { return type_; }
 
-  bool is_integral() const { return internal::is_integral(type_); }
-  bool is_arithmetic() const { return internal::is_arithmetic(type_); }
+  bool is_integral() const { return internal::is_integral_type(type_); }
+  bool is_arithmetic() const { return internal::is_arithmetic_type(type_); }
 };
 
 /**
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -195,6 +195,14 @@
 #  define FMT_USE_EXPERIMENTAL_STRING_VIEW
 #endif
 
+#ifdef FMT_USE_INT128
+// Do nothing.
+#elif defined(__SIZEOF_INT128__)
+#  define FMT_USE_INT128 1
+#else
+#  define FMT_USE_INT128 0
+#endif
+
 FMT_BEGIN_NAMESPACE
 
 // Implementations of enable_if_t and other types for pre-C++14 systems.
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -645,6 +645,8 @@ enum type {
   uint_type,
   long_long_type,
   ulong_long_type,
+  int128_type,
+  uint128_type,
   bool_type,
   char_type,
   last_integer_type = char_type,
@@ -671,6 +673,10 @@ FMT_TYPE_CONSTANT(int, int_type);
 FMT_TYPE_CONSTANT(unsigned, uint_type);
 FMT_TYPE_CONSTANT(long long, long_long_type);
 FMT_TYPE_CONSTANT(unsigned long long, ulong_long_type);
+#if FMT_USE_INT128
+FMT_TYPE_CONSTANT(__int128_t, int128_type);
+FMT_TYPE_CONSTANT(__uint128_t, uint128_type);
+#endif
 FMT_TYPE_CONSTANT(bool, bool_type);
 FMT_TYPE_CONSTANT(Char, char_type);
 FMT_TYPE_CONSTANT(double, double_type);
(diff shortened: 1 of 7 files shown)


include/fmt/core.h
```diff
@@ -652,6 +656,13 @@ FMT_TYPE_CONSTANT(int, int_type);
 FMT_TYPE_CONSTANT(unsigned, uint_type);
 FMT_TYPE_CONSTANT(long long, long_long_type);
 FMT_TYPE_CONSTANT(unsigned long long, ulong_long_type);
+#if defined(__GNUC__) && defined(__SIZEOF_INT128__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wpedantic"
```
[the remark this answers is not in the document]
denizevrenci: When GNU extensions are not used, GCC emits a warning wherever `__int128` used with `-pedantic`.

include/fmt/format.h
```diff
@@ -1702,6 +1798,20 @@ template <typename Range> class basic_writer {
 
 using writer = basic_writer<buffer_range<char>>;
 
+template <typename T>
+struct is_integral_s : std::is_integral<T> {};
```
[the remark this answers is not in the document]
denizevrenci: This was just a placeholder name I put temporarily because of a conflict with the `is_integral` function.

include/fmt/core.h
```diff
@@ -705,6 +723,13 @@ template <typename Context> class value {
   FMT_CONSTEXPR value(unsigned val) : uint_value(val) {}
   value(long long val) : long_long_value(val) {}
   value(unsigned long long val) : ulong_long_value(val) {}
+#if defined(__GNUC__) && defined(__SIZEOF_INT128__)
```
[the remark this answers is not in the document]
denizevrenci: Not sure how you'd do that. The other block is inside the union.

Do you mean like this?:

```cpp
    string_value<char_type> string;
    custom_value<Context> custom;
    const named_arg_base<char_type>* named_arg;
#if FMT_USE_INT128
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wpedantic"
    __int128 int128_value;
    unsigned __int128 uint128_value;
  };

  value(__int128 val) : int128_value(val) {}
  value(unsigned __int128 val) : uint128_value(val) {}
#pragma GCC diagnostic pop
#else
  };
#endif
```

denizevrenci:
Behavior of `std::is_integral` is different between libstdc++ with and without GNU extensions, and libc++. I do not want to add specializations to `std` namespace so I think it is appropriate to have a separate one in the `fmt::internal` namespace.

I noticed that `std::is_integral` is used in `printf.h` as well. I suppose at least some of those usages have to be replaced with `fmt::internal::is_integral`.

I suggest changing the existing functions' name to `is_integral_type` and `is_arithmetic_type` to reserve the names of the standard type traits classes.

Let me know your thoughts.

include/fmt/core.h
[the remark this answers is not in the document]
vitaut: I don't see gcc warning about `__int128_t`: https://godbolt.org/z/mvboT-. These pragmas add too much noise.

denizevrenci:
Okay, the warning is on `__int128`, not `__int128_t` which is what I used in this PR.
https://godbolt.org/z/WYAi6I

Looks like `__int128_t` is aliased to `__int128` and `__uint128_t` is aliased to `unsigned __int128`. Using those aliases, we can avoid warning suppression.

vitaut:
One more thing: please add a test that checks if `format` works with `int128_t` somewhere around https://github.com/fmtlib/fmt/blob/a128b5b2cb28a0a4cb063e5245903cc85c4ff967/test/format-test.cc#L1336.

include/fmt/format.h
```diff
@@ -644,11 +636,20 @@ FMT_CONSTEXPR bool is_negative(T) {
   return false;
 }
 
+#if FMT_USE_INT128
+// Smallest of uint32_t, uint64_t, unsigned __int128 that is large enough to represent all
```
vitaut: nit: This seems like a long line. Could you apply clang-format to your changes or reflow the comment to fit into 80 cols?

test/format-impl-test.cc
```diff
@@ -149,6 +149,17 @@ template <typename T> struct value_extractor {
   template <typename U> FMT_NORETURN T operator()(U) {
     throw std::runtime_error(fmt::format("invalid type {}", typeid(U).name()));
   }
+
+#ifdef __apple_build_version__
+  // Apple Clang does not define typeid for __int128.
+  FMT_NORETURN T operator()(__int128) {
```
vitaut: I suggest using `__int128_t` instead of `__int128` here too for consistency.

test/format-impl-test.cc
```diff
@@ -149,6 +149,17 @@ template <typename T> struct value_extractor {
...
+    throw std::runtime_error(fmt::format("invalid type {}", "__int128"));
+  }
+
+  FMT_NORETURN T operator()(unsigned __int128) {
```
vitaut: ditto

test/format-test.cc
```diff
@@ -1333,6 +1333,14 @@ TEST(FormatterTest, FormatBin) {
             format("{0:b}", std::numeric_limits<uint32_t>::max()));
 }
 
+#if FMT_USE_INT128
+#define INT128_MAX \
+  static_cast<__int128_t>((static_cast<__uint128_t>(1) << ((__SIZEOF_INT128__ * CHAR_BIT) - 1)) - 1)
+#define INT128_MIN (-INT128_MAX - 1)
+
+#define UINT128_MAX (~static_cast<__uint128_t>(0))
```
vitaut: I suggest replacing macros with constants.

test/format-test.cc
```diff
@@ -1923,7 +1942,14 @@ using buffer_range = fmt::buffer_range<char>;
 class mock_arg_formatter
     : public fmt::internal::arg_formatter_base<buffer_range> {
  private:
+#if FMT_USE_INT128
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wpedantic"
+  MOCK_METHOD1(call, void(__int128 value));
+#pragma GCC diagnostic pop
```
vitaut: Let's kill these pragmas too and replace `__int128` with `__int128_t`.

vitaut:
Merged, thanks!

denizevrenci:
Thank you for reviewing and accepting the PR.

PR #1290, opened 2019-08-29
Fix minor clang-tidy warnings

neheb:
using instead of typedef.

climits instead of limits.h

Added missing cast to size_t.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -698,7 +698,7 @@ FMT_API bool grisu_format(Double value, buffer<char>& buf, int precision,
       buf.push_back('0');
     } else {
       exp = -precision;
-      buf.resize(precision);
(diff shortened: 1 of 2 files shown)


include/fmt/format-inl.h
```diff
@@ -698,7 +698,7 @@ FMT_API bool grisu_format(Double value, buffer<char>& buf, int precision,
       buf.push_back('0');
     } else {
       exp = -precision;
-      buf.resize(precision);
+      buf.resize(static_cast<size_t>(precision));
```
vitaut: Please use `to_unsigned` instead of `static_cast`:

```c++
buf.resize(to_unsigned(precision));
```

neheb: Fixed

Issue #1292, opened 2019-08-30
Compile time format string fails for user defined types in 6.0.0


marack:
I'm finding that the compile time format strings via the `_format` UDL are failing for my user defined types in situations where they have previously worked.  Here's a minimal example:

```
#include "fmt/format.h"

struct foo { };·

template <>
struct fmt::formatter<foo> : fmt::formatter<int>
{
  template <typename FmtCtx>
  auto format(foo const& f, FmCtx& ctx) { return ctx.out(); }
};

using namespace fmt::literals;

int main(int argc, char const* argv[])
{
  fmt::format("{}", foo{}); // okay
  "{}"_format(foo{}); // okay

  foo f;
  fmt::format("{}", f); // okay
  "{}"_format(f); // compilation error

  return 0;
}
```

Compilation error on the final UDL format which was passed a l-value foo indicates that `fmt` has not detected the custom formatter and is trying to use the fallback formatter:

```
.../fmt-6.0.0/include/fmt/format.h:2503:7: error: use of deleted function ‘fmt::v6::internal::fallback_formatter<T, Char, Enable>::fallback_formatter() [with T = foo&; Char = char; Enable = void]’
```

Most likely it is failing the `has_formatter<T>` trait.  Since the custom formatter is found when passing a rvalue (`"{}"_format(foo{})`) perhaps there is a `std::decay` missing somewhere?

I'm using gcc 8.3.1, and fmt 6.0.0.

marack:
I've found that changing the call to `do_check_format_string<>` in the `udl_formatter` template so that it decays the `Args` parameter pack seems to resolve the problem:

From:
```
do_check_format_string<Char, error_handler, Args...>(
```

To:
```
do_check_format_string<Char, error_handler, typename std::decay<Args>::type...>(
```

This change still passes the built in unit tests.

vitaut:
Good catch, thanks! Fixed in https://github.com/fmtlib/fmt/commit/422e7b9d700c6d8ee668068e4e9dc8bfb8c724af.

PR #1293, opened 2019-08-30
Decay argument types in call to do_check_format_string

marack:

This change fixes issue #1292 which causes the UDL template based formatter to fail when passed and lvalue to a user defined type despite an appropriate custom formatter being available.


vitaut:
Thanks for the PR. `std::decay` is a bit too strong, so I used `remove_cvref_t` instead and added a regression test in https://github.com/fmtlib/fmt/commit/422e7b9d700c6d8ee668068e4e9dc8bfb8c724af.

Issue #492, opened 2017-04-03
No validation on number of arguments?


bryceschober:
There doesn't seem to be any validation on the number of arguments passed matching the number in the format, at least in my usage of the writer api. Since fmtlib was advertised as a "safe" alternative to printf(), I guess I assumed that it would implement an equivalent of the compile-time checking of `-Wformat`, which does warn about argument count mismatches. Am I doing something wrong, or just expecting too much of fmtlib?

vitaut:
Unfortunately compile-time checks are not supported although there are some tricks in #62 for printf formatting.

As for checking the arguments, the library follows Python's `str.format` conventions and reports an error when trying to access non-existent argument (or argument of a wrong type), but it's OK to pass more arguments than needed.

foonathan:
Have you considered doing something with UDLs?

Like: `"Hello, {}!"_format` which creates a `format_str<1>` object containing the format string itself and the number of arguments. Then simply another `format()` overload accepting this objects that can do `static_assert()` with `sizeof...(Args)`. With C++14 constexpr it's pretty straightforward, with 11 also easy, just a little verbose and ugly.

dean0x7d:
@foonathan Unfortunately, UDLs don't help here because there is no way to get type-system constants out of a `constexpr` function. The following seems like it should work, but it does not:

```c++
#include <type_traits>

constexpr auto operator""_size(char const*, std::size_t size) {
    return std::integral_constant<std::size_t, size>{};
    //                                         ^~~~
    // error: non-type template argument is not a constant expression
}

int main() {
    static_assert("hello"_size.value == 5, "");
}
```

The issue is that `constexpr` functions must be callable with either compiletime or runtime arguments, so no argument-dependent types can be generated. Up to and including C++17, the only way to get around this is by using macros... but it's as ugly as it sounds.

This could be solved if future standards allow overloading on the `constexpr`ness of the arguments or perhaps with static reflection (I haven't kept up with the reflection proposals, so I'm just guessing on the latter).

foonathan:
There's a UDL version that takes the chars at non-type template argument... but not for string literals, damn it, you're right.

vitaut:
Compile-time format string checks are now available: http://zverovich.net/2017/11/05/compile-time-format-strings.html

fulara:
@vitaut but that does not solve mentioned issue right?
there is still no validation whether the number of {} matches the number of specified arguments, if they are without indices?
[...]

PR #1294, opened 2019-08-30
Add compile-time check for too many arguments:

mkurdej:
* Check only for automatic argument indexing.
* Tests not included as I am not sure how compile failures should be tested (I haven't seen any death test in the repo).

All the following lines fail to compile after applying this patch:
```
    fmt::print(FMT_STRING(""), 1); // too many
    fmt::print(FMT_STRING("too many\n"), 1);
    fmt::print(FMT_STRING("too many {}\n"), 1, 2);
    fmt::print(FMT_STRING("too many {}\n"), 1, 2, 3);
    fmt::print(FMT_STRING("too few {}-{}-{}bal\n"), 1, 2);
```

Tested on GCC and clang (https://wandbox.org/permlink/f7Pf9ERBskTWfFQx).

### DIFF
--- a/include/fmt/compile.h
+++ b/include/fmt/compile.h
@@ -9,6 +9,7 @@
 #define FMT_COMPILE_H_
 
 #include <vector>
+
 #include "format.h"
 
 FMT_BEGIN_NAMESPACE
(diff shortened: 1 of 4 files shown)


mkurdej:
Tested on MSVC 2019 as well.

vitaut:
Thanks for the PR but the current behavior is intentional (see #492).

mkurdej:
Ok. Fair enough, but couldn't be an opt-in only error as you suggested yourself in the discussion for #492? Something like a macro `FMT_ENFORCE_ARGUMENT_COUNT`/`FMT_STRICT_ARGUMENT_COUNT_CHECK` or whatever the name, that would be disabled by default?

vitaut:
I think this belongs to something like a lint warning which can be implemented using fmt's format string parser. Even as an opt-in it will just add extra complexity and/or will be very limited (consider handling named arguments and user-defined parsers).

PR #1301, opened 2019-09-06
Fixed vs 2015 warning about unused variable when grisu is off

gabime:
It happens only vs2015 - and this fixes it.


vitaut:
Thanks!

PR #1312, opened 2019-09-16
fmt.pc.in: Fix for cross compilation

neheb:
These variables get set to /usr/lib and /usr/include , which is totally wrong in a cross compilation
environment.

Changed to standard pkgconfig values consistent with most other ones.


vitaut:
Thanks!

PR #1315, opened 2019-09-19
add oss fuzz badge

pauldreik:
Added a nice shiny oss fuzz badge. The link appears to be wrong, but it isn't, there have been no bugs found by oss fuzz so far :-)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -11,6 +11,10 @@
    :alt: Ask questions at StackOverflow with the tag fmt
    :target: http://stackoverflow.com/questions/tagged/fmt
 
+.. image:: https://oss-fuzz-build-logs.storage.googleapis.com/badges/libfmt.svg
+   :alt: fmt is continuously fuzzed att oss-fuzz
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks!

vitaut:
Hey @pauldreik, do you know why it says `coverage failing` but doesn't show any failures?

pauldreik:
Where does it say that? Could you provide a link?

I know these places to look for information:
 - build logs at https://oss-fuzz-build-logs.storage.googleapis.com/index.html#libfmt which work fine to watch, and fmt builds just fine.
 - the bug tracker, https://bugs.chromium.org/p/oss-fuzz/issues/list?colspec=ID%20Type%20Component%20Status%20Proj%20Reported%20Owner%20Summary&q=proj%3Dfmt&can=1 which is empty, could you please introduce some bug so we can verify that oss fuzz actually reports it :-)

And here is on the control panel at https://oss-fuzz.com/ which requires login with the account associated with the project. I find it cumbersome to use, it is not easy to browse around since one often get empty output even if "all" is selected.

vitaut:
The badge itself was yellow for a while with the text in it saying "coverage failing". But it is back to green again so probably some transient issue. Thanks for the links.

PR #1320, opened 2019-09-19
Fix shared build on Solaris

vedranmiletic:


vitaut:
Thanks!

PR #1325, opened 2019-09-24
Compile-time error on too many arguments provided

MartSlot:
Adds a compile time error if the number of arguments provided to the
format function is larger than the number of braces in the format
string.
Only works in automatic argument indexing mode.

This check deliberately only works for compile-time format string
checking, not at runtime. This is because we don't want existing code to
have unexpected runtime errors after upgrades. There are also scenarios
imaginable where either the arguments or the format string is generated
at runtime, and the ability to have less braces in the format string can
actually be a feature. At the same time, compile-time format calls are
guaranteed to be constant. In other words, having too many arguments will
always mean there's a bug in the code.

The new feature works by adding a on_end_of_string() function to
formatting handlers. This function is called after the whole format
string has been parsed, and needs to be implemented by all parsing
handlers.

----------------------

Both a first-time {fmt} contributor and a first-time user, so my knowledge here is limited. I may easily have missed some important code changes or tests. Happy to change the code wherever necessary.

### DIFF
--- a/include/fmt/compile.h
+++ b/include/fmt/compile.h
@@ -82,6 +82,8 @@ template <typename Char> struct part_counter {
     return begin;
   }
 
+  FMT_CONSTEXPR void on_end_of_string() {}
+
   FMT_CONSTEXPR void on_error(const char*) {}
 };
 
@@ -148,6 +150,8 @@ class format_string_compiler : public error_handler {
     handler_(part);
     return it;
   }
+
+  FMT_CONSTEXPR void on_end_of_string() {}
 };
 
 // Compiles a format string and invokes handler(part) for each parsed part.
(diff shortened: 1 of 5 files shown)


MartSlot:
Seems like the g++8 versions failed, and the problem is in the `test/format` file, which I didn't touch. (I compiled on devtoolset-7, so gcc-7.3, C++ 14)

From what I see in this file this deals with the integration of {fmt} in the standard library, and my newly added `on_end_of_string` function is of course not a part of this. Please advise on how to deal with this. Is there a way around this or does this sink the whole idea for my PR?

vitaut:
Thanks for the PR but as discussed in #1294 and elsewhere unused arguments are allowed intentionally.

Issue #1324, opened 2019-09-23
Issues with compiled format


dkavolis:
I've been trying out compiling format strings and run into a few issues when using custom types with examples at the end (I know it's still experimental):

* `get` function fails to compile when using multiple type arguments in `fmt::compile`
https://github.com/fmtlib/fmt/blob/758446c80d64cd03fe04bb12b028c992ed7bb063/include/fmt/compile.h#L332-L339

It should look like:
```c++
// Returns a reference to the argument at index N from [first, rest...].
template <int N, typename T, typename... Args>
const auto& get(const T& first, const Args&... rest) {
  static_assert(N < 1 + sizeof...(Args), "");
  if constexpr (N == 0)
    return first;
  else
    return get<N - 1>(rest...);
}
```

* In header-only mode:
https://github.com/fmtlib/fmt/blob/758446c80d64cd03fe04bb12b028c992ed7bb063/include/fmt/compile.h#L416
clashes with
https://github.com/fmtlib/fmt/blob/758446c80d64cd03fe04bb12b028c992ed7bb063/include/fmt/format-inl.h#L553

* Type safety can be violated quite easily:
```c++
constexpr auto f1 = fmt::compile<double>(FMT_STRING("{}\n"));
std::cout << fmt::format(f1, formattable{2});
```
Maybe the compiled format should contain template arguments that could later be used to enforce type safety at compile time but still allowing implicit conversions?

* Format specifiers don't work with custom types:
```c++
constexpr auto f2 = fmt::compile<formattable>(FMT_STRING("{:05}\n"));
std::cout << fmt::format(f2, formattable{2});
```
crashes at runtime with
```
terminate called after throwing an instance of 'fmt::v6::format_error'
  what():  format specifier requires numeric argument
```
Stack:
```
__GI_raise 0x00007fd078350ed7
__GI_abort 0x00007fd078332535
<unknown> 0x00007fd0785ab672
<unknown> 0x00007fd0785b72b6
std::terminate() 0x00007fd0785b7301
__cxa_throw 0x00007fd0785b7535
fmt::v6::internal::error_handler::on_error(char const*) [clone .cold.573] 0x0000555efd7d2400
fmt::v6::internal::numeric_specs_checker<fmt::v6::internal::error_handler>::require_numeric_argument format.h:2020
[...]

PR #1326, opened 2019-09-24
Fix errors in fmt/compile.h

dkavolis:
Fixes `get` and `struct unknown` name clash in header-only mode (see #1324). Also added a simple test for multiple types in `fmt::compile` since `get` works now.

Also, `get` should probably take universal references of arguments and forward them.


vitaut:
Thanks!

PR #1328, opened 2019-09-24
workaround VS2019 Update 3 compiler bug.

mwinterb:

Bug reported here seven weeks ago:
https://developercommunity.visualstudio.com/content/problem/675854/fmtlibs-safe-duration-cast-doesnt-compile-in-vs201.html
But even calling attention to it by saying "you can't compile a library that's the basis for a C++20 proposal" wasn't enough to get any attention.


vitaut:
Wow, that's quite a regression in MSVC. Thanks for the PR!

TheQwertiest:
PS: a comment about the bug would be nice (that is in-code comment). So that the workaround could be removed in the future, once the bug is fixed.

Issue #1322, opened 2019-09-23
`fmt::join` does not support tuples


jeremyong:
Currently, `fmt::join` appears to only function on containers/ranges that support `begin` and `end`. It would be extremely convenient if `fmt::join` used the `is_tuple_like` trait to format each element with the separator as well.

vitaut:
Sounds reasonable. Will be happy to accept a PR =).

jeremyong:
@vitaut I'm looking at this now. Can I use `if constexpr` or are you trying to maintain backwards compatibility? I'd like to avoid using SFINAE unless absolutely needed

vitaut:
SFINAE is generally preferred for portability reasons but it's OK to use `if constexpr` for new non-essential features.

jeremyong:
I see that `is_tuple_like` is defined in `ranges.h`, but `join` is defined in `format.h`. The options are to put the specialization for tuples of `join` inside `ranges.h` or to move `is_tuple_like` to `format.h`. Thoughts?

vitaut:
I suggest putting it in `ranges.h` together with `is_tuple_like`. Eventually we should move the other overload of `join` there too. It's in `format.h` mostly for historical reasons.

jeremyong:
OK I'll work on this tonight after work, thanks

PR #1330, opened 2019-09-26
#1322 tuple join

jeremyong:

This addresses the enhancement proposed in #1322 .

This is my initial 10 minute pass or so for early review (needs cleanup, documentation, more tests, etc). Note that this code only works with C++17... it's been a while since I've needed to code in the earlier standards. Feel free to comment or leave feedback, thanks.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -10,6 +10,8 @@ The {fmt} library API consists of the following parts:
   facilities and a lightweight subset of formatting functions
 * :ref:`fmt/format.h <format-api>`: the full format API providing compile-time
   format string checks, output iterator and user-defined type support
+* :ref:`fmt/ranges.h <ranges-api>`: additional formatting support for ranges
+  and tuples
 * :ref:`fmt/chrono.h <chrono-api>`: date and time formatting
 * :ref:`fmt/ostream.h <ostream-api>`: ``std::ostream`` support
 * :ref:`fmt/printf.h <printf-api>`: ``printf`` formatting
@@ -316,6 +318,31 @@ custom argument formatter class::
 
 .. doxygenclass:: fmt::arg_formatter
    :members:
+
+.. _ranges-api:
+
+Ranges and Tuple Formatting
+===========================
+
+The library also supports convenient formatting of ranges and tuples::
+
+  #include <fmt/ranges.h>
+
+  std::tuple<char, int, float> t{'a', 1, 2.0f};
+  // Prints a 1 2.0
+  fmt::print("{}", t);
+
+
+NOTE: currently, the overload of ``fmt::join`` for iterables exists in the main
+``format.h`` header, but expect this to change in the future.
+
+Using ``fmt::join``, you can separate tuple elements with a custom separator::
+
+  #include <fmt/ranges.h>
+
+  std::tuple<int, char> t = {1, 'a'};
+  // Prints: "1, a"
+  fmt::print("{}", fmt::join(t, ", "));
 
 .. _chrono-api:
 
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -283,6 +283,85 @@ struct formatter<RangeT, Char,
   }
 };
 
+template <typename Char, typename... T> struct tuple_arg_join : internal::view {
+  const std::tuple<T...>& tuple;
+  basic_string_view<Char> sep;
+
+  tuple_arg_join(const std::tuple<T...>& t, basic_string_view<Char> s) : tuple{t}, sep{s} {}
+};
+
+template <typename Char, typename... T>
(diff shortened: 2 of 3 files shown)


jeremyong:
Things I'm aware of that need to be done:

[ ] Handle other value categories within the tuples (and test them, aka pointers, references, etc).
[ ] Mark things as constexpr and noexcept as appropriate
[ ] Backport functionality or disable overload if the compiler doesn't support it

Note that I opted to use simple ADL instead of template specialization

include/fmt/format.h
```diff
@@ -3305,6 +3306,66 @@ arg_join<internal::iterator_t<const Range>, wchar_t> join(const Range& range,
...
+  }
+
+  template <typename FormatContext>
+  typename FormatContext::iterator format(const tuple_arg_join<Char, T...>& value,
+                                          FormatContext& ctx) {
+    return format(value, ctx, std::make_index_sequence<sizeof...(T)>{});
+  }
+
+  template <typename FormatContext, size_t... N>
+  typename FormatContext::iterator format(const tuple_arg_join<Char, T...>& value,
```
vitaut: This and `format_args` should be private.

CMakeLists.txt
```diff
@@ -68,6 +68,9 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH}
 include(cxx14)
 include(CheckCXXCompilerFlag)
 
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
vitaut: I hope this is only for testing =)

include/fmt/format.h
```diff
@@ -3305,6 +3306,66 @@ arg_join<internal::iterator_t<const Range>, wchar_t> join(const Range& range,
   return join(std::begin(range), std::end(range), sep);
 }
 
+template <typename Char, typename... T> struct tuple_arg_join : internal::view {
```
vitaut: Let's move all of this to `ranges.h`.

CMakeLists.txt
jeremyong: Yup :)

jeremyong:
One thing that wasn't immediately obvious to me was how to do the concatenation using a fold on C++17 enabled compilers (aka without recursion). I think this should be possible but I'm not sure whether it's worth losing the clarity of this approach (which is simpler) for some indeterminate compile time gains. I'm open to suggestions here though.

include/fmt/format.h
jeremyong: For some reason, the test suite wasn't finding the symbols I was defining there in the ranges-test.cc file. Was I just doing something silly? I can try again later.

include/fmt/format.h
jeremyong: Good catch I'll make that change before checkin

vitaut:
I think recursive solution is fine. It can even be ported to pre-C++17. AFACS the only C++17 construct is `if constexpr`.

jeremyong:
OK does the FMT_CONSTEXPR macro expand to constexpr if available?

vitaut:
`FMT_CONSTEXPR` expands to `constexpr` or `inline` (to be used with function definitions).

jeremyong:
@vitaut I believe all feedback is now addressed and I have finished my own TODOs.

jeremyong:
Last push addresses a caught unused-parameter warning on GCC and fixes a documentation build issue I noticed.

doc/api.rst
```diff
@@ -317,6 +319,31 @@ custom argument formatter class::
...
+The library also supports convenient formatting of ranges and tuples::
+
+  #include <fmt/ranges.h>
+
+  std::tuple<char, int, float> t{'a', 1, 2.0f};
+  // Prints a 1 2.0
+  fmt::print("{}", t);
+
+
+NOTE: currently, the overload of `fmt::join` for iterables exists in the main
```
vitaut: Inline code is double-tick delimited in RST, e.g. ``fmt::join``.

vitaut:
Looks great! Just one minor comment regarding RST and please run `clang-format` to format your changes.

doc/api.rst
jeremyong: Ah ok I'll fix that thanks

jeremyong:
I use the clang-format extension which runs on save so we should be good there

jeremyong:
OK done. The change was squashed into the single commit as before.

vitaut:
Merged with minor tweaks in https://github.com/fmtlib/fmt/commit/b4f1988c4b1aca7e15ce32d74d3ec47180022f21. Thanks!

jeremyong:
Thanks!

PR #1331, opened 2019-09-26
Compiled format fixes

dkavolis:
* Added more meaningful names for compiled format SFINAE rather than repeating `FMT_ENABLE_IF(std::is_base_of<internal::basic_compiled_format, CompiledFormat>::value)` everywhere
* Fixed `fmt::format` and `fmt::format_to` ambigous overloads with compiled format
* Added `fmt::print` overloads for compiled format with tests

### DIFF
--- a/include/fmt/compile.h
+++ b/include/fmt/compile.h
@@ -603,6 +603,26 @@ std::size_t formatted_size(const CompiledFormat& cf, const Args&... args) {
       .count();
 }
 
+template <typename CompiledFormat, typename... Args,
+          FMT_ENABLE_IF(internal::is_compiled_format<CompiledFormat>::value)>
+inline void print(std::FILE* f, const CompiledFormat& cf, Args&&... args) {
+  using Char = char_t<CompiledFormat>;
+  using context = buffer_context<Char>;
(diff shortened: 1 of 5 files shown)


include/fmt/compile.h
```diff
@@ -568,14 +594,35 @@ format_to_n_result<OutputIt> format_to_n(OutputIt out, size_t n,
...
 std::size_t formatted_size(const CompiledFormat& cf, const Args&... args) {
   return format_to(
              internal::counting_iterator<typename CompiledFormat::char_type>(),
              cf, args...)
       .count();
 }
 
+template <typename CompiledFormat, typename... Args,
+          FMT_ENABLE_IF(internal::is_compiled_format<CompiledFormat>::value)>
+inline void print(std::FILE* f, const CompiledFormat& cf, Args&&... args) {
```
vitaut: Please don't add `print` here and below, because duplicating formatting functions won't scale. In the future we'll make the original formatting functions work with compiled strings.

dkavolis: It's probably better to handle this in a separate PR.

PR #1334, opened 2019-09-30
Remove no-op destructor definitions to opt-in for compiler generated ones

denizevrenci:

Use rule of 0 and avoid `-Wdeprecated` warnings when the exception objects are copied.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -223,9 +223,6 @@ FMT_FUNC Char internal::decimal_point_impl(locale_ref) {
 }
 #endif
 
-FMT_API FMT_FUNC format_error::~format_error() FMT_NOEXCEPT {}
-FMT_API FMT_FUNC system_error::~system_error() FMT_NOEXCEPT {}
-
 FMT_FUNC void system_error::init(int err_code, string_view format_str,
                                  format_args args) {
   error_code_ = err_code;
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -630,7 +630,6 @@ class FMT_API format_error : public std::runtime_error {
   explicit format_error(const char* message) : std::runtime_error(message) {}
   explicit format_error(const std::string& message)
       : std::runtime_error(message) {}
-  ~format_error() FMT_NOEXCEPT;
 };
 
 namespace internal {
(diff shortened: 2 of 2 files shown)


vitaut:
Thanks for the PR but the destructors are out-of-line intentionally to avoid  `weak-vtables` warnings.

denizevrenci:
How about providing defaulted special members?
Also, I suggest adding `-Wweak-vtables` to CI builds to explicitly check the code since it is not turned on by default or with `-Wall` or `-Wextra`. I also suggest adding `-Wdeprecated` to not rely on deprecated behavior with current code. Let me know if you would accept PR(s) about these three suggestions.

vitaut:
> How about providing defaulted special members?

I don't think it's necessary since they won't be different from compiler-generated, just more syntax noise.

> I suggest adding -Wweak-vtables to CI builds to explicitly check the code since it is not turned on by default or with -Wall or -Wextra. I also suggest adding -Wdeprecated to not rely on deprecated behavior with current code.

Good idea. I'll be happy to accept a PR but please make sure that the new warnings are only added in `FMT_PEDANTIC` is set in fmt's CMake config.

Thanks.

denizevrenci:
> I don't think it's necessary since they won't be different from compiler-generated, just more syntax noise.

The issue with that is compiler-generated copy constructor/assignment is deprecated behaviour (which is why clang's `-Wdeprecated` fires when they are copied.) and move constructor/assignment isn't generated at all.
https://foonathan.net/2019/02/special-member-functions/

Furthermore, here is the core guideline about this topic:
http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c21-if-you-define-or-delete-any-default-operation-define-or-delete-them-all

vitaut:
> The issue with that is compiler-generated copy constructor/assignment is deprecated behaviour (which is why clang's -Wdeprecated fires when objects of this class are copied.)

Could you give a warning example in godbolt?

denizevrenci:
https://godbolt.org/z/CIaoVH

vitaut:
That's a weird warning, but fortunately it doesn't seem to affect `format_error`: https://godbolt.org/z/euNxl6

vitaut:
Ah, it seems godbolt is swallowing the warning. I think it's worth adding defaulted copy ctor to silence the warning.

PR #1337, opened 2019-10-01
Repoint one more Python 2 link to Python 3

jackoalan:
One more python reference, this one acknowledging the `str.format` function.

A grep of the codebase for `python.org` shows no remaining Python 2 references.


vitaut:
Thanks!

PR #396, opened 2016-10-09
Thread-safe time formatting

codicodi:
This adds thread-safe (at least on platforms that provide necessary extensions) replacement functions for std::localtime and std::gmtime. Alternatively they could be placed in a new source file time.cc, but time.h seems so empty right now...

PR #1341, opened 2019-10-04
Make localtime() aware of timezone changes

layus:
According to POSIX.1-2004, localtime() is required to behave as though
tzset(3) was called, while localtime_r() does not have this requirement.
For portable code tzset(3) should be called before localtime_r().

See https://stackoverflow.com/questions/19170721/real-time-awareness-of-timezone-change-in-localtime-vs-localtime-r#answer-19170798 for more.

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -47,6 +47,7 @@ inline std::tm localtime(std::time_t time) {
     dispatcher(std::time_t t) : time_(t) {}
 
     bool run() {
+      tzset();
       using namespace fmt::internal;
       return handle(localtime_r(&time_, &tm_));
     }


layus:
Not too sure if the patch is valid for windows, and whether it is needed.

vitaut:
Thanks for the PR but the intent here is to have the `localtime_r` semantics. `localtime` is only used as a fallback on systems that don't have `localtime_r`.

layus:
Hi, thanks for you response. Please note that the intent is totally unclear from the only specification I could get:

> Thread-safe time formatting #396
> 
> This adds thread-safe (at least on platforms that provide necessary extensions) replacement functions for std::localtime and std::gmtime. Alternatively they could be placed in a new source file time.cc, but time.h seems so empty right now...

If it is intended to replace std::localtime, it should have the same semantics. Why not call it localtime_r instead ? Note that the c++ std::localtime has exactly the same semantics as the c localtime: it calls tzset() (or works just as if it had been called) as per https://en.cppreference.com/w/cpp/chrono/c/localtime#Notes

vitaut:
I agree that it could be clearer. A PR clarifying the behavior in a comment is welcome. I don't think it's worth renaming the function to follow weird C naming convention (it would also be a breaking change).

PR #1342, opened 2019-10-05
Fix handling of types convertible to std::string_view

vitaut:

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -828,6 +828,14 @@ template <typename Context> struct arg_mapper {
   FMT_CONSTEXPR basic_string_view<char_type> map(const T& val) {
     return basic_string_view<char_type>(val);
   }
+  template <typename T,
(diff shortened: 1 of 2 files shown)


denizevrenci:
Should there also be a guard to check that `std::string_view` cannot be constructed from the type at this overload?

```cpp
  template <typename T,
            FMT_ENABLE_IF(!is_string<T>::value && !is_char<T>::value &&
                          !std::is_constructible<basic_string_view<char_type>,
                                                 T>::value &&
                          (has_formatter<T, Context>::value ||
                           has_fallback_formatter<T, Context>::value))>
  FMT_CONSTEXPR const T& map(const T& val) {
    return val;
  }
```

As it stands, a strong type that wraps a `std::string_view` has ambiguous overloads between the newly added overload and the one above.

vitaut:
@denizevrenci probably yes.

PR #1347, opened 2019-10-07
 Add defaulted copy and move operations to format_error and system_error 

denizevrenci:
Another approach to not trigger `-Wdeprecated` and `-Wweak-vtables` after the discussion in #1334.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -98,7 +98,7 @@ endif ()
 
 if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   set(PEDANTIC_COMPILE_FLAGS -Wall -Wextra -pedantic -Wconversion
-      -Wno-sign-conversion)
+      -Wno-sign-conversion -Wweak-vtables)
   check_cxx_compiler_flag(-Wzero-as-null-pointer-constant HAS_NULLPTR_WARNING)
   if (HAS_NULLPTR_WARNING)
     set(PEDANTIC_COMPILE_FLAGS ${PEDANTIC_COMPILE_FLAGS}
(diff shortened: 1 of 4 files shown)


denizevrenci:
Looks like there are other instances of weak vtables in the test code. It could be fixed with some boilerplate but I am not sure if it is worth it.

denizevrenci-quoine:
Yes. They are caused by `-Wweak-vtables` firing in testing code. I will fix them or find a workaround.

denizevrenci:
I noticed it is possible to suppress the warnings and reduce the sizes of the object files by adding an undefined private virtual member function as follows:
```cpp
private:
  virtual avoid_weak_vtable();
```

This can be used in both test code and the two exceptions that I added the defaulted special members to. It avoids the boilerplate associated with adding defaulted members for rule of 5.

vitaut:
> I noticed it is possible to suppress the warnings and reduce the sizes of the object files by adding an undefined private virtual member function as follows

Great find! I think it's a nicer solution than adding defaulted members. Do you know why it results i smaller object files?

vitaut:
Thinking more of it, not emitting `format_error`'s vtable may be problematic.

denizevrenci:
> Great find! I think it's a nicer solution than adding defaulted members. Do you know why it results in smaller object files?
I noticed later that it needs the definition of the virtual function even though it is never used.

So, this "solution" with the definitions included created `core-test.cc.o` 23K bytes bigger than that of master.

vitaut:
The test sizes don't matter but the library size does. There were requests to make it smaller and we'll do it at some point. Also adding an unused function is a bit of a hack. I suggest going to the previous version of the PR with defaulted members.

vitaut:
Thanks!

PR #1349, opened 2019-10-08
Fix undefined in format-test

### LINKED ISSUE #1344: MemoryBufferTest.MoveCtor provokes undefined sanitizer
When [`MoveCtor` performs `check_move_buffer`](https://github.com/fmtlib/fmt/blob/b6011453/test/format-test.cc#L283), the buffer allocator becomes null [as expected](https://github.com/fmtlib/fmt/blob/b6011453/test/format-test.cc#L270), but then `MoveCtor` attempts to use it to [allocate a dynamic buffer](https://github.com/fmtlib/fmt/blob/b6011453/test/format-test.cc#L287). UBSAN reports this as a reference binding to a null pointer at [TestAllocator.allocate](https://github.com/fmtlib/fmt/blob/b6011453/test/mock-allocator.h#L55). I think that this succeeds nevertheless because a typical `std::allocator<char>::allocate` does not use `this`, so it does not crash when `this` is null.

orivej:
When `MoveCtor` performs `check_move_buffer`, the buffer allocator becomes null, but then `MoveCtor` attempts to use it to allocate a dynamic buffer. This succeeds nevertheless because a typical `std::allocator<char>::allocate` does not use `this`, so it does not crash when `this` is null.

Fixes #1344

### DIFF
--- a/test/format-test.cc
+++ b/test/format-test.cc
@@ -271,7 +271,7 @@ static void check_move_buffer(
   EXPECT_EQ(alloc, buffer2.get_allocator().get());
 }
 
-TEST(MemoryBufferTest, MoveCtor) {
+TEST(MemoryBufferTest, MoveCtorInlineBuffer) {
   std::allocator<char> alloc;
   basic_memory_buffer<char, 5, TestAllocator> buffer((TestAllocator(&alloc)));
   const char test[] = "test";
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks!

Issue #1348, opened 2019-10-08
Is fmt::print thread-safe?


tankiJong:
Hi, 

I was testing the code in multiple threads and I observed different behavior for `fmt::print`:
With code
```
      fmt::print( fmt::fg( fmt::color::blue ), "worker init\n" );
```
I got:
![image](https://user-images.githubusercontent.com/8042676/66368411-2d73a380-e966-11e9-9736-96e92a0183f4.png)

With code
```
      fmt::print( fmt::fg( fmt::color::blue ), "worker {} init\n", mId );
```
I got:
![image](https://user-images.githubusercontent.com/8042676/66368472-5f850580-e966-11e9-9c40-c033f3d598c6.png)

Notice that items were missing color. And the only difference in code is just two lines above.

So I am wondering is that because the `print` is not thread-safe? Or actually some other issues going on.

Thanks for the help :)

vitaut:
It is thread-safe in the sense that there are no data races but the output is not atomic meaning that text style changes and the text from different threads may interleave. It can be made atomic by placing both escape sequences that control text style and the formatted text into memory_buffer and then writing them to file in one go. PRs are welcome =).

PR #1351, opened 2019-10-09
use memory_buffer to make color print behave atomic #1348

tankiJong:

I did not have time to read the whole coding standard, but I did try to following the existing style.
Please do point out if there is any formating issue.

Thanks for your time :)

### DIFF
--- a/include/fmt/color.h
+++ b/include/fmt/color.h
@@ -504,24 +504,32 @@ template <typename S, typename Char = char_t<S> >
 void vprint(std::FILE* f, const text_style& ts, const S& format,
             basic_format_args<buffer_context<Char> > args) {
   bool has_style = false;
+  basic_memory_buffer<Char> buf;
   if (ts.has_emphasis()) {
     has_style = true;
-    internal::fputs<Char>(internal::make_emphasis<Char>(ts.get_emphasis()), f);
+    auto emphasis = internal::make_emphasis<Char>(ts.get_emphasis());
+    buf.append(emphasis.begin(), emphasis.end());
   }
   if (ts.has_foreground()) {
     has_style = true;
-    internal::fputs<Char>(
-        internal::make_foreground_color<Char>(ts.get_foreground()), f);
+    auto foreground = internal::make_foreground_color<Char>(ts.get_foreground());
+    buf.append(foreground.begin(), foreground.end());
   }
   if (ts.has_background()) {
     has_style = true;
-    internal::fputs<Char>(
-        internal::make_background_color<Char>(ts.get_background()), f);
+    auto background = internal::make_background_color<Char>(ts.get_background());
+    buf.append(background.begin(), background.end());
   }
-  vprint(f, format, args);
+
+  format_to( buf, "{}", format );
+
   if (has_style) {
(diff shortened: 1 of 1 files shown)


tankiJong:
I do not feel `format_to( buf, "{}", format );` is the right way to put it..? 

I try to format it and `vprint` the whole complete string but the complier complains when I type `vprintf(f, buf.data());`. Please inspire me the proper way to do it. 

Thanks!

include/fmt/color.h
```diff
@@ -504,24 +504,31 @@ template <typename S, typename Char = char_t<S> >
...
   if (ts.has_background()) {
     has_style = true;
-    internal::fputs<Char>(
-        internal::make_background_color<Char>(ts.get_background()), f);
+    auto background = internal::make_background_color<Char>(ts.get_background());
+    buf.append(background.begin(), background.end());
   }
-  vprint(f, format, args);
+
+  format_to( buf, "{}", format );
```
vitaut: This should be something like `vformat_to(buf, format, args)`.

include/fmt/color.h
```diff
@@ -504,24 +504,31 @@ template <typename S, typename Char = char_t<S> >
...
+
   if (has_style) {
-    internal::reset_color<Char>(f);
+    reset_color( buf );
   }
+
```
vitaut: Please apply clang-format and no need to add extra empty lines =).

include/fmt/color.h
```diff
@@ -504,24 +504,31 @@ template <typename S, typename Char = char_t<S> >
...
+
+  vprint(f, buf.data(), args);
```
vitaut: And this `internal::fputs<Char>(...)`.

tankiJong:
Thanks for the feedback. 
I will ping you again when I feel it's in shape and rebase the change into a single commit when you think it looks good:)

include/fmt/color.h
```diff
@@ -504,24 +504,30 @@ template <typename S, typename Char = char_t<S> >
...
+    buf.append(background.begin(), background.end());
   }
-  vprint(f, format, args);
+  vformat_to(buf, format, args);
   if (has_style) {
-    internal::reset_color<Char>(f);
+    internal::reset_color<Char>(buf);
   }
+  buf.push_back(Char(0));
+  internal::fputs(buf.data(), f);
```
orivej: Although it is unusual, the result of formatting args may contain zero bytes, then this fputs will truncate it and will not reset the color.

tankiJong: Hi, thanks for bring that up! But I don't think I understand what you mean... Could you give an example?

tankiJong: Hi, thanks for bring that up! But I don't think I understand what you mean... Could you give an example?

tankiJong: Hi, thanks for bring that up! But I don't think I understand what you mean... Could you give an example?

orivej: See the test in https://github.com/fmtlib/fmt/pull/1356/files

tankiJong: Ah.. Yes, that's a good point, in case there is a terminator in the string, that won't work.

PR #1355, opened 2019-10-10
Fix color stream print example

orivej:


vitaut:
Thanks for the PR but the example looks correct.

orivej:
Please take a look at the signature of the function. This one takes a `FILE*` as the first argument. The one without `FILE*` is defined later.

vitaut:
Ah, OK. I think we should just remove this example because it's basically a copy paste of the one below and doesn't add anything useful.

PR #1356, opened 2019-10-10
Support color printing strings with zero bytes

orivej:
This is a clean up after #1351.

### DIFF
--- a/include/fmt/color.h
+++ b/include/fmt/color.h
@@ -503,8 +503,7 @@ void vprint(std::FILE* f, const text_style& ts, const S& format,
             basic_format_args<buffer_context<Char> > args) {
   basic_memory_buffer<Char> buf;
   internal::vformat_to(buf, ts, to_string_view(format), args);
-  buf.push_back(Char(0));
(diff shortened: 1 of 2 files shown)


include/fmt/color.h
```diff
@@ -503,8 +503,7 @@ void vprint(std::FILE* f, const text_style& ts, const S& format,
             basic_format_args<buffer_context<Char> > args) {
   basic_memory_buffer<Char> buf;
   internal::vformat_to(buf, ts, to_string_view(format), args);
-  buf.push_back(Char(0));
-  internal::fputs(buf.data(), f);
+  std::fwrite(buf.data(), 1, buf.size(), f);
```
vitaut: To handle `wchar_t` correctly on Windows this should be `fputws` if `Char` is `wchar_t` (#1243). This will truncate output on NUL chars but I don't think we should worry about this. Also use `fwrite_fully` for `char` to handle I/O errors.

vitaut:
Closing because, as commented earlier, I think we should continue using `fputws` for `wchar_t`.

PR #1357, opened 2019-10-10
Inline internal single-use fputs template

orivej:
Depends on #1356

### DIFF
--- a/include/fmt/color.h
+++ b/include/fmt/color.h
@@ -444,22 +444,12 @@ FMT_CONSTEXPR ansi_color_escape<Char> make_emphasis(emphasis em) FMT_NOEXCEPT {
   return ansi_color_escape<Char>(em);
 }
 
-template <typename Char>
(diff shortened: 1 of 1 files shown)


include/fmt/color.h
```diff
@@ -444,22 +444,12 @@ FMT_CONSTEXPR ansi_color_escape<Char> make_emphasis(emphasis em) FMT_NOEXCEPT {
...
 }
 
-template <typename Char>
-inline void fputs(const Char* chars, FILE* stream) FMT_NOEXCEPT {
-  std::fputs(chars, stream);
-}
-
-template <>
-inline void fputs<wchar_t>(const wchar_t* chars, FILE* stream) FMT_NOEXCEPT {
-  std::fputws(chars, stream);
```
vitaut: As commented on the other PR, I think we should keep `fputws`. It is still possible to use `fwrite` in the other overload (this will probably require changing the signature to take `basic_string_view` to allow embedded NULs).

PR #1360, opened 2019-10-12
Distinguish float type formatting from double

### LINKED ISSUE #1353: Cannot override float processing
Tried to override format for standard types according to documentation:

`
using arg_formatter = fmt::arg_formatter<fmt::back_insert_range<fmt::internal::buffer>>;

class na_arg_formatter : public arg_formatter {
public:
	na_arg_formatter(fmt::format_context &ctx, fmt::format_specs *spec = nullptr)
		: arg_formatter(ctx, spec),
		m_na_formatter(ctx)
	{}

	using arg_formatter::operator();

	auto operator()(double value) {
		if (IsNA(value))
			return m_na_formatter("NA");
		return arg_formatter::operator()(value);
	}

	auto operator()(float value) {
		if (IsNA(value))
			return m_na_formatter("NA");
		return arg_formatter::operator()(value);
	}

	auto operator()(unsigned value) {
		if (IsNA(value))
			return m_na_formatter("NA");
		return arg_formatter::operator()(value);
	}

	auto operator()(int value) {
		if (IsNA(value))
			return m_na_formatter("NA");
		return arg_formatter::operator()(value);
	}

private:
	arg_formatter m_na_formatter;
};

inline std::string na_vformat(fmt::string_view format_str, fmt::format_args args) {
	fmt::memory_buffer buffer;
	// Pass custom argument formatter as a template arg to vformat_to.
	fmt::vformat_to<na_arg_formatter>(buffer, format_str, args);
	return fmt::to_string(buffer);
}

template <typename ...Args>
inline std::string na_format(
	fmt::string_view format_str, const Args &... args) {
	return na_vformat(format_str, fmt::make_format_args(args...));
}
`

Float overloading is not working.
Double case is chosen.
`
na_format("{}", 4.f);
`

orivej:
This fixes #1353 and is the first half of the fix for #1336.

I suggest reviewing commits one by one.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -980,6 +980,13 @@ FMT_API bool grisu_format(Double value, buffer<char>& buf, int precision,
   return true;
 }
 
+template <>
+char* sprintf_format<float>(float value, internal::buffer<char>& buf,
+                            sprintf_specs specs) {
+  // printf does not have a float format modifier, it only supports double.
+  return sprintf_format<double>(value, buf, specs);
+}
+
 template <typename Double>
 char* sprintf_format(Double value, internal::buffer<char>& buf,
                      sprintf_specs specs) {
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -889,7 +889,11 @@ using mapped_type_constant =
                   typename Context::char_type>;
 
 // Maximum number of arguments with packed types.
-enum { max_packed_args = 12 };
+enum {
+  packed_arg_bitsize = 5,
+  packed_arg_mask = (1 << packed_arg_bitsize) - 1,
+  max_packed_args = 63 / packed_arg_bitsize,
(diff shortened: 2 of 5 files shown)


include/fmt/format-inl.h
```diff
@@ -980,6 +980,13 @@ FMT_API bool grisu_format(Double value, buffer<char>& buf, int precision,
   return true;
 }
 
+template <>
+char* sprintf_format<float>(float value, internal::buffer<char>& buf,
+                            sprintf_specs specs) {
+  // printf does not have a float format modifier, it only supports double.
+  return sprintf_format(double(value), buf, specs);
```
vitaut: I suggest replacing this with 
```
return sprintf_format<double>(value, buf, specs);
```
to avoid cast.

vitaut:
Merged, thanks!

PR #1364, opened 2019-10-14
Fix Visual Studio 2019 pedantic warning C4334: '<<': result of 32-bit…

chronoxor:
warning C4334: '<<': result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?)

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -981,7 +981,7 @@ FMT_FUNC void fallback_format(Double v, buffer<char>& buf, int& exp10) {
     denominator <<= 1 - fp_value.e;
     lower.assign(1);
     if (shift != 0) {
-      upper_store.assign(1 << shift);
+      upper_store.assign(1ull << shift);
       upper = &upper_store;
     }
   }


include/fmt/format-inl.h
```diff
@@ -981,7 +981,7 @@ FMT_FUNC void fallback_format(Double v, buffer<char>& buf, int& exp10) {
     denominator <<= 1 - fp_value.e;
     lower.assign(1);
     if (shift != 0) {
-      upper_store.assign(1 << shift);
+      upper_store.assign(1ull << shift);
```
vitaut: The result must be 32-bit. Does changing `1` to `1u` works around the bogus warning?

chronoxor: Unfortunately no, faliled with the similar warning:
`
format-inl.h(984,1): warning C4334: '<<': result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?)
`

chronoxor: Maybe this fix will be suitable?
`
upper_store.assign((1ull << shift) & 0xFFFFFFFF);
`

vitaut:
Thinking more of it, the type doesn't really matter because `assign` should be inlined and mostly optimized away. So merging as is, thanks!

PR #1370, opened 2019-10-21
Patch compiler error when building using nvcc

kent-tri:
If you compile using `nvcc` and pass the option `--expt-relaxed-constexpr` it will crash with an internal compiler error. This modification prevents using `constexpr` in `fmtlib` when compiling using `nvcc` and prevents the crash.

To clarify, this seems to be necessary on CUDA 10 with `gcc(5->7)` as the host compiler (every `gcc` version I tested). I did not have the chance to validate if other CUDA versions suffer from this same issue.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -58,9 +58,11 @@
 // Check if relaxed C++14 constexpr is supported.
 // GCC doesn't allow throw in constexpr until version 6 (bug 67371).
 #ifndef FMT_USE_CONSTEXPR
+#if !defined(__NVCC__)
 #  define FMT_USE_CONSTEXPR                                           \
     (FMT_HAS_FEATURE(cxx_relaxed_constexpr) || FMT_MSC_VER >= 1910 || \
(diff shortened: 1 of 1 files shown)


include/fmt/core.h
```diff
@@ -58,10 +58,12 @@
 // Check if relaxed C++14 constexpr is supported.
 // GCC doesn't allow throw in constexpr until version 6 (bug 67371).
 #ifndef FMT_USE_CONSTEXPR
+#if !(defined(__NVCC__) || defined(__CUDACC__))
```
vitaut: Is it necessary to check both `__NVCC__` and `__CUDACC__`? Can one be set and another not?

kentsommer: It is not necessary as `__CUDACC__` can be considered a subset of `__NVCC__`. I'll remove the `__CUDACC__` check: https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#nvcc-identification-macro

kentsommer: Done.

vitaut:
Thanks!

PR #1371, opened 2019-10-21
Fix more Visual Studio 2019 pedantic warnings

chronoxor:
…n operation

format.h(2808,1): warning C4127: conditional expression is constant

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -46,7 +46,6 @@
 
 #ifdef _MSC_VER
 #  pragma warning(push)
-#  pragma warning(disable : 4127)  // conditional expression is constant
 #  pragma warning(disable : 4702)  // unreachable code
 #endif
 
@@ -939,7 +938,11 @@ template <int GRISU_VERSION> struct grisu_shortest_handler {
                           uint64_t error, int exp, bool integral) {
     buf[size++] = digit;
     if (remainder >= error) return digits::more;
+#ifdef __cpp_if_constexpr
+    if constexpr (GRISU_VERSION != 3) {
+#else
     if (GRISU_VERSION != 3) {
+#endif
       uint64_t d = integral ? diff : diff * data::powers_of_10_64[-exp];
       round(d, divisor, remainder, error);
       return digits::done;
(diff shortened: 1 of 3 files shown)


include/fmt/format.h
```diff
@@ -71,6 +71,11 @@
 #  define FMT_HAS_BUILTIN(x) 0
 #endif
 
+#ifdef _MSC_VER
+#  pragma warning(push)
+#  pragma warning(disable : 4127)  // conditional expression is constant
```
vitaut: I'd rather avoid suppressing the warnings globally. It's better to fix/suppress in specific places because a warning may point to a genuine issue.

chronoxor: This suppress in `format.h` was made similar to `format-inl.h`

chronoxor: I'll look if this could be made more specific

mwinterb:
If "constexpr if" is available and used, MSVC won't issue a "conditional
expression is constant" warning. It looks like for this particular warning,
it could be used (if C++17 is enabled).

mwinterb:
... which you just now did! sorry for the noise.

include/fmt/format.h
chronoxor: Fixed

include/fmt/format-inl.h
```diff
@@ -939,7 +938,11 @@ template <int GRISU_VERSION> struct grisu_shortest_handler {
                           uint64_t error, int exp, bool integral) {
     buf[size++] = digit;
     if (remainder >= error) return digits::more;
+#ifdef __cpp_if_constexpr
+    if constexpr (GRISU_VERSION != 3) {
+#else
     if (GRISU_VERSION != 3) {
+#endif
```
vitaut: Here and below: I suggest using `const_check` from

https://github.com/fmtlib/fmt/blob/d39ebf3ff251bf199f88f917cecc504230823342/include/fmt/printf.h#L21

which was introduced specifically to silence this warning. It will avoid duplication.

E.g.

```
if (const_check(GRISU_VERSION != 3)) {
```

chronoxor: Fixed

chronoxor:
Do we need something more to fix in the PR to merge it?

vitaut:
No, all looks good. Thanks for the fix!

PR #1382, opened 2019-10-30
typo , an opening bracket was missing

JayatiGoyal:


vitaut:
Thanks for the PR but the opening parenthesis is not missing, it's on the previous line.

PR #1383, opened 2019-10-30
Changed a line for a giving a demo

Jainan-Tandel:
Just changed the sentence a bit to give a demo. Sorry for the inconvenience.


vitaut:
Thanks for the PR but this is not right because `std::string_view` doesn't return anything.

Issue #1380, opened 2019-10-30
`to_string_view`  is not called when using fmt::join() with 6.0.0


AlexandreBossard:
Hi,
Here is a minimal code example:

```
#include <fmt/core.h>
#include <fmt/format.h>

#include <string>
#include <vector>

namespace my_ns {
// Tagged string
template <typename Tag> class StringWrapper : std::string {
  using base_t = std::string;
public:
  StringWrapper(base_t const &s) : base_t(s) {}
  base_t const &string() const noexcept { return *this; }
};

template <typename Tag>
fmt::string_view to_string_view(StringWrapper<Tag> const &s) {
  return s.string();
}
} // namespace my_ns

#if 1 // set to 0 to trigger compilation errors
namespace fmt {
template <typename Tag>
struct formatter<my_ns::StringWrapper<Tag>, char>
    : formatter<fmt::basic_string_view<char>, char> {
  template <typename FormatContext>
  auto format(my_ns::StringWrapper<Tag> const &str, FormatContext &ctx) {
    return formatter<fmt::basic_string_view<char>, char>::format(
        to_string_view(str), ctx);
  }
};
} // namespace fmt
#endif
struct Email {}; // A Tag

void test1() {
  auto const s = my_ns::StringWrapper<Email>{"bob@provider.com"};
  fmt::print("{}\n", s);
}
void test2() {
  std::vector<my_ns::StringWrapper<Email>> emails = {{"bob@provider.com"},
                                                     {"alice@provider.com"}};
  fmt::print("{}\n", fmt::join(emails, ", "));
}

int main(int ac, char **argv) {
  test1();
  test2();
  return 0;
}
```
the `to_string_view` works fine on a single value (test1), but is not called when using `fmt::join()` (test2). Thus, You need to have a custom type formatter, which does called the to_string_view here.

This was working fine with **fmt/5.3.0**, but it doesn't anymore with **fmt/6.0.0**.

vitaut:
The old behavior was accidental. `to_string_view` is not an extension API and to make a type work with `join`, a `formatter` specialization should be provided.

AlexandreBossard:
The documentation is not accurate or at least misleading then:
https://fmt.dev/latest/api.html#_CPPv2N3fmt14to_string_viewEPK4Char

This feature is not very useful if it's contrived to the format string only (probably loosing static check).

Accidental or not it was quite nice to have and easy to understand and implement.

vitaut:
[...]

PR #1386, opened 2019-10-31
Fix is_string<> to allow string like class

AlexandreBossard:
The previous implementation did not compile while formatting a type with private inheritance of std::string.

The problem is that you cannot put SFINAE code in the inheriting declaration of a class/struct. It will trigger an error instead of disabling the overload.

This is the proper fix for  #1380.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -423,11 +423,24 @@ namespace internal {
 void to_string_view(...);
 using fmt::v6::to_string_view;
 
+template <std::size_t N> struct priority_tag : priority_tag<N - 1> {};
+
+template <> struct priority_tag<0> {};
+
+// This is required to handle types privately inheriting from std::string,
+// which triggers a compilation failure.
+template <typename S>
+auto is_string_impl(priority_tag<1>)
+    -> decltype(to_string_view(std::declval<S>()));
+
+template <typename S> void is_string_impl(priority_tag<0>);
+
 // Specifies whether S is a string type convertible to fmt::basic_string_view.
 // It should be a constexpr function but MSVC 2017 fails to compile it in
 // enable_if and MSVC 2015 fails to compile it as an alias template.
-template <typename S>
-struct is_string : std::is_class<decltype(to_string_view(std::declval<S>()))> {
+template <typename S> struct is_string {
+  static constexpr bool value = std::is_class<decltype(
+      is_string_impl<S>(std::declval<priority_tag<1>>()))>::value;
 };
 
 template <typename S, typename = void> struct char_t_impl {};
@@ -827,7 +840,8 @@ template <typename Context> struct arg_mapper {
   FMT_CONSTEXPR const char_type* map(char_type* val) { return val; }
   FMT_CONSTEXPR const char_type* map(const char_type* val) { return val; }
   template <typename T, FMT_ENABLE_IF(is_string<T>::value)>
-  FMT_CONSTEXPR basic_string_view<char_type> map(const T& val) {
+  FMT_CONSTEXPR basic_string_view<char_type> map_impl(const T& val,
+                                                      priority_tag<4>) {
     static_assert(std::is_same<char_type, char_t<T>>::value,
                   "mixing character types is disallowed");
     return to_string_view(val);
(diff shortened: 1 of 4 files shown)


vitaut:
Thanks! Could you add a test case somewhere around https://github.com/fmtlib/fmt/blob/791294d17b965f8750b54fe1ae3b8aebe237ff14/test/core-test.cc#L605?

AlexandreBossard:
As usual, writing tests was not a total waste of time, if not paintful. I had to modify my previous patch in order to make it work properly. I do need to use priority tag pattern in order to avoid deduced context. It might look scary, but it works :)

vitaut:
MSVC looks unhappy.

AlexandreBossard:
So, playing compiler whack a mole this morning, me and @theodelrieu eventually got it to work  with careful application of the 'priority tag' hammer.

Sadly we encounter a unavoidable bug with GCC 4.8. `is_like_std_string_like<>` was testing for `find()` and `data()` and `length()` which does not bowed well with SFINAE and private inheritance.
We choose to remove this test, since IMHO it doesn't mean much and is not very reliable. No tests were harmed in this process.

vitaut:
Thanks again for the PR but looking at the changes I'm not sure if this the added complexity is worth making this corner case work. The problem is not specific to fmt; you won't be able to pass `string_wrapper` to any function taking a `string_view` (and conceptually `format` is one of such functions):

```c++
#include <string>
#include <string_view>

template <typename T> class string_wrapper : std::string {
 public:
  string_wrapper(std::string const& s) : std::string(s) {}

  std::string const& string() const { return *this; }
};

void f(std::string_view) {}

int main() {
  f(string_wrapper<struct Tag>("foo"));
}
```

I recommend replacing inheritance with composition if you control `string_wrapper` or using `string_wrapper::string()` otherwise.

AlexandreBossard:
You seem to not get the issue. 

Right know, I have to use both a custom `formatter` class AND a `to_string_view()` to be able to format a `string_wrapper` through `fmt::format()` AND `fmt::join()`. This is a bug, there is no way around that. It would be curious (a non blocking but still annoying bug) that one way would work with both cases and not the other, but here, I have to use both, because of the way fmtlib is implemented (SFINAE expression in inheritance clause). 

Here is our StringWrapper implementation https://github.com/TankerHQ/sdk-native/blob/8320f01fc6db2b8b65f1de032bdec563fee6e61b/modules/types/include/Tanker/Types/StringWrapper.hpp

Using composition there would be a huge pain, as much as fmtlib is useful, I won't patch our implementation just to work around a fmtlib bug.

Here `StringWrapper` is a string, it completely satisfy its api. But fmtlib failed to detect that, fine, bummer, but fine. Then, after providing fmtlib how to format `StringWrapper` (with `to_string_view()`) as a string it still fails with `fmt::join()`, but the alternative, defining a `formatter` fails to.

Sure I could go the Java way and 'manually` call to_string() when passing `StringWrapper` to format but it defeat the point so much that it's ridiculous.

Complexity avoidance should not be an obstacle to correctness. Also, I believe we could avoid some of the priority tag dispatching with some heavy refactoring. But it's way to much work for me alone.

vitaut:
I'm not opposed to making this case work if this can be done in a less intrusive way. The main problem is that `is_string` doesn't support classes derived from `std::string` privately. Could you provide a PR that just fixes `is_string` with a test case checking its behavior only. Once it's done we can look at the `formatter` separately, if it's still a problem.

AlexandreBossard:
I believe that is exactly what https://github.com/fmtlib/fmt/commit/ecaae506b8b9c01f24a9a08f07312c3f11f72c71 does.

What this patch does is checking if a type `T` has a `to_string_view()` using ADL, just as it was before, but it does not trigger a compilation error due to access checking.

Problem is that it makes MSVC unhappy, because the same kind of bug is triggered down the line.

By the way, the purpose of a tagged string like `string_wrapper` here, is to distinguish business (Email, Address, Name, etc) strings one from another. Functions are expected to take those tagged strings as parameters and not std::string. This is why private inheritance is useful and why having a formatting tool, like fmtlib, that handle them like strings is useful to us.

I'll try to make another PR with only the is_string<> fix without priority tagging if possible, but it won't fix my issue :/

vitaut:
> I believe that is exactly what ecaae50 does.

Yes, except that the test should be checking the behavior of `is_string` not `formatter` which is what triggers another issue. I suggest focusing just on making `is_string` work, testing `is_string` only and then addressing MSVC issue in a separate PR.

AlexandreBossard:
Got it, will do that then.

Issue #1392, opened 2019-11-02
Specifier 'L' does not include thousand separator on floating number


dlaugt:
Floating numbers should have thousand separator with specifier 'n' like python.

Python:
```python
>>> from locale import setlocale, LC_ALL
>>> setlocale(LC_ALL, 'en_US')
'en_US'
>>> '{:n}'.format(1234)
'1,234'
>>> '{:n}'.format(1234.56)
'1,234.56'
```
C++
```cpp
setlocale(LC_ALL, "en-US");
fmt::format("{:n} {:n}", 1234, 1234.56);
// gives 1,234 1234.56
```

eman291:
#1129

vitaut:
Floating point formatting now supports thousand separators, for example:

```c++
std::locale::global(std::locale("en_US"));
fmt::print("{:L} {:L}", 1234, 1234.56);
```
prints
```
1,234 1,234.56
```

PR #1400, opened 2019-11-10
Add thousand separators on floating number with specifier 'n'

dlaugt:
PR for [#1392](https://github.com/fmtlib/fmt/issues/1392)


vitaut:
Thanks for the PR. Unfortunately floating-point formatting is WIP right now and merging this as is will make unifying `grisu` and `sprintf` paths harder and will require major rework afterwards. For this reason I'd prefer not merge this until the internal FP handling APIs are stabilized.

dlaugt:
I understand. For information, I did a quick benchmark with `{:n}` on integer between the current revision on master and this PR. This solution seems to be 15% faster probably because it avoids a modulo between two dynamic values.

PR #1404, opened 2019-11-13
fmt :: internal::do_throw(x) 

DBJDBJ:
FMT_THROW simplified

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -73,7 +73,7 @@
 
 FMT_BEGIN_NAMESPACE
 // 
-[[noreturn]] inline  void fail_fast(const char msg_[BUFSIZ]) {
+inline  void fail_fast(const char msg_[BUFSIZ]) {
 	::puts("\n\n\nFast exit on {fmt} error with message: ");
 	::puts(msg_);
 	::perror("\nPOSIX error message: ");
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -70,15 +70,6 @@
 #else
 #  define FMT_HAS_BUILTIN(x) 0
 #endif
-FMT_BEGIN_NAMESPACE
-// terror == terminating error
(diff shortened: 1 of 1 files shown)


DBJDBJ:
I give up ...  I am curious how will <format> look like ... this here is becoming a "jack of all trades" or in English: "the mess" ...

vitaut:
It's very tricky to do portably so I gave up too in the past.

vitaut:
But tanks for trying.

DBJDBJ:
Vitaut, {fmt}; I was "in", and it is great. I just do not get the release policy. 

In particular, I am looking after the `<format>`  header.   I assume that will be pure C++20 stuff?

I really can not understand what is/was the problem with my obvious simplification of `do_throw()` and `FMT_THROW` ... I assume (as Herb would say) "a wall of macros" will solve that one...

vitaut:
`<format>` could be C++20 but {fmt} must be portable to be useful.

DBJDBJ:
Thank you for your patience. 

Do you expect/think <format> will be asked to have `char16_t`, `char32_t` overloads?

vitaut:
Noone will ask for them =). If there's a proposal these overloads will be added.

DBJDBJ:
Is [this](https://stackoverflow.com/a/58909239/10870835) crucial to implement char8_t C++20? (besides being crucial for  char 16 and 32, it seems)

vitaut:
There will be no char8_t std::format in C++20.

DBJDBJ:
one of the key questions: all the three top compilers latest versions, have ni problems passing `char8_t *` to the `printf()` and printing it out properly. with warnings. And WG14 has no decision on that in the same time. 

I assume there is some compiler vendors clique :) who made that work, without WG21 or WG14 reaching a C++20 decision on `printf` + `char8_t`?

PR #1414, opened 2019-11-19
test: add default constructor for a const value

leonklingele:
This fixed a compilation error of the OS X 10.11.6 C++ compiler:

    ./fmt/test/format-test.cc:1861:16: error: default initialization of an object of const type 'const Answer' without a user-provided default constructor
      const Answer const_answer;
                   ^
                               {}


leonklingele:

vitaut:
Thanks

PR #1418, opened 2019-11-23
fix bad oss fuzz link in the oss-fuzz badge

pauldreik:
Finally a bug crept in, so we could verify that the bug tracker link worked properly :-)


vitaut:
I expected that. Have been working hard on this bug =).

Issue #1422, opened 2019-11-26
buffered_file::operator=(buffered_file&& other) is annotated as FMT_NOEXCEPT but calls close


0x8000-0000:
https://github.com/fmtlib/fmt/blob/4afb39bc242b7b7bdf852936f8fd7495d485695a/include/fmt/posix.h#L161

buffered_file::operator=(buffered_file&& other) is annotated as FMT_NOEXCEPT but calls close, which rightly is not annotated as FMT_NOEXCEPT because it throws .

0x8000-0000:
Found with [cppcheck](https://github.com/danmar/cppcheck) 1.89

0x8000-0000:
Also G++ 9.2 reports it as an error with -pedantic

fmt/src/posix.cc:119:6: error: declaration of ‘void fmt::v6::file::close()’ has a different exception specifier                                      119 | void file::close() {                                                                                                                                                                        |      ^~~~                                                                                                                                                                             fmt/src/posix.cc:13:                                                                                                         fmt/posix.h:238:16: note: from previous declaration ‘void fmt::v6::file::close() noexcept’                                               238 |   FMT_API void close() FMT_NOEXCEPT;                                                                                                                                                        |                ^~~~~

0x8000-0000:
Introduced in bb0c8bfea8608670063e7bd549279f630a5f4ede

PR #1427, opened 2019-11-27
Fix -Wconversion warnings

0x8000-0000:

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -1567,7 +1567,7 @@ template <typename Range> class basic_writer {
         basic_string_view<char_type> s(&sep, sep_size);
         // Index of a decimal digit with the least significant digit having
         // index 0.
-        unsigned digit_index = 0;
+        int digit_index = 0;
         std::string::const_iterator group = groups.cbegin();
         it = format_decimal<char_type>(
             it, abs_value, size,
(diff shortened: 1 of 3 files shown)


0x8000-0000:
This would close #1422

include/fmt/format.h
```diff
@@ -1572,7 +1572,7 @@ template <typename Range> class basic_writer {
         it = format_decimal<char_type>(
             it, abs_value, size,
             [this, s, &group, &digit_index](char_type*& buffer) {
-              if (*group <= 0 || ++digit_index % *group != 0 ||
+              if (*group <= 0 || ++digit_index % static_cast<unsigned int>(*group) != 0 ||
```
vitaut: Please change `digit_index`'s type to `int` instead.

include/fmt/posix.h
```diff
@@ -157,7 +157,7 @@ class buffered_file {
     other.file_ = nullptr;
   }
 
-  buffered_file& operator=(buffered_file&& other) FMT_NOEXCEPT {
+  buffered_file& operator=(buffered_file&& other) /* FMT_NOEXCEPT */ {
```
vitaut: I suggest removing `/* FMT_NOEXCEPT */` altogether.

include/fmt/format.h
0x8000-0000: OK

include/fmt/posix.h
0x8000-0000: OK

vitaut:
Thanks!

PR #1428, opened 2019-11-28
Enable -Wsign-conversion for CI

0x8000-0000:


0x8000-0000:
Well, ahem, 317 build failures. But I learned about "-DFMT_PEDANTIC=On -DFMT_WERROR=On". I'll be back with more fixes, before this can go in.

vitaut:
This warning has too many false positives. I don't think the ones in `format-inl.h` worth addressing since the latter is basically an internal header which is not supposed to be used by users (except the users of the header-only mode but those know what they signed up for =)).

0x8000-0000:
I got it down to 58 warnings for GCC9 and 21 warnings from Clang9. The next part is harder, since it involves templates usage, where the template parameters differ in signedness, mainly around "arg"

check_arg_id(), arg() ...

PR #1433, opened 2019-11-29
Squelch MSVC warning exporting subclasses of runtime_error

marti4d:
When compiling {fmt} as a DLL, MSVC complains that we are exporting
classes that inherit from "std::runtime_error", which we are not
exporting.

In this case, it's not really a problem because that symbol is already
exported via the C++ stdlib. So we just add a pragma to silence the
warning.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -168,9 +168,9 @@
 
 #if !defined(FMT_HEADER_ONLY) && defined(_WIN32)
 #  ifdef FMT_EXPORT
-#    define FMT_API __declspec(dllexport)
+#    define FMT_API __pragma(warning(suppress : 4275)) __declspec(dllexport)
 #  elif defined(FMT_SHARED)
-#    define FMT_API __declspec(dllimport)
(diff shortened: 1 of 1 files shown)


marti4d:
After I looked more at the code and saw how `runtime_error` is used, I strongly agree with your reasoning for closing down my earlier issue. Sorry about that - If I had looked and thought-it-through a bit more, I probably wouldn't have filed the ticket.

What I have here is one potential solution to fix the warning. One of the only things that may-or-may-not matter is that it will suppress this warning now for everything - So in the future if you *actually did* make the mistake of exporting a class without exporting its parent, no warning would be given.

Alternatively, I suppose it's possible to use `#pragma warning(push)` and `#pragma warning(pop)` to only disable the warning for those specific declarations - But that may negatively affect code readability.

codicodi:
I would advise for using `push` and `pop`, so that warning supression doesn't leak into user code.

marti4d:
Indeed - It also appears that MSVC supports this `__pragma(warning())` directive, and that it can even be put inline with the other API errata.

Perhaps you like this solution better?

vitaut:
Thanks!

PR #1435, opened 2019-11-29
Clean-up sign-conversion warnings

0x8000-0000:

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -696,7 +696,7 @@ inline int to_nonnegative_int(T value, int upper) {
 
 template <typename T, FMT_ENABLE_IF(std::is_integral<T>::value)>
 inline T mod(T x, int y) {
-  return x % y;
+  return x % static_cast<T>(y);
 }
 template <typename T, FMT_ENABLE_IF(std::is_floating_point<T>::value)>
 inline T mod(T x, int y) {
@@ -793,11 +793,7 @@ struct chrono_formatter {
 
   explicit chrono_formatter(FormatContext& ctx, OutputIt o,
                             std::chrono::duration<Rep, Period> d)
-      : context(ctx), out(o), val(d.count()), negative(false) {
-    if (d.count() < 0) {
-      val = 0 - val;
-      negative = true;
-    }
+      : context(ctx), out(o), val(static_cast<rep>(std::abs(d.count()))), negative(d.count() < 0) {
 
     // this may overflow and/or the result may not fit in the
     // target type.
@@ -1023,8 +1019,8 @@ struct formatter<std::chrono::duration<Rep, Period>, Char> {
     void on_error(const char* msg) { FMT_THROW(format_error(msg)); }
     void on_fill(Char fill) { f.specs.fill[0] = fill; }
     void on_align(align_t align) { f.specs.align = align; }
-    void on_width(unsigned width) { f.specs.width = width; }
-    void on_precision(unsigned _precision) { f.precision = _precision; }
+    void on_width(int width) { f.specs.width = width; }
+    void on_precision(int _precision) { f.precision = _precision; }
     void end_precision() {}
 
     template <typename Id> void on_dynamic_width(Id arg_id) {
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -795,6 +795,15 @@ struct chrono_formatter {
                             std::chrono::duration<Rep, Period> d)
       : context(ctx), out(o), val(static_cast<rep>(std::abs(d.count()))), negative(d.count() < 0) {
 
+   if (d.count() < 0) {
+      val = static_cast<rep>(- d.count());
+      negative = true;
+   }
+   else {
+      val = static_cast<rep>(d.count());
+      negative = false;
+   }
+
     // this may overflow and/or the result may not fit in the
     // target type.
 #if FMT_SAFE_DURATION_CAST
--- a/include/fmt/compile.h
+++ b/include/fmt/compile.h
@@ -141,7 +141,7 @@ class format_string_compiler : public error_handler {
     auto it = parse_format_specs(begin, end, handler);
     if (*it != '}') on_error("missing '}' in format string");
     repl.arg_id = part_.part_kind == part::kind::arg_index
-                      ? arg_ref<Char>(static_cast<int>(part_.val.arg_index))
+                      ? arg_ref<Char>(part_.val.arg_index)
                       : arg_ref<Char>(part_.val.str);
     auto part = part::make_replacement(repl);
     part.arg_id_end = begin;
--- a/include/fmt/compile.h
+++ b/include/fmt/compile.h
@@ -26,11 +26,11 @@ template <typename Char> struct format_part {
 
   kind part_kind;
   union value {
-    unsigned arg_index;
+    int arg_index;
     basic_string_view<Char> str;
     replacement repl;
 
-    FMT_CONSTEXPR value(unsigned index = 0) : arg_index(index) {}
+    FMT_CONSTEXPR value(int index = 0) : arg_index(index) {}
     FMT_CONSTEXPR value(basic_string_view<Char> s) : str(s) {}
     FMT_CONSTEXPR value(replacement r) : repl(r) {}
   } val;
@@ -40,7 +40,7 @@ template <typename Char> struct format_part {
   FMT_CONSTEXPR format_part(kind k = kind::arg_index, value v = {})
       : part_kind(k), val(v) {}
 
-  static FMT_CONSTEXPR format_part make_arg_index(unsigned index) {
+  static FMT_CONSTEXPR format_part make_arg_index(int index) {
     return format_part(kind::arg_index, index);
   }
   static FMT_CONSTEXPR format_part make_arg_name(basic_string_view<Char> name) {
@@ -62,7 +62,7 @@ template <typename Char> struct part_counter {
   }
 
   FMT_CONSTEXPR void on_arg_id() { ++num_parts; }
-  FMT_CONSTEXPR void on_arg_id(unsigned) { ++num_parts; }
+  FMT_CONSTEXPR void on_arg_id(int) { ++num_parts; }
   FMT_CONSTEXPR void on_arg_id(basic_string_view<Char>) { ++num_parts; }
 
   FMT_CONSTEXPR void on_replacement_field(const Char*) {}
@@ -119,7 +119,7 @@ class format_string_compiler : public error_handler {
     part_ = part::make_arg_index(parse_context_.next_arg_id());
   }
 
-  FMT_CONSTEXPR void on_arg_id(unsigned id) {
+  FMT_CONSTEXPR void on_arg_id(int id) {
     parse_context_.check_arg_id(id);
     part_ = part::make_arg_index(id);
   }
@@ -141,7 +141,7 @@ class format_string_compiler : public error_handler {
     auto it = parse_format_specs(begin, end, handler);
     if (*it != '}') on_error("missing '}' in format string");
     repl.arg_id = part_.part_kind == part::kind::arg_index
-                      ? arg_ref<Char>(part_.val.arg_index)
+                      ? arg_ref<Char>(static_cast<int>(part_.val.arg_index))
                       : arg_ref<Char>(part_.val.str);
     auto part = part::make_replacement(repl);
     part.arg_id_end = begin;
@@ -512,8 +512,6 @@ template <typename CompiledFormat, typename... Args,
                                          CompiledFormat>::value)>
 std::basic_string<Char> format(const CompiledFormat& cf, const Args&... args) {
   basic_memory_buffer<Char> buffer;
-  using range = buffer_range<Char>;
-  using context = buffer_context<Char>;
   cf.format(std::back_inserter(buffer), args...);
   return to_string(buffer);
 }
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -482,8 +482,7 @@ inline fp operator*(fp x, fp y) { return {multiply(x.f, y.f), x.e + y.e + 64}; }
 FMT_FUNC fp get_cached_power(int min_exponent, int& pow10_exponent) {
   const uint64_t one_over_log2_10 = 0x4d104d42;  // round(pow(2, 32) / log2(10))
   int index = static_cast<int>(
-      static_cast<int64_t>(
-          static_cast<uint64_t>(min_exponent + fp::significand_size - 1) * one_over_log2_10 +
+      (static_cast<uint64_t>(min_exponent + fp::significand_size - 1) * one_over_log2_10 +
           ((1ULL << 32) - 1)  // ceil
           ) >>
       32  // arithmetic shift
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -18,6 +18,7 @@
 #include <cstdarg>
 #include <cstring>  // for std::memmove
 #include <cwchar>
+#include <memory>    // for std::uninitialized_fill_n
 #if !defined(FMT_STATIC_THOUSANDS_SEPARATOR)
 #  include <locale>
 #endif
@@ -482,8 +483,8 @@ FMT_FUNC fp get_cached_power(int min_exponent, int& pow10_exponent) {
   const uint64_t one_over_log2_10 = 0x4d104d42;  // round(pow(2, 32) / log2(10))
   int index = static_cast<int>(
       static_cast<int64_t>(
-          (min_exponent + fp::significand_size - 1) * one_over_log2_10 +
-          ((uint64_t(1) << 32) - 1)  // ceil
+          static_cast<uint64_t>(min_exponent + fp::significand_size - 1) * one_over_log2_10 +
+          ((1ULL << 32) - 1)  // ceil
           ) >>
       32  // arithmetic shift
   );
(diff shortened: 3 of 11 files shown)


0x8000-0000:
This first step reduces the number of -Wsign-conversion warnings from 317 to 58 with GCC9.

0x8000-0000:
I have tried to reproduce the test crash on Visual Studio 2019 building in Release Mode for 32 bits (I don't have 2017)

    > "C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\Common7\Tools\VsDevCmd.bat" -arch=x86

    > cmake ..\fmt -DFMT_PEDANTIC=ON -DCMAKE_BUILD_TYPE=Release -G"Visual Studio 16 2019"

    > cmake --build . --config Release

    > ctest -C Release

And I can't reproduce the crash. If somebody has access to that toolchain and could try debugging it, I would be very grateful. Otherwise I would have to do a binary search on all the changes in this commit :(

0x8000-0000:
Made doubly-sure I have produced 32 bit binaries with:

"C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake" ..\fmt -DFMT_PEDANTIC=ON -DCMAKE_BUILD_TYPE=Release -A win32

$ file bin/Release/color-test.exe
bin/Release/color-test.exe: PE32 executable (console) Intel 80386, for MS Windows


Still, no crash when running the 32-bit binary on Windows 64, running on AMD laptop (FWIW)

0x8000-0000:
This still crashes in UtilTest.WriteUIntPtr, even with just some changes in chrono.h ! I suspect the master branch just fails in the same way at this time.

0x8000-0000:
Oh, it "fails" in the same way as #1438 , even though that one just modifies a text document.

vitaut:
Yeah, `WriteUIntPtr` test failure is unrelated. I'm about to fix it (just waiting for see if CI is green).

vitaut:
Merged the fix into master. Please rebase.

0x8000-0000:
Down to 51 errors on GCC9, 18 on Clang9.

0x8000-0000:
Down to 1 warning in fmt code, reported by both Gcc9 and Clang9.

fmt/test/format-test.cc:2026:10: error: implicit conversion changes signedness: 'unsigned __int128' to '::testing::internal::Function<void (__int128_t)>::Argument1' (aka '__int128') [-Werror,-Wsign-conversion]

Gcc9 also reports several warnings stemming from a macro in test/gtest/gtest.h:

   1922 #define GTEST_COMPILE_ASSERT_(expr, msg) \                                                                                                             
   1923   typedef ::testing::internal::CompileAssert<(static_cast<bool>(expr))> \                                                                              
   1924       msg[static_cast<bool>(expr) ? 1 : -1] GTEST_ATTRIBUTE_UNUSED_            

I think there are some cases where Gcc9 sees "msg[-1]" and reports the warning. It might be worth exploring whether upgrading gtest/gmock would clean that up.

include/fmt/chrono.h
```diff
@@ -793,11 +793,7 @@ struct chrono_formatter {
 
   explicit chrono_formatter(FormatContext& ctx, OutputIt o,
                             std::chrono::duration<Rep, Period> d)
-      : context(ctx), out(o), val(d.count()), negative(false) {
-    if (d.count() < 0) {
-      val = 0 - val;
-      negative = true;
-    }
+      : context(ctx), out(o), val(static_cast<rep>(std::abs(d.count()))), negative(d.count() < 0) {
```
vitaut: I suggest keeping the old code where negative case is handled in one place.

include/fmt/compile.h
```diff
@@ -141,7 +141,7 @@ class format_string_compiler : public error_handler {
     auto it = parse_format_specs(begin, end, handler);
     if (*it != '}') on_error("missing '}' in format string");
     repl.arg_id = part_.part_kind == part::kind::arg_index
-                      ? arg_ref<Char>(part_.val.arg_index)
+                      ? arg_ref<Char>(static_cast<int>(part_.val.arg_index))
```
vitaut: Please change the type of `arg_index` in `arg_ref` to `int` and remove the cast.

include/fmt/format-inl.h
```diff
@@ -482,8 +483,8 @@ FMT_FUNC fp get_cached_power(int min_exponent, int& pow10_exponent) {
   const uint64_t one_over_log2_10 = 0x4d104d42;  // round(pow(2, 32) / log2(10))
   int index = static_cast<int>(
       static_cast<int64_t>(
-          (min_exponent + fp::significand_size - 1) * one_over_log2_10 +
-          ((uint64_t(1) << 32) - 1)  // ceil
+          static_cast<uint64_t>(min_exponent + fp::significand_size - 1) * one_over_log2_10 +
```
vitaut: Please replace the cast with `to_unsigned`.

include/fmt/format-inl.h
```diff
@@ -531,14 +532,14 @@ class bigint {
 
   friend struct formatter<bigint>;
 
-  void subtract_bigits(int index, bigit other, bigit& borrow) {
+  void subtract_bigits(size_t index, bigit other, bigit& borrow) {
```
vitaut: In all the bigint and floating-point formatting code let's keep the type of indices `int`. More generally we don't need to address false positives in `format-inl.h` because it's an internal header.

include/fmt/format.h
```diff
@@ -1244,7 +1244,7 @@ int snprintf_float(T value, int precision, float_specs specs,
                    buffer<char>& buf);
 
 template <typename T> T promote_float(T value) { return value; }
-inline double promote_float(float value) { return value; }
+inline double promote_float(float value) { return static_cast<double>(value); }
```
vitaut: Why is this needed?

include/fmt/format.h
```diff
@@ -2627,11 +2627,11 @@ class format_string_checker {
   FMT_CONSTEXPR void on_text(const Char*, const Char*) {}
 
   FMT_CONSTEXPR void on_arg_id() {
-    arg_id_ = context_.next_arg_id();
+    arg_id_ = static_cast<unsigned>(context_.next_arg_id());
```
vitaut: This cast won't be needed with the change mentioned earlier.

include/fmt/format.h
```diff
@@ -2627,11 +2627,11 @@ class format_string_checker {
...
     check_arg_id();
   }
   FMT_CONSTEXPR void on_arg_id(int id) {
-    arg_id_ = id;
+    arg_id_ = static_cast<unsigned>(id);
```
vitaut: ditto

include/fmt/printf.h
```diff
@@ -399,10 +399,10 @@ template <typename OutputIt, typename Char>
 typename basic_printf_context<OutputIt, Char>::format_arg
 basic_printf_context<OutputIt, Char>::get_arg(unsigned arg_index) {
   if (arg_index == internal::max_value<unsigned>())
-    arg_index = parse_ctx_.next_arg_id();
+    arg_index = static_cast<unsigned>(parse_ctx_.next_arg_id());
```
vitaut: Please change `arg_index` to `int` and remove all the casts.

test/format-impl-test.cc
```diff
@@ -224,19 +224,23 @@ TEST(FPTest, Normalize) {
...
+      {static_cast<double>(1.0f), 0.9999999701976776, 1.0000000596046448},
       // min normal
-      {1.1754944e-38f, 1.1754942807573643e-38, 1.1754944208872107e-38},
+      {static_cast<double>(1.1754944e-38f), 1.1754942807573643e-38, 1.1754944208872107e-38},
       // max subnormal
-      {1.1754942e-38f, 1.1754941406275179e-38, 1.1754942807573643e-38},
+      {static_cast<double>(1.1754942e-38f), 1.1754941406275179e-38, 1.1754942807573643e-38},
       // min subnormal
-      {1e-45f, 7.006492321624085e-46, 2.1019476964872256e-45},
+      {static_cast<double>(1e-45f), 7.006492321624085e-46, 2.1019476964872256e-45},
```
vitaut: This is unrelated to sign conversion and adds unnecessary noise - please revert.

test/format-test.cc
```diff
@@ -2005,9 +2005,11 @@ class mock_arg_formatter
     : public fmt::internal::arg_formatter_base<buffer_range> {
  private:
 #if FMT_USE_INT128
-  MOCK_METHOD1(call, void(__int128_t value));
+  MOCK_METHOD1(callSigned, void(__int128_t value));
+  MOCK_METHOD1(callUnsigned, void(__uint128_t value));
 #else
-  MOCK_METHOD1(call, void(long long value));
+  MOCK_METHOD1(callSigned, void(long long value));
+  MOCK_METHOD1(callUnsigned, void(unsigned long long value));
```
vitaut: This is unnecessary for the test purposes. Just static cast to the largest signed type.

test/printf-test.cc
```diff
@@ -589,7 +590,8 @@ class custom_printf_arg_formatter : public formatter_t {
       if (round(value * pow(10, specs()->precision)) == 0.0) value = 0;
   return formatter_t::operator()(value);
 }
-};
+}
+;
```
vitaut: Could you remove ; while at it? =)

test/std-format-test.cc
```diff
@@ -110,7 +110,7 @@ template <> struct std::formatter<S> {
     char c = get_char();
     if (!isdigit(c) || (++iter, get_char()) != '}')
       throw format_error("invalid format");
-    width_arg_id = c - '0';
+    width_arg_id = static_cast<size_t>(c - '0');
```
vitaut: This tests an example from the standard and should remain unchanged.

include/fmt/chrono.h
0x8000-0000: You are trying to initialize 'val' which is unsigned with a signed value (d.count()).

include/fmt/compile.h
0x8000-0000: It seems to be already 'int' in struct format_part.value . Will remove the cast.

include/fmt/format-inl.h
0x8000-0000: Your suggested change causes four tests failures (building with Gcc9 in Debug mode):

          5 - grisu-test (Child aborted)
          7 - format-test (Child aborted)
          8 - format-impl-test (Failed)
         12 - printf-test (Child aborted)

It trips this:

    fmt/include/fmt/core.h:266: assertion failed: negative value

include/fmt/format-inl.h
0x8000-0000: The code is trying to subscript into bigits object (of type buffer) with a signed value, when the operator [] expects std::size_t, thus the warning.

Do you want me to convert the buffer implementation and interface from size_t to int? This, after all, what Scott Meyers and Bjarne Stroustroup are recommending:
     http://open-std.org/JTC1/SC22/WG21/docs/papers/2019/p1428r0.pdf
     https://www.aristeia.com/Papers/C++ReportColumns/sep95.pdf

Converting to int from size_t *should* clean up a lot of casts.

include/fmt/format.h
0x8000-0000: Seems it is not needed. Removed.

include/fmt/printf.h
0x8000-0000: This will be a bit tricky because of line 401. Should I change the sentinel value to internal::max_value<int>() ?

test/std-format-test.cc
0x8000-0000: Should width_arg_id become signed then?

test/format-test.cc
0x8000-0000: The test is calling the mocked method with both signed and unsigned values.

I don't care as much about the tests being -Wsign-conversion clean, but...

I want my own code to be -Wsign-conversion clean, because we got bitten in the past by (unsigned)-1, and other bit fields confusion (somebody shoving 32 bits into a 16 bit register) and I want to use {fmt}. Which means {fmt} needs to be -Wsign-conversion clean, which means its test suite needs to be -Wsign-conversion clean, otherwise we can't enforce it via the automated regression suite.

test/format-impl-test.cc
0x8000-0000: OK

test/printf-test.cc
0x8000-0000: Can't, since it follows a class definition :)  Not sure why clang-format wants to put it on a separate line.

include/fmt/chrono.h
vitaut: Adding a cast is fine. I'm just saying that having negation and setting `negative` in one if block makes the intent a bit clearer, so let's keep that part.

include/fmt/format-inl.h
vitaut: This suggests that the cast is misplaced.

include/fmt/format.h
0x8000-0000: This is a different arg_id_ ; do you want me to convert it to 'int' as well?

include/fmt/chrono.h
0x8000-0000: OK

include/fmt/format-inl.h
vitaut: Changing `buffer`'s index to a signed type might be cleaner but let's not do this in the current PR. For now let's just keep floating-point related code as is.

include/fmt/printf.h
vitaut: I think sentinel can be changed to -1 once `arg_index` is signed.

include/fmt/format-inl.h
0x8000-0000: Yes, in fact it is not needed at all.

0x8000-0000: Removing the cast also silences the warning, but still trips the assertion.

include/fmt/format-inl.h
0x8000-0000: OK, will leave changing buffer for a follow-up change.

include/fmt/printf.h
0x8000-0000: I was thinking of wrapping the sentinel value in a constant, introduced via an enum so it is portable to older compilers.

test/format-test.cc
vitaut: This is a simple mock which is only used here: https://github.com/fmtlib/fmt/blob/62da1db62affe82dc3ff5e5396e7b5cc235eca3d/test/format-test.cc#L2052

so casting is fine.

0x8000-0000: I changed as suggested and I got back the dreaded:

fmt/test/format-test.cc:2026:10: error: implicit conversion changes signedness: 'unsigned __int128' to '::testing::internal::Function<void (__int128_t)>::Argument1' (aka '__int128') [-Werror,-Wsign-conversion]

test/std-format-test.cc
0x8000-0000: width_arg_id is passed on the next line to check_arg_id which requires a size_t .

test/format-test.cc
0x8000-0000: This is the long story, threaded through format.h and core.h:

> /home/florin/libraries/fmt/test/format-test.cc: In instantiation of ‘typename std::enable_if<fmt::v6::internal::is_integral<T>::value, std::back_insert_iterator<fmt::v6::internal::buffer<char> > >::type mock_arg_formatter::operator()(T) [with T = __int128 unsigned; typename std::enable_if<fmt::v6::internal::is_integral<T>::value, std::back_insert_iterator<fmt::v6::internal::buffer<char> > >::type = std::back_insert_iterator<fmt::v6::internal::buffer<char> >]’:

> /home/florin/libraries/fmt/include/fmt/core.h:1024:15:   required from ‘constexpr decltype (vis(0)) fmt::v6::visit_format_arg(Visitor&&, const fmt::v6::basic_format_arg<Context>&) [with Visitor = mock_arg_formatter; Context = fmt::v6::basic_format_context<std::back_insert_iterator<fmt::v6::internal::buffer<char> >, char>; decltype (vis(0)) = std::back_insert_iterator<fmt::v6::internal::buffer<char> >]’

> /home/florin/libraries/fmt/include/fmt/format.h:3172:25:   required from ‘void fmt::v6::format_handler<ArgFormatter, Char, Context>::on_replacement_field(const Char*) [with ArgFormatter = mock_arg_formatter; Char = char; Context = fmt::v6::basic_format_context<std::back_insert_iterator<fmt::v6::internal::buffer<char> >, char>]’

> /home/florin/libraries/fmt/include/fmt/format.h:2583:7:   required from ‘constexpr void fmt::v6::internal::parse_format_string(fmt::v6::basic_string_view<Char>, Handler&&) [with bool IS_CONSTEXPR = false; Char = char; Handler = fmt::v6::format_handler<mock_arg_formatter, char, fmt::v6::basic_format_context<std::back_insert_iterator<fmt::v6::internal::buffer<char> >, char> >&]’

> /home/florin/libraries/fmt/include/fmt/format.h:3205:39:   required from ‘typename Context::iterator fmt::v6::vformat_to(typename ArgFormatter::range, fmt::v6::basic_string_view<Char>, fmt::v6::basic_format_args<Context>, fmt::v6::internal::locale_ref) [with ArgFormatter = mock_arg_formatter; Char = char; Context = fmt::v6::basic_format_context<std::back_insert_iterator<fmt::v6::internal::buffer<char> >, char>; typename Context::iterator = std::back_insert_iterator<fmt::v6::internal::buffer<char> >; typename ArgFormatter::range = fmt::v6::buffer_range<char>]’

> /home/florin/libraries/fmt/test/format-test.cc:2043:63:   required from here
> /home/florin/libraries/fmt/test/format-test.cc:2026:10: error: conversion to ‘testing::internal::Function<void(__int128)>::Argument1’ {aka ‘__int128’} from ‘__int128 unsigned’ may change the sign of the result [-Werror=sign-conversion]
 2026 |     call(value);
      |          ^~~~~

0x8000-0000:
Once this is approved I can refactor the top commit and push the changes into the earlier commits. I didn't want to force-push during the review.

include/fmt/format-inl.h
0x8000-0000: Actually the static_cast<int64_t> is not needed.

include/fmt/format.h
0x8000-0000: Done.

vitaut:
I guess this can be closed now in favor of #1440.

PR #1437, opened 2019-11-29
Fixing installation directory of '*.dll' files on Windows

robquant:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -219,6 +219,9 @@ if (FMT_INSTALL)
   set(FMT_LIB_DIR ${CMAKE_INSTALL_LIBDIR} CACHE STRING
       "Installation directory for libraries, relative to ${CMAKE_INSTALL_PREFIX}.")
 
+  set(FMT_BIN_DIR ${CMAKE_INSTALL_BINDIR} CACHE STRING
+      "Installation directory for libraries, relative to ${CMAKE_INSTALL_PREFIX}.")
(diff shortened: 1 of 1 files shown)


tdegeus:
For future reference:

* https://stackoverflow.com/a/21592765/2646505
* https://cmake.org/cmake/help/git-master/command/install.html

CMakeLists.txt
```diff
@@ -219,6 +219,9 @@ if (FMT_INSTALL)
   set(FMT_LIB_DIR ${CMAKE_INSTALL_LIBDIR} CACHE STRING
       "Installation directory for libraries, relative to ${CMAKE_INSTALL_PREFIX}.")
 
+  set(FMT_BIN_DIR ${CMAKE_INSTALL_BINDIR} CACHE STRING
+      "Installation directory for libraries, relative to ${CMAKE_INSTALL_PREFIX}.")
```
vitaut: Does this need to be configurable? I suggest removing this option and just using `${CMAKE_INSTALL_BINDIR}`.

vitaut:
Merged without the new option in https://github.com/fmtlib/fmt/commit/159f89e2b87a821e01f8761e04da1c7373f29042.

CMakeLists.txt
robquant: No, this does not need to be configurable. I just copied what was there for `${FMT_LIB_DIR}`.

PR #1438, opened 2019-11-29
Adding shield for conda-forge to readme

tdegeus:


tdegeus:
Feel free to close if you don't like it!

vitaut:
Thanks for the PR but I don't think this badge adds much for users or contributors and we don't have such badges for other package management systems. Installation information belongs to the Usage section of the docs.

PR #1440, opened 2019-12-02
Clean-up sign-conversion warnings (rebased)

0x8000-0000:

This pull request contains the reviewed and agreed-upon portion of #1435, rebased on top of current master.

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -696,7 +696,7 @@ inline int to_nonnegative_int(T value, int upper) {
 
 template <typename T, FMT_ENABLE_IF(std::is_integral<T>::value)>
 inline T mod(T x, int y) {
-  return x % y;
+  return x % static_cast<T>(y);
 }
 template <typename T, FMT_ENABLE_IF(std::is_floating_point<T>::value)>
 inline T mod(T x, int y) {
@@ -793,11 +793,14 @@ struct chrono_formatter {
 
   explicit chrono_formatter(FormatContext& ctx, OutputIt o,
                             std::chrono::duration<Rep, Period> d)
-      : context(ctx), out(o), val(d.count()), negative(false) {
-    if (d.count() < 0) {
-      val = 0 - val;
-      negative = true;
-    }
+      : context(ctx), out(o), negative(d.count() < 0) {
+
+   if (negative) {
+      val = static_cast<rep>(- d.count());
+   }
+   else {
+      val = static_cast<rep>(d.count());
+   }
 
     // this may overflow and/or the result may not fit in the
     // target type.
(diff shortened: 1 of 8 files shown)


Danniiix:
What is a pull request

0x8000-0000:
> What is a pull request

Welcome to GitHub - you might want to start with https://guides.github.com/activities/hello-world/

include/fmt/format.h
```diff
@@ -2660,7 +2660,9 @@ class format_string_checker {
   // Format specifier parsing function.
   using parse_func = const Char* (*)(parse_context_type&);
 
-  unsigned arg_id_;
+  enum { ARG_ID_SENTINEL = -1 };
```
vitaut: This is not needed, negative value is a clear enough indication of an invalid index.

include/fmt/printf.h
```diff
@@ -183,6 +183,12 @@ internal::truncating_iterator<OutputIt> printf(
     basic_format_args<Context> args) {
   return Context(it, format, args).format();
 }
+
+enum
+{
+   ARG_INDEX_SENTINEL = -1
+};
```
vitaut: Ditto.

include/fmt/chrono.h
```diff
@@ -793,11 +793,16 @@ struct chrono_formatter {
...
+
+   if (d.count() < 0) {
+      val = static_cast<rep>(- d.count());
       negative = true;
-    }
+   }
+   else {
+      val = static_cast<rep>(d.count());
+      negative = false;
+   }
```
vitaut: There is too much redundancy here. Please keep the old code and just add a cast.

test/scan.h
```diff
@@ -159,7 +159,7 @@ struct scan_handler : error_handler {
   const char* pos() const { return scan_ctx_.begin(); }
 
   void on_text(const char* begin, const char* end) {
-    auto size = end - begin;
+    auto size = static_cast<size_t>(end - begin);
```
vitaut: Please use `to_unsigned` here and below.

0x8000-0000:
> Thanks for the PR. As discussed in #1435, please keep the internal header `include/fmt/format-inl.h` as it is now i.e. revert the changes.

I will update as requested; this means I have to keep maintaining my fork that builds with -Wsign-conversion for a while, presumably until we get C++20 conformant compilers.

include/fmt/format.h
0x8000-0000: OK

include/fmt/printf.h
0x8000-0000: OK

include/fmt/chrono.h
0x8000-0000: I apologize for the line 796, it was mismerged by me. I intended to remove the member initialization of val and negative.

But we can't keep the old code, since it assigns a possibly signed value "d.count()" into an unsigned variable "val". You need "val" to acquire the absolute value of d.count, which can be cleanly done either with "std::abs()" like on line 796, or with the test on 798:804.

Which one would you prefer?

test/scan.h
0x8000-0000: OK

include/fmt/chrono.h
vitaut: I think it should be enough to add `static_cast` to the initializer list.

vitaut:
Merged with minor tweaks, thanks.

PR #1443, opened 2019-12-02
Fix compilation with MinGW

bbolli:
Commit 3bc28fcc6b2a ("Squelch MSVC warning exporting subclasses of
runtime_error", 2019-11-29) silenced a MSVC warning under. The MinGW
compiler also defines _WIN32, but does not support the "warning" pragma.

Introduce a helper macro to squelch the MSVC warning only when using the
Microsoft compiler.

Signed-off-by: Beat Bolli <dev@drbeat.li>


vitaut:
Thanks!

Issue #1445, opened 2019-12-03
Link error on 6.1.0


egorpugin:
Custom build system, worked fine on 6.0.0 and before, error after 6.1.0 update.
Dll build + build with sources (non header only).
Is there something on my side or it's a bug?

```
error LNK2019: unresolved external symbol "void __cdecl fmt::v6::internal::assert_fail(char const *,int,char const *)" (?assert_fail@internal@v6@fmt@@YAXPEBDH0@Z) referenced in function "class stdext::checked_array_iterator<char *> __cdecl fmt::v6::internal::format_decimal<char,unsigned int,class stdext::checked_array_iterator<char *>,class <lambda_18a81b6c877b6827ad6aa98de427f027> >(class stdext::checked_array_iterator<char *>,unsigned int,int,class <lambda_18a81b6c877b6827ad6aa98de427f027>)" (??$format_decimal@DIV?$checked_array_iterator@PEAD@stdext@@V<lambda_18a81b6c877b6827ad6aa98de427f027>@@@internal@v6@fmt@@YA?AV?$checked_array_iterator@PEAD@stdext@@V34@IHV<lambda_18a81b6c877b6827ad6aa98de427f027>@@@Z)
fatal error LNK1120: 1 unresolved externals
```

egorpugin:
Added FMT_API in core.h and FMT_API in format-inl.h and it worked.

vitaut:
Could you submit a PR?

egorpugin:
Sure.
https://github.com/fmtlib/fmt/pull/1446

egorpugin:
Probably it's worth it to create test that reproduces this issue (dll build). But I don't have time to work on it.

PR #1446, opened 2019-12-03
Export assert_fail with FMT_API. This fixes dll build.

egorpugin:
See https://github.com/fmtlib/fmt/issues/1445

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -55,7 +55,7 @@ inline fmt::internal::null<> strerror_s(char*, std::size_t, ...) { return {}; }
 FMT_BEGIN_NAMESPACE
 namespace internal {
 
-FMT_API FMT_FUNC void assert_fail(const char* file, int line, const char* message) {
+FMT_FUNC void assert_fail(const char* file, int line, const char* message) {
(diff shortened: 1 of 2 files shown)


include/fmt/format-inl.h
```diff
@@ -55,7 +55,7 @@ inline fmt::internal::null<> strerror_s(char*, std::size_t, ...) { return {}; }
 FMT_BEGIN_NAMESPACE
 namespace internal {
 
-FMT_FUNC void assert_fail(const char* file, int line, const char* message) {
+FMT_API FMT_FUNC void assert_fail(const char* file, int line, const char* message) {
```
vitaut: I think `FMT_API` is not needed here.

vitaut:
Thanks!

include/fmt/format-inl.h
egorpugin: Oh, maybe. First, I thought header goes to users, but it seems it is fmt internal stuff?

egorpugin: Yep, tested, works without it.

vitaut: It's for the opt-in header-only mode and effectively an internal header.

PR #1451, opened 2019-12-04
Fix handling of types with custom formatters that are convertible to std::string_view

denizevrenci:

I commented at the PR (https://github.com/fmtlib/fmt/pull/1342) as well but types that are convertible to `std::string_view` have ambiguous lookup after that change.

I am not sure if this is the right solution but I wanted to start the discussion here. And this change fixes the ambiguous lookup in my case.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -878,7 +878,9 @@ template <typename Context> struct arg_mapper {
       FMT_ENABLE_IF(
           std::is_constructible<std_string_view<char_type>, T>::value &&
           !std::is_constructible<basic_string_view<char_type>, T>::value &&
-          !is_string<T>::value)>
+          !is_string<T>::value &&
+          !has_formatter<T, Context>::value &&
+          !has_fallback_formatter<T, Context>::value)>
   FMT_CONSTEXPR basic_string_view<char_type> map(const T& val) {
     return std_string_view<char_type>(val);
   }
--- a/test/format-test.cc
+++ b/test/format-test.cc
@@ -1685,6 +1685,28 @@ TEST(FormatterTest, FormatStdStringView) {
   EXPECT_EQ("test", format("{}", std::string_view("test")));
   EXPECT_EQ("foo", format("{}", string_viewable()));
 }
+
+struct explicitly_convertible_to_std_string_view {
+  explicit operator std::string_view() const { return "foo"; }
+};
+
+namespace fmt {
+
+template <>
+struct formatter<explicitly_convertible_to_std_string_view>
+    : formatter<std::string_view> {
+  auto format(const explicitly_convertible_to_std_string_view& v,
+              format_context& ctx) {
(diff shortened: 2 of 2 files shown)


vitaut:
Thanks! Could you add a test case similar to https://github.com/fmtlib/fmt/blob/123e7f7fc33381a07a9d4d28ea2006f5132c503d/test/core-test.cc#L620-L627 ?

denizevrenci:
I dug a little deeper into this issue and understood that the conflict is caused by the existence of a custom formatter. My earlier fix disables the custom formatter and uses the default `std::string_view` formatter. So the test that I just added does not pass.

denizevrenci:
I pushed a new fix that I think addresses this issue.

test/custom-formatter-test.cc
```diff
@@ -54,3 +54,29 @@ TEST(CustomFormatterTest, Format) {
...
+  }
+};
+
+}  // namespace fmt
+
+TEST(CustomFormatterTest, FormatExplicitlyConvertibleToStdStringView) {
+  EXPECT_EQ("'foo'",
+            fmt::format("{}", explicitly_convertible_to_std_string_view()));
+}
+#endif
```
vitaut: Please move the test to `format-test.cc` since `custom-formatter` tests a different API.

test/custom-formatter-test.cc
```diff
@@ -54,3 +54,29 @@ TEST(CustomFormatterTest, Format) {
...
+
+#ifdef FMT_USE_STRING_VIEW
+struct explicitly_convertible_to_std_string_view {
+  explicit operator std::string_view() const { return "foo"; }
+};
+
+namespace fmt {
+
+template <>
+class formatter<explicitly_convertible_to_std_string_view>
```
vitaut: `formatter` must be a struct.

test/custom-formatter-test.cc
```diff
@@ -54,3 +54,29 @@ TEST(CustomFormatterTest, Format) {
...
+    : public formatter<std::string_view> {
+ public:
```
vitaut: Then both `public` keywords can be removed.

test/custom-formatter-test.cc
```diff
@@ -54,3 +54,29 @@ TEST(CustomFormatterTest, Format) {
...
+  template <typename format_context>
```
vitaut: `template <...>` can be removed.

test/custom-formatter-test.cc
```diff
@@ -54,3 +54,29 @@ TEST(CustomFormatterTest, Format) {
...
+  auto format(const explicitly_convertible_to_std_string_view& v,
+              format_context& ctx) {
+    return format_to(ctx.out(), "'{}'", static_cast<std::string_view>(v));
```
vitaut: No need for `static_cast`: `static_cast<std::string_view>(v)` -> `std::string_view(v)`

vitaut:
LGTM, thanks!

PR #1454, opened 2019-12-04
Allow for different begin/end types

parnmatt:
Different types for begin/end iterators are possible for the same range. Multiple libraries have such containers/ranges defined; including range-v3 and C++20 standard ranges.

This issue presented itself as a ghastly template compile error when I tried to use a specific range-v3 view with the fmt library; whereas other views worked just fine. On inspection, it was due to begin/end having different types; and fmt not being able to hand it.

This patch adds the ability to pass a "range" to `fmt::join` and `fmt::arg_join` which has different types.

Whilst making the change, I noticed a couple of structs using `begin` and `end` as member variables rather than member functions. This is not ideal, and breaks compatibility with many libraries, including the standard. Keeping them as structs, would require a different name; to keep consistent naming conventions, `first` and `last` were chosen.

Member functions `begin()` and `end()` were added for clarity and consistency, and to improve usability of the objects with other libraries.

This technically is a breaking change because these were structs, and thus these members were public. If some poor soul is using the `begin` and `end` member variables directly (which they are allowed to do); this change would result in a compile error; which can be fixed by adding parens.

Feel free to cherry pick around the patches that make that change; however, I suggest it is kept, as it improves usability with most libraries that follow the standard's practice.

My suggestion would be to make them classes, and to make `first` and `last` private; and thus can be changed to `begin_` and `end_` to be consistent with the class member variable naming, and be true to their original names. This of course has the added benefit of allowing internal changes, without disrupting the public API.

Technically the addition of allowing a different type is a breaking change, for any code that has explicitly stored `fmt::arg_join` without an `auto`. This *can* be "fixed", either by moving the template parameter to the end of the list, and adding a default:

    template <typename It, typename Char, typename S = It>

The alternative of:

    template <typename It, typename S = It, typename Char = char>

fixes that, but alas, still will break the code of someone who has hardcoded the type.

The only suggestion I would have would be to `std::move` the iterators into `fmt::arg_join` (both from `fmt::join` and `fmt::arg_join::arg_join`. These days iterators may not be as light weight as a pointer (or two); especially smart iterators. Thus copying may be a little expensive. fmt forces these copies with it's value interface; this is fine (and preferred), for sinks; but `std::move`ing the copy into the structure can prevent unnecessary copies (up to 2 if using `fmt::join`).

I played with it, but it caused issues with some of the tests, and wouldn't compile; which is perplexing; and indicates to me that the tests might be relying on something they ought not.

On a similar note, `fmt::arg_join` should not need a constructor, at least as it stands. It should match the criteria of an aggregate type, without the constructor; thus not needing it. However, returning the braced-initializer-lists from `fmt::join` cause compile errors when the constructor is removed. This should not occur for aggregate types; but I cannot determine why.

Touched files were run through `clang-format`.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -3217,22 +3217,22 @@ template <typename T> inline const void* ptr(const std::shared_ptr<T>& p) {
   return p.get();
 }
 
-template <typename It, typename Char>
+template <typename It, typename S, typename Char>
 struct arg_join : internal::view {
   It first;
-  It last;
+  S last;
   basic_string_view<Char> sep;
 
-  arg_join(It b, It e, basic_string_view<Char> s) : first(b), last(e), sep(s) {}
+  arg_join(It f, S l, basic_string_view<Char> s) : first(f), last(l), sep(s) {}
   It begin() const { return first; }
-  It end() const { return last; }
+  S end() const { return last; }
 };
 
-template <typename It, typename Char>
-struct formatter<arg_join<It, Char>, Char>
+template <typename It, typename S, typename Char>
+struct formatter<arg_join<It, S, Char>, Char>
     : formatter<typename std::iterator_traits<It>::value_type, Char> {
   template <typename FormatContext>
-  auto format(const arg_join<It, Char>& value, FormatContext& ctx)
+  auto format(const arg_join<It, S, Char>& value, FormatContext& ctx)
       -> decltype(ctx.out()) {
     using base = formatter<typename std::iterator_traits<It>::value_type, Char>;
     auto it = value.begin();
@@ -3253,13 +3253,13 @@ struct formatter<arg_join<It, Char>, Char>
   Returns an object that formats the iterator range `[begin, end)` with elements
   separated by `sep`.
  */
-template <typename It>
-arg_join<It, char> join(It begin, It end, string_view sep) {
+template <typename It, typename S = It>
+arg_join<It, S, char> join(It begin, S end, string_view sep) {
   return {begin, end, sep};
 }
 
-template <typename It>
-arg_join<It, wchar_t> join(It begin, It end, wstring_view sep) {
+template <typename It, typename S = It>
+arg_join<It, S, wchar_t> join(It begin, S end, wstring_view sep) {
   return {begin, end, sep};
 }
 
@@ -3275,13 +3275,15 @@ arg_join<It, wchar_t> join(It begin, It end, wstring_view sep) {
   \endrst
  */
 template <typename Range>
-arg_join<internal::iterator_t<const Range>, char>
+arg_join<internal::iterator_t<const Range>, internal::iterator_t<const Range>,
+         char>
 join(const Range& range, string_view sep) {
   return join(std::begin(range), std::end(range), sep);
 }
 
 template <typename Range>
-arg_join<internal::iterator_t<const Range>, wchar_t>
+arg_join<internal::iterator_t<const Range>, internal::iterator_t<const Range>,
+         wchar_t>
 join(const Range& range, wstring_view sep) {
   return join(std::begin(range), std::end(range), sep);
 }
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -33,8 +33,6 @@
 #ifndef FMT_FORMAT_H_
 #define FMT_FORMAT_H_
 
-#include "core.h"
-
 #include <algorithm>
 #include <cerrno>
 #include <cmath>
@@ -42,6 +40,8 @@
 #include <limits>
 #include <memory>
 #include <stdexcept>
+
+#include "core.h"
 
 #ifdef __clang__
 #  define FMT_CLANG_VERSION (__clang_major__ * 100 + __clang_minor__)
@@ -3217,12 +3217,15 @@ template <typename T> inline const void* ptr(const std::shared_ptr<T>& p) {
   return p.get();
 }
 
-template <typename It, typename Char> struct arg_join : internal::view {
-  It begin;
-  It end;
+template <typename It, typename Char>
+struct arg_join : internal::view {
+  It first;
+  It last;
   basic_string_view<Char> sep;
 
-  arg_join(It b, It e, basic_string_view<Char> s) : begin(b), end(e), sep(s) {}
+  arg_join(It b, It e, basic_string_view<Char> s) : first(b), last(e), sep(s) {}
+  It begin() const { return first; }
+  It end() const { return last; }
 };
 
 template <typename It, typename Char>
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR. Looks great but please revert changes to `fmt/chrono.h` - the struct there is internal and add a test case for different iterator/sentinel types near https://github.com/fmtlib/fmt/blob/071794ec654a7c503b6214de4032c8ec8a07440b/test/format-test.cc#L1861.

include/fmt/format.h
```diff
@@ -43,6 +41,8 @@
 #include <memory>
 #include <stdexcept>
 
+#include "core.h"
```
vitaut: Please revert this as well.

include/fmt/format.h
```diff
@@ -3217,26 +3217,29 @@ template <typename T> inline const void* ptr(const std::shared_ptr<T>& p) {
...
+template <typename It, typename S, typename Char>
+struct arg_join : internal::view {
+  It first;
+  S last;
   basic_string_view<Char> sep;
 
-  arg_join(It b, It e, basic_string_view<Char> s) : begin(b), end(e), sep(s) {}
+  arg_join(It f, S l, basic_string_view<Char> s) : first(f), last(l), sep(s) {}
+  It begin() const { return first; }
+  S end() const { return last; }
```
vitaut: `begin()/end()` accessors are not needed either - we should discourage use `arg_join` as a range.

include/fmt/format.h
parnmatt: This is following the project's style as per [CONTRIBUTING.md](https://github.com/fmtlib/fmt/blob/ebec00138bd10e949ff34955bcb096387cd32acc/CONTRIBUTING.md), which states to follow the Google Style with a few listed exceptions.

The Google Style [order of includes](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) has project headers at the bottom.

I did not manually do this; this was done by `clang-format` which used the [.clag-format](https://github.com/fmtlib/fmt/blob/071794ec654a7c503b6214de4032c8ec8a07440b/.clang-format) file for this project; where [line 4](https://github.com/fmtlib/fmt/blob/071794ec654a7c503b6214de4032c8ec8a07440b/.clang-format#L4) indicates the Google style which made that change.

If it is intended, for the includes to have a different order; then it should be noted in `CONTRIBUTING.md`, and the relevant options set in the `.clang-format` file (unfortunately I cannot link directly to the options, in the [docs](https://clang.llvm.org/docs/ClangFormatStyleOptions.html), but I believe a mix of `IncludeBlocks` and `IncludeCategories` will allow the customisation).

The final files can then be re-run through `clang-format`, once the project has decided and implemented the format.

include/fmt/format.h
parnmatt: I disagree with this; the object itself is useful, and I can even think of situations where it would be beneficial to be able to use it as a range.

Currently it is a struct, and thus it's internals are in the public API (intended or otherwise); it currently *can* be used as range, and probably is. There is no harm in allowing it to act as one. This change just makes it more standard compiling. (Only suggesting the data to be private)

If you do not intend for the user to "use" it, other than an intermediate type; then please consider making it a class with only private data.

If you do not wish the user to use the structure, why bother having the type, and just return a `std::string`? Or, better still, to allow for different `std::basic_string` types, have implicit conversion operators defined. With the data private.

parnmatt:
> Looks great but please revert changes to `fmt/chrono.h` - the struct there is internal

The struct within chrono.h is in the public API; if it is not intended to be used by the user; it may be best to put it in a `detail` namespace; which is a common practice in many projects. Alternatively, or in addition, make the member variables private; thus preventing the user from accessing them.

> add a test case for different iterator/sentinel types

It is not 'trivial' to add a test case. There is nothing currently in the fmt library, or within any current C++ standard that has two different iterator types. Considering the documentation states C++11 (and the cmake file suggests C++14) standard compliance; this is not possible to test; unless I specifically think and construct a 'range' like type which does use two separate types.

parnmatt:
On second thought; it should be possible to check, for example, `std::vector<T, A>::iterator` vs `std::vector<T, A>::const_iterator`; it is not uncommon for a modifying iterator to be compared to the const iterator (specifically the `cend()`).

This shouldn't be possible with the current version, but possible with this patch.

I will look into it.

parnmatt:
If you *really* do not wish for the changes to the iterators; and just wish for the sentinel change; you are welcome to simply `cherry-pick` the current HEAD of this pull request branch.

include/fmt/format.h
vitaut: Looks like some `clang-format` issue. Which version do you use?

vitaut:
> The struct within chrono.h is in the public API

`parse_range` is private to the `formatter` specialization for durations and not a a part of public API.

vitaut:
Regarding the test, I think a null terminated string range is a simple option because you can use pointer as iterator an empty struct with comparisons against iterator as a sentinel. The ranges paper might even have an example.

include/fmt/format.h
vitaut: It's public only for historical reasons. We should probably make it internal in the next major release.

include/fmt/format.h
parnmatt: I do not believe this is an issue with clang format (I am using `clang-format 9.0.0`)
The Google Style is to have the project includes after all others; and the clang-format in the project root indicates that it is based off of the Google Style, and does not override those options.

The only way I can see there being a conflict, is if you have an untracked `.clang-format` in your copy of the repo at `include/.clang-format` and/or `include/fmt/.clang-format`, with different options, if you are using the automatic detection; or you a pointing it manually to the wrong style file (which I doubt is the case)

parnmatt:
> `parse_range` is private to the `formatter` specialization for durations and not a a part of public API.

Ah yes, you are indeed right, I missed that. I agree that it is not needed; but I still think it is better practice. The first commit in this request can therefore be safely discarded.

> I think a null terminated string range is a simple option

very interesting and simple approach; I will look into that and try to implement the test accordingly

include/fmt/format.h
parnmatt: as it stands, and how it is used; that might be a good shout.

include/fmt/format.h
vitaut: Looks like they added include reordering recently. Anyway, I think it's fine since, as you correctly pointed out, this does follow the style guide.

vitaut:
> I will look into that and try to implement the test accordingly

Any updates?

vitaut:
Closing but another PR with comments addressed will be welcome.

PR #1475, opened 2019-12-12
warning C4468: 'fallthrough': attribute must be followed by a case la…

chronoxor:
…bel or a default label


vitaut:
Good catch, thanks!

PR #1477, opened 2019-12-13
Eliminate unreachable code warnings

gsjaardema:
The code following a `FMT_THROW` is unreachable since the exception does not return.  Some compilers (e.g. Nvidia) emit "Unreachable Code" warnings for this construct.  Since this is in an include file, the warning can be emitted many many times.

### DIFF
--- a/include/fmt/printf.h
+++ b/include/fmt/printf.h
@@ -45,10 +45,7 @@ class printf_precision_handler {
 
   template <typename T, FMT_ENABLE_IF(!std::is_integral<T>::value)>
   int operator()(T) {
-    FMT_THROW(format_error("precision is not integer"));
-#if FMT_MSC_VER
-    return 0;
-#endif
+    throw format_error("precision is not integer");
   }
 };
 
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR. Looks like MSVC builds are failing.

vitaut:
Could you post the full warning and compiler version? Instead of fixing the callsites the compiler should be added to the check here: https://github.com/fmtlib/fmt/blob/9e450911fa21af2271707ab3f554dcbf51ac1750/include/fmt/format.h#L89

vitaut:
Unfortunately this won't work with exceptions disabled. Please see my earlier comment on the recommended solution instead.

gsjaardema:
But if exceptions are disabled, don't you end up with the following:
```
  #define FMT_THROW(x)                                                                               \
    do {                                                                                             \
      static_cast<void>(sizeof(x));                                                                  \
      assert(false);                                                                                 \
    } while (false)
```
And in a release code if `NDEBUG` is defined, this will result in just skipping the error all together won't it?

Also, there are some unprotected throws in compile.h; should those be surrounded by a `FMT_THROW()`?

I will redo the PR

vitaut:
Yes, that's why disabling exceptions is not recommended. Users can also override the default `FMT_THROW`.

vitaut:
> Also, there are some unprotected throws in compile.h; should those be surrounded by a FMT_THROW()?

The whole compile.h is experimental, so I wouldn't bother with changing it yet.

PR #1479, opened 2019-12-15
Template memory_buffer aliases

sylveon:
This allows consumers to write something like `fmt::memory_buffer<50>` instead of `fmt::basic_memory_buffer<char, 50>` to save stack space when they know that the result of the formatting cannot be bigger than 50 characters.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -700,8 +700,11 @@ void basic_memory_buffer<T, SIZE, Allocator>::grow(std::size_t size) {
   if (old_data != store_) Allocator::deallocate(old_data, old_capacity);
 }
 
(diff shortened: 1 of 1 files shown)


sylveon:
Unfortunately I just figured out this is API breaking until P1814 is implemented by compilers, and nobody does yet: https://godbolt.org/z/JjX8hK

Fortunately, we can detect this using `__cpp_deduction_guides >= 201907L`

vitaut:
Thanks for the PR. Considering that `basic_memory_buffer` is a polymorphic class I think it's better not to encourage creating variations of it at call sites. If someone wants a smaller buffer they can create a single alias like `small_memory_buffer` and use that in multiple places.

PR #1480, opened 2019-12-16
Mitigate MSVC issue with min/max macros

parkertomatoes:


vitaut:
Thanks!

PR #1485, opened 2019-12-18
Eliminate NVCC NVidia compiler emits unreachable code warnings

gsjaardema:
Similar to the MSC Compiler, the NVidia NVCC compiler also
emits unreachable code warnings when there is a return
statement following an exception.  These changes eliminate
those warnings.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -117,8 +117,8 @@
 #  endif
 #endif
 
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks

gsjaardema:
Thanks for the great library and guiding me to the correct way to fix the issue.  Took me awhile, but now understand the problems with my first attempt.

PR #1489, opened 2019-12-24
Add FMT_FPIC flag

robinchrist:
In this PR, the `FMT_FPIC` CMake option is added.

If this option is set to `OFF` (default), nothing happens (backwards compatible).
If the option is set to `ON`, the fmt lib is built as position independent code.

This is necessary if you want to built fmt as a static lib and link it into a dynamic lib later

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,7 @@ if (MASTER_PROJECT AND NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE Release CACHE STRING ${doc})
 endif ()
 
+option(FMT_FPIC "If ON, force FMT to be built as position independent code" OFF)
 option(FMT_PEDANTIC "Enable extra warnings and expensive tests." OFF)
 option(FMT_WERROR "Halt the compilation with an error on compiler warnings."
        OFF)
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR but this is already possible to do via [`CMAKE_POSITION_INDEPENDENT_CODE`](https://cmake.org/cmake/help/v3.0/variable/CMAKE_POSITION_INDEPENDENT_CODE.html).

robinchrist:
@vitaut And how am I supposed to use this if I add `fmt` as a library and cannot set this globally?

vitaut:
Why can't you set it globally? Are you compiling some targets with -fPIC and some without?

robinchrist:
Yes, exactly....

I could technically use something like 
```
set_target_properties(fmt PROPERTIES POSITION_INDEPENDENT_CODE ON)
```
at some point (I can't use `fmt::fmt`, CMake does not allow setting target properties for alias targets), however, this is very error prone if you change something else in the buildfile for some reason.

vitaut:
But why? =) Normally you either want all code to be compiled with -fPIC or none.

robinchrist:
We have some static libs which will later get linked into a shared lib, `-fPIC` is necessary for those static libs.
We also have some static libs which are linked directly against an executable which is not position independent.

vitaut:
In that case
```
set_target_properties(fmt PROPERTIES POSITION_INDEPENDENT_CODE ON)
```
is a reasonable solution. The target name is stable.

PR #1492, opened 2019-12-26
warning C4267: 'argument': conversion from 'size_t' to 'DWORD', possi…

chronoxor:
…ble loss of data


vitaut:
Thanks!

PR #1505, opened 2020-01-04
Convert 'char8_t' character sequences to 'char' sequences

DanielaE:
Otherwise, Google Test will insist on inserting 'char8_t' NTBS into 'char' streams, but basic_ostream<char>::operator<< overloads taking 'char8_t' arguments are defined as deleted by P1423.
Handling individual 'char8_t's is done inline.

This fixes the compilation errors seen in C++20 mode beginning with VS2019 Update 2.


vitaut:
Thanks!

PR #1512, opened 2020-01-09
check if _SECURE_SCL is defined not equal to 0

torsten48:


vitaut:
Thanks!

PR #1513, opened 2020-01-10
Add Components for CMake Installation Rules

jwillikers:

Adding components for the CMake installation rules allows users to selectively install different pieces of the library. These are especially useful by parent projects which wish to only install / package the shared library and not the header files, and do not want to have to write their own install rule(s) for the ```fmt``` target in their project.

All components are still installed by default so this behavior is unchanged. For those who wish to only install a particular component, they can leverage CMake commands to do so. As an example, to install only the shared library:
```bash
$ cmake -DCOMPONENT=FMT_Runtime -P cmake_install.cmake
```
If CMake is at least version 3.15, a built-in cmake command is available:
```bash
$ cmake --install . --component FMT_Runtime
```

The other proposed components are ```FMT_Development``` for headers, the static library, pdb file, and package configuration files and ```FMT_Documentation``` for the html documentation. These CMake components follow standard conventions, which include using a project-specific prefix and the common idiom of ```Runtime``` and ```Development``` components.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -247,20 +247,20 @@ if (FMT_INSTALL)
   # Install version, config and target files.
   install(
     FILES ${project_config} ${version_config}
-    DESTINATION ${FMT_CMAKE_DIR})
+    DESTINATION ${FMT_CMAKE_DIR} COMPONENT FMT_Development)
   install(EXPORT ${targets_export_name} DESTINATION ${FMT_CMAKE_DIR}
-          NAMESPACE fmt::)
+          NAMESPACE fmt:: COMPONENT FMT_Development)
 
   # Install the library and headers.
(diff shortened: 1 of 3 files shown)


vitaut:
Thanks for the PR. {fmt} is a small library normally linked statically so I don't think it's worth the trouble further dividing it into components. Moreover it's already possible to skip the tests and docs via CMake config variables which has the advantage that the targets won't be generated at all:

```
cmake -DFMT_DOC=OFF -DFMT_TEST=OFF ...
```

jwillikers:
@vitaut I must be one of those few people that link to {fmt} dynamically, then, and my proposed feature is really only helpful for me in that case. Thanks for taking the time to review my PR 👍 and love the library, really helps me figure out what my error messages are actually saying. 😌

PR #1516, opened 2020-01-13
Use FMT_NOEXCEPT instead of noexcept directly

tohammer:
Otherwise breaks on compilers without noexcept support


vitaut:
Thanks!

PR #1520, opened 2020-01-14
Use C++11-compatible operations

gsjaardema:
The `std::is_base_of<T,U>()` and `std::is_reference<T>()` member functions were added in C++14.  To maintain C++11 compatibility, use the `::value` instead.

Current code fails on intel-17 and other compilers if using strict C++11


vitaut:
Technically this code is C++11-compatible because these are ctor not `operator()` calls but I'm fine merging this as a workaround for the Intel compiler.

vitaut:
Thanks!

PR #1522, opened 2020-01-15
Avoid possible namespace clash for fmt

lefticus:
## Problem

In the case of an existing `fmt` namespace (in my project this looks like `Project::fmt`) it is possible to get a namespace clash in debug builds (MSVC 2017)

## Proposed Solution

When referencing `fmt` internally, be explicit that it is relative to the global namespace using `::fmt`


vitaut:
Thanks!

PR #1523, opened 2020-01-16
Add vcpkg installation instructions

JackBoosY:
fmt is available as a port in vcpkg, a C++ library manager that simplifies installation for fmt and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build fmt , ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/fmt/portfile.cmake). We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -236,6 +236,20 @@ Folly Format             79.9                  445                430
 compare formatting function overhead only. Boost Format is a
 header-only library so it doesn't provide any linkage options.
 
+
+Quick build
+~~~~~~~~~~~~~~~~~
+
+You can download and install fmt using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
+
+    git clone https://github.com/Microsoft/vcpkg.git
+    cd vcpkg
+    ./bootstrap-vcpkg.sh
+    ./vcpkg integrate install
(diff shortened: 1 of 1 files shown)


JackBoosY:
Sorry, I don’t know about the rst format, I submitted too many times.

vitaut:
Thanks! LGTM but please break long lines for readability. {fmt} uses 80 character limit in code and documentation.

README.rst
```diff
@@ -236,6 +236,20 @@ Folly Format             79.9                  445                430
 compare formatting function overhead only. Boost Format is a
 header-only library so it doesn't provide any linkage options.
 
+
+Quick build
+~~~~~~~~~~~~~~~~~
```
vitaut: `~` acts as an underline, so you can remove extra characters =)

vitaut:
Merged. Thinking more of it though this should have gone to https://fmt.dev/latest/usage.html where all installation/build instructions live.

PR #1528, opened 2020-01-20
Fixed format.h(2351,1): warning C4127: conditional expression is constant

chronoxor:


vitaut:
Thanks for the PR. The warning can be suppressed a bit simpler using `const_check` which is done in https://github.com/fmtlib/fmt/commit/fd1cabe464d7bed9739e7b52ef81eea936646098.

chronoxor:
Thanks for the information!

PR #1530, opened 2020-01-21
Fix so can work without locale defined

gsjaardema:
If `FMT_STATIC_THOUSANDS_SEPARATOR` defined, then locale is not included or defined, so this call will be unresolved.  I think this is the correct fix based on the code in `format-inl.h` and `format.h`


vitaut:
Thank you!

Issue #1531, opened 2020-01-22
Internal compiler error when building with MinGW on Windows


ryusakki:
When i tried to build a console app using the header-only version, got the following error:

> In file included from ../fmt/include/fmt/format.h:3536,
                 from ../main.cpp:4:
../fmt/include/fmt/format-inl.h: In function 'int fmt::v6::internal::snprintf_float(T, int, fmt::v6::internal::float_specs, fmt::v6::internal::buffer<char>&)':
../fmt/include/fmt/format-inl.h:69:25: internal compiler error: _Segmentation fault_
 **#define FMT_SNPRINTF snprintf**
../fmt/include/fmt/format-inl.h:1163:25: note: in expansion of macro 'FMT_SNPRINTF'
     auto snprintf_ptr = FMT_SNPRINTF;

This was caused by the snprintf assignment to snprintf_ptr variable. I was able to build successfully after a few changes:

Line 64, change:
`#define FMT_SNPRINTF snprintf`

To:
`#if defined(__MINGW32__)`
`    int (*psnprintf)(char*, std::size_t, char const*, ...) = std::snprintf;`
`    #define FMT_SNPRINTF psnprintf`
`#else`
`    #define FMT_SNPRINTF snprintf`
`#endif`

vitaut:
Does changing

https://github.com/fmtlib/fmt/blob/25d6916b3ae5602a5cb1cf1bb8be1de3d2ecbd0f/include/fmt/format-inl.h#L1150

to 

```c++
int (*snprintf_ptr)(char*, std::size_t, char const*, ...) = FMT_SNPRINTF;
```
and similarly in `format.cc` also "fix" the internal compiler error?

ryusakki:
Yes, it works as well.

vitaut:
Great, thanks for checking. Could you submit a PR with these two small changes and a comment that it's a workaround for MinGW?

ryusakki:
Sure

vitaut:
@ryusakki, could you also post your compiler version here for future reference?

PR #1532, opened 2020-01-22
Workaround for building with MinGW on Windows

ryusakki:
"Solves" #1531 

G++ Version: 8.2.0


vitaut:
Thanks! LGTM but I'd suggest changing "segfault" to "internal compiler error" in the commit message for clarity.

vitaut:
Merged with minor tweaks in https://github.com/fmtlib/fmt/commit/9fc4161f5e905aef3cf8932334f4d5cf440c407a.

PR #1533, opened 2020-01-22
Fix formatting std::chrono::duration types to wide strings

zeffy:
I was getting several compiler errors when trying to format a `std::chrono::duration` to a wide string, and found that various parts of `chrono.h` had a hard dependency on narrow chars. I tried to make it agnostic as possible, and it seems to work fine for me now, although I haven't tested it with anything besides `std::chrono::duration<double, std::milli>`.

If anything needs to be changed or fixed to get this merged just let me know.

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -441,49 +441,26 @@ template <typename Char> struct formatter<std::tm, Char> {
 
 namespace internal {
 
-template <typename Char, typename Period> FMT_CONSTEXPR const Char* get_units() { return nullptr; }
-template <> FMT_CONSTEXPR const char* get_units<char, std::atto>() { return "as"; }
-template <> FMT_CONSTEXPR const wchar_t* get_units<wchar_t, std::atto>() { return L"as"; }
-template <> FMT_CONSTEXPR const char* get_units<char, std::femto>() { return "fs"; }
-template <> FMT_CONSTEXPR const wchar_t* get_units<wchar_t, std::femto>() { return L"fs"; }
-template <> FMT_CONSTEXPR const char* get_units<char, std::pico>() { return "ps"; }
-template <> FMT_CONSTEXPR const wchar_t* get_units<wchar_t, std::pico>() { return L"ps"; }
-template <> FMT_CONSTEXPR const char* get_units<char, std::nano>() { return "ns"; }
-template <> FMT_CONSTEXPR const wchar_t* get_units<wchar_t, std::nano>() { return L"ns"; }
-template <> FMT_CONSTEXPR const char* get_units<char, std::micro>() { 
-  return "µs"; 
-}
-template <> FMT_CONSTEXPR const wchar_t* get_units<wchar_t, std::micro>() { 
-  return L"\u00B5s"; 
-}
-template <> FMT_CONSTEXPR const char* get_units<char, std::milli>() { return "ms"; }
-template <> FMT_CONSTEXPR const wchar_t* get_units<wchar_t, std::milli>() { return L"ms"; }
-template <> FMT_CONSTEXPR const char* get_units<char, std::centi>() { return "cs"; }
-template <> FMT_CONSTEXPR const wchar_t* get_units<wchar_t, std::centi>() { return L"cs"; }
-template <> FMT_CONSTEXPR const char* get_units<char, std::deci>() { return "ds"; }
-template <> FMT_CONSTEXPR const wchar_t* get_units<wchar_t, std::deci>() { return L"ds"; }
-template <> FMT_CONSTEXPR const char* get_units<char, std::ratio<1>>() { return "s"; }
-template <> FMT_CONSTEXPR const wchar_t* get_units<wchar_t, std::ratio<1>>() { return L"s"; }
-template <> FMT_CONSTEXPR const char* get_units<char, std::deca>() { return "das"; }
-template <> FMT_CONSTEXPR const wchar_t* get_units<wchar_t, std::deca>() { return L"das"; }
-template <> FMT_CONSTEXPR const char* get_units<char, std::hecto>() { return "hs"; }
-template <> FMT_CONSTEXPR const wchar_t* get_units<wchar_t, std::hecto>() { return L"hs"; }
-template <> FMT_CONSTEXPR const char* get_units<char, std::kilo>() { return "ks"; }
-template <> FMT_CONSTEXPR const wchar_t* get_units<wchar_t, std::kilo>() { return L"ks"; }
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR! Looks good but could you add a test case checking that `format` compiles when passing a wide format string an a chrono object?

zeffy:
I have committed a wide version of the FormatDefault case, will that work?

zeffy:
Hmm, seems to be failing with the 42µs case. I might need to look at this more.

```C:\projects\fmt\test\chrono-test.cc(157): error: Value of: fmt::format(L"{}", std::chrono::microseconds(42))
  Actual: L"42\xFFC2\xFFB5s"
Expected: L"42µs"
Which is: L"42\xC2\xB5s"```

**Comment by zeffy** [comment]
After some trial and error, I believe the µ issue was from the encoding of the source file itself, so I specialized all of the get_units for wchar_t and in the wide version I used `\u00B5` instead of the µ symbol, to make sure it is encoded correctly.

I also added `Char` template overloads everywhere necessary to make sure everything is being done with wide characters. I think this should pass all the tests.

File: include/fmt/chrono.h [hunk 1]
```diff
@@ -759,18 +777,23 @@ inline std::chrono::duration<Rep, std::milli> get_milliseconds(
...
 }
 
-template <typename Period, typename OutputIt>
+template <typename Char, typename Period, typename OutputIt>
 OutputIt format_duration_unit(OutputIt out) {
-  if (const char* unit = get_units<Period>()) return format_to(out, "{}", unit);
-  if (Period::den == 1) return format_to(out, "[{}]s", Period::num);
-  return format_to(out, "[{}/{}]s", Period::num, Period::den);
+  const Char format[] { '{', '}', 0 };
+  if (const Char* unit = get_units<Char, Period>()) return format_to(out, format, unit);
```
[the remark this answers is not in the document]
vitaut: Correction: [`utf8_to_utf16`](https://github.com/fmtlib/fmt/blob/09a13244c85de35471ff7a660b17360111b14430/include/fmt/format.h#L948) + copy because of micro.

include/fmt/chrono.h
```diff
@@ -871,13 +894,13 @@ struct chrono_formatter {
   void write_pinf() { std::copy_n("inf", 3, out); }
   void write_ninf() { std::copy_n("-inf", 4, out); }
 
-  void format_localized(const tm& time, const char* format) {
+  void format_localized(const tm& time, const char_type* format) {
```
vitaut: Is this actually needed? Locale is not parameterized by char type.

zeffy: Good point, I've reverted it.

zeffy: After looking at this a bit more and understanding it better I think using the `put(iter_type, std::ios_base&, char_type fill, const std::tm* t, char, char)` overload is preferable here, as the overload that takes iterators for the format string needs to match the output as I did previously.

include/fmt/chrono.h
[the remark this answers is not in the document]
zeffy: Still need a way to determine if it should just do `copy` if `Char` is narrow, instead `utf8_to_utf16` then `copy`. How would you like me to approach this? Is `if FMT_CONSTEXPR (std::is_same<Char, wchar_t>::value)` allowed?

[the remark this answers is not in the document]
zeffy: I went with the `if constexpr` solution, since it was the most compact way I could think of. It seems to work well. Let me know if it is satisfactory.

include/fmt/chrono.h
```diff
@@ -440,9 +440,7 @@ template <typename Char> struct formatter<std::tm, Char> {
 };
 
 namespace internal {
-template <typename Period> FMT_CONSTEXPR const char* get_units() {
-  return nullptr;
-}
+template <typename Period> FMT_CONSTEXPR const char* get_units() { return nullptr; }
```
vitaut: You should probably format this and the rest of file with `clang-format`.

include/fmt/chrono.h
```diff
@@ -871,13 +877,13 @@ struct chrono_formatter {
...
 
-  void format_localized(const tm& time, const char* format) {
+  void format_localized(const tm& time, char format, char modifier = 0) {
     if (isnan(val)) return write_nan();
     auto locale = context.locale().template get<std::locale>();
     auto& facet = std::use_facet<std::time_put<char_type>>(locale);
     std::basic_ostringstream<char_type> os;
     os.imbue(locale);
-    facet.put(os, os, ' ', &time, format, format + std::strlen(format));
+    facet.put(os, os, ' ', &time, format, modifier);
```
vitaut: Good idea.

include/fmt/chrono.h
```diff
@@ -759,18 +753,30 @@ inline std::chrono::duration<Rep, std::milli> get_milliseconds(
...
 }
 
-template <typename Period, typename OutputIt>
+template <typename Char, typename Period, typename OutputIt>
 OutputIt format_duration_unit(OutputIt out) {
-  if (const char* unit = get_units<Period>()) return format_to(out, "{}", unit);
-  if (Period::den == 1) return format_to(out, "[{}]s", Period::num);
-  return format_to(out, "[{}/{}]s", Period::num, Period::den);
+  if (const char* unit = get_units<Period>()) {
+    string_view s(unit);
```
vitaut: I suggest making `unit` a `string_view` and removing `s`.

include/fmt/chrono.h
```diff
@@ -759,18 +753,30 @@ inline std::chrono::duration<Rep, std::milli> get_milliseconds(
...
+    if (std::is_same<Char, wchar_t>::value) {
```
vitaut: Please wrap the condition in the call to `const_check` to prevent a warning about constant condition.

include/fmt/chrono.h
```diff
@@ -759,18 +753,30 @@ inline std::chrono::duration<Rep, std::milli> get_milliseconds(
...
+      utf8_to_utf16 u(s);
+      return std::copy(u.c_str(), u.c_str() + u.size(), out);
+    } else {
```
vitaut: You can remove else thanks to the early return above.

include/fmt/chrono.h
zeffy: I'll need to rework some things to do this properly since `get_units` can return a `nullptr`, which fmt's `string_view` doesn't handle properly as far as I know. Since it passes the pointer to `std::char_traits<T>::length()` which will dereference it without checking.

So it either needs to stay as-is, or I can return some other magic value from `get_units` default overload, like an empty string.

vitaut: Let's leave as is this then.

vitaut:
Merged, thanks!

Issue #1264, opened 2019-08-12
Wrong library name in fmt.pc when building Debug.


ambitslix:
The wrong library name is configured in fmt.pc file when with CMAKE_BUILD_TYPE=Debug. 

As a result

$pkg-config --libs fmt

Returns -lfmt however the library name is actually "fmtd". The solution is to configure fmt.pc via variable. For example
...
prefix=@CMAKE_INSTALL_PREFIX@
exec_prefix=@CMAKE_INSTALL_PREFIX@
libdir=@CMAKE_INSTALL_FULL_LIBDIR@
includedir=@CMAKE_INSTALL_FULL_INCLUDEDIR@
library=@FMT_LIBRARY_NAME@

Name: fmt
Description: A modern formatting library
Version: @FMT_VERSION@
Libs: -L${libdir} -l${library}
Cflags: -I${includedir}

Alternately change:

CMakeLists.txt:183

set_target_properties(fmt PROPERTIES
  VERSION ${FMT_VERSION} SOVERSION ${CPACK_PACKAGE_VERSION_MAJOR}
  **DEBUG_POSTFIX d**)

It seems unnecessary to have different names for library builds?

vitaut:
`fmt.pc` currently expects that the release version of the library is installed. If you want to add support for debug libraries there, please submit a PR.

> It seems unnecessary to have different names for library builds?

It is convenient to distinguish between debug and release versions of the library and it allows placing them in one directory.

ambitslix:
Debug libraries are easily generated separately by changing
CMAKE_INSTALL_PREFIX. The issue is in my view in generating wrong pc file.
It should derive from target file name.

ambitslix:
I'm glad to add support for this are you ok with my solution in the ticket?

vitaut:
> are you ok with my solution in the ticket?

I'm fine with the solution "to configure fmt.pc via variable". Thanks!

PR #1534, opened 2020-01-25
Proposed fix for issue #1264

ambitslix:
Fixes for issue:
https://github.com/fmtlib/fmt/issues/1264


CMakeLists.txt
```diff
@@ -178,6 +178,15 @@ set_target_properties(fmt PROPERTIES
   VERSION ${FMT_VERSION} SOVERSION ${CPACK_PACKAGE_VERSION_MAJOR}
   DEBUG_POSTFIX d)
 
+# Configure pkg-config fmt.pc properly
+string(TOLOWER ${PROJECT_NAME} PROJECT_NAME_LOWER)
```
vitaut: Please use `fmt` instead of `PROJECT_NAME_LOWER` because the project name is not related to the library target or output name (although they happen to be the same).

vitaut:
Thanks for the PR!

CMakeLists.txt
```diff
@@ -178,6 +178,15 @@ set_target_properties(fmt PROPERTIES
...
+if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+  get_target_property(_FMT_LIB_POSTFIX ${PROJECT_NAME_LOWER} DEBUG_POSTFIX)
+  set(FMT_LIB_NAME ${PROJECT_NAME_LOWER}${_FMT_LIB_POSTFIX})
```
vitaut: I suggest defining `FMT_DEBUG_POSTFIX` and using it here and to set target postfix in line 177 above instead of getting it from the properties.

ambitslix: Ok that sounds good.

ambitslix: Like so?
```
set(FMT_DEBUG_POSTFIX d)

set_target_properties(fmt PROPERTIES
  OUTPUT_NAME fmt
  VERSION ${FMT_VERSION}
  SOVERSION ${CPACK_PACKAGE_VERSION_MAJOR}
  DEBUG_POSTFIX ${FMT_DEBUG_POSTFIX})

# Configure pkg-config fmt.pc properly
get_target_property(FMT_LIB_NAME fmt OUTPUT_NAME)
if(CMAKE_BUILD_TYPE STREQUAL "Debug")
  set(FMT_LIB_NAME ${FMT_LIB_NAME}${FMT_DEBUG_POSTFIX})
endif()
```

vitaut: Exactly. This simplifies the logic a bit.

ambitslix: Pushed.

vitaut:
Merged in https://github.com/fmtlib/fmt/commit/6c30f41443629ddc647875887768d1af17815409, thanks!

PR #1535, opened 2020-01-27
Fix symbol visibility on Linux when compiling with -fvisibility=hidden

milianw:
Make FMT_API symbols use the default visibility on non-Windows
platforms. Otherwise, one cannot use the generated fmt library when
compiling globally with -fvisibility=hidden.

Fixes compile errors like:

```
../3rdParty/fmt/include/fmt/core.h:757: error: undefined reference to 'fmt::v6::internal::assert_fail(char const*, int, char const*)'
```

Note that the symbol exists, but is local:

```
$ nm -C libfmtd.so.6.1.3  | grep assert_fail
                 U __assert_fail
0000000000233ffa t fmt::v6::internal::assert_fail(char const*, int, char const*)
```

With this patch, the compile error is gone and the symbol is properly
exported:

```
$ nm -a bin/libfmtd.so -C | grep assert_fail
                 U __assert_fail
00000000002366ba T fmt::v6::internal::assert_fail(char const*, int, char const*)
```

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -185,10 +185,19 @@
 #  define FMT_CLASS_API
 #endif
 #ifndef FMT_API
-#  define FMT_API
+#  if FMT_GCC_VERSION || FMT_CLANG_VERSION
+#    define FMT_API __attribute__((visibility("default")))
+#    define FMT_EXTERN_TEMPLATE_API FMT_API
+#    define FMT_INSTANTIATION_DEF_API
+#  else
+#    define FMT_API
+#  endif
 #endif
 #ifndef FMT_EXTERN_TEMPLATE_API
 #  define FMT_EXTERN_TEMPLATE_API
+#endif
+#ifndef FMT_INSTANTIATION_DEF_API
+#  define FMT_INSTANTIATION_DEF_API FMT_API
 #endif
 
 #ifndef FMT_HEADER_ONLY
--- a/src/format.cc
+++ b/src/format.cc
@@ -121,7 +121,7 @@ template FMT_API char* internal::sprintf_format(long double,
                                                 internal::buffer<char>&,
                                                 sprintf_specs);
 
-template struct FMT_API internal::basic_data<void>;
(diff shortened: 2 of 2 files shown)


vitaut:
Thanks for the PR. Looks good but please fix the CI errors.

milianw:
I'm quite puzzled by the CI errors, since I cannot reproduce them locally. Apparently it's testing some different compile configuration there than what I have locally. I'll have to find the time to figure out what's going on here. If you have any hints that would be appreciated.

milianw:
Ah, `-Werror` explains it - I do see it locally after all...

milianw:
this should fix the linux build, but I wonder if the `FMT_API` in the `.cpp` file was required for Windows?

vitaut:
> I wonder if the FMT_API in the .cc file was required for Windows?

AFAIK yes, it is required by MSVC. I suggest introducing another macro (sigh) `FMT_INSTANTIATION_DEF` and using it on the instantiation definition instead of `FMT_API`.

milianw:
With the AppVeyor build passing... do we really need this? Shouldn't that test this API too, just like the Linux CI on Travis did?

vitaut:
> do we really need this?

We do. CI doesn't test all possible configurations, just several common ones.

include/fmt/core.h
```diff
@@ -185,11 +185,20 @@
 #  define FMT_CLASS_API
 #endif
 #ifndef FMT_API
-#  define FMT_API
+#  if defined(_WIN32)
+#    define FMT_API
+#  else
+#    define FMT_API __attribute__((visibility("default")))
```
vitaut: One more thing: since `__attribute__` is a non-standard construct it should be conditionally compiled, e.g.

```c++
#   if FMT_GCC_VERSION || FMT_CLANG_VERSION
#     define FMT_API __attribute__((visibility("default")))
#     define FMT_EXTERN_TEMPLATE_API FMT_API
#     define FMT_INSTANTIATION_DEF_API
#   else
#     define FMT_API
#   endif
```

vitaut:
Thanks!

PR #1546, opened 2020-02-07
Respect CMAKE_RUNTIME_OUTPUT_DIRECTORY if already set. Fixes #1545

### LINKED ISSUE #1545: Respect Parent Project's use of CMAKE_RUNTIME_OUTPUT_DIRECTORY
The use of `CMAKE_RUNTIME_OUTPUT_DIRECTORY` in the CMakeLists.txt file does not allow parent projects to override it.
Would it be possible to only set that variable if it is not already set?
The CMake code to accomplish follows below.
```cmake
if(NOT CMAKE_RUNTIME_OUTPUT_DIRECTORY)
  set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
endif()
```

jwillikers:

Simply check that `CMAKE_RUNTIME_OUTPUT_DIRECTORY` is not set before modifying it, otherwise leave it alone.

See Issue #1545 for details.


vitaut:
Thanks!

PR #1553, opened 2020-02-15
add const begin and end overload to buffer

dominicpoeschko:
since there is a const overload for data I think there should also be one for begin and end


vitaut:
Thanks. Could you please apply clang-format to your changes?

PR #1554, opened 2020-02-17
Fix FMT_FORMAT_AS const specifier position

fghzxm:
The current `FMT_FORMAT_AS` macro will make `formatter<Char *>::format`
have the first argument type `const Char *&` which is incorrect an
should be `Char *const &`.  This pull request fixes that by changing the
first argument type in the macro definition body from `const Type &` to
`Type const &`.


vitaut:
Good catch, thanks!

PR #1560, opened 2020-02-20
improved use of find

IkarusDeveloper:
*begin is supposed to be different from '{' when this find is used, so we can avoid checking it.


vitaut:
LGTM but please apply clang-format. Thanks!

IkarusDeveloper:
I m Sorry for the bad format, i have changed it anyway.

vitaut:
Thank you

PR #1568, opened 2020-02-29
Move FMT_CLANG_VERSION definition to core.h

refnum:
Previously format.h defined FMT_CLANG_VERSION after including core.h, however core.h tests FMT_CLANG_VERSION when it defines FMT_API.

format.h still defines FMT_ICC_VERSION and FMT_CUDA_VERSION, however given that core.h now defines clang/gcc/msc versions it may be worth moving that version detection into core.h too.


vitaut:
Good catch, thanks.

PR #1569, opened 2020-02-29
Add FMT_NORETURN to assert_fail prototype.

refnum:
When building with -Werror,-Wmissing-noreturn clang identifies that assert_fail could be declared with the 'noreturn' attribute.


vitaut:
Thanks!

PR #1570, opened 2020-03-01
Make underlying type of align_t / sign_t unsigned.

refnum:
When building with -Werror,-Wsigned-enum-bitfield clang reports that "enums in the Microsoft ABI are signed integers by default; consider giving the enum type an unsigned underlying type to make this code portable" for these two types.

Under this ABI one bit of an enum bitfield is used by the sign bit. Enum values large enough to overlap that bit would be seen as negative integers if read out into an integer.

There may not be any such reading out at present, but making the underlying type unsigned ensures consistent results across platforms.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -997,12 +997,12 @@ template <typename Char> struct fill_t {
 // We cannot use enum classes as bit fields because of a gcc bug
 // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61414.
 namespace align {
-enum type { none, left, right, center, numeric };
+enum type : uint8_t { none, left, right, center, numeric };
 }
 using align_t = align::type;
 
 namespace sign {
-enum type { none, minus, plus, space };
+enum type : uint8_t { none, minus, plus, space };
 }
 using sign_t = sign::type;
 


refnum:
As per the [bug 61414](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61414) comment above these enums, it looks like assigning an underlying type generates an incorrect "is too small to hold all values" warning regardless of wether the values would fit the underlying type or not.

Godbolt shows this has been fixed on gcc-trunk, but is still present in gcc 9.2.0, so realistically there is no way to avoid the gcc error if you normally build with `-Werror=signed-enum-bitfield` in clang.

As such I'll just suppress that clang warning locally when building {fmt}, so you can ignore this PR.

It might be worth adding a link to [bug 51242](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=51242#c28) to the comment as well, as that includes a comment from a clang dev comparing the behaviour with their flag.

vitaut:
Closing per

> As such I'll just suppress that clang warning locally when building {fmt}, so you can ignore this PR.

> It might be worth adding a link to bug 51242 to the comment as well, as that includes a comment from a clang dev comparing the behaviour with their flag.

There is an indirect link to this bug via https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61414 which is similar.

PR #1571, opened 2020-03-01
Fix simple -Wsign-conversion cases.

refnum:
In both of these commits the variables can only hold positive values and are only passed to methods that take size_t parameters.


vitaut:
Thanks!

PR #1572, opened 2020-03-02
Fix clang -Wsign-conversion warning in bigint::subtract_bigits.

refnum:
subtract_bigits's index parameter is only passed to basic_memory_buffer::operator[] (size_t).

subtract_bigits is only called from subtract_aligned, which passes it 'i' for the index.

subtract_aligned initialises 'i' to 'other.exp_ - exp_', after asserting that other.exp_ >= exp_. As such the result of this subtraction must always be positive, and so can safely be cast to a size_t.

As 'i' is only ever incremented after its assignment it too can be a size_t.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -508,7 +508,7 @@ class bigint {
 
   friend struct formatter<bigint>;
 
-  void subtract_bigits(int index, bigit other, bigit& borrow) {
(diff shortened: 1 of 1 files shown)


vitaut:
As in the other diff:

> Signed integers are normally used for indexing in {fmt} unless wrap around behavior is needed.

PR #1573, opened 2020-03-02
Fix clang -Wsign-conversion warning in grisu_count_digits.

refnum:
grisu_count_digits is only used by grisu_gen_digits, which assigns the unsigned result to a (signed) int.

Although grisu_count_digits always returns a positive integer this keeps its return type in sync with the type its result is assigned to.


vitaut:
Thanks!

PR #1574, opened 2020-03-02
Fix clang -Wsign-conversion warning in fallback_format.

refnum:
num_digits is initialised to 0, and is always incremented by 1 prior to usage where at most 1 is subtracted from it.


vitaut:
Signed integers are normally used for indexing in {fmt} unless wrap around behavior is needed.

PR #1575, opened 2020-03-02
Fix clang -Wdisabled-macro-expansion warning from FMT_STRING_IMPL.

refnum:
FMT_STRING_IMPL has an internal helper named FMT_STRING, however FMT_STRING is also the name of the macro that invokes FMT_STRING_IMPL.

Renaming this helper avoids the appearance of a recursive macro.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -3547,7 +3547,7 @@ FMT_END_NAMESPACE
 #define FMT_STRING_IMPL(s, ...)                              \
   [] {                                                       \
     /* Use a macro-like name to avoid shadowing warnings. */ \
-    struct FMT_STRING : fmt::compile_string {                \
+    struct FMT_COMPILE_STRING : fmt::compile_string {        \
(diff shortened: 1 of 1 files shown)


include/fmt/format.h
```diff
@@ -3504,15 +3504,15 @@ FMT_END_NAMESPACE
 #define FMT_STRING_IMPL(s, ...)                              \
   [] {                                                       \
     /* Use a macro-like name to avoid shadowing warnings. */ \
-    struct FMT_STRING : fmt::compile_string {                \
+    struct INVOKE_FMT_STRING : fmt::compile_string {         \
```
vitaut: I suggest renaming to `FMT_COMPILE_STRING` to follow the naming convention for {fmt} macros as per comment above.

refnum: Updated to use FMT_COMPILE_STRING as suggested.

vitaut:
Thanks!

PR #1576, opened 2020-03-02
Fix clang -Wunused-member-function warnings with FMT_MAYBE_UNUSED.

refnum:
FMT_MAYBE_UNUSED currently expands to the [[maybe_unused]] attribute on C++17.

It could be extended to previous language versions with a compiler-specific attribute if required.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -36,6 +36,12 @@
 #  define FMT_HAS_CPP_ATTRIBUTE(x) 0
 #endif
 
+#define FMT_HAS_CPP14_ATTRIBUTE(attribute) \
+  (__cplusplus >= 201402L && FMT_HAS_CPP_ATTRIBUTE(attribute))
+
+#define FMT_HAS_CPP17_ATTRIBUTE(attribute) \
+  (__cplusplus >= 201703L && FMT_HAS_CPP_ATTRIBUTE(attribute))
+
 #ifdef __clang__
 #  define FMT_CLANG_VERSION (__clang_major__ * 100 + __clang_minor__)
 #else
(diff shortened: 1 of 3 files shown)


include/fmt/core.h
```diff
@@ -132,6 +132,14 @@
 #  define FMT_NORETURN
 #endif
 
+#ifndef FMT_MAYBE_UNUSED
+#  if (__cplusplus >= 201703L)
```
vitaut: I suggest replacing this check with `FMT_HAS_CPP_ATTRIBUTE(maybe_unused)`.

refnum: Updated to use FMT_HAS_CPP_ATTRIBUTE as requested.

vitaut: > Perhaps a more general approach, to test both language version and compiler support within that version

Good idea.

refnum: Great, thanks - I've added a FMT_HAS_CPP14_ATTRIBUTE to cover FMT_DEPRECATED as well so all of the version+attribute checks now use this approach.

include/fmt/core.h
```diff
@@ -36,6 +36,9 @@
 #  define FMT_HAS_CPP_ATTRIBUTE(x) 0
 #endif
 
+#define FMT_HAS_CPP14_ATTRIBUTE(_attribute) (__cplusplus >= 201402L && FMT_HAS_CPP_ATTRIBUTE(_attribute))
```
vitaut: Why underscore in `_attribute`?

vitaut:
LGTM but please apply clang-format to your changes. Thanks!

include/fmt/core.h
refnum: I normally underscore macro parameters to help distinguish them from regular variable names.

That's not the normal fmt style though so I've removed them and run clang-format as requested.

PR #1577, opened 2020-03-02
Fix clang -Wsign-conversion warning in bigint::remove_leading_zeros.

refnum:
Rather than subtracting 1 from the size before the loop, and potentially going negative, the loop can test there is at least one element before subtracting for the dereference.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -515,9 +515,12 @@ class bigint {
   }
 
   void remove_leading_zeros() {
-    int num_bigits = static_cast<int>(bigits_.size()) - 1;
-    while (num_bigits > 0 && bigits_[num_bigits] == 0) --num_bigits;
-    bigits_.resize(num_bigits + 1);
+    size_t old_bigits = bigits_.size();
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR but I'd prefer to keep signed integers for consistency.

refnum:
No problem. I prefer unsigned to avoid UB, however I appreciate that they do sometimes make indexing logic a bit more cumbersome vs signed!

That said, I did notice the size_t approach produced smaller code. That might be useful to consider if this is a significant routine performance-wise.

A comparison is at https://godbolt.org/z/WgPB8S.

The int approach compiles to 27 instructions on gcc and 24 on clang.

The size_t approach compiles to 22 instructions on gcc and 24 on clang, or 19 and 18 respectively if you simply call bigits_.resize() regardless.

vitaut:
> That might be useful to consider if this is a significant routine performance-wise.

I don't think it is. It doesn't show up in perf profile.

Looks like the code size difference is mostly due to change in size (`int` is 32-bit and `size_t` is 64-bit on this platform).

PR #1578, opened 2020-03-04
Rename MASTER_PROJECT to use FMT_ prefix

ClausKlein:
This prevent name conflicts with other projects

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,9 +9,9 @@ endif()
 
 # Determine if fmt is built as a subproject (using add_subdirectory)
 # or if it is the master project.
-set(MASTER_PROJECT OFF)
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR but I don't think there is a conflict because `MASTER_PROJECT` is local to the fmt directory scope: https://cmake.org/cmake/help/latest/command/set.html#set-normal-variable

vitaut:
Closing per comment above.

ClausKlein:
Yes, if you see you project isolated.

We use i.e. fmt as a subproject and what happens if we use MASTER_PROJECT too?

vitaut:
it shouldn't affect your `MASTER_PROJECT` because CMake variables are local to the directory by default (unless `PARENT_SCOPE` is specified).

Issue #1455, opened 2019-12-05
FMT_STRING does not compile with "if constexpr()"


rianquinn:
If you attempt to compile FMT_STRING with an if constexpr, you get the following with GCC 9

```
if constexpr (contracts_throw && !autosar_compliant) {
    std::string s = fmt::format(FMT_STRING("{}"), "foo");
}
```

```
/home/rianquinn/working/bsl/tests/int/../../include/bsl/contracts.hpp:120:68: error: invalid use of void expression
  120 |                 std::string s = fmt::format(FMT_STRING("{}"), "foo");
```

rianquinn:
If I remove the constexpr, this works fine.

eliaskosunen:
Looks like a bug with gcc: https://godbolt.org/z/krYNm9
If you look at the error messages, gcc thinks that the return type of the lambda is `void`, while clang correctly detects it to be `str`. This only happens if the `if constexpr` condition evaluates to `false`.

PR #1580, opened 2020-03-06
Workaround for issue #1455 (gcc error on FMT_STRING within if constexpr).

refnum:
This fixes issue #1455 by making the return type explicit, and adds a C++17-specific test to demonstrate the bug.

This allows usage like:

```
#define LOG(_messsage, ...)                                     \
    do                                                          \
    {                                                           \
        if constexpr(_uses_format_specifiers(_message))         \
        {                                                       \
            LogWithFormat(FMT_STRING(_message), ##__VA_ARGS__)  \
        }                                                       \
        else                                                    \
        {                                                       \
            LogWithPrintf(_message, ##__VA_ARGS__)              \
        }                                                       \
    }                                                           \
    while (false)
```
Where _uses_format_specifiers is a constexpr function that uses some heuristic to decide if a formatting string is fmt-style or printf-style.

We are using this technique to move a large codebase from printf-based logging to fmt-based and hit this same bug with gcc 9.

Edit: Nothing's ever easy! I will check the other compilers on Travis and update.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -3544,18 +3544,18 @@ FMT_CONSTEXPR internal::udl_arg<wchar_t> operator"" _a(const wchar_t* s,
 #endif  // FMT_USE_USER_DEFINED_LITERALS
 FMT_END_NAMESPACE
 
-#define FMT_STRING_IMPL(s, ...)                              \
-  [] {                                                       \
-    /* Use a macro-like name to avoid shadowing warnings. */ \
-    struct FMT_COMPILE_STRING : fmt::compile_string {        \
-      using char_type = fmt::remove_cvref_t<decltype(*s)>;   \
-      __VA_ARGS__ FMT_CONSTEXPR                              \
-      operator fmt::basic_string_view<char_type>() const {   \
-        /* FMT_STRING only accepts string literals. */       \
-        return fmt::internal::literal_to_view(s);            \
-      }                                                      \
-    };                                                       \
-    return FMT_COMPILE_STRING();                             \
+#define FMT_STRING_IMPL(s, ...)                                       \
+  []() -> fmt::basic_string_view<fmt::remove_cvref_t<decltype(*s)>> { \
+    /* Use a macro-like name to avoid shadowing warnings. */          \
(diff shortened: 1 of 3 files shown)


include/fmt/format.h
```diff
@@ -3544,8 +3544,14 @@ FMT_CONSTEXPR internal::udl_arg<wchar_t> operator"" _a(const wchar_t* s,
 #endif  // FMT_USE_USER_DEFINED_LITERALS
 FMT_END_NAMESPACE
 
+#if FMT_GCC_VERSION >= 840
+#  define FMT_STRING_RESULT(s)   ()->fmt::basic_string_view<fmt::remove_cvref_t<decltype(*s)>>
```
vitaut: I think this will break compile-time string detection: https://github.com/fmtlib/fmt/blob/1e8493196efa4e05baa22461ba61b636ab2e662b/include/fmt/core.h#L460

vitaut:
Closing per comment above.

refnum:
Hi,

Yes, unfortunately I think that's the case.

The is_compile_string version of compiled_format_base() relies on being able to obtain the string by creating a new instance of its type with to_string_view(S()).

That instance has to return the same text as seen by FMT_STRING, so effectively each FMT_STRING needs to return a unique type.

I experimented with adding a <bool IsCompiled> to basic_string_view, to replace the compile_string type, but couldn't get a constexpr-preserving conversion between the two types of basic_string_view working.


It looks like the bug exists from gcc 7-9 but has now been fixed in gcc 10.

For our purposes we can work around it by moving the common FMT_STRING() out of the constexpr if:

```
#define LOG(_messsage, ...)                                     \
    do                                                          \
    {                                                           \
        auto _fmt_message = FMT_STRING(_message);               \
                                                                \
        if constexpr(_uses_format_specifiers(_message))         \
        {                                                       \
            LogWithFormat(_fmt_message, ##__VA_ARGS__)          \
        }                                                       \
        else                                                    \
        {                                                       \
            LogWithPrintf(_fmt_message, ##__VA_ARGS__)          \
        }                                                       \
    }                                                           \
    while (false)
```

PR #1582, opened 2020-03-09
Fix empty debug postfix

stac47:

Passing an empty string to FMT_DEBUG_POSTFIX leads to the following error:

```
CMake Error at CMakeLists.txt:179 (set_target_properties):
  set_target_properties called with incorrect number of arguments.
```
This PR fixes this.


vitaut:
Thanks

Issue #1170, opened 2019-05-21
Add an API for dynamic construction of argument lists


vitaut:
Currently the number and types (but not values) of formatting arguments has to be known at compile time. Consider adding an API to allow fully dynamic construction of argument lists to support use cases like the one mentioned in #1168:

> I have a few use cases, but they all boil down to wanting to externalize the format of whatever needs to be produced from the C++ code. So, for example, a config file might have a format string in it and that format string wants to extract zero or more named values from some data dictionary in the program.

vincentlaucsb:
Adding on to #1168, I was working on a resume generator and ran into a similar issue.

One of the features I was working on was the ability for the user to define HTML templates like below:

```
        <header>
          <h1 id="name">Vincent La</h1>
          <p>
            <span>{Address}</span>
            <span>{Email}</span>
            <span>{LinkedIn}</span>
            <span>{Website}</span>
          </p>
        </header>
```

The main problem is that I want to be able to define the templates without having to touch C++ code. That means the parameters have to been parsed from some sort of data file, so the formatting arguments have to be constructed at runtime. In Python, I would just build a `dict` of arguments and pass them into `str.format()` directly. I was disappointed I couldn't do that here, and ended up writing my own function to do the formatting.

With that being said, I really liked a lot of the other features this library provided. I'm willing to help implement this if you have good pointers on where to start.

vitaut:
I think the easiest solution is to implement something like `dynamic_arg_store` which is similar to [`format_arg_store`](https://github.com/fmtlib/fmt/blob/afdbbac75c60b5f5176e99d9bc27454446c5d77d/include/fmt/core.h#L1084) but owns the arguments and allows adding them dynamically. It can provide a conversion operator to `basic_format_args` that will allow passing it to `vformat` and other formatting functions once it's populated.

joyqat:
Qt has a format API which looks like `QString("%1, %2!").arg("hello").arg("world")`, this operation will look and replace the placeholder `%1` and `%2` inside `str`.
Seems not so difficult to realize?
But in this way performance may not be as good.

abrownsword:
Adding dynamic_arg_store as described would be extremely useful.  My typical use case is that I have a std::map<std::string, my_type> and I want to add one or more of these to the dynamic_arg_store and then pass this to vformat.  Ideally I could keep the dynamic_arg_store around and update it in place, and also used the new compiled format object with it to speed up the operation.

slonm:
[...]

PR #1584, opened 2020-03-09
Dynamic arguments storage. Implementation of enhancement from issue #1170.

vsolontsov-ll:

(Hope I didn't miss any guidelines.)


vsolontsov-ll:
Travis builds are fixed. Let's give a try with appveyor (don't know how to configure it for my fork yet)

include/fmt/dyn-args.h
```diff
@@ -0,0 +1,193 @@
...
+  \rst
+  A dynamic version of `fmt::format_arg_store<>`.
+  It's equipped with a storage to potentially temporary objects which lifetime
+  could be shorter than the format arguments object.
+
+  It can be implicitly converted into `~fmt::basic_format_args` for passing
+  into type-erased formatting functions such as `~fmt::vformat`.
+  \endrst
+ */
+template <typename Context, typename... Args>
```
vitaut: The main point of dynamic argument store is that argument types may not be known at compile time, so this shouldn't be parameterized on `Args`. Custom args should be type erased at the time of insertion.

Also I don't think it's worth adding a new header just for this class. Please merge into `fmt/format.h`.

include/fmt/dyn-args.h
```diff
@@ -0,0 +1,193 @@
...
+class dynamic_format_arg_store
+#if FMT_GCC_VERSION < 409
+    // Workaround a GCC template argument substitution bug.
+    : public basic_format_args<Context>
+#endif
+{
+ private:
+  using char_type = typename Context::char_type;
+  static const bool is_packed = false;
```
vitaut: Unused.

include/fmt/core.h
```diff
@@ -969,6 +970,11 @@ template <typename Context> struct arg_mapper {
     return val;
   }
 
+  FMT_CONSTEXPR const named_arg_base<char_type>& map(
+      const named_arg_base<char_type>& val) {
+    return val;
+  }
```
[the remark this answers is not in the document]
vsolontsov-ll: That's a place I actually wanted you to consult me...

`void push_back(const internal::named_arg<T, char_type>& arg)` calls `internal::make_arg()` on a serialized `named_arg_base` (internal::named_arg::value copied to named_arg_base::data).

I started with adding overload `internal::make_arg(const named_arg_base&)` but it's ambiguous, VC2015 doesn't consume it. Actually I didn't get why it works with other compilers.

Alternative options I can see are
1. add a wrapper for a reference to serialized named_arg, replace this overload of map() with the one taking the wrapper
2. add another template internal::make_arg_serialized_named_arg() as a friend for `basic_format_arg`

The chosen one is probably the smallest but probably less supportable than the others.

include/fmt/dyn-args.h
vsolontsov-ll: I head to mention this in the comment -- it's required to support custom types.
But you're right, I can avoid it! 

Re separate header -- I didn't want to add a dependency from std::variant<> into the core.h.
But removing `Args` variant dependency will go away.

include/fmt/core.h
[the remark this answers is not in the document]
vitaut: I'd recommend first making dynamic arg storage work without named arguments and then adding them in a separate PR.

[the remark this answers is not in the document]
vsolontsov-ll: Done

include/fmt/core.h
```diff
@@ -259,7 +259,8 @@ namespace internal {
 // A workaround for gcc 4.8 to make void_t work in a SFINAE context.
 template <typename... Ts> struct void_t_impl { using type = void; };
 
-FMT_NORETURN FMT_API void assert_fail(const char* file, int line, const char* message);
+FMT_NORETURN FMT_API void assert_fail(const char* file, int line,
+                                      const char* message);
```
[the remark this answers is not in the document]
vsolontsov-ll: Done

include/fmt/core.h
```diff
@@ -1357,6 +1365,17 @@ template <typename Context> class basic_format_args {
     set_data(store.data_);
   }
 
+  /**
+   \rst
+   Constructs a `dynamic_basic_format_args` object from
+   `~fmt::format_arg_store`. \endrst
```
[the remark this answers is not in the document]
vsolontsov-ll: Fixed

include/fmt/dyn-args.h
vsolontsov-ll: Done

include/fmt/dyn-args.h
vsolontsov-ll: Removed

vsolontsov-ll:
Maybe worth adding 
1. `dynamic_format_arg_store::reserve(size_t)` 
2. static_assert into `void push_back(std::reference_wrapper<T> arg)` to prevent passing primitive types by ref

But I'd leave those for the a PR with named args.

include/fmt/core.h
```diff
@@ -1613,6 +1629,194 @@ inline void print(const S& format_str, Args&&... args) {
...
+template <typename Traits, typename Char>
+struct is_string_view<std::basic_string_view<Char, Traits>, Char>
+    : std::true_type {};
+#endif
+
+#ifdef FMT_USE_EXPERIMENTAL_STRING_VIEW
+template <typename Traits, typename Char>
+struct is_string_view<std::experimental::basic_string_view<Char, Traits>, Char>
+    : std::true_type {};
+#endif
```
vitaut: This can be replaced with 

```
template <typename Char>
struct is_string_view<std_string_view<Char>, Char> : std::true_type {};
```
I wouldn't bother with traits at this time since they are not handled elsewhere in {fmt}.

include/fmt/core.h
```diff
@@ -1613,6 +1629,194 @@ inline void print(const S& format_str, Args&&... args) {
...
+
+template <typename T> struct is_ref_wrapper : std::false_type {};
+
+template <typename T>
+struct is_ref_wrapper<std::reference_wrapper<T>> : std::true_type {};
```
vitaut: Why do we need to handle `std::reference_wrapper`?

include/fmt/core.h
```diff
@@ -1613,6 +1629,194 @@ inline void print(const S& format_str, Args&&... args) {
...
+
+template <typename T, typename Context> struct need_dyn_copy {
+  using mapped_type = mapped_type_constant<T, Context>;
```
vitaut: Since you only use `mapped_type::value` I suggest replacing this with

```c++
static internal::type mapped_type = mapped_type_constant<T, Context>::value;
```
and using `mapped_type` in place of  `mapped_type::value`.

include/fmt/core.h
```diff
@@ -1613,6 +1629,194 @@ inline void print(const S& format_str, Args&&... args) {
...
+  using type = std::integral_constant<
+      bool, !(is_ref_wrapper<T>::value ||
+              is_string_view<T, typename Context::char_type>::value ||
+              (mapped_type::value != internal::type::cstring_type &&
+               mapped_type::value != internal::type::custom_type &&
+               mapped_type::value != internal::type::string_type))>;
+};
+
+template <typename T, typename Context>
+using need_dyn_copy_t = typename need_dyn_copy<T, Context>::type;
```
vitaut: Let's kill this alias since `need_dyn_copy` is only used in one place.

include/fmt/core.h
```diff
@@ -1613,6 +1629,194 @@ inline void print(const S& format_str, Args&&... args) {
...
+
+class dyn_arg_storage {
+  // Workaround clang's -Wweak-vtables. For templates (unlike regular classes
+  // doesn't complain about inability to deduce translation unit to place vtable
+  // So dyn_arg_node_base is made a fake template
```
vitaut: Did you mean `storage_node_base` instead of `dyn_arg_node_base`?

include/fmt/core.h
```diff
@@ -1613,6 +1629,194 @@ inline void print(const S& format_str, Args&&... args) {
...
+
+  template <typename = void> struct storage_node_base {
+    using owning_ptr = std::unique_ptr<storage_node_base<>>;
+    virtual ~storage_node_base() = default;
+    owning_ptr next_;
```
vitaut: `next_` -> `next` since it is public.

include/fmt/core.h
```diff
@@ -1613,6 +1629,194 @@ inline void print(const S& format_str, Args&&... args) {
...
+  };
+
+  using owning_ptr = storage_node_base<>::owning_ptr;
```
vitaut: This can probably be killed too.

include/fmt/core.h
```diff
@@ -1613,6 +1629,194 @@ inline void print(const S& format_str, Args&&... args) {
...
+
+  template <typename T> struct storage_node : storage_node_base<> {
+    T value_;
+    FMT_CONSTEXPR explicit storage_node(const T& arg, owning_ptr&& next)
+        : value_{arg} {
+      // Must be initialised after value_
```
vitaut: Why?

include/fmt/core.h
```diff
@@ -1613,6 +1629,194 @@ inline void print(const S& format_str, Args&&... args) {
...
+              is_string_view<T, typename Context::char_type>::value ||
+              (mapped_type::value != internal::type::cstring_type &&
+               mapped_type::value != internal::type::custom_type &&
+               mapped_type::value != internal::type::string_type))>;
+};
+
+template <typename T, typename Context>
+using need_dyn_copy_t = typename need_dyn_copy<T, Context>::type;
+
+class dyn_arg_storage {
```
vitaut: I'd merge this class into `dynamic_format_arg_store` unless there is a strong reason to have it separate.

include/fmt/core.h
```diff
@@ -1613,6 +1629,194 @@ inline void print(const S& format_str, Args&&... args) {
...
+class dynamic_format_arg_store
+#if FMT_GCC_VERSION < 409
+    // Workaround a GCC template argument substitution bug.
+    : public basic_format_args<Context>
+#endif
+{
+ private:
+  using char_type = typename Context::char_type;
+  using string_type = std::basic_string<char_type>;
+  using value_type = basic_format_arg<Context>;
```
vitaut: I'd kill these two aliases since they are only used once each (and the notion of value_type is somewhat unclear here).

include/fmt/core.h
```diff
@@ -1613,6 +1629,194 @@ inline void print(const S& format_str, Args&&... args) {
...
+
+  template <typename T>
+  using stored_type =
+      conditional_t<internal::is_string<T>::value, string_type, T>;
+
+  // Storage of basic_format_arg must be contiguous
+  // Required by basic_format_args::args_ which is just a pointer
```
vitaut: nit: missing punctuation at the end of sentenses (.)

include/fmt/core.h
```diff
@@ -1613,6 +1629,194 @@ inline void print(const S& format_str, Args&&... args) {
...
+
+  template <typename T>
+  const T& stored_value(const std::reference_wrapper<T>& arg, std::false_type) {
+    return arg.get();
+  }
+
+  template <typename T>
+  const stored_type<T>& stored_value(const T& arg, std::true_type) {
+    return storage_.push<stored_type<T>>(arg);
+  }
```
vitaut: I suggest replacing tag dispatch with a normal if in `push_back`.

include/fmt/core.h
```diff
@@ -1613,6 +1629,194 @@ inline void print(const S& format_str, Args&&... args) {
...
+      store.push_back("abc1");
+      store.push_back(1.5f);
+      std::string result = fmt::vformat("{} and {} and {}", store);
+    \endrst
+  */
+  template <typename T> void push_back(const T& arg) {
+    static_assert(
+        !std::is_base_of<internal::named_arg_base<char_type>, T>::value,
+        "Named arguments are not supported yet");
+    emplace_arg(stored_value(arg, internal::need_dyn_copy_t<T, Context>{}));
```
vitaut: i.e. replacing this with something like

```c++
if (const_check(internal::need_dyn_copy_t<T, Context>()))
  emplace_arg(storage_.push<stored_type<T>>(arg));
else
  emplace_arg(arg);
```
You'll need to move `const_check` from `fmt/format.h` to `fmt/core.h`. In the future this `if` will become `constexpr if`.

include/fmt/core.h
vsolontsov-ll: Hm, for me it's the main use-case :- ).
It's to let one to avoid dynamic memory allocation when the argument lifetime is guaranteed.

Actually I would want to make the `dynamic_format_arg_store` reusable and more friendly for those who concerned about memory allocations:
1. add reserve(); -- reserves 
2. add clean(); -- drops the `storage_` and resizes the `data_` to zero, but doesn't de-allocate the `storage_`.

vitaut:
> Maybe worth adding
> 
> 1. `dynamic_format_arg_store::reserve(size_t)`
> 2. static_assert into `void push_back(std::reference_wrapper<T> arg)` to prevent passing primitive types by ref
> 
> But I'd leave those for the a PR with named args.

Yes, let's leave these for a separate PR.

include/fmt/core.h
vitaut: Makes sense. Thanks for clarification.

include/fmt/core.h
vsolontsov-ll: Well, I'd create an issue to track it then...
There's a constructor for `internal::basic_string_view` from `std::basic_string<>` with all the template parameters. But `std_string_view` is recognized with only defaults...

include/fmt/core.h
vsolontsov-ll: yes, thx

vsolontsov-ll:
sorry, seems like I missed a few comments

vsolontsov-ll:
Ok, seems to be done. Sorry for that false-start.

include/fmt/core.h
vsolontsov-ll: done

include/fmt/core.h
vitaut: Also let's rename this trait to `is_reference_wrapper` for consistency with the class name itself.

include/fmt/core.h
```diff
@@ -1613,6 +1633,180 @@ inline void print(const S& format_str, Args&&... args) {
...
+      bool, !(is_ref_wrapper<T>::value ||
+              is_string_view<T, typename Context::char_type>::value ||
+              (mapped_type != internal::type::cstring_type &&
+               mapped_type != internal::type::custom_type &&
+               mapped_type != internal::type::string_type))>;
+};
+
+class dyn_arg_storage {
+  // Workaround clang's -Wweak-vtables. For templates (unlike regular classes
+  // doesn't complain about inability to deduce translation unit to place vtable
```
vitaut: I had problems parsing the above. How about rephrasing to something like:

```
  // Workaround clang's -Wweak-vtables for templates (unlike regular classes)
  // don't complain about inability to deduce translation unit to place vtable.
```

include/fmt/core.h
```diff
@@ -1613,6 +1633,180 @@ inline void print(const S& format_str, Args&&... args) {
...
+      emplace_arg(arg);
+  }
+
+  /**
+    \rst
+    Adds an argument into the dynamic store for later passing to a formating
+    function without copying into type-erasing list.
+
+    Note that primitive type values are copied into basic_format_arg<>.
+    \endrst
```
vitaut: I think we should just disallow passing reference wrappers of built-in types to prevent confusion.

test/format-dyn-args-test.cc
```diff
@@ -0,0 +1,87 @@
...
+  store.push_back(1.5f);
+
+  std::string result = fmt::vformat("{} and {} and {}", store);
+
+  EXPECT_EQ("42 and abc1 and 1.5", result);
+}
+
+TEST(FormatDynArgsTest, StringsAndRefs) {
+  // Unfortunately the tests are compiled with old ABI
+  // So strings use COW.
```
vitaut: Why is this a problem?

include/fmt/core.h
vsolontsov-ll: done

include/fmt/core.h
vsolontsov-ll: done

include/fmt/core.h
vsolontsov-ll: Because copying of arg may throw. If by that time we transfer ownership... you know.

include/fmt/core.h
vsolontsov-ll: done

include/fmt/core.h
vsolontsov-ll: done

include/fmt/core.h
vsolontsov-ll: done

include/fmt/core.h
vsolontsov-ll: done

test/format-dyn-args-test.cc
vsolontsov-ll: Imagine a `std::string` with COW in shared state passed into dyn_arg_store by ref.
It takes fmt::string_view of the string. Next non-const subscription operator leads to COW, so string_view keeps pointing to an old version. So the test below won't work with old ABI std::string.

include/fmt/core.h
vsolontsov-ll: Will fix.. Not exactly like you wrote but somehow...

vitaut:
Merged, thanks!

PR #1589, opened 2020-03-12
Enable FMT_STRING() use with types other than string literals

scramsby:
This enables using FMT_STRING() with raw character arrays, like:
```
constexpr char msg[5] = {'h', 'e', 'l', 'l', 'o'};
FMT_STRING(msg);
```
...and other string view types like:
```
constexpr std::string_view msg{"hello"};
FMT_STRING(msg);
```

Also fixes some `constexpr` evaluation build errors when building with C++17 with `arg_ref` internal class.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -371,6 +371,9 @@ template <typename Char> class basic_string_view {
     the size with ``std::char_traits<Char>::length``.
     \endrst
    */
+#if __cplusplus >= 201703L  // C++17's char_traits::length() is constexpr.
+  FMT_CONSTEXPR
+#endif
   basic_string_view(const Char* s)
       : data_(s), size_(std::char_traits<Char>::length(s)) {}
 
@@ -381,11 +384,12 @@ template <typename Char> class basic_string_view {
       : data_(s.data()),
         size_(s.size()) {}
 
-  template <
-      typename S,
-      FMT_ENABLE_IF(std::is_same<S, internal::std_string_view<Char>>::value)>
-  FMT_CONSTEXPR basic_string_view(S s) FMT_NOEXCEPT : data_(s.data()),
-                                                      size_(s.size()) {}
+#if __cplusplus >= 201703L  // C++17's std::string_view::size() is constexpr.
+  FMT_CONSTEXPR
+#endif
+  basic_string_view(internal::std_string_view<Char> s)
+      FMT_NOEXCEPT : data_(s.data()),
+                     size_(s.size()) {}
 
   /** Returns a pointer to the string data. */
   FMT_CONSTEXPR const Char* data() const { return data_; }
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -2236,6 +2236,7 @@ enum class arg_id_kind { none, index, name };
 // An argument reference.
 template <typename Char> struct arg_ref {
   FMT_CONSTEXPR arg_ref() : kind(arg_id_kind::none), val() {}
+
   FMT_CONSTEXPR explicit arg_ref(int index)
       : kind(arg_id_kind::index), val(index) {}
   FMT_CONSTEXPR explicit arg_ref(basic_string_view<Char> name)
@@ -2266,8 +2267,8 @@ struct dynamic_format_specs : basic_format_specs<Char> {
   arg_ref<Char> precision_ref;
 };
 
-// Format spec handler that saves references to arguments representing dynamic
-// width and precision to be resolved at formatting time.
+// Format spec handler that saves references to arguments representing
+// dynamic width and precision to be resolved at formatting time.
 template <typename ParseContext>
 class dynamic_specs_handler
     : public specs_setter<typename ParseContext::char_type> {
@@ -2579,8 +2580,9 @@ FMT_CONSTEXPR void parse_format_string(basic_string_view<Char> format_str,
   auto begin = format_str.data();
   auto end = begin + format_str.size();
   while (begin != end) {
-    // Doing two passes with memchr (one for '{' and another for '}') is up to
-    // 2.5x faster than the naive one-pass implementation on big format strings.
+    // Doing two passes with memchr (one for '{' and another for '}') is up
+    // to 2.5x faster than the naive one-pass implementation on big format
+    // strings.
     const Char* p = begin;
     if (*begin != '{' && !find<IS_CONSTEXPR>(begin + 1, end, '{', p))
       return write(begin, end);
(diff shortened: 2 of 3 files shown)


vitaut:
Thanks for the PR. Could you explain the motivation for this change?

scramsby:
Our coding standards use string_view and equivalent types for defining shared string constants as they are generally more type safe. This enables us to continue to use such types for use in our logging system which uses fmt. The change in the terminating null character detection is mostly just because I noticed it while working in the same area, we don't have concrete need for it currently.

include/fmt/format.h
```diff
@@ -3484,9 +3484,18 @@ template <typename Char> struct udl_arg {
   }
 };
 
+// Converts string literals to basic_string_view.
 template <typename Char, size_t N>
-FMT_CONSTEXPR basic_string_view<Char> literal_to_view(const Char (&s)[N]) {
-  return {s, N - 1};
+FMT_CONSTEXPR basic_string_view<Char> fmt_string_to_view(const Char (&s)[N]) {
```
vitaut: `fmt_string` is a bit ambiguous. I suggest renaming `fmt_string_to_view` to `compile_string_to_view`, since it's for converting `fmt::compile_string` subclass object into a string view.

include/fmt/format.h
```diff
@@ -3484,9 +3484,18 @@ template <typename Char> struct udl_arg {
...
+  // Remove trailing null character if needed. Won't be present if this is used with
+  // raw character array (i.e. not defined as a string).
```
vitaut: Please apply `clang-format` to your changes.

include/fmt/format.h
```diff
@@ -3484,9 +3484,18 @@ template <typename Char> struct udl_arg {
...
+  return {s, N - (std::char_traits<Char>::to_int_type(s[N - 1]) == 0 ? 1 : 0)};
+}
+
+// Converts string_view-like types to basic_string_view.
+template <typename Char, typename String>
+FMT_CONSTEXPR basic_string_view<Char> fmt_string_to_view(const String& s) {
```
vitaut: This needs to be more constrained. I suggest using `std_string_view` (an alias to `std::string_view`) here.

scramsby: It's not desirable that we constrain it to just std::string_view for us, since there are other compatible types outside the standard we use, such as folly::StringPiece, that std::string_view doesn't have implicit constructors for. Do you have any suggestion for how to better constrain it that retains use with compatible types?

vitaut: `folly::StringPiece` is morally obsolete and we should actively discourage new uses of it. Since this is a new API there is no reason to make it work with legacy types.

include/fmt/format.h
scramsby: Attempting to clang-format this file is resulting in a massive number of changes for me. Can you please clarify how my changes are not in conformance?

scramsby: I'm seeing a ton of reformatting when using clang-format on this file. I'm going to leave for now and this can be addressed separately.

scramsby:
Added unit test when building with C++17.

include/fmt/format.h
```diff
@@ -3484,9 +3484,19 @@ template <typename Char> struct udl_arg {
...
 
+// Converts string literals to basic_string_view.
 template <typename Char, size_t N>
-FMT_CONSTEXPR basic_string_view<Char> literal_to_view(const Char (&s)[N]) {
-  return {s, N - 1};
+FMT_CONSTEXPR basic_string_view<Char> compile_string_to_view(
+    const Char (&s)[N]) {
+  // Remove trailing null character if needed. Won't be present if this is used
+  // with raw character array (i.e. not defined as a string).
+  return {s, N - (std::char_traits<Char>::to_int_type(s[N - 1]) == 0 ? 1 : 0)};
```
vitaut: Since you are extending this to more than string literals, please handle the case of N == 0 (and add a test).

include/fmt/format.h
```diff
@@ -3484,9 +3484,19 @@ template <typename Char> struct udl_arg {
...
+}
+
+// Converts string_view-like types to basic_string_view.
+template <typename Char, typename String>
+FMT_CONSTEXPR basic_string_view<Char> compile_string_to_view(const String& s) {
```
vitaut: As commented earlier, please replace `String` with `std_string_view`.

scramsby: You've convinced me. I found workarounds for our internal code as well to not use folly::StringPiece, so I'm fine with making this more narrow.

include/fmt/format.h
scramsby: I added the bounds check, but as far as I can tell there's no way to declare a compile-time 0 length array in C++ without relying upon a non-standard compiler extension, so I was unable to add a test for it.

include/fmt/core.h
```diff
@@ -373,6 +376,11 @@ template <typename Char> class basic_string_view {
       : data_(s.data()),
         size_(s.size()) {}
 
+  /** Constructs a string reference from a ``fmt::basic_string_view`` object. */
+  FMT_CONSTEXPR basic_string_view(const basic_string_view<Char>& s) FMT_NOEXCEPT
+      : data_(s.data()),
+        size_(s.size()) {}
```
vitaut: This shouldn't be necessary because the copy ctor is defined implicitly.

include/fmt/core.h
```diff
@@ -363,6 +363,9 @@ template <typename Char> class basic_string_view {
     the size with ``std::char_traits<Char>::length``.
     \endrst
    */
+#ifdef FMT_USE_STRING_VIEW  // C++17's char_traits::length() is constexpr.
```
vitaut: `FMT_USE_STRING_VIEW` is not a good indicator of whether `char_traits::length()` is `constexpr` because libc++ supports `string_view` in pre-c++17.

scramsby: While that may be true, it's not guaranteed by the standard to be constexpr prior to the C++17 language standard as far as I can tell. Prior to my change this function was _never_ marked constexpr, so I think this change is still an improvement. If you have a more direct suggestion I'd be happy to wrap it in, as I'm not sure how to evaluate based on what C++ standard library version is being used and what particular version they each decided to support this.

vitaut: Since it's a C++17 feature, a better check would be `__cplusplus >= 201703L`.

scramsby:
Finally got the changes trimmed down to what is needed and fixed the last few niggling platform-specific build errors, so this should be good to go absent further feedback.

vitaut:
Merged with minor tweaks in https://github.com/fmtlib/fmt/commit/664dd88e3175b92cdac96cb811dab94b6f3bdc62. Thanks!

PR #1590, opened 2020-03-12
Allow disabling floating point support

albaguirre:
Add FMT_USE_FLOAT, FMT_USE_DOUBLE and FMT_USE_LONG_DOUBLE to allow a
user of the library to configure the float types they want to allow.
This is specially useful in embedded environements where code size is
important.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -1095,23 +1095,11 @@ FMT_CONSTEXPR auto visit_format_arg(Visitor&& vis,
   case internal::type::char_type:
     return vis(arg.value_.char_value);
   case internal::type::float_type:
-#if FMT_USE_FLOAT
     return vis(arg.value_.float_value);
-#else
-    break;
-#endif
   case internal::type::double_type:
-#if FMT_USE_DOUBLE
     return vis(arg.value_.double_value);
-#else
-    break;
-#endif
   case internal::type::long_double_type:
-#if FMT_USE_LONG_DOUBLE
     return vis(arg.value_.long_double_value);
-#else
-    break;
-#endif
   case internal::type::cstring_type:
     return vis(arg.value_.string.data);
   case internal::type::string_type:
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -296,6 +296,18 @@ struct int128_t {};
 struct uint128_t {};
 #endif
 
+#ifndef FMT_USE_FLOAT
+#  define FMT_USE_FLOAT 1
+#endif
+
+#ifndef FMT_USE_DOUBLE
+#  define FMT_USE_DOUBLE 1
+#endif
+
+#ifndef FMT_USE_LONG_DOUBLE
+#  define FMT_USE_LONG_DOUBLE 1
+#endif
+
 // Casts a nonnegative integer to unsigned.
 template <typename Int>
 FMT_CONSTEXPR typename std::make_unsigned<Int>::type to_unsigned(Int value) {
@@ -1083,11 +1095,23 @@ FMT_CONSTEXPR auto visit_format_arg(Visitor&& vis,
   case internal::type::char_type:
     return vis(arg.value_.char_value);
   case internal::type::float_type:
+#if FMT_USE_FLOAT
     return vis(arg.value_.float_value);
+#else
+    break;
+#endif
   case internal::type::double_type:
+#if FMT_USE_DOUBLE
     return vis(arg.value_.double_value);
+#else
+    break;
+#endif
   case internal::type::long_double_type:
+#if FMT_USE_LONG_DOUBLE
     return vis(arg.value_.long_double_value);
+#else
+    break;
+#endif
   case internal::type::cstring_type:
     return vis(arg.value_.string.data);
   case internal::type::string_type:
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -726,6 +726,13 @@ FMT_CONSTEXPR bool is_negative(T) {
   return false;
 }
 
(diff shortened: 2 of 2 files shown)


vitaut:
Thanks for the PR. Have you verified that this indeed results in smaller binary code?

albaguirre:
> Thanks for the PR. Have you verified that this indeed results in smaller binary code? Could you provide any specific numbers?

Yeah on our current image, stripped and compiled with -0s using the GCC 4.8 bare metal toolchain (available at https://launchpad.net/gcc-arm-embedded/+download) we save the following:

FMT_USE_FLOAT=1 FMT_USE_DOUBLE=1 FMT_USE_LONG_DOUBLE=0
Saves 3264 bytes

FMT_USE_FLOAT=1 FMT_USE_DOUBLE=0 FMT_USE_LONG_DOUBLE=0
Saves 3696 bytes

FMT_USE_FLOAT=0 FMT_USE_DOUBLE=0 FMT_USE_LONG_DOUBLE=0
saves 12664 bytes

vitaut:
Do you need partial float support? If not I'd suggest having just one control to enable/disable all floating-point formatting instead of doing it type-by-type.

albaguirre:
> Do you need partial float support? If not I'd suggest having just one control to enable/disable all floating-point formatting instead of doing it type-by-type.

Yes, on some platforms we need single precision float formatting, but not double or long double.

include/fmt/core.h
```diff
@@ -1083,11 +1095,23 @@ FMT_CONSTEXPR auto visit_format_arg(Visitor&& vis,
...
     return vis(arg.value_.double_value);
+#else
+    break;
+#endif
   case internal::type::long_double_type:
+#if FMT_USE_LONG_DOUBLE
     return vis(arg.value_.long_double_value);
+#else
+    break;
+#endif
```
vitaut: Please remove conditional compilation here. `FMT_USE_*` should be handled on the visitor side instead.

include/fmt/format.h
```diff
@@ -1685,8 +1685,9 @@ template <typename Range> class basic_writer {
     handle_int_type_spec(spec.type, int_writer<T, Spec>(*this, value, spec));
   }
 
+#if FMT_USE_FLOAT || FMT_USE_DOUBLE || FMT_USE_LONG_DOUBLE
   template <typename T, FMT_ENABLE_IF(std::is_floating_point<T>::value)>
-  void write(T value, format_specs specs = {}) {
+  void write_float(T value, format_specs specs = {}) {
```
vitaut: Instead of renaming and conditionally compiling this function, I suggest adding something like:

```c++
    if (const_check(
      (std::is_same<T, float>::value && !FMT_USE_FLOAT) ||
      (std::is_same<T, double>::value && !FMT_USE_DOUBLE) ||
      (std::is_same<T, long double>::value && !FMT_USE_LONG_DOUBLE))) {
      return;
    }
```

include/fmt/format.h
```diff
@@ -1743,6 +1744,17 @@ template <typename Range> class basic_writer {
...
+
+#if FMT_USE_FLOAT
+  void write(float value, format_specs specs = {}) { write_float(value, specs); }
+#endif
+#if FMT_USE_DOUBLE
+  void write(double value, format_specs specs = {}) { write_float(value, specs); }
+#endif
+#if FMT_USE_LONG_DOUBLE
+  void write(long double value, format_specs specs = {}) { write_float(value, specs); }
+#endif
```
vitaut: Then these overloads will become unnecessary.

include/fmt/format.h
```diff
@@ -2925,9 +2937,25 @@ struct formatter<T, Char,
...
+#else
+      FMT_ASSERT(false, "double support disabled");
+#endif
+      break;
     case internal::type::long_double_type:
+#if FMT_USE_LONG_DOUBLE
       internal::parse_float_type_spec(specs_, eh);
+#else
+      FMT_ASSERT(false, "long double support disabled");
+#endif
```
vitaut: Conditional compilation is not really needed here, just use normal `if` statements.

include/fmt/core.h
albaguirre: Makes sense, done.

include/fmt/format.h
albaguirre: Done.

include/fmt/format.h
albaguirre: Yep, done

include/fmt/format.h
albaguirre: Done.

include/fmt/format.h
```diff
@@ -1885,6 +1891,13 @@ class arg_formatter_base {
 
   template <typename T, FMT_ENABLE_IF(std::is_floating_point<T>::value)>
   iterator operator()(T value) {
+    if (const_check(
+            (std::is_same<T, float>::value && !FMT_USE_FLOAT) ||
+            (std::is_same<T, double>::value && !FMT_USE_DOUBLE) ||
+            (std::is_same<T, long double>::value && !FMT_USE_LONG_DOUBLE))) {
```
vitaut: This can be factored out into a constexpr function and reused here and above.

albaguirre: OK, is the latest commit sort of what you had in mind?

vitaut: Yes, thank you.

grisharav:
I see the macros implemented here e.g. `FMT_USE_FLOAT` are not in the current ()  code base, @vitaut - any reason for this?

vitaut:
Those have been superseded by `FMT_BUILTIN_TYPES`.

Issue #1594, opened 2020-03-15
Color formatting doesn't work for wide strings


Spirrwell:
Hello there!

I don't know if wide strings are intended to not be supported by color formatting, but it seems like a bug.

To put simply, this does not work:

`fmt::print( fmt::fg( fmt::color::red ), L"This is a wide formatted string\n" );`

If you look at this line here: https://github.com/fmtlib/fmt/blob/master/include/fmt/color.h#L415

It uses std::strlen and if you try to format a wide string with this, it fails to compile as std::strlen assumes 'char' is the type.

I've tried replacing it with this:

`return buffer + std::char_traits<Char>::length(buffer);`

This seems to work just fine for me. But I haven't done a whole lot of testing.

vitaut:
Thanks for reporting. Could you submit a PR with your suggested fix and a test in `test/color-test.cc`?

PR #1596, opened 2020-03-16
Color formatting fixed for wide strings (fixes issue #1594)

Spirrwell:

Along with the fix I added tests that mirror their non-wide counterparts. I did NOT add tests for file writing (EXPECT_WRITE) as there doesn't seem to be an implementation using wide strings in the gtest-extra code. This is something I don't really know how to tackle, so I'd leave it up to somebody else to do.

### DIFF
--- a/test/color-test.cc
+++ b/test/color-test.cc
@@ -54,62 +54,32 @@ TEST(ColorsTest, Format) {
             L"\x1b[38;2;255;020;030mrgb(255,20,30) wide\x1b[0m");
   EXPECT_EQ(fmt::format(fg(fmt::color::blue), "blue"),
             "\x1b[38;2;000;000;255mblue\x1b[0m");
-  EXPECT_EQ(fmt::format(fg(fmt::color::blue), L"blue wide"),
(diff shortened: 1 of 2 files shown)


test/color-test.cc
```diff
@@ -50,34 +50,66 @@ TEST(ColorsTest, ColorsPrint) {
 TEST(ColorsTest, Format) {
   EXPECT_EQ(fmt::format(fg(fmt::rgb(255, 20, 30)), "rgb(255,20,30)"),
             "\x1b[38;2;255;020;030mrgb(255,20,30)\x1b[0m");
+  EXPECT_EQ(fmt::format(fg(fmt::rgb(255, 20, 30)), L"rgb(255,20,30) wide"),
+            L"\x1b[38;2;255;020;030mrgb(255,20,30) wide\x1b[0m");
```
vitaut: I think one test case for wide strings is enough since the code paths are shared - it either all works or all doesn't. Let's remove all the ones below.

Spirrwell: Alright, should be done!

PR #1598, opened 2020-03-18
Add back missing OUTPUT_NAME in target properties.

ambitslix:


vitaut:
Thanks!

PR #1602, opened 2020-03-23
Enable correct __cplusplus macro when building with MSVC

scramsby:
This enables proper evaluation of the C++ standard version preprocessor macro when building with MSVC, which gates some logic and tests.


vitaut:
Thanks for the PR but if there is something wrong with the conditional compilation logic it should be fixed in code, not build config, because not everyone uses CMake.

PR #1603, opened 2020-03-23
Add .vs to .gitignore

scramsby:
This suppresses check-in of Visual Studio cache files.


vitaut:
Thanks

PR #1606, opened 2020-03-25
Use FMT_THROW macro where applicable

scramsby:
This enables the code to properly build without errors with Clang/GCC when `-fno-exceptions` is specified.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -1371,7 +1371,7 @@ FMT_FUNC void vprint(std::FILE* f, string_view format_str, format_args args) {
     if (!WriteConsoleW(reinterpret_cast<HANDLE>(_get_osfhandle(fd)),
                        u16.c_str(), static_cast<DWORD>(u16.size()), &written,
                        nullptr)) {
-      throw format_error("failed to write to console");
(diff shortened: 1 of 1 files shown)


scramsby:
For some reason GitHub is reporting that the Travis builds are still pending but clicking through they all succeeded.

include/fmt/format-inl.h
```diff
@@ -1371,7 +1371,7 @@ FMT_FUNC void vprint(std::FILE* f, string_view format_str, format_args args) {
     if (!WriteConsoleW(reinterpret_cast<HANDLE>(_get_osfhandle(fd)),
                        u16.c_str(), static_cast<DWORD>(u16.size()), &written,
                        nullptr)) {
-      throw format_error("failed to write to console");
+      FMT_THROW(format_error("failed to write to console"));
```
vitaut: Thanks for the PR! I think this is the only case that needs conversion to `FMT_THROW` because the ones in `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` are specifically for the fuzzer config and the ones `fmt/compile.h` for compile-time errors.

vitaut:
Thanks!

PR #1616, opened 2020-04-06
Avoid windows issue with min() max() macros

gsjaardema:
Including the ``windows.h`` file without defining ``NOMINMAX`` will define the `min()` and `max()` macros which will result in issues compiling any C++ code that uses any variant of `max`, for example `std::numeric_limits<std::streamsize>::max()` and many others.  Although max() isn't used in Fmt anywhere, it is often used in codes that include a format include file so simply upgrading to the current version of lib::fmt will break the windows build which worked prior to the update...

The fix here is pulled from the 6.1.2 release.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -22,8 +22,14 @@
 #endif
 
 #ifdef _WIN32
+#  if defined(NOMINMAX)
+#    include <windows.h>
+#  else
+#    define NOMINMAX
+#    include <windows.h>
+#    undef NOMINMAX
+#  endif
 #  include <io.h>
-#  include <windows.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR. `fmt/format-inl.h` is an internal header so it shouldn't affect users unless they opt into the header-only mode.

gsjaardema:
Yes, we use the header-only mode and our code started failing the windows build right after the update to 6.2.0 with no other changes.  Not something that we expected; also not something that is easy to protect in our code since difficult to tell when we include a fmt include file whether it will be including `fmt/format-inl.h`

gsjaardema:
Thanks.  Great library.

PR #1629, opened 2020-04-11
Fix warnings with -Wshadow in gcc 7.5

johnor:


vitaut:
Merged with minor tweaks, thanks.

PR #1633, opened 2020-04-14
Add check for CompiledFormat to avoid ambiguous call

Kurkin:
Add a check that CompileFormat is derived from basic_compiled_format to avoid compilation error in case if both compile.h and format.h are included.
https://godbolt.org/z/BuhQCh


vitaut:
Thanks!

PR #1635, opened 2020-04-16
Use UDL template for GCC 9.3 and newer

LarsGullik:
It seems that this was fixed for GCC 9.3 with
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=88095


vitaut:
Thanks for the PR. The issue is unrelated GCC bug 88095 but I suppressed the warning and reenabled compile-time `_format` in https://github.com/fmtlib/fmt/commit/f46f5ecaf0157749a6ac8c876cac98198b9f4489.

PR #1641, opened 2020-04-21
Add a disable wchar support option

APokorny:
On MCUs with a static libstdc++ disabling wchar in the standard library can
safe a lot of space. This change tries to autodetect the wchar_t support
on gcc systems. But also allows disabling wchar with a compiler flag.


vitaut:
Thanks for the PR. 4.x is a very old version that is no longer developed, all new features should go to master instead.

PR #1650, opened 2020-04-27
Use FMT_THROW in fuzzing build mode

asraa:
Make code properly build in fuzz mode with clang/gcc when `-fno-exceptions` is specified by using `FMT_THROW` in format files when building with FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION.

Signed-off-by: Asra Ali <asraa@google.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -308,6 +308,7 @@ endif ()
 # Control fuzzing independent of the unit tests.
 if (FMT_FUZZ)
   add_subdirectory(test/fuzzing)
+  target_compile_definitions(fmt PUBLIC FMT_FUZZ)
 endif ()
 
 set(gitignore ${PROJECT_SOURCE_DIR}/.gitignore)
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -1151,7 +1151,7 @@ int snprintf_float(T value, int precision, float_specs specs,
   for (;;) {
     auto begin = buf.data() + offset;
     auto capacity = buf.capacity() - offset;
-#ifdef FUZZ_MODE
+#ifdef FMT_FUZZ
     if (precision > 100000)
       throw std::runtime_error(
           "fuzz mode - avoid large allocation inside snprintf");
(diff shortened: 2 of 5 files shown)


vitaut:
Thanks for the PR. Could you elaborate why this is needed? The fuzz tests are always run with exceptions enabled in oss-fuzz: https://bugs.chromium.org/p/oss-fuzz/issues/list?colspec=ID%20Type%20Component%20Status%20Proj%20Reported%20Owner%20Summary&q=proj%3Dlibfmt&can=1

asraa:
Thanks! We have projects (Envoy, and others) that depend on fmt which are built with no exception mode, and are also fuzzed. These project's fuzz targets cannot build with the same production fmt lib dependency with exceptions disabled. Ideally, those project's fuzzers should model their production environment, they do not expect throws in their dependency.

vitaut:
I think a better fix would be to replace `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` with an {fmt}-specific macro (e.g. `FMT_FUZZING`) set when the `FMT_FUZZ` CMake compiler option is enabled: https://github.com/fmtlib/fmt/blob/9dde9f0131eb86245e547d6b5984b205d3319cb5/CMakeLists.txt#L309

asraa:
Thank you, that would work well for us! I will update the PR and fuzzing README to reflect that.

vitaut:
Is `FUZZ_MODE` documented anywhere?

asraa:
Sorry about that -- I mixed up the macros. I tested the fuzz targets in this branch and in master branch and didn't hit any OOMs at least while running `fuzzer_one_arg`

asraa:
Both master and this branch start around 100k+ exec/sec, but my branch will go down to 40k sometimes. Any idea why that's happening?

vitaut:
Looks good but please

1. Define `FMT_FUZZ` in CMake: https://github.com/fmtlib/fmt/blob/f2ed03b9191a629d728f9ee47e4fd81f9dd1931c/CMakeLists.txt#L309
```
    target_compile_definitions(fmt PUBLIC FMT_FUZZ)
```
2. Use `#ifdef` instead of `#if` to prevent warnings when `FMT_FUZZ` is not defined.

> Any idea why that's happening?

Not sure.

vitaut:
Thank you!

Issue #1654, opened 2020-04-29
6.2.0 is less embedded friendly in comparison to 5.3.0


kwesolowski:
Hi,

Playing with IAR and GHS compilers I see some places (mostly os.h/os.cpp) where for example:

EINTR retry is done for "not windows", FILE * operations are included etc. Most of the functionality is perfectly "in RAM", but those few pieces leak os related items. Would you like to reduce such "heave OS" dependencies? What would be preferred Pull Request?

kwesolowski:
Example patch (definitions would need to move outside lib). Or maybe just better mechanism to exclude os dependant features? 

+#define FMT_RETRY(result, expression) result = (expression)
+#define FMT_RETRY_VAL(result, expression, error_result) result = (expression)
+
+#ifndef FMT_RETRY
 // Retries the expression while it evaluates to error_result and errno
 // equals to EINTR.
 #ifndef _WIN32
 #  define FMT_RETRY_VAL(result, expression, error_result) \
     do {                                                  \
@@ -69,10 +73,11 @@
 #else
 #  define FMT_RETRY_VAL(result, expression, error_result) result = (expression)
 #endif
 
 #define FMT_RETRY(result, expression) FMT_RETRY_VAL(result, expression, -1)
+#endif

vitaut:
`os.h/os.cc` are non-essential and can be excluded from the build. You only need `format.cc` and headers. I am open to a PR to add a CMake flag (e.g. `FMT_OS`) to control whether to include OS-specific APIs.

kwesolowski:
Will make a PR.

PR #1656, opened 2020-04-30
Allow to avoid inclusion of os.cc in fmt target

kwesolowski:
Allows to avoid part of #1654

Shudl be fine, but if i missed to update some docs etc - please mention it.


vitaut:
Merged in https://github.com/fmtlib/fmt/commit/e2ef12a8c0ca40a80001f20421fe5bf40fa59a98 with minor tweaks, thanks.

PR #1657, opened 2020-04-30
build: Fix installation paths

jtojnar:
It is not generally true that `CMAKE_INSTALL_<dir>` variables are relative paths:
    
https://github.com/jtojnar/cmake-snips#concatenating-paths-when-building-pkg-config-files

This lead to `includedir=/nix/store/9axh4pacy7l4s7ii9j0ij54nmdslj7yy-fmt-6.2.0-dev//nix/store/9axh4pacy7l4s7ii9j0ij54nmdslj7yy-fmt-6.2.0-dev/include`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,6 +81,7 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH}
 
 include(cxx14)
 include(CheckCXXCompilerFlag)
+include(JoinPaths)
 
 list(FIND CMAKE_CXX_COMPILE_FEATURES "cxx_variadic_templates" index)
 if (${index} GREATER -1)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -235,8 +236,7 @@ if (FMT_INSTALL)
   include(GNUInstallDirs)
   include(CMakePackageConfigHelpers)
   set_verbose(FMT_CMAKE_DIR ${CMAKE_INSTALL_LIBDIR}/cmake/fmt CACHE STRING
-              "Installation directory for cmake files, relative to "
```
vitaut, line 238: I'd change "relative to ${CMAKE_INSTALL_PREFIX}" to "absolute or relative to ${CMAKE_INSTALL_PREFIX}" instead of removing this part completely. Same in other places.

support/cmake/JoinPaths.cmake
```diff
@@ -0,0 +1,23 @@
+# This module provides function for joining paths
+# known from from most languages
+#
+# SPDX-License-Identifier: (MIT OR CC0-1.0)
```
vitaut, line 4: Can this be distributed on {fmt}'s license (MIT with an optional exception for binary distribution): https://github.com/fmtlib/fmt/blob/master/LICENSE.rst? If yes I suggest replacing this with "# For the license information refer to LICENSE.rst."

CMakeLists.txt
jtojnar: Would wording like “absolute path, or relative path that will be joined to ${CMAKE_INSTALL_PREFIX}”? I want to make it clear that the absolute paths can be arbitrary, not only based on `${CMAKE_INSTALL_PREFIX}`.

support/cmake/JoinPaths.cmake
jtojnar: I would expect the CC0 (public domain like) license to be compatible with anything but I am not a lawyer. I can relicense it here under {fmt}’s license so we do not have to consider this.

vitaut: Ah, public domain is fine.

vitaut:
Thank you!

PR #1660, opened 2020-05-02
Update date formatting example to use threadsafe localtime

lsr0:
As a bonus, this hints the user to the existence of `fmt::localtime()`/`fmt::gmtime()`.

----


vitaut:
Thank you!

PR #1661, opened 2020-05-02
Exclude std::abort from compilation when compiling CUDA with Clang

### LINKED ISSUE #1659: std::abort causes error when compiling CUDA with Clang in device code
Compiling with Clang 9 in CUDA mode results in the following compilation errors during its device code pass:
```cpp
.../fmt-master/include/fmt/format-inl.h:50:8: error: expected identifier
  std::abort();
       ^
.../fmt-master/include/fmt/format-inl.h:50:3: error: unexpected namespace name 'std': expected expression
  std::abort();
  ^
```
One possible fix is to use ```std::terminate``` instead:
```cpp
FMT_FUNC void assert_fail(const char* file, int line, const char* message) {
  print(stderr, "{}:{}: assertion failed: {}", file, line, message);
#if defined(__clang__) && defined(__CUDA__) && defined(__CUDA_ARCH__)
  std::terminate();
#else
  std::abort();
#endif
}
```

invexed:
Excludes ```std::abort``` for device code only. Fixes #1659.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -48,10 +48,9 @@ namespace internal {
 
 FMT_FUNC void assert_fail(const char* file, int line, const char* message) {
   print(stderr, "{}:{}: assertion failed: {}", file, line, message);
-#if defined(__clang__) && defined(__CUDA__) && defined(__CUDA_ARCH__)
+  // Chosen instead of std::abort to satisfy Clang in CUDA mode during device
+  // code pass.
   std::terminate();
-#else
-  std::abort();
 #endif
 }
 
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -15,6 +15,7 @@
 #include <cstdarg>
 #include <cstring>  // for std::memmove
 #include <cwchar>
+#include <exception>
 
 #include "format.h"
 #if !defined(FMT_STATIC_THOUSANDS_SEPARATOR)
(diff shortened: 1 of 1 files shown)


include/fmt/format-inl.h
```diff
@@ -47,7 +48,11 @@ namespace internal {
 
 FMT_FUNC void assert_fail(const char* file, int line, const char* message) {
   print(stderr, "{}:{}: assertion failed: {}", file, line, message);
+#if defined(__clang__) && defined(__CUDA__) && defined(__CUDA_ARCH__)
+  std::terminate();
```
vitaut, line 52: I think we can use `std::terminate` unconditionally but please add a short comment clarifying that it was chosen instead of `abort` because of CUDA.

include/fmt/format-inl.h
```diff
@@ -48,10 +48,9 @@ namespace internal {
 
 FMT_FUNC void assert_fail(const char* file, int line, const char* message) {
   print(stderr, "{}:{}: assertion failed: {}", file, line, message);
-#if defined(__clang__) && defined(__CUDA__) && defined(__CUDA_ARCH__)
+  // Chosen instead of std::abort to satisfy Clang in CUDA mode during device
+  // code pass.
   std::terminate();
-#else
-  std::abort();
 #endif
```
vitaut, line 54: Extra `#endif`.

invexed: My bad. I Tested the wrong branch locally!

vitaut:
Thanks!

PR #1667, opened 2020-05-06
Fix some warnings

peterbell10:
This fixes a few warnings I was seeing compiling with clang:
- `-Wsigned-enum-bitfield` from using standard enums in a bitfield, fixed by giving them an unsigned underlying type.
- `-Wundefined-func-template` from templates without definitions, fixed by adding `extern template` declarations.
- `-Wsign-conversion` from constructing a `string_view` with an `int` size, fixed with `internal::to_unsigned`.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -849,6 +849,10 @@ inline int count_digits(uint32_t n) {
 #endif
 
 template <typename Char> FMT_API std::string grouping_impl(locale_ref loc);
+#ifndef FMT_HEADER_ONLY
+extern template FMT_API std::string grouping_impl<char>(locale_ref loc);
+extern template FMT_API std::string grouping_impl<wchar_t>(locale_ref loc);
+#endif
 template <typename Char> inline std::string grouping(locale_ref loc) {
   return grouping_impl<char>(loc);
 }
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR. Please move all the extern template declarations to https://github.com/fmtlib/fmt/blob/59fe455f36f2dca71369b5d1eeca63daf3ba7558/include/fmt/format.h#L3337-L3340 to avoid introducing extra #ifdef blocks.

vitaut:
Thanks!

include/fmt/format.h
```diff
@@ -1022,12 +1022,12 @@ template <typename Char> struct fill_t {
 // We cannot use enum classes as bit fields because of a gcc bug
 // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61414.
 namespace align {
-enum type { none, left, right, center, numeric };
+enum type : uint8_t { none, left, right, center, numeric };
```
vitaut, line 1025: Reverted because of
```
include/fmt/format.h:1039:19: error: ‘fmt::v6::basic_format_specs<char>::align’ is too small to hold all values of ‘using align_t = enum fmt::v6::align::type {aka enum fmt::v6::align::type}’ [-Werror]
```

peterbell10: That's unfortunate; it looks like there isn't a nice way to satisfy both compilers. I can switch the struct data members to use `uint8_t` directly, but then we lose the type safety of enums.

PR #1669, opened 2020-05-07
Make implicit capture explicit for C++20

### LINKED ISSUE #1668: Implicit capture of ‘this’ via ‘[=]’ is deprecated in C++20
Compiling with `g++ (Ubuntu 9.2.1-17ubuntu1~18.04.1) 9.2.1 20191102` I obtain warnings of the kind

```bash
fmt/include/fmt/format.h:1504:59: warning: implicit capture of ‘this’ via ‘[=]’ is deprecated in C++20 [-Wdeprecated]
 1504 |     out = write_int(out, num_digits, get_prefix(), specs, [=](iterator it) {
``` 

IMHO the solution is to explicitly specify `this` in the capture list, e.g.

```c++
out = write_int(out, num_digits, get_prefix(), specs, [=, this](iterator it) {
``` 

PR follows in a minute.

daixtrose:
- Implicit capture of ‘this’ via ‘[=]’ is deprecated in C++20, therefore this change.
- fixes #1668

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -172,7 +172,7 @@ FMT_END_NAMESPACE
 #  define FMT_BUILTIN_CLZLL(n) __builtin_clzll(n)
 #endif
 
-#if (__cplusplus >= 201703L)
+#if (__cplusplus > 201703L)
 #  define FMT_CAPTURE_OF_THIS , this
 #else
 #  define FMT_CAPTURE_OF_THIS
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -172,6 +172,12 @@ FMT_END_NAMESPACE
 #  define FMT_BUILTIN_CLZLL(n) __builtin_clzll(n)
 #endif
 
+#if (__cplusplus >= 201703L)
+#  define FMT_CAPTURE_OF_THIS , this
+#else
+#  define FMT_CAPTURE_OF_THIS
+#endif
+
 // Some compilers masquerade as both MSVC and GCC-likes or otherwise support
 // __builtin_clz and __builtin_clzll, so only define FMT_BUILTIN_CLZ using the
 // MSVC intrinsics if the clz and clzll builtins are not available.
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR. There are some CI build failures: https://travis-ci.org/github/fmtlib/fmt/jobs/684199343

daixtrose:
OK, it seems that mitigating the warning for a C++20 compiler is mutually exclusive to a C++14 compiler. Are you generally OK with [the `MACRO` solution](https://github.com/fmtlib/fmt/blob/91251b90696ab3769b31937e5716abaf33e68da1/include/fmt/format.h#L175)? Waiting for CI ...

vitaut:
> Are you generally OK with the MACRO solution?

Sure but I suggest renaming `FMT_CAPTURE_OF_THIS` to `FMT_CAPTURE_THIS`.

vitaut:
Another option is to change `=` to explicit capture - it might be slightly better than a macro.

daixtrose:
> Another option is to change `=` to explicit capture - it might be slightly better than a macro.

I am not sure how to do this such that it is portable since C++14 compilers reject any mention of `this` as error.

> I suggest renaming `FMT_CAPTURE_OF_THIS` to `FMT_CAPTURE_THIS`.

ACK. Done.

timsong-cpp:
`[this, num_digits]`. I think that's the only thing you need to capture in all cases?

vitaut:
> I am not sure how to do this such that it is portable since C++14 compilers reject any mention of this as error.

This seems to work (https://godbolt.org/z/jy-YYn):

```c++
struct S {
  void f() {
    int i;
    auto g = [i, this]() {};
    g();
  }
};

int main() {}
```

daixtrose:
Ok, I will give it a try.

daixtrose:
@vitaut explicit capture done. Now go [fix your CI](https://travis-ci.org/github/fmtlib/fmt/jobs/685084125) :repeat_one:

vitaut:
Thanks! Looks like a transient failure, I restarted the job.

PR #1670, opened 2020-05-08
clang-tidy cleanups

neheb:


vitaut:
Thanks for the PR but this is inconsistent with library conventions and breaks the build: https://travis-ci.org/github/fmtlib/fmt/jobs/684517793

neheb:
Which commits?

vitaut:
All of them for different reasons maybe except for making members const.

vitaut:
Actually no, that's problematic too because those methods are semantically non-const.

neheb:
The last three should be fine. The first two I understand.

neheb:
The build breakage was caused by the third commit. I've fixed it locally.

Issue #1674, opened 2020-05-09
Make dynamic_format_arg_store reusable


vsolontsov-ll:
It must be easy to make the dynamic_format_arg_store reusable by adding method `clear()`. Also adding method `reserve(num_args, num_named_args)` can improve memory management.

vitaut:
PRs are welcome =).

vsolontsov-ll:
Will do. The only trouble - I see no simple way to test reserve() method.

vitaut:
https://github.com/fmtlib/fmt/blob/7f723fbcb8eccc50f6efe988a97db6c9d7f18b28/test/mock-allocator.h if you feel adventurous.

vsolontsov-ll:
Right....:-)
Well, as of now dynamic_format_arg_store<> is not parameterized by allocator. I'm not infected by TDD that much. Do you see a reason for adding allocator there besides testing trivial method?

vitaut:
No, I'm OK not testing whether reserve does anything useful since it's just an optimization and doesn't affect correctness. It's enough to test that it doesn't break things.

PR #1677, opened 2020-05-11
Make dynamic_format_arg_store reusable and add reserve()

vsolontsov-ll:
Implementation of #1674.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -1586,7 +1586,7 @@ class dynamic_format_arg_store
   */
   void reserve(size_t new_cap, size_t new_cap_named) {
     FMT_ASSERT(new_cap >= new_cap_named,
-               "Set of argument includes set of named arguments");
+               "Set of arguments includes set of named arguments");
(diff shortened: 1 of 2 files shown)


include/fmt/core.h
```diff
@@ -1572,6 +1572,24 @@ class dynamic_format_arg_store
...
+    dynamic_args_ = internal::dynamic_arg_list();
+  }
+
+  /**
+    Reserves space to store at least *new_cap* arguments including
+    *new_cap_named* named arguments.
+  */
+  void reserve(size_t new_cap, size_t new_cap_named) {
+    FMT_ASSERT(new_cap >= new_cap_named,
+               "Set of argument includes set of named arguments");
```
vitaut, line 1589: argument -> arguments

vitaut:
One nit, otherwise LGTM. Thanks for the PR!

include/fmt/core.h
vsolontsov-ll: done. thanks for the review.

PR #1681, opened 2020-05-13
nested replacement fields may omit arg_id

puetzk:
syntax.html already has examples like `fmt::format("{:.{}f}", 3.14, 1)` using this, 
and https://en.cppreference.com/w/cpp/utility/format/formatter#width_and_precision shows that this is the case for the C++20 std::format

The Format Specification Mini-Language grammar seems to be the only one not showing this; update it to match.

Closes fmtlib/fmtlib.github.io#10


vitaut:
Good catch, thanks!

PR #1687, opened 2020-05-16
fix interaction of space flag and '+' flag, as well as '-' flag and '…

rimathia:
…0' flag

### DIFF
--- a/include/fmt/printf.h
+++ b/include/fmt/printf.h
@@ -405,7 +405,9 @@ void basic_printf_context<OutputIt, Char>::parse_flags(format_specs& specs,
       specs.fill[0] = '0';
       break;
     case ' ':
-      specs.sign = sign::space;
+      if (specs.sign != sign::plus) {
+        specs.sign = sign::space;
+      }
       break;
     case '#':
       specs.alt = true;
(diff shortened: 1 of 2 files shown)


include/fmt/printf.h
```diff
@@ -405,7 +405,9 @@ void basic_printf_context<OutputIt, Char>::parse_flags(format_specs& specs,
       specs.fill[0] = '0';
       break;
     case ' ':
-      specs.sign = sign::space;
+      if (specs.sign != sign::plus) {
```
rimathia, line 408: from the opengroup description: "This means that if the <space> and '+' flags both appear, the <space> flag shall be ignored."

include/fmt/printf.h
```diff
@@ -517,10 +519,11 @@ OutputIt basic_printf_context<OutputIt, Char>::format() {
     if (specs.alt && visit_format_arg(detail::is_zero_int(), arg))
       specs.alt = false;
     if (specs.fill[0] == '0') {
-      if (arg.is_arithmetic())
+      if (arg.is_arithmetic() && specs.align != align::left)
         specs.align = align::numeric;
       else
-        specs.fill[0] = ' ';  // Ignore '0' flag for non-numeric types.
+        specs.fill[0] = ' ';  // Ignore '0' flag for non-numeric types or if '-'
+                              // flag is also present.
```
rimathia, line 526: opengroup description: "If the `0` and `-` flags both appear, the `0` flag is ignored."

rimathia:
Hi, since the behaviour changes implemented here seem pretty clearly in line with the opengroup description of the printf familiy I have skipped opening an issue.

vitaut:
This is great, thanks!

PR #1689, opened 2020-05-17
Adding sentinel support to fmt::join().

BRevzin:

This PR adds support to invoke `fmt::join()` on ranges that have a sentinel type that differs from its iterator type. This allows you to use `fmt::join()` on range-v3/C++20 ranges adapters. This shouldn't affect any existing code (whose sentinel is just the same as the iterator type).

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -291,6 +291,7 @@ template <> constexpr int num_bits<fallback_uintptr>() {
 // An approximation of iterator_t for pre-C++20 systems.
 template <typename T>
 using iterator_t = decltype(std::begin(std::declval<T&>()));
+template <typename T> using sentinel_t = decltype(std::end(std::declval<T&>()));
 
 // Detect the iterator category of *any* given type in a SFINAE-friendly way.
 // Unfortunately, older implementations of std::iterator_traits are not safe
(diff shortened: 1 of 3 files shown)


vitaut:
Thank you!

puetzk:
This did technically break any code that instantiated `fmt::arg_join<Iterator, char16_t>` explicitly. I had one place that was doing this, because it wanted `Char = char16_t`, but fmt::join is only overloaded for fmt::string_view and fmt::wstring_view, and cannot be partially specialized or order to pull out `Char` from other options (in this case u16string_view).

The breakage could have been avoided if `Sentinel` had been introduced as the last template argument, and defaulted, i.e.

```c++
template <typename It, typename Char, typename Sentinel = It>
struct arg_join : detail::view {
```

OTOH, `arg_join` isn't in c++20 to be compatible with, and changing the template param order to that now would be a breaking change for anyone who has used it its is current form (since this change hit 7.0). Probably not worth trying to change it back...

vitaut:
Oops, sorry. It is indeed too late to change it back.

puetzk:
Not a problem, I had exactly one use, and it was trivial to fix, and you did label 7 as a major version (so potentially-breaking changes were not a surprise).

PR #1691, opened 2020-05-19
Knuth is using fmt library

fpelliccioni:


vitaut:
Thanks!

fpelliccioni:
Thank you for this excellent library!

PR #1693, opened 2020-05-21
Fix Mingw support

petrutlucian94:
If _POSIX_ is set, mingw will no longer prefix *some* function names
with an underscore, which is expected by libfmt. Rather than handling
them case by case, we'll just avoid the POSIX macros.

Signed-off-by: Lucian Petrut <lpetrut@cloudbasesolutions.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,16 @@ if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
   message(STATUS "CMake version: ${CMAKE_VERSION}")
 endif ()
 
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR but these should be fixed in source

https://github.com/fmtlib/fmt/blob/297c3b2ed551a4989826fc8c4780bf533e964bd9/include/fmt/os.h#L50-L60

not in CMake.

petrutlucian94:
Right, ``_fdopen`` seems to be the only function that's missing the "_" prefix. I've submitted another PR: https://github.com/fmtlib/fmt/pull/1696.

Thanks for the quick feedback.

PR #1696, opened 2020-05-22
Fix Mingw support

petrutlucian94:
If the ``_POSIX_`` flag is set, _fdopen will not be defined by
Mingw headers, which is addressed by this PR.

### DIFF
--- a/include/fmt/os.h
+++ b/include/fmt/os.h
@@ -50,7 +50,7 @@
 #ifdef FMT_SYSTEM
 #  define FMT_POSIX_CALL(call) FMT_SYSTEM(call)
 #else
-#  define FMT_SYSTEM(call) call
+#  define FMT_SYSTEM(call) ::call
 #  ifdef _WIN32
 // Fix warnings about deprecated symbols.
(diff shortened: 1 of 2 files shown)


include/fmt/os.h
```diff
@@ -59,6 +59,14 @@
 #  endif
 #endif
 
+#ifdef __MINGW32__
+// The following will be undefined if the _POSIX_ flag is set.
+// For the rest of the functions, we'll use the underscore prefix.
+#  ifndef _fdopen
+#     define _fdopen fdopen
```
vitaut, line 66: We shouldn't define any global macros. Perhaps the `#ifdef` should go to https://github.com/fmtlib/fmt/blob/8069265373f999ba31dd200e7dd090104a4236ad/src/os.cc#L292 and choose between `FMT_POSIX_CALL` and `FMT_SYSTEM` based on whether `_POSIX_` is defined.

petrutlucian94: Makes sense. I've updated the commit accordingly.

vitaut:
Thank you!

PR #1697, opened 2020-05-22
fix max/min macro

Neargye:


vitaut:
Thanks

PR #1698, opened 2020-05-22
FMT_IF_CONSTEXPR

Neargye:

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -44,7 +44,7 @@ FMT_CONSTEXPR To lossless_integral_conversion(const From from, int& ec) {
   static_assert(T::is_integer, "To must be integral");
 
   // A and B are both signed, or both unsigned.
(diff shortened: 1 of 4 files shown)


vitaut:
Thanks for the PR but I'm not sure if it's necessary. The compiler will optimize the dead branches away. `if constexpr` is only needed when the code in the branch wouldn't compile otherwise.

Neargye:
I decided to do it because I got such a compiler message
```
warning C4127: conditional expression is constant
note: consider using 'if constexpr' statement instead
```

Neargye:
But overall you are right

vitaut:
The warning can be suppressed with `const_check`, e.g. https://github.com/fmtlib/fmt/blob/a265e25b7924d7edec77c28ee76182e69bfcc3fe/include/fmt/format.h#L1612

vitaut:
It's not great but a bit more readable than a macro.

PR #1699, opened 2020-05-23
fix zero flag for char types and make zero flag ignored if a precisio…

rimathia:
…n is specified


vitaut:
Thank you!

PR #1702, opened 2020-05-26
Support custom FMT_INC_DIR in pkgconfig and cmake configs

orivej:
When `CMAKE_INSTALL_INCLUDEDIR` or `FMT_INC_DIR` override the header installation directory, they should be used instead of `${CMAKE_INSTALL_PREFIX}/include` in `fmt-targets.cmake` and `fmt.pc`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -197,8 +197,7 @@ endif ()
 target_compile_features(fmt INTERFACE ${FMT_REQUIRED_FEATURES})
 
 target_include_directories(fmt PUBLIC
-  $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
-  $<INSTALL_INTERFACE:include>)
+  $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>)
 
 set(FMT_DEBUG_POSTFIX d CACHE STRING "Debug library postfix.")
 
@@ -232,8 +231,7 @@ target_compile_definitions(fmt-header-only INTERFACE FMT_HEADER_ONLY=1)
 target_compile_features(fmt-header-only INTERFACE ${FMT_REQUIRED_FEATURES})
 
 target_include_directories(fmt-header-only INTERFACE
-  $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
-  $<INSTALL_INTERFACE:include>)
+  $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>)
 
 # Install targets.
 if (FMT_INSTALL)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -247,11 +245,6 @@ if (FMT_INSTALL)
   set(pkgconfig ${PROJECT_BINARY_DIR}/fmt.pc)
   set(targets_export_name fmt-targets)
 
-  set (INSTALL_TARGETS fmt)
-  if (TARGET fmt-header-only)
-    set(INSTALL_TARGETS ${INSTALL_TARGETS} fmt-header-only)
-  endif ()
```
orivej, line 253: fmt-header-only has been always enabled since 691a7a91.

CMakeLists.txt
```diff
@@ -197,8 +197,7 @@ endif ()
 target_compile_features(fmt INTERFACE ${FMT_REQUIRED_FEATURES})
 
 target_include_directories(fmt PUBLIC
-  $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
-  $<INSTALL_INTERFACE:include>)
```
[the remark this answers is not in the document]
vitaut: I suggest updating `INSTALL_INTERFACE:include` here instead of splitting `target_include_directories` into two.

CMakeLists.txt
```diff
@@ -232,8 +231,7 @@ target_compile_definitions(fmt-header-only INTERFACE FMT_HEADER_ONLY=1)
 target_compile_features(fmt-header-only INTERFACE ${FMT_REQUIRED_FEATURES})
 
 target_include_directories(fmt-header-only INTERFACE
-  $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
-  $<INSTALL_INTERFACE:include>)
```
vitaut, line 236: Ditto

CMakeLists.txt
vitaut: Good catch.

CMakeLists.txt
```diff
@@ -264,14 +257,22 @@ if (FMT_INSTALL)
               "Installation directory for pkgconfig (.pc) files, a relative path "
               "that will be joined to ${CMAKE_INSTALL_PREFIX}, or an arbitrary absolute path.")
 
+  get_filename_component(FMT_INC_DIR_PARENT "${FMT_INC_DIR}" DIRECTORY)
```
vitaut, line 260: I think we should remove `/fmt` from `FMT_INC_DIR` for consistency with `FMT_LIB_DIR` and `CMAKE_INSTALL_INCLUDEDIR`.

orivej:
Done.

vitaut:
Thanks!

PR #1703, opened 2020-05-27
Fix target file include paths

tobim:
If the project is generated with a custom `CMAKE_INSTALL_INCLUDEDIR`, the generated target properties would point to the wrong location.


tobim:
I didn't notice this would already be fixed by #1702. Sorry for the inconvenience.

PR #1705, opened 2020-05-27
Indicate that vectors are a contiguous sequence

### LINKED ISSUE #1704: support vformat_to into an external buffer?
I have tried formatting using vformat_to into a buffer that I manage:

```
std::vector<char> myBuf;
fmt::vformat_to(std::back_inserter(myBuf), format, args);
```

But this fails to compile since there is no such specialization. Is it missing intentionally, or was just a "nobody asked for it yet" case?

0x8000-0000:

Closes #1704

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -1719,6 +1719,9 @@ template <typename Char>
 struct is_contiguous<std::basic_string<Char>> : std::true_type {};
 
 template <typename Char>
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR but unfortunately this specialization cannot live in `fmt/core.h` because dependency on `<vector>` is an implementation detail and may be removed in the future. However, you can add this specialization to your code.

PR #1706, opened 2020-05-29
Don't install sphinx cache files

jelly:
When building documentation sphinx creates cached files in the .doctrees
directory and aren't required for viewing documentation only for
building. As added benefit this makes fmt reprodcubile as the cached
files are different when the build environment is varied.


vitaut:
Thank you!

PR #1717, opened 2020-06-05
fix typo which caused the loss of the counting information when using…

rimathia:
… a printf context with a truncating_iterator


vitaut:
Thank you!

PR #1729, opened 2020-06-15
Added  #define WIN32_LEAN_AND_MEAN before including windows.h

gabime:
This prevents winsock2 redefinition warnings (e.g. see issue https://github.com/gabime/spdlog/issues/1589) and minimizes the stuff that windows.h includes.

I also reversed  the `#ifdef`condition to avoid redefinition warnings if either  `NOMINMAX` or `WIN32_LEAN_AND_MEAN` is already defined.


vitaut:
Thanks!

PR #1734, opened 2020-06-19
6.2.1 to string fix

ashutosh108:
PR after that question: https://stackoverflow.com/q/62423981/19905

This only applies cleanly to tags/6.2.1. I can't find a stable branch (or equivalent) in fmt, and github doesn't seem to allow creating PRs for tags, so applying for master branch. If there is a better way for me to submit this commit https://github.com/ashutosh108/fmt/commit/e8b47c3efb17609659645c0dfdab949cb5ecdc00 for tags/6.2.1, please let me know.

I can't create working PR for the most current version as it seems to be in the process of re-working, with to_string() apparently failing for user-defined classes (after commit https://github.com/fmtlib/fmt/commit/7431165f38bc608c8367f5511ad670e1afa9aefa ) and I can't fix that.
By the way, this PR contains a test for to_string()-ing user-defined class so this kind of regressions should be caught in the future.

P.S. I agree that my contributions are licensed under the {fmt} license, and agree to future changes to the licensing.

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -1,10 +1,11 @@
 6.2.1 - TBD
 -----------
 
-* Fix ostream support in sprintf
+* Fixed ostream support in ``sprintf``
   (`#1631 <https://github.com/fmtlib/fmt/issues/1631>`_).
 
-* Fix inconsistent type detection
+* Fixed type detection when using implicit conversion to ``string_view`` and
+  ostream ``operator<<`` inconsistently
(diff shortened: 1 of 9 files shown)


vitaut:
Thanks for the PR. I applied a slightly different fix since as you correctly pointed out `to_string` is being refactored: https://github.com/fmtlib/fmt/commit/2cac8a9d2e36cd920a9a60ec5b776c187e843fbb, so this should work in master.

Issue #1736, opened 2020-06-21
Emscripten > wasm-ld: error: unknown argument: --as-needed


ArthurSonzogni:
When using fmt with emscripten, I got my build broken when updating fmt:
```
wasm-ld: error: unknown argument: --as-needed
```
This is caused by the lines:
```cmake
if (BUILD_SHARED_LIBS)
  if (UNIX AND NOT APPLE AND NOT ${CMAKE_SYSTEM_NAME} MATCHES "SunOS")
    # Fix rpmlint warning:
    # unused-direct-shlib-dependency /usr/lib/libformat.so.1.1.0 /lib/libm.so.6.
    target_link_libraries(fmt -Wl,--as-needed)
  endif ()
  target_compile_definitions(fmt PRIVATE FMT_EXPORT INTERFACE FMT_SHARED)
endif ()
```
We probably should just exclude the emscripten. I will make a patch soon to fix this.

PR #1737, opened 2020-06-21
Do not use -Wl,--as-needed with emscripten.

ArthurSonzogni:
On some platform, fmt is using the following link flag:
```cmake
  target_link_libraries(fmt -Wl,--as-needed)
```

This is not supported by wasm-ld as of today.
This patch stop using it when compiling using emscripten.

Bug: https://github.com/fmtlib/fmt/issues/1736


vitaut:
Thank you!

ArthurSonzogni:
Thanks for the review!

PR #1749, opened 2020-07-02
Add CE6 support

escherstair:
I was able to compile 4.x branch for CE6 and run a demo exmple
`fmt::print("Hello, {}!", "world");`
with the changes in this PR.
I'm not sure if the workaround for the lack of `errno` and `<cerrno>` are ok.
CE6 doesn't implement `errno`, `<cerrno>` and `strerror()` and so I return 0.

It's really useful if someone can review this PR and give impressions.

### DIFF
--- a/fmt/format.cc
+++ b/fmt/format.cc
@@ -30,7 +30,9 @@
 #include <string.h>
 
 #include <cctype>
-#include <cerrno>
+#if !defined(UNDER_CE)
+# include <cerrno>
+#endif
 #include <climits>
 #include <cmath>
 #include <cstdarg>
@@ -98,7 +100,11 @@ namespace {
 inline int fmt_snprintf(char *buffer, size_t size, const char *format, ...) {
   va_list args;
   va_start(args, format);
+# if !defined(UNDER_CE)
   int result = vsnprintf_s(buffer, size, _TRUNCATE, format, args);
+# else
+  int result = _vsnprintf_s(buffer, size, _TRUNCATE, format, args);
+# endif
   va_end(args);
   return result;
 }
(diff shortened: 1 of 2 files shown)


fmt/format.cc
```diff
@@ -137,11 +147,13 @@ int safe_strerror(
     // A noop assignment operator to avoid bogus warnings.
     void operator=(const StrError &) {}
 
+#if !defined(UNDER_CE)
```
vitaut, line 150: If the problem is lack of `errno` then I think an easier workaround is to define a dummy `errno` as a member of `StrError` itself. Then you'll only need one ifdef.

escherstair: The problem is the lack of `errno` and `strerror()`.
So I can add a dummy `errno` member (and I must initialize it in the constructor) and a dummy `strerror()` implementation.

Do you suggest to proceed in this way?

vitaut: I guess adding `errno` won't help then.

fmt/format.cc
```diff
@@ -171,9 +183,13 @@ int safe_strerror(
 
     // Fallback to strerror if strerror_r and strerror_s are not available.
     int fallback(internal::Null<>) {
+#if !defined(UNDER_CE)
       errno = 0;
       buffer_ = strerror(error_code_);
       return errno;
+#else
+	  return 0;
```
vitaut, line 191: Would it make sense to write some generic error message if strerror if not available?

Also the indentation is off.

escherstair: I don't know.
Which function in fmt sets `errno` in case of failure?
I can chack what happens in this kind of situations in CE6 and see how to handle in the best way.

vitaut: Thinking more of it, I think it's fine as is since the strerror message is normally used only as a part of error message.

vitaut:
Merged, thanks.

PR #1750, opened 2020-07-02
Fix a conversion warning with Clang10 on Windows

### LINKED ISSUE #1747: -Wsign-conversion warning with Clang10 on Windows
include\fmt/format.h(560,71): error: implicit conversion changes signedness: 'std::size_t' (aka 'unsigned long long') to 'stdext::checked_array_iterator<char *>::difference_type' (aka 'long long') [-Werror,-Wsign-conversion]
  std::uninitialized_copy(begin, end, make_checked(ptr_, capacity_) + size_);

0x8000-0000:
Fixes #1747

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -566,7 +566,7 @@ template <typename U>
 void buffer<T>::append(const U* begin, const U* end) {
   size_t new_size = size_ + to_unsigned(end - begin);
   reserve(new_size);
-  std::uninitialized_copy(begin, end, make_checked(ptr_, capacity_) + size_);
+  std::uninitialized_copy(begin, end, make_checked(ptr_ + size_, capacity_ - size_));
   size_ = new_size;
 }
 }  // namespace detail


include/fmt/format.h
```diff
@@ -566,7 +566,7 @@ template <typename U>
 void buffer<T>::append(const U* begin, const U* end) {
   size_t new_size = size_ + to_unsigned(end - begin);
   reserve(new_size);
-  std::uninitialized_copy(begin, end, make_checked(ptr_, capacity_) + size_);
+  std::uninitialized_copy(begin, end, make_checked(ptr_ + size_, capacity_));
```
vitaut, line 569: Thanks for the PR. I think you also need to subtract `size_` from `capacity_`.

0x8000-0000: Will do, but why do all the test pass? :)

vitaut: Because we don't write over capacity. The whole checked iterator machinery in MSVC is completely useless and is here mostly to avoid warnings but let's provide correct bounds anyway.

vitaut:
Thank you!

PR #1751, opened 2020-07-04
Add ClickHouse to the list of projects

alexey-milovidov:

I think it's good to have a link to a project with 10k+ stars.
It aligns perfectly with some other databases in the list.

Thank you for the great library!


vitaut:
Thank you!

PR #1757, opened 2020-07-07
Fix unresolved reference error

Kurkin:
I'm using {fmt} in not header-only mode. After upgrade to 7.0.0 build is broken due to unresolved reference.
```
error LNK2001: unresolved external symbol "class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > __cdecl fmt::v6::detail::vformat(class fmt::v6::basic_string_view<char>,struct fmt::v6::format_args)"
```


vitaut:
Thank you!

Issue #1758, opened 2020-07-07
Default formats for user defined objects based on reflection


martinitus:
Hey there,
after posting on [SO](https://stackoverflow.com/questions/62679099/extension-plans-for-formatting-of-trivial-structs-in-c-fmt)  and putting some more thoughts into it, it may make sense to move the discussion here...

Essentially, this is a feature request for something as follows:
```c++

struct Inner {...};
struct Outer {...};

template<>
struct reflection<Outer> {
    /*definition of class name and field names has to provided manually...*/
};

template<>
struct reflection<Inner> {
    /*definition of class name and field names has to provided manually...*/
};

auto outer = Outer{.a=1,.b="hello",.inner={.x=3.12,.y=" ",.z=2}};

std::string simple = fmt::format("{:s}", outer); // :s means format as simple
assert(simple == "a|hello|3.12| |2");

std::string  extended = fmt::format("{:e}",outer); // :e means format as extended
assert(extended == "Outer{.a=1, .b=hello, .inner=Inner{.x=3.12, .y= , .z=2}}");
```

I tried to structure my thoughts... hope its not to boring/wrong!

# Current State
Currently, `{fmt}` supports formatting of user defined types by specialization of the `fmt::formatter` template.
Formatting of custom types does not require any specific format string, however,
custom types can support custom format strings via the a `auto parse(ParseContext& ctx)` method in the specialized 
`formatter`. Examples:

```c++
fmt::format("{}", CustomType{...}); // format according to default format for custom type.
fmt::format("{:foobar}", CustomType{...}); // use "foobar" to configure the formatter before formatting.
```

Further, `{fmt}` allows formatting of tuples and vectors when including the `<fmt/ranges.h>` header:

```c++
#include <fmt/ranges.h>
fmt::format("{}", fmt::join(std::vector<int>{1,2,3},";")); // results in "1;2;3"
// using custom types in a join:
fmt::format("{}", fmt::join(std::tuple{1,"hello",CustomType{}},";")); // equivalent to 
fmt::format("1;hello;{}", CustomType{}); // i.e. both lines will use the default custom formatter
```

Here, there are two little conceptual inconsistencies:
 - the formatter to be used is not selected via the type of the argument directly, instead the formatter for the result 
   type of the join function is used (some `JoinType`).
 - The configuration for this formatter is not part of the format string (i.e. _not_ `fmt::format("{sep=;}",...)`) but 
   passed in via the custom join type. 

(please don't take that part as criticism!)
[...]

PR #1761, opened 2020-07-07
Feature/structure formatting

martinitus:

Poc code integrated into {fmt} repository and test framework.


martinitus:
Ahh, sorry for the spam, I wanted to open this in my own fork.... 🤦

vitaut:
Commented in #1758.

martinitus:
By the way, I found some comment in your cmake files with regard of compilation of tests with c++17 on msvc because of gtest/gmock.
The first commit here replaces all the `std::tr1::tuple` usage with `std::tuple` that fixed compilation for me on current msvc with c++17. No idea if it also works for older versions -.-

vitaut:
> The first commit here replaces all the std::tr1::tuple usage with std::tuple that fixed compilation for me on current msvc with c++17.

This might have been fixed in newer gtest. A PR to update it is welcome.

PR #1763, opened 2020-07-07
Remove allocating containers from on_num

### LINKED ISSUE #1762: Remove potentially allocating containers from integer_writer::on_num()
## Setting

In the process of attempting to reduce the binary size impact of this library for an embedded project, I noticed that malloc/free and operator new/delete were being linked into the binary via the std::basic_string constructor. 

The sample code I used to make sure that no potentially allocating container was being used was the following:

```C++
#define FMT_HEADER_ONLY
#define FMT_STATIC_THOUSANDS_SEPARATOR ','
#define FMT_USE_FLOAT 0
#define FMT_USE_DOUBLE 0
#define FMT_USE_LONG_DOUBLE 0
#include "third_party/fmt-7.0.0/include/fmt/format.h"

#include <array>
#include <cstddef>

template <class T, size_t buffer_size>
class FormatBuffer
{
 public:
  using value_type = T;

  size_t size()
  {
    return size_;
  }

  size_t max_size()
  {
    return buffer_size;
  }

  void push_back(T & value)
  {
    buffer_[size_++] = value;
  }

  void push_back(const T & value)
  {
    buffer_[size_++] = value;
  }

  void push_back(T && value)
  {
    buffer_[size_++] = value;
  }

  T * data()
  {
    return buffer_.data();
  }

  size_t size_ = 0;
  std::array<T, buffer_size> buffer_{};
};

int main()
{
  FormatBuffer<char, 256> buffer;
  fmt::format_to_n(std::back_inserter(buffer), buffer.max_size(),
                    "Argument = {}", 5);
  return 0;
}
```

## Problem

This code shouldn't require the allocating functions to be pulled in, but it does. This is due to the fact that `integer_writer::on_num()` uses a `std::string` to hold its locale set of characters. Also, `basic_memory_buffer` uses `std::allocator` which will cause the allocating functions to be linked in as well. By switching `std::string` to `std::string_view` and from `basic_memory_buffer` to `std::array`, the linking of malloc and free can be eliminated, given the above approach.

kammce:
Use std::string_view and std::array containers when
FMT_STATIC_THOUSANDS_SEPARATOR is defined.

This will help in the case for embedded system where even using
containers that can potentially allocate results in malloc/free and
new/delete being linked into the binary. In many embedded cases dynamic
library allocation is prohibited and even the cost of linking in these
libraries costs too high of a cost in flash size.

Resolves #1762

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -34,6 +34,7 @@
 #define FMT_FORMAT_H_
 
 #include <algorithm>
+#include <array>
 #include <cerrno>
 #include <cmath>
 #include <cstdint>
@@ -1552,13 +1553,18 @@ template <typename OutputIt, typename Char, typename UInt> struct int_writer {
   enum { sep_size = 1 };
 
   void on_num() {
+#if !defined(FMT_STATIC_THOUSANDS_SEPARATOR)
     std::string groups = grouping<Char>(locale);
+#else
+    constexpr const char kDefaultGroups[] = "\03";
+    basic_string_view<Char> groups(kDefaultGroups, sizeof(kDefaultGroups));
+#endif
     if (groups.empty()) return on_dec();
     auto sep = thousands_sep<Char>(locale);
     if (!sep) return on_dec();
     int num_digits = count_digits(abs_value);
     int size = num_digits, n = num_digits;
-    std::string::const_iterator group = groups.cbegin();
+    auto group = groups.cbegin();
     while (group != groups.cend() && num_digits > *group && *group > 0 &&
            *group != max_value<char>()) {
       size += sep_size;
@@ -1568,9 +1574,9 @@ template <typename OutputIt, typename Char, typename UInt> struct int_writer {
     if (group == groups.cend()) size += sep_size * ((n - 1) / groups.back());
     char digits[40];
     format_decimal(digits, abs_value, num_digits);
-    basic_memory_buffer<Char> buffer;
+    // TODO(kammce): remove the * 2!
+    std::array<Char, inline_buffer_size * 2> buffer;
     size += prefix_size;
-    buffer.resize(size);
     basic_string_view<Char> s(&sep, sep_size);
     // Index of a decimal digit with the least significant digit having index 0.
     int digit_index = 0;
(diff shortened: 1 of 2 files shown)


include/fmt/format.h
```diff
@@ -1568,9 +1574,9 @@ template <typename OutputIt, typename Char, typename UInt> struct int_writer {
     if (group == groups.cend()) size += sep_size * ((n - 1) / groups.back());
     char digits[40];
     format_decimal(digits, abs_value, num_digits);
-    basic_memory_buffer<Char> buffer;
+    // TODO(kammce): remove the * 2!
+    std::array<Char, inline_buffer_size * 2> buffer;
```
vitaut, line 1578: This won't work because there is no limit on the separator size. What you could do instead is pass an allocator to memory buffer that throws/aborts whenever it wants to allocate.

include/fmt/format.h
```diff
@@ -1552,13 +1553,18 @@ template <typename OutputIt, typename Char, typename UInt> struct int_writer {
   enum { sep_size = 1 };
 
   void on_num() {
+#if !defined(FMT_STATIC_THOUSANDS_SEPARATOR)
     std::string groups = grouping<Char>(locale);
+#else
+    constexpr const char kDefaultGroups[] = "\03";
+    basic_string_view<Char> groups(kDefaultGroups, sizeof(kDefaultGroups));
```
vitaut, line 1560: You can probably remove https://github.com/fmtlib/fmt/blob/60c43e870387affb07de32ed4fc2794ac7a9f60d/include/fmt/format-inl.h#L209-L212 with this change.

include/fmt/format.h
kammce: I'll make the change for this this weekend.

kammce:
Closing this for now. 

After some close inspect of the code, when I was attempting to remove all references to `std::allocator`, I actually removed it directly from `basic_memory_buffer`. Doing this and switching away from `std::string` as shown in this PR, where the key to removing references to `malloc`/`free` and `std::allocator`. But that is a pretty hefty change to this library.

I'm not sure if the right move is to try and template or strip out `std::allocator` everywhere in this library at this time. I'm going to do more experimentation with my codebase (arm-none-eabi-gcc-9 on Cortex-M3, Cortex-M4, etc) to see if I can figure out a method for doing this that doesn't require a large rewrite of the code base. At least for now.

PR #1773, opened 2020-07-14
Do not pass empty format_str to UDT formatters

fghzxm:
Basically changed `format.h:1837` to actually pass the format string instead of `{}`.

For consistency I did a `grep -r. -nFwe on_replacement_field` and added an additional `const Char*` to every function with that name.  All the ones except `format_handler::on_replacement_field` does not appear to be used in any way by any code however.  Should those functions be touched?

### DIFF
--- a/include/fmt/compile.h
+++ b/include/fmt/compile.h
@@ -94,7 +94,7 @@ template <typename Char> struct part_counter {
     return ++num_parts, 0;
   }
 
-  FMT_CONSTEXPR void on_replacement_field(int, const Char*) {}
+  FMT_CONSTEXPR void on_replacement_field(int, const Char*, const Char*) {}
 
   FMT_CONSTEXPR const Char* on_format_specs(int, const Char* begin,
                                             const Char* end) {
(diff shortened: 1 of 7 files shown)


vitaut:
Could you elaborate what problem are you trying to solve?

fghzxm:
Ugh, sorry for not clarifying that.

User-defined type formatters are currently broken because `format.h:1837` passes an empty string view as the format string to the formatter.  This commit fixes that by actually passing the part of format string beginning at the format specs to the formatter.

I've added a test for user-defined type formatters, which fails without that patch ([Travis fghzxm/compare/patch-2](https://travis-ci.org/github/fghzxm/fmt/jobs/708192857))

vitaut:
Thanks for the PR but the current behavior is correct, quoting the [documentation](https://fmt.dev/latest/api.html#formatting-user-defined-types):

> The formatter should parse specifiers until '}' or the end of the range.

Note the "end of the range" part. The parse function shouldn't assume existence of `}`.

Issue #1772, opened 2020-07-14
FMT_COMPILE vs FMT_STRING vs ""_format()


leolchat:
The documentation is not extremely clear about the difference between the 3 ways I found to have compile-time format string checks. I would like to try to PR an improvement to the documentation, but first I need to be sure

- Is there a difference as to what they can check? In my quick survey I could not find an example rejected by one but not the others.
- Is there a preferred one? Compiler verbose output seems shorter with the `FMT_STRING` version, but not using a macro is nice, or is the speed boost from `FMT_COMPILE` usually worth it?

Bonus question, is there one which has a chance to land in `std::format`?

vitaut:
> Is there a difference as to what they can check?

Yes, `FMT_STRING` always does compile-time checks while `_format` only if a compiler supports UDL templates extensions.

> Is there a preferred one?

`FMT_STRING` and `FMT_COMPILE` serve different purposes. If you only need checks then use `FMT_STRING`. If you want to avoid format string parsing overhead then use `FMT_COMPILE` (this is rarely needed) and note that:

> Format string compilation can generate more binary code compared to the default API and is only recommended in places where formatting is a performance bottleneck.

leolchat:
Thanks for clarifying. I would say the documentation could be more explicit. I gave it a try.

PR #1775, opened 2020-07-15
Trying to clear ambiguous compile time claims

leolchat:
Documentation was a bit misleading. Many people quickly assume that fmt does compile time checks by default, while it requires the use of `FMT_STRING`. It was also unclear that `FMT_COMPILE` does the same checks.

See initial discussion here: https://github.com/fmtlib/fmt/issues/1772

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -114,9 +114,12 @@ string checks, wide string, output iterator and user-defined type support.
 Compile-time Format String Checks
 ---------------------------------
 
-Compile-time checks are supported for built-in and string types as well as
-user-defined types with ``constexpr`` ``parse`` functions in their ``formatter``
-specializations.
+Compile-time checks are enabled when using ``FMT_STRING``. It supports built-in
+and string types as well as user-defined types with ``constexpr`` ``parse``
+functions in their ``formatter`` specializations.
+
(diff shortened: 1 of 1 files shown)


doc/api.rst
```diff
@@ -114,9 +114,12 @@ string checks, wide string, output iterator and user-defined type support.
 Compile-time Format String Checks
 ---------------------------------
 
-Compile-time checks are supported for built-in and string types as well as
-user-defined types with ``constexpr`` ``parse`` functions in their ``formatter``
-specializations.
+Compile-time checks are enabled when using ``FMT_STRING``. It supports built-in
```
vitaut, line 117: It supports -> They support

doc/api.rst
```diff
@@ -114,9 +114,12 @@ string checks, wide string, output iterator and user-defined type support.
...
+and string types as well as user-defined types with ``constexpr`` ``parse``
+functions in their ``formatter`` specializations.
+
+When the compiler supports UDL templates extensions, using the litteral string
+``""_format()`` will also enable compile-time checks.
```
vitaut, line 122: I suggest dropping this part because the extension is obsolete (gcc even warns on it now). `FMT_STRING` is the only recommended API for compile-time checks.

vitaut:
Thanks

PR #1777, opened 2020-07-16
Optimize bloat generated by write_padding

### LINKED ISSUE #1774: Reduce Binary Size by Optimizing write_padded
## Current Situation

The following is the full function signature for the `write_padded` function:

```C++
template <align::type align = align::left, typename OutputIt, typename Char,
          typename F>
inline OutputIt write_padded(OutputIt out,
                             const basic_format_specs<Char>& specs, size_t size,
                             size_t width, const F& f) {
```

This templated function generates 11 separate instances of itself, with an average implementation size of around 325 bytes each, based on its usage within the fmt::details namespace. This leaves a great opportunity for optimizing the binary requirements of fmt by reducing the number of instances of this function generated. This can be done by eliminating some template parameters.

## Proposal

Remove the `align` and `typename F` templates parameters.

### Removing `align`

Simply make align a function parameter. This means that parts of the function that use align in what could be a `constexpr` context, no longer will get that speedup. But that speed up results in just a few assignments and look ups to produce, so its not that big of a performance hit.

### Removing `typename F`

This one is a bit harder and will require an API change to `write_padded`. This should be fine since this is just the internal details of the library and is not apart of the public API.

It would look like this:

```C++
template <typename OutputIt, typename Char>
struct WritePadding
{

  constexpr WritePadding(OutputIt out,
                             const basic_format_specs<Char>& specs, 
                             size_t size, 
                             size_t width, 
                             align::type align = align::left): out_(out), specs_(specs)
  {
    unsigned spec_width = to_unsigned(specs.width);
    size_t padding = spec_width > width ? spec_width - width : 0;
    auto* shifts = align == align::left ? data::left_padding_shifts
                                        : data::right_padding_shifts;
    left_padding_ = padding >> shifts[specs.align];
    right_padding_ = padding - left_padding_;
    it_ = reserve(out, size + padding * specs.fill.size());
  }

  constexpr WritePadding(OutputIt out,
                             const basic_format_specs<Char>& specs, 
                             size_t size, 
                             align::type align = align::left): 
                             WritePadding(out, specs, size, size, align)
  {
  }

  constexpr WritePadding& left()
  {
    it_ = fill(it_, left_padding_, specs_.fill);
    return *this;
  }

  template <typename F>
  constexpr WritePadding& content(const F& f)
  {
    it_ = f(it_);
    return *this;
  }

  constexpr OutputIt right()
  {
    it_ = fill(it_, right_padding_, specs_.fill);
    return base_iterator(out_, it_);
  }

  OutputIt out_;
  const basic_format_specs<Char>& specs_;
  decltype(reserve(out_, 0)) it_ = reserve(out_, 0);
  size_t left_padding_ = 0;
  size_t right_padding_ = 0;
};

// Usage:

  auto it = WritePadding(out, specs, data.size, align::right).left().content([=](iterator it) {
    if (prefix.size() != 0)
      it = copy_str<Char>(prefix.begin(), prefix.end(), it);
    it = std::fill_n(it, data.padding, static_cast<Char>('0'));
    return f(it);
  }).right();
```

So two things are happening here, where (1) invocation is split into 3 parts, `left()`, `content()`, `right()` which allows the `template <typename F>` template parameter to be isolated to just 1 single 1 liner method call. This reduces the baggage generated from templating the whole implementation and that 1 line method in general gets inlined.

## Conclusion

This resulted in a small 1105 byte reduction in my binary. With all of the configurations used in my app, fmt accounts for about 16kB, so bringing it down by 1k is nice, but I felt it would reduce it by 2kB. I'll continue to investigate.

In the mean time I'm currently cleaning up a PR for this. Wondering if I'm missing something with this change?

## Project Tested On

- arm-none-eabi-gcc 
- cortex-m3
- -Os -fno-lto
- float/double/long double all disabled

kammce:
Reimplment write_padded function as a structure that removes align
and <typename F> template parameters in order to reduce the number of
template specializations generated. This results in a binary size
reduction.

Caveat, static_assert to detect invalid alignment has been removed as
the aligment parameter has been removed from the template arguments and
added to the constructor parameters.

Resolves #1774

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -1397,43 +1397,81 @@ FMT_NOINLINE OutputIt fill(OutputIt it, size_t n, const fill_t<Char>& fill) {
   return it;
 }
 
-// Writes the output of f, padded according to format specifications in specs.
-// size: output size in code units.
-// width: output display width in (terminal) column positions.
-template <align::type align = align::left, typename OutputIt, typename Char,
-          typename F>
-inline OutputIt write_padded(OutputIt out,
-                             const basic_format_specs<Char>& specs, size_t size,
-                             size_t width, const F& f) {
-  static_assert(align == align::left || align == align::right, "");
-  unsigned spec_width = to_unsigned(specs.width);
-  size_t padding = spec_width > width ? spec_width - width : 0;
-  auto* shifts = align == align::left ? data::left_padding_shifts
-                                      : data::right_padding_shifts;
-  size_t left_padding = padding >> shifts[specs.align];
-  auto it = reserve(out, size + padding * specs.fill.size());
-  it = fill(it, left_padding, specs.fill);
-  it = f(it);
-  it = fill(it, padding - left_padding, specs.fill);
-  return base_iterator(out, it);
-}
+// Helper class to handle writing padding to an an output function according to
+// a format specification.
+// Typically Usage:
+//
+//  return write_padding(iterator, specs, size)
+//      .left()
+//      .content([](iterator it) {
+//        /* do something */
+//      })
+//      .right()
+//      .get_base_iterator();
+//
+template <typename OutputIt, typename Char> struct write_padding {
+  // Writes the output of f, padded according to format specifications in specs.
+  //
+  //  size: output size in code units.
+  // width: output display width in (terminal) column positions.
+  write_padding(OutputIt out, const basic_format_specs<Char>& specs,
+                size_t size, size_t width, align::type align = align::left)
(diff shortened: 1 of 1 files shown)


kammce:
I'm not sure what exactly has gone wrong here:

https://ci.appveyor.com/project/vitaut/fmt/builds/34121568/job/grhm9hn1ph9bdf7s

vitaut:
Thanks for the PR but I don't think it's worth splitting `write_padded` like this because the resulting code is too repetitive.

PR #1781, opened 2020-07-17
Add FMT_REDUCE_INT_INSTANTIATIONS flag

### LINKED ISSUE #1778: Reduce Binary Size by Removing <typename Uint> from int_writer
## Current Situation

Current class template parameters for int_writer:

```C++
template <typename OutputIt, typename Char, typename UInt> struct int_writer {}
```

## Proposal

Removing `<typename UInt>` from the structure template parameters. This would eliminate a few of the template instances generated of the int_writer structure. Simply replacing the Uint with uint64_t showed a 2476 byte reduction for the following:

- Compiler: arm-none-eabi-gcc
- Processor: cortex-m3
- Optimization level: -Os -fno-lto
- float/double/long double all disabled

What do you all think? I'm thinking of either using the largest data size or something along those lines in order to eliminate this template parameter.

kammce:
Reduce code bloat by removing multiple instantiation of int_writer based
on the <typename UInt> parameter.

Rationale:
- The only functions that gains a speedup by int size would be
  int_writer::on_dec()'s call to count_digits which uses CLZ. Thus to
  still take advantage of this speedup, we store the size of the int
  so we can use a switch statement to call the correct count_digits.
- All other implementations of count_digits require some sort of looping
  that terminates when the value hits zero regardless of what sized int
  it is.

Caveats:
- There is a performance hit when dealing with and passing around
  64-bit/128-bit values compared to 32-bit values on 32-bit platforms,
  and with 64-bit values on 64-bit systems. But this should not reduce the
  performance that dramatically.
- There is also a performance hit for on_dec() due to the addition of a
  switch case. But, due to it size, this should reduce to a jump table.

Resolves #1778

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -311,6 +311,10 @@ struct int128_t {};
 struct uint128_t {};
 #endif
 
+#if !defined(FMT_USE_SMALLEST_INT)
+#define FMT_USE_SMALLEST_INT 1
+#endif
+
 // Casts a nonnegative integer to unsigned.
 template <typename Int>
 FMT_CONSTEXPR typename std::make_unsigned<Int>::type to_unsigned(Int value) {
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -311,8 +311,12 @@ struct int128_t {};
 struct uint128_t {};
 #endif
 
-#if !defined(FMT_USE_SMALLEST_INT)
-#define FMT_USE_SMALLEST_INT 1
+// Defining FMT_REDUCE_INT_INSTANTIATIONS to 1, will reduce the number of
+// int_writer template instances to just one by only using the largest integer
+// type. This results in a reduction in binary size but will cause a decrease in
+// integer formatting performance.
+#if !defined(FMT_REDUCE_INT_INSTANTIATIONS)
+#  define FMT_REDUCE_INT_INSTANTIATIONS 0
 #endif
 
 // Casts a nonnegative integer to unsigned.
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -311,14 +311,6 @@ struct int128_t {};
 struct uint128_t {};
 #endif
 
-// Defining FMT_REDUCE_INT_INSTANTIATIONS to 1, will reduce the number of
-// int_writer template instances to just one by only using the largest integer
-// type. This results in a reduction in binary size but will cause a decrease in
-// integer formatting performance.
-#if !defined(FMT_REDUCE_INT_INSTANTIATIONS)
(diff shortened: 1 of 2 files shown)


kammce:
@vitaut is there a make target to run a performance test?

vitaut:
> There is a performance hit when dealing with and passing around
64-bit/128-bit values compared to 32-bit values on 32-bit platforms,
and with 64-bit values on 64-bit systems.

This is a non-starter.

A better option would be to keep `int_writer` parameterized on integer type and add a configuration option to make `uint32_or_64_or_128_t` only map to the largest type. This won't add overhead for the common case.

kammce:
Hmmm, I like it. I can make that change.

vitaut:
Looks good but I suggest renaming `FMT_USE_SMALLEST_INT` to something that more clearly describes the effect of this option. Maybe `FMT_REDUCE_INT_INSTANTIATIONS`?

include/fmt/core.h
```diff
@@ -311,6 +311,14 @@ struct int128_t {};
 struct uint128_t {};
 #endif
 
+// Defining FMT_REDUCE_INT_INSTANTIATIONS to 1, will reduce the number of
+// int_writer template instances to just one by only using the largest integer
+// type. This results in a reduction in binary size but will cause a decrease in
+// integer formatting performance.
+#if !defined(FMT_REDUCE_INT_INSTANTIATIONS)
+#  define FMT_REDUCE_INT_INSTANTIATIONS 0
+#endif
```
vitaut, line 320: Please move this to `format.h` since it's only used there.

vitaut:
Thanks!

vitaut:
@kammce, BTW for future reference, what platform are you targeting and what are the savings with `-DFMT_REDUCE_INT_INSTANTIATIONS=1`?

kammce:
TL;DR: Saved **2024 bytes**

I'm surprised I didn't mention that at the start.

**Target CPU**: Cortex M3 (although I test on M4F, M7 etc) 
**Target MCU**: stm32f103 (64kB of flash)
**Compiler**: gcc-arm-none-eabi-9-2020-q2-update
**Optimization Level**: -Os

Following flags were used to get the binary as small as possible during testing: 

```C++
#define FMT_STATIC_THOUSANDS_SEPARATOR ','
#define FMT_USE_FLOAT 0
#define FMT_USE_DOUBLE 0
#define FMT_USE_LONG_DOUBLE 0
#define FMT_REDUCE_INT_INSTANTIATIONS 1
```

Binary increased by **10,812 bytes** when using fmt with the above flags turned on.
**12,836 bytes** is the increase with `-D FMT_REDUCE_INT_INSTANTIATIONS 0`.

Thus the binary amount saved is **2024 bytes**.

PR #1783, opened 2020-07-18
Bump tested CMake version to 3.18

tambry:
Use the version range feature introduced in 3.12. On CMake <3.12 the extra dots are simply interpreted as extra version number separators.  
The fallback for ancient CMake versions is kept.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,10 +1,8 @@
-cmake_minimum_required(VERSION 3.1.0)
+cmake_minimum_required(VERSION 3.1...3.18)
 
-# Use newer policies if available, up to most recent tested version of CMake.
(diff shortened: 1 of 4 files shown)


vitaut:
Thanks for the PR. Is there any benefit to using a version range?

tambry:
> Thanks for the PR. Is there any benefit to using a version range?

Compared to the current approach of an `if` it's simpler.

Otherwise, using a range enables policies for versions up to the max version (usually the latest tested version).  
So with `cmake_minimum_required(VERSION 3.1...3.18)` CMake 3.1 is the minimum version required. Using CMake 3.14 you get the `NEW` policies up to version 3.14. Using CMake 3.20 you'd get policies up to version 3.18.

Issue #1784, opened 2020-07-18
join() forces const-ness on the range which prevents it accepting some range-v3 ranges


tonyelewis:
Thanks so much for all your fantastic work on fmt, standardisation and Tweeting.

The following attempt to combine range-v3 with fmt fails to compile under Clang or GCC (`-std=c++17 -fsyntax-only`) ([Godbolt](https://godbolt.org/z/9shjWc)) :

~~~cpp
#include <fmt/core.h>
#include <fmt/ranges.h>

#include <range/v3/view/filter.hpp>

using ::std::array;
using ::std::literals::string_literals::operator""s;
using ::std::string;

namespace views = ::ranges::views;

int main() {
    const array some_strs{ "a"s, "bb"s, "ccc"s };
    ::fmt::join(
        some_strs | views::filter(
            [] (const string &x) { return x.length() != 2; }
        )
        ,
        ""
    );
}
~~~

I'm pretty sure the problem is caused by the `const`s around [`format.h:3388`](https://github.com/fmtlib/fmt/blob/7a01c9c52320e7f42a31882b00378fd940b68d31/include/fmt/format.h#L3388). The range-v3 design involves some views (such as `filter` in the example above) updating internal state (eg a cached iterator) so they have `const`-only `begin()` / `end()`, which means a `const` reference to such a view is not a valid range. See "View const-ness" on https://ericniebler.github.io/range-v3/ for more info.

vitaut:
Thanks for the suggestion. A PR to make the range requirements more compatible with ranges-v3 would be welcome.

tonyelewis:
Thanks. I've submitted #1786 as you'll see above.

BTW, I'm seeing various test failures in GCC and under sanitizers. Are these know issues or should I submit issues for them?

vitaut:
> Are these know issues or should I submit issues for them?

Please submit an issue.

tonyelewis:
Thanks.

PR #1786, opened 2020-07-20
Make join() handle non-const-only begin/end ranges

tonyelewis:
See fmtlib/fmt#1784.

Add tests that demonstrate the problem and check obvious variations.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -3420,15 +3420,15 @@ arg_join<It, Sentinel, wchar_t> join(It begin, Sentinel end, wstring_view sep) {
   \endrst
  */
 template <typename Range>
-arg_join<detail::iterator_t<const Range>, detail::sentinel_t<const Range>, char>
-join(const Range& range, string_view sep) {
+arg_join<detail::iterator_t<Range>, detail::sentinel_t<Range>, char>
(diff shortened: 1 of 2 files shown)


tonyelewis:
Roger - I've applied clang-format.

mattgodbolt:
Will there be a release including this fix any time? I just hit it and saw that it's been fixed for a while, but doesn't seem to be any 7.0.x release (although the fix is a few months old by 7.0.x.) Thanks!

vitaut:
I plan to make a new minor release in the next two months or so. This is more than a fix which is why it's not included in patch releases. BTW master is stable so you don't need to wait for release, just pin to a specific commit.

mattgodbolt:
Thanks @vitaut - understood completely :) I'm using `conan` and the public conan repo in the project that hit this issue: it's not deadly enough to try and get a new packaging system in; I can wait for 7.1.x :) Thanks for all your hard work!

tonyelewis:
> I'm using conan and the public conan repo in the project that hit this issue

Yep &ndash; FWIW, I'm in this same situation as @mattgodbolt. I wish conan provided some way of accessing a dependency's `master`. But I too can wait for 7.1.x.

>Thanks for all your hard work!

:+1:

PR #1790, opened 2020-07-22
Disabled __attribute__((deprecated)) usage for LCC

TheQwertiest:
This fixes compilation error when compiling with `lcc`.

From Google:
> LCC is a compiler for Elbrus (E2K) CPU. (...) The LCC tries to mimic GCC by user interface (both supported options and GCC-specific features)

 but it fails in this case: it doesn't support `__attribute__((deprecated))` for type declarations.


vitaut:
Thank you!

Issue #1798, opened 2020-07-30
Linker error when compiling with clang-cl: undefined symbol: __udivti3


Kingcom:
This is very likely to be the fault of Clang, but perhaps something can be done to work around it. 

I'm using the official Windows build of LLVM 10.0.0 from a Visual Studio 2017 x64 environment. Calling any formatting function (even something trivial like ```fmt::print("");```) results in the linker error above. This seems to be because Clang mishandles 128 bit integers and incorrectly generates a call to a function not implemented by the Microsoft runtime.

I was able to work around the issue with the following patch:
```
diff --git a/include/fmt/format.h b/include/fmt/format.h
index 336d2c67..5e5b87dd 100644
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -759,8 +759,7 @@ using uint32_or_64_or_128_t =
 // represent all values of T.
 template <typename T>
 using uint32_or_64_or_128_t = conditional_t<
-    num_bits<T>() <= 32, uint32_t,
-    conditional_t<num_bits<T>() <= 64, uint64_t, uint128_t>>;
+    num_bits<T>() <= 32, uint32_t, uint64_t>;
 #endif
```

I've tried defining FMT_REDUCE_INT_INSTANTIATIONS to 1, but with that it fails to compile due to an ambiguous call:
```
..\ext\fmt\include\fmt/format-inl.h(639,5): error: call to member function 'multiply' is ambiguous
    multiply(uint32_or_64_or_128_t<Int>(value));
    ^~~~~~~~
..\ext\fmt\include\fmt/format-inl.h(566,8): note: candidate function
  void multiply(uint32_t value) {
       ^
..\ext\fmt\include\fmt/format-inl.h(577,8): note: candidate function
  void multiply(uint64_t value) {
```

PR #1800, opened 2020-07-30
Don't use 128 bit integers with clang-cl

Kingcom:
clang-cl currently has a long-standing bug that using 128 bit integers generates references to symbols that are provided neither by its own nor by the Microsoft runtime: https://bugs.llvm.org/show_bug.cgi?id=25305

Visual Studio (2017, I was unable to test 2019) also does not satisfy the conditions to use 128 bit integers in {fmt}, so this effectively causes Clang to match its behavior when targetting the Microsoft ABI. The condition could therefore be simplified to just checking FMT_MSC_VER, but also checking for Clang expresses the intent better as it does advertise its support for 128 bit integers while Visual Studio does not (at the very least not in the way tested here).

This fixes the linker error described in #1798.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -298,7 +298,7 @@ template <typename T> struct std_string_view {};
 
 #ifdef FMT_USE_INT128
 // Do nothing.
-#elif defined(__SIZEOF_INT128__) && !FMT_NVCC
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks

PR #1802, opened 2020-08-05
include/fmt/format.h: explicit cast to std::size_t for parameter to buffer.resize() in order to get rid of warning 'implicit conversion changes signedness:' in clang-8

medithe:

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -1593,7 +1593,7 @@ template <typename OutputIt, typename Char, typename UInt> struct int_writer {
     format_decimal(digits, abs_value, num_digits);
     basic_memory_buffer<Char> buffer;
     size += prefix_size;
-    buffer.resize(size);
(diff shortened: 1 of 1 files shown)


include/fmt/format.h
```diff
@@ -1593,7 +1593,7 @@ template <typename OutputIt, typename Char, typename UInt> struct int_writer {
     format_decimal(digits, abs_value, num_digits);
     basic_memory_buffer<Char> buffer;
     size += prefix_size;
-    buffer.resize(size);
+    buffer.resize(static_cast<std::size_t>(size));
```
vitaut, line 1596: I suggest using to_unsigned instead of a cast.

medithe: Yupp, I agree.
Is done in a2c668fcaafd92afddb3837bdd36573006c4a69c.

vitaut:
Thanks!

PR #1806, opened 2020-08-07
Adding convenience append(range)

BRevzin:
Just a convenience overload of append() that can be invoked with contiguous ranges (notably `string_view`s).


vitaut:
Thanks for the PR. Looks good but I suggest moving `append` to `basic_memory_buffer` because it's the only one available to users.

vitaut:
Merged in https://github.com/fmtlib/fmt/commit/d0dd6786938500b42c7f05118d1aae76c0951cd6.

PR #1808, opened 2020-08-08
small changes to reduce clang-9 warnings

medithe:


include/fmt/format.h
```diff
@@ -1613,9 +1613,9 @@ template <typename OutputIt, typename Char, typename UInt> struct int_writer {
                               make_checked(p, s.size()));
     }
     if (prefix_size != 0) p[-1] = static_cast<Char>('-');
-    using iterator = remove_reference_t<decltype(reserve(out, 0))>;
+    using out_iterator = remove_reference_t<decltype(reserve(out, 0))>;
```
vitaut, line 1616: This can be removed completely.

include/fmt/format.h
```diff
@@ -1593,7 +1593,7 @@ template <typename OutputIt, typename Char, typename UInt> struct int_writer {
     format_decimal(digits, abs_value, num_digits);
     basic_memory_buffer<Char> buffer;
     size += prefix_size;
-    buffer.resize(size);
+    buffer.resize(to_unsigned(size));
```
vitaut, line 1596: I suggest storing `to_unsigned(size)` in a variable (`usize`?) to avoid copy-pasta.

include/fmt/format.h
medithe: Is addressed with 412a658.

include/fmt/format.h
medithe: Is adressed with e6159ed.

PR #1818, opened 2020-08-13
nvcc compiler should be EDG-based, but fails test

gsjaardema:
Our nvcc compilers (10.1.243 and 9.2.X) do not define the correct value for `FMT_USE_UDL_TEMPLATE` and then end up with an error later on in the build.  Explicitly search for `__NVCC__` symbol not being defined.   Might want to instead use `FMT_NVCC` or some other check, but the raw `__EDG_VERSION__` check is not working correctly for nvcc.


vitaut:
Thank you!

PR #1821, opened 2020-08-14
CMakeLists.txt: Added -Wshadow warning to clang and gcc.

medithe:


vitaut:
Thanks for the PR but shadowing warnings are mostly noise, I don't think we need to have them enabled in CI. That said a PR to fix them in headers (but not in tests) is welcome.

PR #1822, opened 2020-08-14
Fixing buffer_appender's ++ slicing.

BRevzin:
`buffer_appender`'s pre- and post-fix increment operators slice down to `back_inserter`, which means it actually breaks the iterator requirements (which would be actually checked in C++20, it is not `weakly_incrementable`) and also breaks any code that assumes that `++it` and `it` have the same type (as illustrated in the test case, a reduced form of how we format our optional type, which no longer compiles).

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -845,11 +845,23 @@ template <typename T = char> class counting_buffer : public buffer<T> {
 // It is used to reduce symbol sizes for the common case.
 template <typename T>
 class buffer_appender : public std::back_insert_iterator<buffer<T>> {
+  using base = std::back_insert_iterator<buffer<T>>;
  public:
   explicit buffer_appender(buffer<T>& buf)
-      : std::back_insert_iterator<buffer<T>>(buf) {}
-  buffer_appender(std::back_insert_iterator<buffer<T>> it)
-      : std::back_insert_iterator<buffer<T>>(it) {}
+      : base(buf) {}
+  buffer_appender(base it)
+      : base(it) {}
+
+  buffer_appender& operator++() {
+    base::operator++();
+    return *this;
+  }
+
+  buffer_appender operator++(int) {
+    buffer_appender tmp = *this;
+    ++*this;
+    return tmp;
+  }
 };
 
 // Maps an output iterator into a buffer.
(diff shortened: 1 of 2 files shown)


include/fmt/core.h
```diff
@@ -845,11 +845,23 @@ template <typename T = char> class counting_buffer : public buffer<T> {
...
+  buffer_appender& operator++() {
+    base::operator++();
+    return *this;
+  }
+
+  buffer_appender operator++(int) {
+    buffer_appender tmp = *this;
+    ++*this;
+    return tmp;
+  }
```
BRevzin, line 864: Given that `back_insert_iterator` only has like... 4 member functions, and I've already implemented 2 of them (and over-implemented them too, these could both just be `return *this;`), maybe we should just implement the whole thing in here without `std`?

test/format-test.cc
```diff
@@ -2473,3 +2473,37 @@ TEST(FormatTest, FormatUTF8Precision) {
   EXPECT_EQ(result.size(), 5);
   EXPECT_EQ(from_u8str(result), from_u8str(str.substr(0, 5)));
 }
+
+#ifdef __cpp_decltype_auto
```
vitaut, line 2477: Can we make the test work without this, e.g. checking if the types are the same manually?

test/format-test.cc
```diff
@@ -2473,3 +2473,37 @@ TEST(FormatTest, FormatUTF8Precision) {
...
+struct lazy_optional {
+  bool engaged;
+  std::string value;
+};
```
vitaut, line 2481: I'd minimize this even further into an empty struct and make the name reflect what we are trying to test here.

include/fmt/core.h
vitaut: I think it's fine as is.

test/format-test.cc
BRevzin: Done. Just have a `static_assert`.

test/format-test.cc
```diff
@@ -2473,3 +2473,28 @@ TEST(FormatTest, FormatUTF8Precision) {
...
+  template <typename ParseContext>
+  auto parse(ParseContext& ctx) -> decltype(ctx.begin()) {
+    return ctx.begin();
+  }
+
+  template <typename Context>
+  auto format(check_back_appender, Context& ctx) -> decltype(ctx.out()) {
+    // needs to satisfy weakly_incrementable
+    auto out = ctx.out();
+    static_assert(std::is_same<decltype(++out), decltype(out)&>::value);
```
vitaut, line 2490: Please add a message to fix contbuild.

BRevzin: Oops, done.

vitaut:
Thanks!

PR #1823, opened 2020-08-14
CMakeLists.txt: Added Wundef warning to clang and gcc.

medithe:


vitaut:
Thanks

PR #1827, opened 2020-08-18
fix warnings (-Wconversion, -Wsign-conversion, -Wshadow)

ITotalJustice:
fixes:
-Wconversion, -Wsign-conversion, -Wshadow, -Wnon-virtual-dtor

Hi, thanks for the great library!

I noticed a few compiler warnings when using the above flags. I've fixed the warnings that I was getting, however I wasn't really sure what to name the [oflag](https://github.com/fmtlib/fmt/blob/f9f02df719329ef7449c2b9e356f99d2f960cf3e/include/fmt/os.h#L364) so i just prefixed it with an underscore. Let me know if you'd prefer something else, or to remove it

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -669,7 +669,7 @@ template <typename T> class buffer {
         size_(sz),
         capacity_(cap) {}
 
-  ~buffer() = default;
+  virtual ~buffer() = default;
 
   /** Sets the buffer data and capacity. */
   void set(T* buf_data, size_t buf_capacity) FMT_NOEXCEPT {
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -669,7 +669,7 @@ template <typename T> class buffer {
         size_(sz),
         capacity_(cap) {}
 
-  virtual ~buffer() = default;
+  ~buffer() = default;
 
   /** Sets the buffer data and capacity. */
   void set(T* buf_data, size_t buf_capacity) FMT_NOEXCEPT {
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -800,10 +800,10 @@ template <typename T = void> struct FMT_EXTERN_TEMPLATE_API basic_data {
 // This is a function instead of an array to workaround a bug in GCC10 (#1810).
 FMT_INLINE uint16_t bsr2log10(int bsr) {
   constexpr uint16_t data[] = {
-    1,  1,  1,  2,  2,  2,  3,  3,  3,  4,  4,  4,  4,  5,  5,  5,
-    6,  6,  6,  7,  7,  7,  7,  8,  8,  8,  9,  9,  9,  10, 10, 10,
-    10, 11, 11, 11, 12, 12, 12, 13, 13, 13, 13, 14, 14, 14, 15, 15,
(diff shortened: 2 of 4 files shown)


include/fmt/core.h
```diff
@@ -669,7 +669,7 @@ template <typename T> class buffer {
         size_(sz),
         capacity_(cap) {}
 
-  ~buffer() = default;
+  virtual ~buffer() = default;
```
vitaut, line 672: This is intentional.

include/fmt/format.h
```diff
@@ -3482,7 +3482,7 @@ inline std::string to_string(T value) {
   // The buffer should be large enough to store the number including the sign or
   // "false" for bool.
   constexpr int max_size = detail::digits10<T>() + 2;
-  char buffer[max_size > 5 ? max_size : 5];
+  char buffer[static_cast<size_t>(max_size > 5 ? max_size : 5)];
```
vitaut, line 3485: static_cast -> to_unsigned

src/os.cc
```diff
@@ -92,11 +92,11 @@ int detail::utf16_to_utf8::convert(wstring_view s) {
 
   int length = WideCharToMultiByte(CP_UTF8, 0, s.data(), s_size, nullptr, 0,
                                    nullptr, nullptr);
-  if (length == 0) return GetLastError();
-  buffer_.resize(length + 1);
+  if (length == 0) return static_cast<int>(GetLastError());
```
vitaut, line 95: I suggest adding an inline function:

```
inline int get_windows_error() {
  static_cast<int>(GetLastError());
}
```
and using it to avoid casts in multiple places.

src/os.cc
```diff
@@ -212,10 +212,10 @@ long long file::size() const {
   if (size_lower == INVALID_FILE_SIZE) {
     DWORD error = GetLastError();
     if (error != NO_ERROR)
-      FMT_THROW(windows_error(GetLastError(), "cannot get file size"));
+      FMT_THROW(windows_error(static_cast<int>(GetLastError()), "cannot get file size"));
   }
   unsigned long long long_size = size_upper;
-  return (long_size << sizeof(DWORD) * CHAR_BIT) | size_lower;
+  return static_cast<long long int>((long_size << sizeof(DWORD) * CHAR_BIT) | size_lower);
```
vitaut, line 218: `long long int` -> `long long` and please apply clang-format.

src/os.cc
```diff
@@ -306,7 +306,7 @@ long getpagesize() {
 #  ifdef _WIN32
   SYSTEM_INFO si;
   GetSystemInfo(&si);
-  return si.dwPageSize;
+  return static_cast<long int>(si.dwPageSize);
```
vitaut, line 309: nit: `long int` -> `long` for consistency with other code.

ITotalJustice:
Thanks for the comments. I've ran clang-format, but it changed a few other things in the format.h file that I hadn't edited for this pr. Haven't used clang format before so not sure if I did it correctly.

vitaut:
> it changed a few other things in the format.h file that I hadn't edited for this pr.

I applied clang-format in master, please rebase to avoid unrelated changes.

vitaut:
Also there are some compile errors.

vitaut:
Unfortunately this diff cannot be applied due to CI failures. Closing for now but feel free to resubmit with a fix.

PR #1828, opened 2020-08-18
CMakeLists.txt: Added fno-common flag to clang and gcc.

medithe:


vitaut:
Please provide more details.

medithe:
> Please provide more details.

According to https://gcc.gnu.org/gcc-10/porting_to.html, the flag `-fno-common` is the default behavior in gcc10 and above.
Therefore, it may be a good idea to check for this flag in order to be complient with this flag in the future.

Issue #1815, opened 2020-08-13
fmt::to_string doesn't work with fallback_formatter


alex-700:
[godbolt](https://godbolt.org/z/dj6s4a)

It [worked](https://godbolt.org/z/babsd9) for `fmt=6.2.1`.

PR #1829, opened 2020-08-18
Fix detail::write with fallback formatter

alexezeder:
Recently I was trying to use both `FMT_COMPILE()` and struct with ostream operator, but with no luck. After first look on issues and pull requests I found one issue that describes the same problem, here it is - https://github.com/fmtlib/fmt/issues/1815. 

I'm not sure that this solution is clean enough, but I am open to change whatever you want.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -1841,13 +1841,26 @@ OutputIt write(OutputIt out, const void* value) {
   return write_ptr<Char>(out, to_uintptr(value), nullptr);
 }
 
-template <typename Char, typename OutputIt, typename T>
+template <typename Char, typename OutputIt, typename T,
+          FMT_ENABLE_IF(
+              has_formatter<T, basic_format_context<OutputIt, Char>>::value)>
 auto write(OutputIt out, const T& value) -> typename std::enable_if<
     mapped_type_constant<T, basic_format_context<OutputIt, Char>>::value ==
         type::custom_type,
     OutputIt>::type {
   basic_format_context<OutputIt, Char> ctx(out, {}, {});
   return formatter<T>().format(value, ctx);
+}
+
+template <typename Char, typename OutputIt, typename T,
+          FMT_ENABLE_IF(has_fallback_formatter<
+                        T, basic_format_context<OutputIt, Char>>::value)>
+auto write(OutputIt out, const T& value) -> typename std::enable_if<
+    mapped_type_constant<T, basic_format_context<OutputIt, Char>>::value ==
+        type::custom_type,
+    OutputIt>::type {
+  basic_format_context<OutputIt, Char> ctx(out, {}, {});
+  return fallback_formatter<T>().format(value, ctx);
 }
 
 // An argument visitor that formats the argument and writes it via the output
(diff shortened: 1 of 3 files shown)


include/fmt/format.h
```diff
@@ -1850,6 +1852,17 @@ auto write(OutputIt out, const T& value) -> typename std::enable_if<
...
+template <typename Char, typename OutputIt, typename T,
+          FMT_ENABLE_IF(has_fallback_formatter<
+                        T, basic_format_context<OutputIt, Char>>::value)>
+auto write(OutputIt out, const T& value) -> typename std::enable_if<
+    mapped_type_constant<T, basic_format_context<OutputIt, Char>>::value ==
+        type::custom_type,
+    OutputIt>::type {
+  basic_format_context<OutputIt, Char> ctx(out, {}, {});
+  return fallback_formatter<T>().format(value, ctx);
+}
```
vitaut, line 1864: This overload can be merged into the one above replacing `formatter<T>` with

```c++
conditional_t<has_formatter<T, Context>::value,
                         typename Context::template formatter_type<T>,
                         fallback_formatter<T, char_type>>
```

test/compile-test.cc
```diff
@@ -23,6 +23,7 @@
 #endif
 
 #include "fmt/compile.h"
+#include "fmt/ostream.h"
```
vitaut, line 26: compile-test shouldn't depend on ostream. I think the test in ostream-test is enough.

include/fmt/format.h
alexezeder: done

test/compile-test.cc
alexezeder: commit with this changes reverted

include/fmt/format.h
```diff
@@ -1841,13 +1841,18 @@ OutputIt write(OutputIt out, const void* value) {
...
 
-template <typename Char, typename OutputIt, typename T>
-auto write(OutputIt out, const T& value) -> typename std::enable_if<
-    mapped_type_constant<T, basic_format_context<OutputIt, Char>>::value ==
-        type::custom_type,
-    OutputIt>::type {
-  basic_format_context<OutputIt, Char> ctx(out, {}, {});
-  return formatter<T>().format(value, ctx);
+template <typename Char, typename OutputIt, typename T,
+          typename Context = basic_format_context<OutputIt, Char>>
```
vitaut, line 1845: I don't think context should be a template parameter.

alexezeder: I did that to have one definition for `Context`, so it can be used in trailing return type too. But if it's a problem then I have 2 solution:

- by using `basic_format_context<OutputIt, Char>` at least in return type (or maybe in entire function)
- probably return type can be removed, but `enable_if` in it should be moved to template parameter and I'm not sure how old compilers will handle that

vitaut: I think it's better to have `basic_format_context<OutputIt, Char>` in two places as before.

alexezeder: Ok, will do, but with conditional for `formatter_type` there will be more than 2 places

vitaut: I suggest adding an alias in a function. I just don't think it should be a template parameter.

alexezeder: done

vitaut:
Thank you!

PR #1836, opened 2020-08-23
Fix usage of override for clang

AlexanderLanin:
It seems fmt is checking cxx_override which seems to be a [CMake related term](https://cmake.org/cmake/help/latest/manual/cmake-compile-features.7.html). At least for clang it's  [cxx_override_control](https://clang.llvm.org/docs/LanguageExtensions.html#c-11-override-control).
In clangs current trunk fmt is reported as having no override keyword. This change silences the warning in clangs trunk as expected.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -99,7 +99,7 @@
 #endif
 
 #ifndef FMT_OVERRIDE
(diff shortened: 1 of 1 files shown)


vitaut:
Good catch, thanks!

Issue #1842, opened 2020-08-30
fmt/color.h: fmt::format_to not implemented for fmt::text_style


Naios:
e66ba16923be2a1bcb6b841df7a64a0b3ac4be4d

Currently it is not possible to format a colorized string to a `fmt::memory_buffer` using `fmt::format_to(memory_buffer, fmt::text_style(...` or through any other function because the overload is missing.

It would be great if the missing overload for `fmt::format_to(memory_buffer, fmt::text_style` could be added.

Overall the color API is inconsistent with the format API and not all format functions support a `fmt::text_style` overload. Maybe the color API could be aligned in general with the format API.

Andersama:
I also just ran into this issue. My work around though not completely templated looks like this:
```c
template <size_t N> struct ansi_esc_sequence {
	uint8_t tail;
	uint8_t opts[N];
};

template <size_t N> struct fmt::formatter<ansi_esc_sequence<N>> : public fmt::formatter<string_view> {
	template <typename FormatContext>
	auto format(const ansi_esc_sequence<N>& p, FormatContext& ctx) noexcept {
		if constexpr (N > 0) {
			return fmt::format_to(ctx.out(), "\033[{}{}"sv, fmt::join(&p.opts[0], &p.opts[0] + N, ";"),
				(char)p.tail);
		}
		else {
			return fmt::format_to(ctx.out(), "\033[{}"sv, (char)p.tail);
		}
	}
};
```
It's a bit of a pain to specify the colors w/ the ansi sequence this way, definitely in full agreement it was strange to try to avoid this by using color.h only to find format_to wasn't implemented.

vitaut:
The color API is for terminal output thus it supports only a subset of functions.

Issue #1593, opened 2020-03-14
fmt/color.h missing format_to overload


Sarcasm:
Hello,

I was wondering if there was any reason for fmt/color.h to not provide a format_to() overload?
I'd like to print messages to stderr using a style similar to GCC/Clang:

1. subject in bold
2. diagnostics kind in bold and colored: e.g. red for errors
3. message in normal text

However, due to multiprocessing issues, I'd like to write the messages "atomically" to stderr so output does not get mixed up.

Right now I can do it that way:

```cpp
std::string subject = fmt::format(fmt::emphasis::bold, "{}:", p);
std::string error = fmt::format(fg(fmt::color::red) | fmt::emphasis::bold, "error:");
fmt::print(stderr, "{} {} {}\n", subject, error, strerror(errno));
```

But it bothers me to allocate so many `std::string`, a color-supporting `format_to()` would be an improvement.

vitaut:
It was proposed before (#1161) but noone was interested enough in implementing it probably because there is little value in optimizing allocations when you are dealing with I/O.

vitaut:
PRs are welcome =).

Sarcasm:
I see, I missed that ticket when searching.
I saw that there was a `vformat_to`, so I thought it will be peanuts to add, but I realize now that the signature takes a memory_buffer, not an output iterator.

It would be nice for completeness to have it, but I agree that colors are mostly used when printing to a tty, in which case I doubt the extra strings matter much, at least they do not for my use case.

Closing, I will let a future brave soul do it.

Thank you

PR #1843, opened 2020-08-31
Add color format_to overloads

Naios:
* Fix variable size basic_memory_buffer colorization
* Rewrite the color unit tests to test the same colors against multiple backends
* Ref #1842
* Ref #1593

---

- Compile time formatted strings are not supported (yet)
- The underlying buffer type the colorization is working on was changed to `detail::buffer` to also accept `iterator_buffer`, and non default internal size memory buffers. Size limited buffers are not supported (and also not accepted by the public API). This is why `format_to_n` is not added as part of this change. Additionally since a colorized string is not an atomic unit per character anymore (because it can't be split in the middle of the ansi color code effectively) it doesnt make sence to add support for `format_to_n` anyway.

---

### DIFF
--- a/include/fmt/color.h
+++ b/include/fmt/color.h
@@ -463,16 +463,16 @@ template <> inline void reset_color<wchar_t>(FILE* stream) FMT_NOEXCEPT {
 }
 
 template <typename Char>
-inline void reset_color(basic_memory_buffer<Char>& buffer) FMT_NOEXCEPT {
+inline void reset_color(buffer<Char>& buffer) FMT_NOEXCEPT {
   const char* begin = data::reset_color;
   const char* end = begin + sizeof(data::reset_color) - 1;
   buffer.append(begin, end);
 }
 
 template <typename Char>
-void vformat_to(basic_memory_buffer<Char>& buf, const text_style& ts,
+void vformat_to(buffer<Char>& buf, const text_style& ts,
                 basic_string_view<Char> format_str,
-                basic_format_args<buffer_context<Char>> args) {
+                basic_format_args<buffer_context<type_identity_t<Char>>> args) {
   bool has_style = false;
   if (ts.has_emphasis()) {
     has_style = true;
@@ -561,6 +561,49 @@ inline std::basic_string<Char> format(const text_style& ts, const S& format_str,
                                       const Args&... args) {
   return vformat(ts, to_string_view(format_str),
                  fmt::make_args_checked<Args...>(format_str, args...));
+}
+
+/** Formats a string with the given text_style and writes the output to ``out``.
+ */
+template <typename OutputIt, typename S, typename Char = char_t<S>,
+          FMT_ENABLE_IF(detail::is_output_iterator<OutputIt>::value)>
+OutputIt vformat_to(
+    OutputIt out, const text_style& ts, const S& format_str,
+    basic_format_args<buffer_context<type_identity_t<Char>>> args) {
+  decltype(detail::get_buffer<Char>(out)) buf(detail::get_buffer_init(out));
+  detail::vformat_to(buf, ts, to_string_view(format_str), args);
+  return detail::get_iterator(buf);
+}
+
+/**
+ \rst
+ Formats arguments with the given text_style, writes the result to the output
+ iterator ``out`` and returns the iterator past the end of the output range.
+
+ **Example**::
+
+   std::vector<char> out;
+   fmt::format_to(std::back_inserter(out),
+                  fmt::emphasis::bold | fg(fmt::color::red), "{}", 42);
+   \endrst
+ */
+template <typename OutputIt, typename S, typename... Args,
+          FMT_ENABLE_IF(detail::is_output_iterator<OutputIt>::value&&
+                            detail::is_string<S>::value)>
+inline OutputIt format_to(OutputIt out, const text_style& ts,
+                          const S& format_str, Args&&... args) {
+  return vformat_to(out, ts, to_string_view(format_str),
+                    fmt::make_args_checked<Args...>(format_str, args...));
+}
+
+template <typename S, typename... Args, size_t SIZE = inline_buffer_size,
+          typename Char = enable_if_t<detail::is_string<S>::value, char_t<S>>>
+inline typename buffer_context<Char>::iterator format_to(
+    basic_memory_buffer<Char, SIZE>& buf, const text_style& ts,
+    const S& format_str, Args&&... args) {
+  detail::vformat_to(buf, ts, to_string_view(format_str),
+                     fmt::make_args_checked<Args...>(format_str, args...));
+  return detail::buffer_appender<Char>(buf);
 }
 
 FMT_END_NAMESPACE
--- a/test/color-test.cc
+++ b/test/color-test.cc
@@ -7,6 +7,10 @@
 
 #include "fmt/color.h"
 
+#include <iterator>
+#include <string>
+#include <utility>
+
 #include "gtest-extra.h"
 
 TEST(ColorsTest, ColorsPrint) {
@@ -83,4 +87,22 @@ TEST(ColorsTest, Format) {
             "\x1b[105mtbmagenta\x1b[0m");
   EXPECT_EQ(fmt::format(fg(fmt::terminal_color::red), "{}", "foo"),
             "\x1b[31mfoo\x1b[0m");
+}
+
+TEST(ColorsTest, FormatToOutAcceptsTextStyle) {
+  fmt::text_style const ts = fg(fmt::rgb(255, 20, 30));
+  std::string out;
+  fmt::format_to(std::back_inserter(out), ts, "rgb(255,20,30){}{}{}", 1, 2, 3);
+
+  EXPECT_EQ(fmt::to_string(out),
+            "\x1b[38;2;255;020;030mrgb(255,20,30)123\x1b[0m");
+}
+
+TEST(ColorsTest, FormatToAcceptsTextStyle) {
+  fmt::text_style const ts = fg(fmt::rgb(255, 20, 30));
+  fmt::basic_memory_buffer<char> out;
+  fmt::format_to(out, ts, "rgb(255,20,30){}{}{}", 1, 2, 3);
+
+  EXPECT_EQ(fmt::to_string(out),
+            "\x1b[38;2;255;020;030mrgb(255,20,30)123\x1b[0m");
 }
--- a/test/gtest-extra.h
+++ b/test/gtest-extra.h
@@ -145,7 +145,13 @@ std::string read(fmt::file& f, size_t count);
               read(file, fmt::string_view(expected_content).size()))
 
 #else
-#  define EXPECT_WRITE(file, statement, expected_output) SUCCEED()
+#  define EXPECT_WRITE(file, statement, expected_output) \
+    do {                                                 \
+      (void)(file);                                      \
+      (void)(statement);                                 \
+      (void)(expected_output);                           \
+      SUCCEED();                                         \
+    } while (false)
 #endif  // FMT_USE_FCNTL
 
 template <typename Mock> struct ScopedMock : testing::StrictMock<Mock> {


include/fmt/color.h
```diff
@@ -563,6 +563,52 @@ inline std::basic_string<Char> format(const text_style& ts, const S& format_str,
                  fmt::make_args_checked<Args...>(format_str, args...));
 }
 
+/** Formats a string with the given text_style and writes the output to ``out``.
+ */
+// GCC 8 and earlier cannot handle std::back_insert_iterator<Container> with
+// vformat_to<ArgFormatter>(...) overload, so SFINAE on iterator type instead.
```
vitaut, line 569: Is this a copy-paste? I don't think it is correct.

include/fmt/color.h
```diff
@@ -563,6 +563,52 @@ inline std::basic_string<Char> format(const text_style& ts, const S& format_str,
...
+/**
+ \rst
+ Formats arguments with the given text_style, writes the result to the output
+ iterator ``out`` and returns the iterator past the end of the output range.
+
+ **Example**::
+
+   std::vector<char> out;
+   fmt::format_to(std::back_inserter(out),
+                  fmt::emphasis::bold | fg(fmt::color::red), "{}", 42); \endrst
```
vitaut, line 589: Please put `\endrst` on a separate line.

include/fmt/color.h
```diff
@@ -563,6 +563,52 @@ inline std::basic_string<Char> format(const text_style& ts, const S& format_str,
...
 }
 
+/** Formats a string with the given text_style and writes the output to ``out``.
+ */
+// GCC 8 and earlier cannot handle std::back_insert_iterator<Container> with
+// vformat_to<ArgFormatter>(...) overload, so SFINAE on iterator type instead.
+template <typename OutputIt, typename S, typename Char = char_t<S>,
+          FMT_ENABLE_IF(detail::is_output_iterator<OutputIt>::value)>
+OutputIt vformat_to(
+    OutputIt out, const text_style& ts, const S& format_str,
```
vitaut, line 573: S should be replaced with basic_string_view

include/fmt/color.h
```diff
@@ -563,6 +563,52 @@ inline std::basic_string<Char> format(const text_style& ts, const S& format_str,
...
+template <typename S, typename... Args, size_t SIZE = inline_buffer_size,
+          typename Char = enable_if_t<detail::is_string<S>::value, char_t<S>>>
+inline typename buffer_context<Char>::iterator format_to(
+    basic_memory_buffer<Char, SIZE>& buf, const text_style& ts,
+    const S& format_str, Args&&... args) {
+  basic_format_args<buffer_context<type_identity_t<Char>>> store =
+      fmt::make_args_checked<Args...>(format_str, args...);
+  detail::vformat_to(buf, ts, to_string_view(format_str), store);
+  return detail::buffer_appender<Char>(buf);
+}
```
vitaut, line 610: I don't think this is necessary.

test/color-test.cc
```diff
@@ -7,80 +7,126 @@
...
+template <> struct ColorsTest<ColorFunctionPrintTag> : public testing::Test {
+  virtual ~ColorsTest();
+
+  template <typename... Args>
+  void expect_color(const std::string& expected, const fmt::text_style& ts,
+                    Args&&... args) {
+    EXPECT_WRITE(stdout, fmt::print(stdout, ts, std::forward<Args>(args)...),
+                 expected);
+  }
+};
```
vitaut, line 32: This is an overkill - please leave the old tests as is and add two test cases that check the new APIs.

include/fmt/color.h
Naios: S can't be replaced because `basic_string_view` is missing an implicit constructor for a fixed size array like `const char(&)[5]`

vitaut: It doesn't matter. `format_to` should convert `S` into `basic_string_view` when calling this function.

Naios: It seems like `to_string_view` can convert more types than the `basic_string_view` constructor.
In that case the `basic_string_view` constructor needs to be aligned with `to_string_view` or we need to rely on `to_string_view` to convert it for us. Replacing `const S&` with `basic_string_view` will lead to build failures otherwise.

include/fmt/color.h
Naios: Thank you for your feedback. I corrected my PR accordingly.
I really would like to keep this overload though, mainly it is the main reason why I submitted this PR.

This overload integrates well into a `struct fmt::formatter<` specializations for custom types,
it is also possible to colorize only parts of a string and for multiple colorized parts a memory allocation adds up fast.

Otherwise we would have to use the output iterator overload for writing into a `memory_buffer` although the private color API accepts a `memory_buffer` already but is inaccessible. This would add unneccessary code overhead (and probably adds a performance overhead too).

include/fmt/color.h
vitaut: We should use `to_string_view`.

Naios:
@vitaut I'm encountering some strange unit test failures on travis with some GCC compilers (GCC-6+) that fail to format `fmt::format_to(std::back_inserter(out), fg(fmt::terminal_color::red), "{}", "foo")`. Do you have an idea about this?

```
/home/travis/build/fmtlib/fmt/test/color-test.cc:140: Failure
Value of: "\x1b[31mfoo\x1b[0m"
  Actual: "\x1B[31mfoo\x1B[0m"
Expected: format_to_out_helper(fg(fmt::terminal_color::red), "{}", "foo")
Which is: "\x1B[31m{}\x1B[0m"
```

https://travis-ci.org/github/fmtlib/fmt/jobs/723128246#L747-L761

I'm using a helper funtion as following:

```cpp
template <typename S, typename... Args,
          typename String = std::basic_string<fmt::char_t<S>>>
String format_to_out_helper(const fmt::text_style& ts, const S& format_str,
                            Args&&... args) {
  String out;
  fmt::format_to(std::back_inserter(out), ts, format_str,
                 std::forward<Args>(args)...);
  return out;
}

// ....

EXPECT_EQ(format_to_out_helper(fg(fmt::terminal_color::red), "{}", "foo"),
          "\x1b[31mfoo\x1b[0m");
```

The other functions for colorization seem fine.

Naios:
I reduced the test cases as requested and now it only does basic tests whether the API accepts `fmt::text_style` through the `format_to` function.
Also I could fix the format failures across some compiler with additional format args.

If I understand you right about `basic_string_view` you mean that I shall replace 

```cpp
template <typename S, typename... Args, size_t SIZE = inline_buffer_size,
          typename Char = enable_if_t<detail::is_string<S>::value, char_t<S>>>
inline typename buffer_context<Char>::iterator format_to(
    basic_memory_buffer<Char, SIZE>& buf, const text_style& ts,
    const S& format_str, Args&&... args) {
  basic_format_args<buffer_context<type_identity_t<Char>>> store =
      fmt::make_args_checked<Args...>(format_str, args...);
  detail::vformat_to(buf, ts, to_string_view(format_str), store);
  return detail::buffer_appender<Char>(buf);
}
```
by
```cpp
template <typename Char, typename... Args, size_t SIZE = inline_buffer_size>
inline typename buffer_context<Char>::iterator format_to(
    basic_memory_buffer<Char, SIZE>& buf, const text_style& ts,
    basic_string_view<Char> format_str, Args&&... args) {
  basic_format_args<buffer_context<type_identity_t<Char>>> store =
      fmt::make_args_checked<Args...>(format_str, args...);
  detail::vformat_to(buf, ts, format_str, store);
  return detail::buffer_appender<Char>(buf);
}
```

which is, from my point of view, not possible because as mentioned above the `basic_string_view` constructor is not capable to accept all string like types (`char const (&)[21]`) as `to_string_view` is. Thus overload resolution will fail for this case and prefers the non `text_style` overload instead.
Originally I just copied both `format_to` overloads from `format.h` and adapted it to accept `text_style`. I would prefer if this PR could be merged and then afterwards the `format_str` type can be changed in all `format_to` overloads.

I attached you my compiler output which could further explain my arguments:
```
In file included from /home/src/test/color-test.cc:8:
In file included from /home/src/include/fmt/color.h:11:
/home/src/include/fmt/format.h:3562:23: error: no matching function for call to 'make_args_checked'
  const auto& vargs = fmt::make_args_checked<Args...>(format_str, args...);
                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/src/test/color-test.cc:107:8: note: in instantiation of function template specialization 'fmt::v7::format_to<fmt::v7::text_style, char const (&)[21], int, int, int, 500, char>' requested here
  fmt::format_to(out, ts, "rgb(255,20,30){}{}{}", 1, 2, 3);
       ^
/home/src/include/fmt/core.h:1631:13: note: candidate template ignored: substitution failure [with Args = <char const (&)[21], int, int, int>, S = fmt::v7::text_style]: no type named 'type' in 'fmt::v7::detail::char_t_impl<fmt::v7::text_style, void>'
inline auto make_args_checked(const S& format_str,
            ^
In file included from /home/src/test/color-test.cc:8:
In file included from /home/src/include/fmt/color.h:11:
/home/src/include/fmt/format.h:3563:34: error: no matching function for call to 'to_string_view'
  return detail::vformat_to(buf, to_string_view(format_str), vargs);
                                 ^~~~~~~~~~~~~~
/home/src/include/fmt/core.h:459:32: note: candidate template ignored: could not match 'const Char *' against 'fmt::v7::text_style'
inline basic_string_view<Char> to_string_view(const Char* s) {
                               ^
/home/src/include/fmt/core.h:464:32: note: candidate template ignored: could not match 'basic_string<type-parameter-0-0, type-parameter-0-1, type-parameter-0-2>' against 'const fmt::v7::text_style'
inline basic_string_view<Char> to_string_view(
                               ^
/home/src/include/fmt/core.h:470:32: note: candidate template ignored: could not match 'basic_string_view<type-parameter-0-0>' against 'fmt::v7::text_style'
inline basic_string_view<Char> to_string_view(basic_string_view<Char> s) {
                               ^
/home/src/include/fmt/core.h:476:32: note: candidate template ignored: could not match 'std_string_view<type-parameter-0-0>' against 'fmt::v7::text_style'
inline basic_string_view<Char> to_string_view(detail::std_string_view<Char> s) {
                               ^
/home/src/include/fmt/core.h:488:52: note: candidate template ignored: requirement 'is_compile_string<fmt::v7::text_style>::value' was not satisfied [with S = fmt::v7::text_style]
constexpr basic_string_view<typename S::char_type> to_string_view(const S& s) {
                                                   ^
2 errors generated.
[6/8] Linking CXX static library libfmtd.a
ninja: build stopped: subcommand failed.

```

include/fmt/color.h
vitaut: This overload is redundant because it's equivalent to passing back_insert_iterator to memory_buffer. There should be no overhead.

vitaut:
>  I just copied both format_to overloads from format.h and adapted it to accept text_style.

Unfortunately those are not good examples and need to change too eventually. We shouldn't repeat the same mistakes in new code.

vitaut:
Merged in https://github.com/fmtlib/fmt/commit/bff4d18efba37c0ffd44f036a7adaff1064c840f with minor tweaks. Thanks again.

Naios:
Thank you for merging this PR partially.

The `back_insert_iterator` overload is sadly by far not equivalent to the suggested `fmt:.memory_buffer` overload which you have dropped. Passing a `back_insert_iterator` always will insert content character-wise whereas using a `fmt::memory_buffer` overload will use `append` and passes the data in chunks.

The transfer to `iterator_buffer` might be buffered, but its flush to the back-inserted container is not, and always happens based on the `value_type` and will happen character-wise in case of strings.

Calling the API for colorization is now also unintentional because the non `fmt::text_style` overload accepts a `fmt::memory_buffer` whereas the `fmt::text_style` overload now requires a `std::back_inserter(...)`

Additionally colorization can't be used easily anymore in `struct fmt::formatter<` without this overload.

vitaut:
> The back_insert_iterator overload is sadly by far not equivalent to the suggested fmt::memory_buffer overload which you have dropped.

It can easily be made equivalent if it's not already. The main point is that extra overload is redundant and shouldn't be there.

>  the non fmt::text_style overload accepts a fmt::memory_buffer whereas the fmt::text_style overload now requires a std::back_inserter(...)

That's fine. The other overload is there mostly for historical reasons.

> Additionally colorization can't be used easily anymore in struct fmt::formatter< without this overload.

This is not the case because formatter specialization use iterators.

Naios:
Ok, I see your point now.

Overall the overload is not equal for `std::back_inserter` because the insertion will happen per character.
But it is possible to supply a custom output iterator which possibly could insert the content chunk-wise.

vitaut:
> the insertion will happen per character.

This is only because there is a missing `get_buffer[_init]` overload that should map `back_insert_iterator` to the underlying buffer.

Naios:
I have to revert my previous statement. `iterator_buffer::flush` is currently implemented as following:

```cpp
template <typename OutputIt, typename T, typename Traits>
void iterator_buffer<OutputIt, T, Traits>::flush() {
  out_ = std::copy_n(data_, this->limit(this->size()), out_);
  this->clear();
}
```

The (buffered) data is handed to the output one by one theoretically.
Maybe the standard library can do an optimization for common standard library containers wrapped inside a `std::back_insert_iterator` like `std::vector` but it is likely that it will fall back to `fmt::memory_buffer::push_back` when encountering `std::back_insert_iterator<fmt::memory_buffer>`.

Without it, every output iterator accepting API will never be equal to the same API using a `fmt::memory_buffer directly`.

Mapping a `std::back_insert_iterator<T>` to `T` is probably not possible due to a missing container getter in the standard, and everything else will be implementation dependent.

vitaut:
> Mapping a std::back_insert_iterator<T> to T is probably not possible due to a missing container getter in the standard, and everything else will be implementation dependent.

It is possible and already implemented in {fmt} with a minor caveat mentioned earlier.

Issue #1844, opened 2020-08-31
fmt::ostream move constructor does not properly initialize buffer


Boostibot:
The following code causes a segmentation fault: 

```cpp
fmt::ostream temp_file = fmt::output_file("file.txt");
fmt::ostream move_constructed(std::move(temp_file));
move_constructed.print("Hello world"); //SIGSEGV - Segmentation fault
```
This is because the internal buffer pointer and data is left uninitialised and left pointing to null.

The following change to fmt::ostream::ostream(ostream&& ) solves this issue
```cpp
ostream(ostream&& other) : file_(std::move(other.file_)) {
    set(other.data(), other.capacity()); //Added - Initialises the data
    other.set(nullptr, 0);
  }
```

vitaut:
Could you submit a PR?

vitaut:
Fixed in https://github.com/fmtlib/fmt/commit/6cccdc24bcb8605965c1e2497c61504cd388bde9. Thanks for catching this.

PR #1845, opened 2020-08-31
Bugfix with ostream move constructors in os.h

Boostibot:
A pull request as requested in #1844  


vitaut:
Thanks a lot for the PR. Could you also add a test case to https://github.com/fmtlib/fmt/blob/master/test/os-test.cc?

vitaut:
On a second thought, we should invoke the correct ctor instead of `set`. Fixed in https://github.com/fmtlib/fmt/commit/6cccdc24bcb8605965c1e2497c61504cd388bde9.

PR #1847, opened 2020-08-31
Add append mode.

t-wiser:

### DIFF
--- a/include/fmt/os.h
+++ b/include/fmt/os.h
@@ -278,7 +278,8 @@ class file {
     RDONLY = FMT_POSIX(O_RDONLY),  // Open for reading only.
     WRONLY = FMT_POSIX(O_WRONLY),  // Open for writing only.
     RDWR = FMT_POSIX(O_RDWR),      // Open for reading and writing.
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks!

t-wiser:
Thanks for the great library!

Would you be willing to consider a change to os.h to allow FMT_USE_FCNTL to
be overridden by a command-line '-DFMT_USE_FCNTL' (or equivalent on
Windows)? I'm trying to use fmt on CenOS/RHEL 7.8 which uses gcc 4.8.5
which does not have __has_include. This prevents me from using anything
that depends on having the <fcntl.h> header, which is available. So
basically the existing check could also pass (and thus use the header) if
FMT_USE_FCNTL was already defined.

vitaut:
`FMT_USE_FCNTL` is a bit too low-level, I recommend adding `|| defined(__linux__)` or similar to

 https://github.com/fmtlib/fmt/blob/1edd38b96e9c1b139c58995397d7cfb4bcd3d1ee/include/fmt/os.h#L32

instead.

PR #1848, opened 2020-09-01
Allow use of <fcntl.h> in Linux when __has_include is not available.

t-wiser:

This should allow older Linux compilers that do not have `__has_include` to use classes conditioned on the presence of the fcntl.h header.


vitaut:
Thanks!

PR #1851, opened 2020-09-02
Fix string_view detection with libstdc++ (#1850)

### LINKED ISSUE #1850: Incorrect FMT_USE_STRING_VIEW logic results in libstdc++ compilation errors
The following linked code in `core.h` results in compilation errors compiling with GCC (libstdc++) in our (C++14) library (https://github.com/rapidsai/cudf) which depends on `spdlog`, which bundles a version of `fmt`.

https://github.com/fmtlib/fmt/blob/6cccdc24bcb8605965c1e2497c61504cd388bde9/include/fmt/core.h#L228-L237

Specifically, while libc++ (clang) provides `string_view` before C++17, libstdc++ does not, it provides `experimental/string_view`.

Thus, I believe the `||` on line 230 should be `&&`, and making this change fixes our compilation issues.

harrism:
Fixes #1850 

Simple fix to correct finding `experimental/string_view` when the `std` library is `libstdc++` and C++ version is < C++17.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -847,10 +847,11 @@ template <typename T = char> class counting_buffer : public buffer<T> {
 template <typename T>
 class buffer_appender : public std::back_insert_iterator<buffer<T>> {
   using base = std::back_insert_iterator<buffer<T>>;
- 
  public:
-  explicit buffer_appender(buffer<T>& buf) : base(buf) {}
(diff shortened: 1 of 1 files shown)


harrism:
Clang-format is messing with me. I don't want to clutter the commit history so will start again.

PR #1853, opened 2020-09-03
use memchr for finding '%' in formatting string in printf.h

rimathia:

### DIFF
--- a/include/fmt/printf.h
+++ b/include/fmt/printf.h
@@ -474,8 +474,18 @@ OutputIt basic_printf_context<OutputIt, Char>::format() {
   const Char* end = parse_ctx_.end();
   auto it = start;
   while (it != end) {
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR but could you elaborate what this is for?

rimathia:
The goal here is to mirror the use of detail::find in format.h lines 2715 and 2730 with the goal of making printf with long format strings (order 1000 characters) faster. However, I don't have data to prove that this makes a difference, at this point the argument is purely by analogy: If it makes sense to use it in forma.h to search '{' and '}', it should make sense here as well.

vitaut:
It doesn't really make sense here since the parsing logic is completely different. Also microoptimizations should be backed by data.

rimathia:
Makes sense to me, thanks for getting to it so quickly.

PR #1854, opened 2020-09-03
treat public includes as system includes to avoid warning from headers

noizefloor:
Include directories for target fmt was splitted into INTERFACE and PRIVATE.
INTERFACE includes was marked as SYSTEM to suppress warnings in targets that
are linking fmt:fmt. PRIVATE includes are left as normal includes so that
warnings still occur when building fmt.

Includes for fmt-header-only are now marked as SYSTEM.

This technique is explained [here](https://foonathan.net/2018/10/cmake-warnings/.).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -211,7 +211,9 @@ endif ()
 
 target_compile_features(fmt INTERFACE ${FMT_REQUIRED_FEATURES})
 
-target_include_directories(fmt PUBLIC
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR but I don't think the headers should be marked as system. If some users want to do this they can do it on their own.

foonathan:
Well, adopting it would eliminate the constant PRs that do various workarounds to fix warnings...

vitaut:
Not sure if it's a good thing. Some of the fixes are actually useful.

PR #1855, opened 2020-09-03
NOMINMAX not handled properly

gcerretani:
Related to https://github.com/fmtlib/fmt/pull/1729. Details on https://github.com/gabime/spdlog/issues/1663.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -23,18 +23,14 @@
 #endif
 
 #ifdef _WIN32
-#  if !defined(NOMINMAX) && !defined(WIN32_LEAN_AND_MEAN)
+#  ifdef FMT_BIG_WIN
+#    include <windows.h>
+#  else
 #    define NOMINMAX
 #    define WIN32_LEAN_AND_MEAN
 #    include <windows.h>
 #    undef WIN32_LEAN_AND_MEAN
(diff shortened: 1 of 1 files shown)


include/fmt/format-inl.h
```diff
@@ -29,6 +29,10 @@
 #    include <windows.h>
 #    undef WIN32_LEAN_AND_MEAN
 #    undef NOMINMAX
+#  elif defined(NOMINMAX) && !defined(WIN32_LEAN_AND_MEAN) 
+#    define WIN32_LEAN_AND_MEAN 
+#    include <windows.h> 
+#    undef WIN32_LEAN_AND_MEAN 
 #  else
 #    include <windows.h>
 #  endif
```
vitaut, line 38: I think this whole block should be replaced with something like

```c++
#  ifdef FMT_BIG_WIN
#    include <windows.h>
#  else
#    define NOMINMAX
#    define WIN32_LEAN_AND_MEAN
#    include <windows.h>
#    undef WIN32_LEAN_AND_MEAN
#    undef NOMINMAX
#  endif
```
Then the default will be less broken windows headers and whoever wants all the garbage can define `FMT_BIG_WIN`.

vitaut (changes requested):
Thanks for the PR.

PR #1856, opened 2020-09-03
fix compiler warnings in public header files

noizefloor:
core.h: buffer had virtual functions but the dtor was not virtual.

format.f: fix two sign conversions warnings

This warnings occurred on clang and gcc with -Wsign-conversion and -Wnon-virtual-dtor.  

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -669,7 +669,7 @@ template <typename T> class buffer {
         size_(sz),
         capacity_(cap) {}
 
-  ~buffer() = default;
+  virtual ~buffer() = default;
 
   /** Sets the buffer data and capacity. */
   void set(T* buf_data, size_t buf_capacity) FMT_NOEXCEPT {
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -1609,7 +1609,7 @@ template <typename OutputIt, typename Char, typename UInt> struct int_writer {
     char digits[40];
     format_decimal(digits, abs_value, num_digits);
     basic_memory_buffer<Char> buffer;
-    size += prefix_size;
+    size += static_cast<int>(prefix_size);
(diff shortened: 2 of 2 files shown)


include/fmt/core.h
```diff
@@ -669,7 +669,7 @@ template <typename T> class buffer {
         size_(sz),
         capacity_(cap) {}
 
-  ~buffer() = default;
+  virtual ~buffer() = default;
```
vitaut, line 672: This is intentional.

include/fmt/format.h
```diff
@@ -3493,7 +3493,7 @@ inline std::string to_string(T value) {
   // The buffer should be large enough to store the number including the sign or
   // "false" for bool.
   constexpr int max_size = detail::digits10<T>() + 2;
-  char buffer[max_size > 5 ? max_size : 5];
+  char buffer[max_size > 5 ? static_cast<size_t>(max_size) : 5];
```
vitaut, line 3496: static_cast -> to_unsigned

vitaut (changes requested):
Thanks for the PR.

include/fmt/core.h
noizefloor: Is the idea behind that to keep the vtable as small as possible? If that's the case maybe by using CRTP we could get rid of the vtable at all and satisfy my compiler ;)

vitaut: http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-dtor-virtual. CRTP is not needed here, I suggest suppressing the useless warning.

vitaut:
Partially merged in https://github.com/fmtlib/fmt/commit/45da432d60535219284a970be3addbcc1d86b780. Thanks!

PR #1857, opened 2020-09-04
Support printing of filtered views

dvhwgumby:
Filtered views cannot be const and have no const iterators
Note: Transformed views can be const but don't have const iterators either.

The change in ranges.h `formatter<RangeT, Char,
                         enable_if_t<fmt::is_range<RangeT, Char>::value>>::format`
is relatively benign for this reason.

However constness is cast away from the argument to `value<typename
Context>::format_custom_arg` which is in core.h -- seemingly pretty harmless in
a private method without a callback.

It is possibly a bug in gcc-10 that `std::view::tranform` views can be const as the standard only permits `std::ranges::enable_view` to be specialized for cv-_un_qualified objects [https://en.cppreference.com/w/cpp/ranges/view](url)

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -1097,7 +1097,10 @@ template <typename Context> class value {
                                 Context& ctx) {
     Formatter f;
     parse_ctx.advance_to(f.parse(parse_ctx));
-    ctx.advance_to(f.format(*static_cast<const T*>(arg), ctx));
+    // We cast away any constness in order to print filtered views (ranges)
+    // which cannot be const
+    // This is a private method and we know that won't modify the object
+    ctx.advance_to(f.format(*const_cast<T*>(static_cast<const T*>(arg)), ctx));
   }
 };
 
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -257,7 +257,7 @@ struct formatter<RangeT, Char,
   }
 
   template <typename FormatContext>
-  typename FormatContext::iterator format(const RangeT& values,
+  typename FormatContext::iterator format(RangeT& values,
                                           FormatContext& ctx) {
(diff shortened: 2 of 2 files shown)


dvhwgumby:
Here's a brief test program:
```
#include <iostream>
#include <ranges>
#include <vector>

#include <fmt/format.h>
#include <fmt/ranges.h>

int main () {

    fmt::print("\n");

    std::vector<int> const ints{0,1,2,3,4,5};

    std::cout << "Streamed raw vector: ";
    for (int i : ints)
      std::cout << i << ' ';
    std::cout << std::endl;

    fmt::print("fmt joined raw vector: {}\n", fmt::join(ints, ", "));
    fmt::print("fmt raw vector: {}\n", ints);

    fmt::print("\n ========================================\n");


    auto square = [](int i) { return i * i; };
    auto const transformed { ints | std::views::transform(square) };

    std::cout << "Streamed transformed vector: ";
    for (int i : transformed)
      std::cout << i << ' ';
    std::cout << std::endl;

    fmt::print("fmt joined transformed vector: {}\n", fmt::join(transformed, ", "));
    fmt::print("fmt transformed vector: {}\n", transformed);

    fmt::print("\n ========================================\n");


    auto even = [](int i){ return 0 == i % 2; };
    auto /* const */ filtered    { transformed | std::views::filter(even) };

    std::cout << "Streamed filtered vector: ";
    for (int i : filtered)
      std::cout << i << ' ';
    std::cout << std::endl;

    fmt::print("fmt joined filtered vector: {}\n", fmt::join(filtered, ", "));

    fmt::print("fmt filtered vector: {}\n", filtered);
}
```

vitaut:
Thanks for the PR but this is a wrong approach. Instead of removing constness the `formatter` specialization for weird ranges should make a copy.

PR #1858, opened 2020-09-04
include/fmt/core.h: detail::buffer: declare destructor as virtual, as it already contains virtual methods.

medithe:


vitaut:
Thanks for the PR but this is intentional.

PR #1859, opened 2020-09-04
CMakeLists.txt: clang-pedantic-compiler-flags: added -Wnon-virtual-dtor

medithe:


vitaut:
Thanks for the PR but I don't see a good reason adding more exotic warnings there.

PR #1767, opened 2020-07-11
Add format_to_n overload that accepts FMT_COMPILE

Kurkin:
Resolves #1764 

I agree that my contributions are licensed under the {fmt} license, and agree to future changes to the licensing.

PR #1869, opened 2020-09-14
Add format_to_n overload that accepts FMT_COMPILE (from #1767)

alexezeder:
Extends https://github.com/fmtlib/fmt/pull/1767 by adding requested test.


vitaut:
Thank you!

Issue #1813, opened 2020-08-12
FMT_COMPILE does not support char8_t


Othereum:
The following code snippet isn't compiled:

```cpp
fmt::format(FMT_COMPILE(u8"{}.json"), u8"filename");
```

It gives compile error:

```
error C2440: 'return': cannot convert from 'std::string' to 'std::basic_string<char8_t,std::char_traits<char8_t>,std::allocator<char8_t>>'
message : No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called
message : see reference to function template instantiation 'std::basic_string<char8_t,std::char_traits<char8_t>,std::allocator<char8_t>> fmt::v7::format<main::<lambda_1>::()::FMT_COMPILE_STRING,const char8_t(&)[9],0>(const S &,const char8_t (&)[9])' being compiled
    with
    [
        S=main::<lambda_1>::()::FMT_COMPILE_STRING
    ]
```

I use char8_t string throughout my project. If FMT_COMPILE supports char8_t, it would be great. Thank you.

vitaut:
Added support for exotic character types including `char8_t` in https://github.com/fmtlib/fmt/commit/1d696dc280e9d2f7bf89be6819a558eacac754fe.

PR #1871, opened 2020-09-15
Fix char8_t with FMT_COMPILE()

alexezeder:
Resolves https://github.com/fmtlib/fmt/issues/1813

Signature of to_string was changed a bit.

### DIFF
--- a/include/fmt/compile.h
+++ b/include/fmt/compile.h
@@ -639,7 +639,7 @@ FMT_INLINE std::basic_string<typename S::char_type> format(const S&,
                                                            Args&&... args) {
   constexpr basic_string_view<typename S::char_type> str = S();
   if (str.size() == 2 && str[0] == '{' && str[1] == '}')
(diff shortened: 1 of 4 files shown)


vitaut:
Thanks for the PR. `to_string` shouldn't be parameterized on character type so I implemented a simpler fix that avoids it altogether in https://github.com/fmtlib/fmt/commit/1d696dc280e9d2f7bf89be6819a558eacac754fe.

alexezeder:
> `to_string` shouldn't be parameterized on character type

BTW, is there any reason to have separate `to_wstring`, but not templated `to_string`?

vitaut:
> is there any reason to have separate to_wstring, but not templated to_string?

It's basically a replacement for `std::to_[w]string`.

alexezeder:
> It's basically a replacement for `std::to_[w]string`.

wow, lucky me, I didn't know `std` has that...

PR #1877, opened 2020-09-17
add forgotten template argument to make_format_args which made some u…

rimathia:
…ses of FMT_COMPILE not work anymore after 54daa0864afb57e9d1, add more elaborate test cases to compile-test as regression tests


vitaut:
Thanks!

PR #1878, opened 2020-09-17
Improve FMT_ALWAYS_INLINE

jk-jeon:
1. `FMT_ALWAYS_INLINE` should imply `inline`; otherwise, there might be  linkage problems
2. Add specialization for MSVC (`__forceinline`)


vitaut:
Thank you!

PR #1879, opened 2020-09-18
Disable fallthrough attributes for the Intel compilers on Linux and MacOS

akohlmey:
On MacOS and Linux the Intel compilers may be identified as the host compilers (Clang or GNU) but do not support the corresponding compiler specific fallthrough attributes. This pull request adds conditions to the preprocessor tests to skip using those in those cases.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -70,9 +70,9 @@
 #endif
 
 #if __cplusplus == 201103L || __cplusplus == 201402L
-#  if defined(__clang__)
+#  if defined(__clang__) && !defined(__INTEL_COMPILER)
 #    define FMT_FALLTHROUGH [[clang::fallthrough]]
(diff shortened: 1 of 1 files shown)


include/fmt/format.h
```diff
@@ -70,9 +70,9 @@
 #endif
 
 #if __cplusplus == 201103L || __cplusplus == 201402L
-#  if defined(__clang__)
+#  if defined(__clang__) && !defined(__INTEL_COMPILER)
```
vitaut, line 73: I suggest moving the Intel compiler check in a branch of its own, something like

```c++
#  ifdef __INTEL_COMPILER
#    define FMT_FALLTHROUGH
#  elif ...
```
to avoid repetition.

akohlmey: how about something like this? looks like PGI falls into the same category.

vitaut: LGTM

Issue #1426, opened 2019-11-27
Optimize floating-point formatting


vitaut:
{fmt} is now ~10% faster than double-conversion and 15x faster than libc++ `std::ostringstream` on floating-point formatting according to the dtoa-benchmark but there is still a lot of room for improvement.

<img width="248" alt="Screen Shot 2019-11-27 at 7 53 15 AM" src="https://user-images.githubusercontent.com/576385/69738490-0d04d180-10eb-11ea-9683-18f11f597375.png">

#1097 gives some optimization ideas.

erthink:
Related to https://github.com/fmtlib/fmt/issues/1559 (There is a three times faster double-to-string implementation).

SanderBouwhuis:
Has this materialized? I use a LOT of those conversions, so this would be good news.
Having said that, I'm currently using MSVC printf, so presumably converting to fmt would give me a speed boost(?)

vitaut:
No but {fmt}'s floating-point formatter is already quite fast (see e.g. the benchmark results in https://github.com/fmtlib/fmt/releases/tag/6.1.0) so if you switch from `printf` you should see perf improvement.

SanderBouwhuis:
Thanks.
Should I use v6.1.0 which is 'somewhat' old, or can I use the nightlies?
I.e., are the nightlies safe, or are there regressions or known bugs?

vitaut:
> Should I use v6.1.0 which is 'somewhat' old, or can I use the nightlies?
> I.e., are the nightlies safe, or are there regressions or known bugs?

I recommend using 6.1.2 or master which includes the same optimization as 6.1.0 + some fixes.

> Is fmt still fast if I supply a wchar_t buffer that fmt should use?

Yes.

SanderBouwhuis:
@Vitaut, maybe fmt could use the new charconv in STL? They use Ryu which is (according to Stephan T. Lavavej, the implementer) 2x to 3x as fast as doubleconv (which you seem to be more or less on par with).
This would make the implementation quite clean, albeit a bit slower than Erthink's optimization.

data-man:
Some related projects:
[Grisu-Exact](https://github.com/jk-jeon/Grisu-Exact)
[C++ Ryu version](https://github.com/expnkx/fast_io/tree/master/include/fast_io_freestanding_impl/ryu) in [fast_io](https://github.com/expnkx/fast_io) library.

vitaut:
[...]

PR #1880, opened 2020-09-18
Fix msvc version of clz & clzll

jk-jeon:
See https://github.com/fmtlib/fmt/issues/1426#issuecomment-694920646


vitaut:
Thank you!

PR #1888, opened 2020-09-20
qkw: generalizing aliasing | using fmt library and it's features

ravijanjam:
I make extensive use of this library for building qkw targeting linux users. Please add this to your list.
https://github.com/ravijanjam/qkw


vitaut:
Merged, thanks.

PR #1882, opened 2020-09-18
Implement Dragonbox

jk-jeon:
From https://github.com/fmtlib/fmt/issues/1426#issuecomment-692282618.

(Sorry for the repetition, I realized a bug so have closed the previous one and now I'm opening this.)

All tests in the repo (except for os-test and format-test, which I believe are not related to this PR) passed, but there might be bugs. I'm not sure which branch I should choose for this kind of large PR, so I just have chose the master branch.

Performance is as measured as before.
(clang-cl)
milo1
milo2

(msvc)
milo1_msvc
milo2_msvc

Currently, I didn't touch `write` function, but I think it would be wiser to do that for the following reasons. First, as now `format_float` follows completely different paths for the shortest formatting and the fixed-precision formatting, I think there is little reason to unify them into one function. Second, it seems that currently the template argument `T` for `format_float` is always chosen to be `double`, but this introduces a needless branch inside `format_float`. Hence, I think it would be wiser to thus make separate functions for three different cases: the shortest formatting for `float`, the shortest formatting for `double` (and possibly `long double` when it is no different from `double`), and the fixed formatting.

I agree that my contributions are licensed under the {fmt} license, and agree to future changes to the licensing.

PR #1894, opened 2020-09-22
Fix an assertion error in Dragonbox

jk-jeon:
Fixes the issue https://github.com/fmtlib/fmt/pull/1882#issuecomment-696823912.
Sorry, it was indeed a very stupid mistake.


vitaut:
Thanks for the quick fix!

PR #1897, opened 2020-09-23
visual studio 2019 gives error C2049 

rglarix:
visual studio 2019 gives error C2049 ('fmt::v7': non-inline namespace cannot be reopened as inline) when compiling fmt with /clr)


vitaut:
Thanks

rglarix:
thanks to you for wonderful fmt!

Chronial:
Thanks for the fix, this has been bugging us for quite a while now. 🥳

PR #1898, opened 2020-09-24
Update format.h

moiwi:
Avoid MSVC 2019 warnings C4251: 'struct 'fmt::v7::detail::divtest_table_entry<uint32_t>' needs to have dll-interface to be used by clients of struct 'fmt::v7::detail::basic_data'


vitaut:
Thank you

PR #1900, opened 2020-09-24
remove dependency to windows.h

bernd5:
Because windows.h is full of macros (not just min and max) it is best to not include this header.
In format-inl.h we need not really much from it...

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -23,14 +23,15 @@
 #endif
 
 #ifdef _WIN32
-#include <io.h>
+#  include <io.h>
 FMT_BEGIN_NAMESPACE
-    extern "C" {
-        int __stdcall WriteConsoleW(void* hConsoleOutput, const void* lpBuffer, uint32_t nNumberOfCharsToWrite, uint32_t* lpNumberOfCharsWritten, void* lpReserved);
-    }
+extern "C" {
+int __stdcall WriteConsoleW(void* hConsoleOutput, const void* lpBuffer,
+                            uint32_t nNumberOfCharsToWrite,
+                            uint32_t* lpNumberOfCharsWritten, void* lpReserved);
+}
 FMT_END_NAMESPACE
 #endif
-
 
 // Dummy implementations of strerror_r and strerror_s called if corresponding
 // system functions are not available.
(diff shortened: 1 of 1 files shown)


include/fmt/format-inl.h
```diff
@@ -23,16 +23,15 @@
...
-#    define NOMINMAX
-#    define WIN32_LEAN_AND_MEAN
-#    include <windows.h>
-#    undef WIN32_LEAN_AND_MEAN
-#    undef NOMINMAX
-#  endif
 #  include <io.h>
+FMT_BEGIN_NAMESPACE
+  using DWORD = uint32_t;
+  using HANDLE = void*;
```
vitaut, line 29: I suggest replacing the uses of `DWORD` and `HANDLE` with `uint32_t` and `void*` respectfully. Then `FMT_BEGIN/ED_NAMESPACE` can go away.

include/fmt/format-inl.h
```diff
@@ -23,16 +23,15 @@
...
+
+  extern "C" {
+    int __stdcall WriteConsoleW(void* hConsoleOutput, const void* lpBuffer, uint32_t nNumberOfCharsToWrite, uint32_t* lpNumberOfCharsWritten, void* lpReserved);
```
vitaut, line 32: Please apply clang-format.

vitaut (changes requested):
Thanks for the PR. Two minor comments, otherwise looks good.

include/fmt/format-inl.h
bernd5: In general I would avoid to define something in the global namespace. This could cause for example a conflict if you have windows.h included.
Removing DWORD and HANDLE is fine.

include/fmt/format-inl.h
```diff
@@ -23,16 +23,14 @@
...
-#    include <windows.h>
-#  else
-#    define NOMINMAX
-#    define WIN32_LEAN_AND_MEAN
-#    include <windows.h>
-#    undef WIN32_LEAN_AND_MEAN
-#    undef NOMINMAX
-#  endif
 #  include <io.h>
+FMT_BEGIN_NAMESPACE
```
vitaut, line 27: This is not needed.

bernd5: This is needed. See here: https://godbolt.org/z/nPG544

vitaut: But why is it overloaded? Shouldn't the types match?

include/fmt/format-inl.h
vitaut: Looks like `uint32_t` should be replaced with `unsigned long` because that's how `DWORD` is defined. Then the namespace won't be necessary.

include/fmt/format-inl.h
bernd5: On Windows (x86, x64) DWORD is a 32 Bit unsigned integer. But in Windows.h it is defined as unsigned long / unsigned int. If we would use uint32 and include "windows.h" we would get an "overload extern C" error.

bernd5: A namespace helps just partially, it allows only C++ overloads for that function.

vitaut:
Merged, thanks again.

PR #1908, opened 2020-10-01
Fix warning C4018: '<=': signed/unsigned mismatch

heckad:

```
fmt\include\fmt/format-inl.h(1797): warning C4018: '<=': signed/unsigned mismatch
```

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -1790,8 +1790,8 @@ bool check_divisibility_and_divide_by_pow5(uint32_t& n) FMT_NOEXCEPT {
 template <int N> uint32_t small_division_by_pow10(uint32_t n) FMT_NOEXCEPT {
   static constexpr struct {
     uint32_t magic_number;
-    int shift_amount;
-    int divisor_times_10;
+    uint32_t shift_amount;
+    uint32_t divisor_times_10;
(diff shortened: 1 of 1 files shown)


include/fmt/format-inl.h
```diff
@@ -1790,8 +1790,8 @@ bool check_divisibility_and_divide_by_pow5(uint32_t& n) FMT_NOEXCEPT {
 template <int N> uint32_t small_division_by_pow10(uint32_t n) FMT_NOEXCEPT {
   static constexpr struct {
     uint32_t magic_number;
-    int shift_amount;
-    int divisor_times_10;
+    uint32_t shift_amount;
```
vitaut, line 1793: Is this necessary? The warning only refers to `divisor_times_10`.

vitaut (commented):
Thanks for the PR.

include/fmt/format-inl.h
heckad: Can `shift_amount` be less then 0?

vitaut: No but if this doesn't cause a warning, please keep it as is.

heckad: Are there any guarantees that it won't give out in the future?

Issue #1917, opened 2020-10-05
Wrong output for fixed-precision floating-point format


jk-jeon:
Please look at the following code:

```cpp
#include <iostream>
#define FMT_HEADER_ONLY 1
#include "fmt/format.h"

int main () {
 std::cout << fmt::format("{:.1e}", 1e-34);
}
```

The output of the code is
```
1.00e-34
```
but the expected output is
```
1.0e-34
```
So fmt is putting one more `0`. The same is happening when the precision is set to larger values. It seems that this issue only happens for certain floating-point inputs.

vitaut:
Good catch, thanks. Fixed in https://github.com/fmtlib/fmt/commit/a18b3fbbdceb6332c59341a44ad80bbe55f9a2dd.

PR #1920, opened 2020-10-06
Fix 1917

gggrafff:
This is my first time contributing.

Fix fmtlib/fmt#1917

Problem:
format("{:.1e}", 1e-12)
Actual: 1.00e-12
Expected: 1.0e-12

Solution:
Added EXPECT expressions for the situation described in fmtlib/fmt#1917.
Added an if expression for the individual case.


vitaut:
Thanks for the PR. This should be handled on a lower lever which is done in https://github.com/fmtlib/fmt/commit/a18b3fbbdceb6332c59341a44ad80bbe55f9a2dd.

Issue #1928, opened 2020-10-10
Unrecognized option `-mbig-obj`


MarcDirven:
The docs says:

---
# Portability

The library is highly portable and relies only on a small set of C++11 features:

- variadic templates
- type traits
- rvalue references
- decltype
- trailing return types
- deleted functions
- alias templates

These are available since GCC 4.8, Clang 3.0 and MSVC 19.0 (2015). For older compilers use {fmt} version 4.x which continues to be maintained and only requires C++98.

---

If you mean GCC 4.8.0, I think this is a small mistake because the library is not able to build because GCC 4.8.0 does not support method [ref qualifiers](https://akrzemi1.wordpress.com/2014/06/02/ref-qualifiers/) (on line 831 format.h):

Compiler | rvalue refs | ref-qualifiers | source
-- | -- | -- | --
GCC | 4.3 | 4.8.1 | here
Clang | 2.9 | 2.9 | here
Intel | 12.0 | 14.0 | here
Visual C++ | VC10 | Nov 2013 CTP | here

---

With GCC 4.8.1 installed however, building on Windows 10 with x86_64-4.8.1-release-posix-seh-rt_v3-rev2 gives: 
```
c:/users/marcd/downloads/x86_64-4.8.1-release-posix-seh-rt_v3-rev2/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/4.8.1/../../../../x86_64-w64-mingw32/bin/as.exe: unrecognized option `-mbig-obj'
mingw32-make.exe[3]: *** [fmt/CMakeFiles/fmt.dir/src/os.cc.obj] Error 1
mingw32-make.exe[3]: *** Waiting for unfinished jobs....
fmt\CMakeFiles\fmt.dir\build.make:96: recipe for target 'fmt/CMakeFiles/fmt.dir/src/os.cc.obj' failed
c:/users/marcd/downloads/x86_64-4.8.1-release-posix-seh-rt_v3-rev2/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/4.8.1/../../../../x86_64-w64-mingw32/bin/as.exe: unrecognized option `-mbig-obj'
mingw32-make.exe[3]: *** [fmt/CMakeFiles/fmt.dir/src/format.cc.obj] Error 1
mingw32-make.exe[2]: *** [fmt/CMakeFiles/fmt.dir/all] Error 2
mingw32-make.exe[1]: *** [CMakeFiles/untitled.dir/rule] Error 2
[...]

PR #1929, opened 2020-10-11
Added check for `-mbig-obj` and ref qualifier check

MarcDirven:

https://github.com/fmtlib/fmt/issues/1928


vitaut:
Thanks for the PR! There is no need to check for ref qualifier support, just remove them. Otherwise looks good.

MarcDirven:
Removed the ref qualifiers now

PR #1934, opened 2020-10-14
buffer should declare virtual dtor

Finkman:

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -670,7 +670,7 @@ template <typename T> class buffer {
         size_(sz),
         capacity_(cap) {}
 
-  ~buffer() = default;
+  virtual ~buffer() = default;
 
   /** Sets the buffer data and capacity. */
   void set(T* buf_data, size_t buf_capacity) FMT_NOEXCEPT {


vitaut:
Thanks for the PR but the destructor is non-virtual (and protected) intentionally.

Finkman:
hm.. interessting.
It will result in:
`fmt/include/fmt/format.h:678:3: error: 'fmt::v7::basic_memory_buffer<wchar_t, 500, std::allocator<wchar_t> >' has virtual functions but non-virtual destructor [-Werror,-Wnon-virtual-dtor]`

Compiled with clang [6,7,10] and gcc [7.5]
Is it an issue?

vitaut:
It's a GCC bug 7302 but a PR to suppress the warning without making the dtor virtual is welcome.

Finkman:
> It's a GCC bug 7302 but a PR to suppress the warning without making the dtor virtual is welcome.

I tried it with gcc 7.5, clang 9/10.
It seams that #1936 fixed it.

vitaut:
Thanks for checking.

PR #1937, opened 2020-10-15
Make classes derived from buffer<T> final to silence the virtual destructor warning.

BartSiwek:


vitaut:
Thank you!

PR #1939, opened 2020-10-17
added position independent documentation

udnaan:
Just a documentation update since not everyone who may come across this library would be familiar with all the CMAKE variables.

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -52,6 +52,13 @@ To build a `shared library`__ set the ``BUILD_SHARED_LIBS`` CMake variable to
 
 __ http://en.wikipedia.org/wiki/Library_%28computing%29#Shared_libraries
 
+
+To build a `static library`__ with position independent code (required if the main consumer of the fmt library is a shared library i.e. python extension) set the ``CMAKE_POSITION_INDEPENDENT_CODE`` CMake variable to
+``TRUE``::
+
+  cmake -DCMAKE_POSITION_INDEPENDENT_CODE=TRUE ...
+
+
 Installing the Library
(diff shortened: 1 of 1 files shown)


doc/usage.rst
```diff
@@ -52,6 +52,13 @@ To build a `shared library`__ set the ``BUILD_SHARED_LIBS`` CMake variable to
 
 __ http://en.wikipedia.org/wiki/Library_%28computing%29#Shared_libraries
 
+
+To build a `static library`__ with position independent code (required if the main consumer of the fmt library is a shared library i.e. python extension) set the ``CMAKE_POSITION_INDEPENDENT_CODE`` CMake variable to
```
vitaut, line 56: Why is "static library" a link?

python extension -> a Python extension

vitaut (changes requested):
Thanks for the PR. Looks good but please address minor comments and break long line (we use 80 character limit when possible).

doc/usage.rst
```diff
@@ -52,6 +52,13 @@ To build a `shared library`__ set the ``BUILD_SHARED_LIBS`` CMake variable to
 
 __ http://en.wikipedia.org/wiki/Library_%28computing%29#Shared_libraries
 
+
+To build a `static library` with position independent code (required if the main consumer of the fmt library is a shared library i.e. a Python extension) set the ``CMAKE_POSITION_INDEPENDENT_CODE`` CMake variable to
```
vitaut, line 56: Please break long line (we use 80 character limit when possible).

vitaut (changes requested):

PR #1940, opened 2020-10-17
Support std::ostream formatter in ranges

ProfDiesel:


vitaut:
Thanks for the PR but the ranges API intentionally uses only the formatter specializations.

ProfDiesel:
The only alternative I see to use ``fmt`` ostream support functionality on classes stored in ranges is to declare:
``template <> struct fmt::formatter<my_class> : fmt::detail::fallback_formatter<my_class> {};``
What would you think about exposing the functionality in the public API ?

vitaut:
Moving `fallback_formatter` to the public API seems reasonable.

PR #1941, opened 2020-10-19
Suppress warnings on gcc 9

Finkman:


vitaut:
Thanks for the PR but the warnings should be fixed, not suppressed.

Finkman:
First warns about, that a private non-virtual dtor is used, which is your intention.

The second warns about private-only methods, which is also your intention.

So... there is no room for a fix, in my Point of view

vitaut:
`is_output_iterator` has a generated ctor so the warning doesn't make much sense. That said the trait can be significantly simplified which is done in https://github.com/fmtlib/fmt/commit/e3b4c22ec9b215bba32cac922eb4503642004f16. As a side effect this eliminates the bogus warning.

The second looks like a shadowing warning which is also easy to fix.

PR #1942, opened 2020-10-19
Fix compilation for systems without fcntl.h

darklukee:


vitaut:
Thank you!

PR #1945, opened 2020-10-21
Add option to force usage of inline namespaces

darklukee:
Detection of inline namespaces is imperfect as some compilers
don't provide __has_feature

This option allows to override it if needed.


vitaut:
Thank you

PR #1953, opened 2020-10-24
Range support

kitegi:

this PR adds support for ranges when the `begin/end` members are not const. this is done by copying the range, since in those cases the range is likely to be a view, so a copy is cheap. if we want to be more restrictive, we can instead only do so when the object is `nothrow_copy_constructible`.

instead of using `begin/end` member functions, i've chosen to extend support for when they're found by argument dependent lookup, with a special case for C arrays. this is to be more consistent with the rest of the language, as the same is done for c++20 ranges and range-for loops.

an alternative to copying the range would be to forward the arguments, rather than capturing them as `const&`. but i'm not sure if that would be preferable.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -1198,13 +1198,22 @@ template <typename Context> struct arg_mapper {
   FMT_CONSTEXPR const void* map(void* val) { return val; }
   FMT_CONSTEXPR const void* map(const void* val) { return val; }
   FMT_CONSTEXPR const void* map(std::nullptr_t val) { return val; }
-  template <typename T> FMT_CONSTEXPR int map(const T*) {
+
+  // We use SFINAE instead of a const T* parameter to avoid conflicting with
+  // the C array overload.
+  template <typename T>
+  FMT_CONSTEXPR auto map(T) -> enable_if_t<std::is_pointer<T>::value, int> {
     // Formatting of arbitrary pointers is disallowed. If you want to output
     // a pointer cast it to "void *" or "const void *". In particular, this
     // forbids formatting of "[const] volatile char *" which is printed as bool
     // by iostreams.
     static_assert(!sizeof(T), "formatting of non-void pointers is disallowed");
     return 0;
+  }
+
+  template <typename T, std::size_t N>
+  FMT_CONSTEXPR auto map(const T (&values)[N]) -> const T (&)[N] {
+    return values;
   }
 
   template <typename T,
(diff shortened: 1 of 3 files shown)


include/fmt/ranges.h
```diff
@@ -95,12 +95,96 @@ template <typename... Ts> struct conditional_helper {};
...
+struct has_mutable_begin_end<
+    T, void_t<decltype(detail::range_begin(std::declval<T>())),
+              decltype(detail::range_begin(std::declval<T>())),
+              enable_if_t<std::is_copy_constructible<T>::value>>>
+    : std::true_type {};
+
+template <typename T>
+struct is_range_<T, void>
+    : std::integral_constant<bool, ((has_const_begin_end<T>::value) ||
+                                    (has_mutable_begin_end<T>::value))> {};
```
vitaut, line 162: nit: I don't think parentheses are needed here

vitaut (changes requested):
Looks great, thanks for the PR! Just a few minor comments inline.

include/fmt/ranges.h
```diff
@@ -95,12 +95,96 @@ template <typename... Ts> struct conditional_helper {};
 template <typename T, typename _ = void> struct is_range_ : std::false_type {};
 
 #if !FMT_MSC_VER || FMT_MSC_VER > 1800
+
+#  define FMT_DECLTYPE_RETURN(val)  \
+    ->decltype(val) { return val; } \
+    static_assert(true, "a")
```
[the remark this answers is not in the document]
kitegi: the static assert makes it so that a semi colon is required after the macro, which mostly just helps clang-format handle the code better. the "a" is to silence a clang-tidy warning to replace the statement with a [unary static assert](https://clang.llvm.org/extra/clang-tidy/checks/modernize-unary-static-assert.html). mostly a habit of mine but i'm not sure if it's needed here

include/fmt/ranges.h
kitegi: the inner parentheses are useless, yeah. the outer ones help with the code formatting, otherwise clang-format adds a level of indentation to the second line, but we can remove them if that's no big deal

include/fmt/ranges.h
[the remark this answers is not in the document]
vitaut: That's what I thought. Looks like clang-format does indeed make a mess without semicolon so let's keep it but replace "a" with an empty string and add a brief comment explaining what's static_assert for.

include/fmt/core.h
```diff
@@ -1207,6 +1211,11 @@ template <typename Context> struct arg_mapper {
     return 0;
   }
 
+  template <typename T, std::size_t N>
+  FMT_CONSTEXPR auto map(T const (&values)[N]) -> T const (&)[N] {
```
vitaut, line 1215: T const -> const T

vitaut (changes requested):

vitaut:
Thank you!

Issue #1957, opened 2020-10-26
Multiple errors on gcc 8.2/8.3


phprus:
gcc: 8.2.1
os: openSUSE 15.1

First of error:

```
[ 21%] Building CXX object test/CMakeFiles/locale-test.dir/locale-test.cc.o
/home/phprus/tmp/fmt/fmt-7.1.0/test/locale-test.cc: In member function ‘virtual void LocaleTest_Format_Test::TestBody()’:
/home/phprus/tmp/fmt/fmt-7.1.0/test/locale-test.cc:59:61: error: no matching function for call to ‘format_to(std::back_insert_iterator<std::__cxx11::basic_string<char> >, std::locale&, const char [5], int)’
   fmt::format_to(std::back_inserter(s), loc, "{:L}", 1234567);
                                                             ^
In file included from /home/phprus/tmp/fmt/fmt-7.1.0/include/fmt/format.h:44,
                 from /home/phprus/tmp/fmt/fmt-7.1.0/include/fmt/locale.h:13,
                 from /home/phprus/tmp/fmt/fmt-7.1.0/test/locale-test.cc:8:
/home/phprus/tmp/fmt/fmt-7.1.0/include/fmt/core.h:1994:13: note: candidate: ‘template<class OutputIt, class S, class ... Args, bool enable> typename std::enable_if<enable, OutputIt>::type fmt::v7::format_to(OutputIt, const S&, Args&& ...)’
 inline auto format_to(OutputIt out, const S& format_str, Args&&... args) ->
             ^~~~~~~~~
/home/phprus/tmp/fmt/fmt-7.1.0/include/fmt/core.h:1994:13: note:   template argument deduction/substitution failed:
/home/phprus/tmp/fmt/fmt-7.1.0/include/fmt/core.h:1993:74: error: no type named ‘type’ in ‘struct fmt::v7::detail::char_t_impl<std::locale, void>’
           bool enable = detail::is_output_iterator<OutputIt, char_t<S>>::value>
                                                                          ^~~~~
/home/phprus/tmp/fmt/fmt-7.1.0/include/fmt/core.h:1993:74: error: could not convert template argument ‘<expression error>’ from ‘<type error>’ to ‘bool’
In file included from /home/phprus/tmp/fmt/fmt-7.1.0/include/fmt/locale.h:13,
                 from /home/phprus/tmp/fmt/fmt-7.1.0/test/locale-test.cc:8:
[...]

PR #1959, opened 2020-10-27
Workaround bugs in gcc 8

phprus:
Workaround bugs in gcc 8.2/8.3 (issue #1957)


vitaut:
Thank you!

PR #1960, opened 2020-10-27
Allocator::max_size support in basic_memory_buffer

phprus:
The allocator may have a limit on the size of the allocated memory.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -742,9 +742,11 @@ void basic_memory_buffer<T, SIZE, Allocator>::grow(size_t size) {
 #ifdef FMT_FUZZ
   if (size > 5000) throw std::runtime_error("fuzz mode - won't grow that much");
 #endif
+  const size_t max_size = std::allocator_traits<Allocator>::max_size(alloc_);
   size_t old_capacity = this->capacity();
   size_t new_capacity = old_capacity + old_capacity / 2;
   if (size > new_capacity) new_capacity = size;
+  else if (new_capacity > max_size) new_capacity = (std::max)(size, max_size);
   T* old_data = this->data();
   T* new_data =
       std::allocator_traits<Allocator>::allocate(alloc_, new_capacity);
--- a/test/format-test.cc
+++ b/test/format-test.cc
@@ -354,6 +354,53 @@ TEST(MemoryBufferTest, ExceptionInDeallocate) {
   EXPECT_CALL(alloc, deallocate(&mem2[0], 2 * size));
 }
 
+template <typename Allocator, size_t MaxSize>
+class allocator_max_size: public Allocator {
+ public:
+  using typename Allocator::value_type;
+  size_t max_size() const FMT_NOEXCEPT {
+    return MaxSize;
+  }
+  value_type* allocate(size_t n) {
+    if (n > max_size()) {
+      throw std::length_error("size > max_size");
+    }
+    return std::allocator_traits<Allocator>::allocate(
+      *static_cast<Allocator *>(this), n);
+  }
+  void deallocate(value_type* p, size_t n) {
+    std::allocator_traits<Allocator>::deallocate(
+      *static_cast<Allocator *>(this), p, n);
+  }
+};
+
+TEST(MemoryBufferTest, AllocatorMaxSize) {
+  // 160 = 128 + 32
+  using TestAllocator = allocator_max_size<std::allocator<char>, 160>;
+  basic_memory_buffer<char, 10, TestAllocator> buffer;
+  buffer.resize(128);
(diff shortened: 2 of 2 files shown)


vitaut:
Thanks for the PR but I don't think it's necessary because `allocate` should throw if the size exceeds `max_size`.

phprus:
If "size <= max_size && new_capacity > max_size" then exception thrown from allocate() will not be correctly.

include/fmt/format.h
```diff
@@ -742,9 +742,14 @@ void basic_memory_buffer<T, SIZE, Allocator>::grow(size_t size) {
 #ifdef FMT_FUZZ
   if (size > 5000) throw std::runtime_error("fuzz mode - won't grow that much");
 #endif
+  const size_t max_size = std::allocator_traits<Allocator>::max_size(alloc_);
+  if (size > max_size) {
+    throw std::length_error("fmt::basic_memory_buffer::grow");
+  }
```
vitaut, line 748: Please remove this to reduce the number of checks.

include/fmt/format.h
```diff
@@ -742,9 +742,14 @@ void basic_memory_buffer<T, SIZE, Allocator>::grow(size_t size) {
...
   size_t old_capacity = this->capacity();
   size_t new_capacity = old_capacity + old_capacity / 2;
-  if (size > new_capacity) new_capacity = size;
+  if (new_capacity > max_size) new_capacity = max_size;
+  if (size > new_capacity) new_capacity = size; // size is never greater max_size.
```
vitaut, line 752: We only need to check `new_capacity` if it is not greater than `size`. The rest `allocate` will take care of. Something like:

```
if (size > new_capacity) new_capacity = size;
else if (new_capacity > max_size) new_capacity = (std::max)(size, max_size);
```

vitaut (changes requested):
This looks reasonable. Please add a unit test and address the inline comments.

phprus:
Commit updated.

include/fmt/format.h
phprus: Removed.

test/mock-allocator.h
```diff
@@ -57,4 +57,24 @@ template <typename Allocator> class allocator_ref {
...
+      throw std::length_error("size > max_size");
+    }
+    return std::allocator_traits<Allocator>::allocate(
+      *static_cast<Allocator *>(this), n);
+  }
+  void deallocate(value_type* p, size_t n) {
+    std::allocator_traits<Allocator>::deallocate(
+      *static_cast<Allocator *>(this), p, n);
+  }
+};
```
vitaut, line 78: Please move this to the `format-test.cc` since it's only used there and not a mock.

test/format-test.cc
```diff
@@ -354,6 +354,53 @@ TEST(MemoryBufferTest, ExceptionInDeallocate) {
   EXPECT_CALL(alloc, deallocate(&mem2[0], 2 * size));
 }
 
+template <typename Allocator, size_t MaxSize>
+class allocator_max_size: public Allocator {
+ public:
+  typedef typename Allocator::value_type value_type;
```
vitaut, line 360: Here and below: typedef -> using

test/format-test.cc
```diff
@@ -354,6 +354,53 @@ TEST(MemoryBufferTest, ExceptionInDeallocate) {
...
+  basic_memory_buffer<char, 10, TestAllocator> buffer;
+  buffer.resize(128);
+  bool throws_on_resize = false;
+  try {
+    // new_capacity = 128 + 128/2 = 192 > 160
+    buffer.resize(160);
+  } catch (const std::exception &) {
+    throws_on_resize = true;
+  }
+  EXPECT_EQ(throws_on_resize, false);
```
vitaut, line 389: EXPECT_TRUE

test/format-test.cc
```diff
@@ -354,6 +354,53 @@ TEST(MemoryBufferTest, ExceptionInDeallocate) {
...
+TEST(MemoryBufferTest, AllocatorMaxSizeOverflow) {
+  typedef allocator_max_size<std::allocator<char>, 160> TestAllocator;
+  basic_memory_buffer<char, 10, TestAllocator> buffer;
+  bool throws_on_resize = false;
+  try {
+    buffer.resize(161);
+  } catch (const std::exception &) {
+    throws_on_resize = true;
+  }
+  EXPECT_EQ(throws_on_resize, true);
```
vitaut, line 401: EXPECT_TRUE

vitaut (commented):
A few more nits, otherwise LGTM

vitaut:
Thank you!

PR #1962, opened 2020-10-28
:new: [CMake] Enable generation of compile_commands.json

jgopel:
Problem:
- `compile_commands.json` is not being generated, which hurts editor
  integration with tooling.

Solution:
- Generate the `compile_commands.json` file with CMake, if the CMake
  version is sufficient.


vitaut:
Thanks for the PR but I don't think we should generate `compile_commands.json` by default and you can already control it with a CMake option.

PR #1964, opened 2020-10-28
Removed [-Wsign-conversion] warning in GCC

OptoCloud:


vitaut:
Thank you

PR #1965, opened 2020-10-28
Rename all shadowed types and variables and enable Wshadow when in pedantic mode

jgopel:

### DIFF
--- a/include/fmt/os.h
+++ b/include/fmt/os.h
@@ -363,8 +363,8 @@ struct ostream_params {
   ostream_params() {}
 
   template <typename... T>
-  ostream_params(T... params, int oflag) : ostream_params(params...) {
-    this->oflag = oflag;
+  ostream_params(T... params, int new_oflag) : ostream_params(params...) {
+    oflag = new_oflag;
   }
 
   template <typename... T>
--- a/test/core-test.cc
+++ b/test/core-test.cc
@@ -83,8 +83,8 @@ template <typename T> struct mock_buffer final : buffer<T> {
 
   void grow(size_t capacity) { this->set(this->data(), do_grow(capacity)); }
 
-  mock_buffer(T* data = nullptr, size_t capacity = 0) {
-    this->set(data, capacity);
+  mock_buffer(T* data = nullptr, size_t buf_capacity = 0) {
+    this->set(data, buf_capacity);
     ON_CALL(*this, do_grow(_)).WillByDefault(Invoke([](size_t capacity) {
       return capacity;
     }));
@@ -224,9 +224,9 @@ struct custom_context {
   };
 
   bool called;
-  fmt::format_parse_context ctx;
+  fmt::format_parse_context parse_ctx;
 
-  fmt::format_parse_context& parse_context() { return ctx; }
+  fmt::format_parse_context& parse_context() { return parse_ctx; }
   void advance_to(const char*) {}
 };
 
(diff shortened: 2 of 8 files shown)


include/fmt/os.h
```diff
@@ -363,8 +363,8 @@ struct ostream_params {
   ostream_params() {}
 
   template <typename... T>
-  ostream_params(T... params, int oflag) : ostream_params(params...) {
-    this->oflag = oflag;
+  ostream_params(T... params, int new_oflag) : ostream_params(params...) {
+    this->oflag = new_oflag;
```
vitaut, line 367: Please remove `this->`.

test/core-test.cc
```diff
@@ -213,12 +213,12 @@ struct custom_context {
 
   template <typename T> struct formatter_type {
     template <typename ParseContext>
-    auto parse(ParseContext& ctx) -> decltype(ctx.begin()) {
-      return ctx.begin();
+    auto parse(ParseContext& parse_ctx) -> decltype(parse_ctx.begin()) {
```
vitaut, line 216: Let's keep argument names as is but rename the data member to `parse_ctx` since it has wider scope.

test/format-test.cc
```diff
@@ -219,12 +219,12 @@ TEST(AllocatorTest, allocator_ref) {
   check_forwarding(alloc, ref3);
 }
 
-typedef allocator_ref<std::allocator<char>> TestAllocator;
+typedef allocator_ref<std::allocator<char>> StdAllocator;
```
vitaut, line 222: Let's follow the naming conventions while at it:

```c++
using std_allocator = allocator_ref<std::allocator<char>>;
```

vitaut (changes requested):
Thanks for the PR. Mostly looks good but please fix CI build failures and address inline comments.

include/fmt/os.h
jgopel: Done

test/core-test.cc
jgopel: Done - I'm glad you looked at that one, I had a feeling that the situation might be something like this but I didn't want to be too aggressive so I tried my best to keep the scope very local.

test/format-test.cc
jgopel: 👍 Done

vitaut:
A few more build failures: https://travis-ci.org/github/fmtlib/fmt/jobs/740386845

jgopel:
I can't get this last set of errors to reproduce locally for reasons that are not clear to me, so there might be a bit of churn with using CI to find the errors.

jgopel:
Just in case anyone else encounters the issue of not being able to reproduce errors locally - the C++ standard seems to matter to at least GCC for whether or not ` -Wshadow` generates warnings in certain narrow contexts.

test/format-test.cc
```diff
@@ -105,7 +105,7 @@ struct uint32_pair {
 TEST(UtilTest, BitCast) {
   auto s = fmt::detail::bit_cast<uint32_pair>(uint64_t{42});
   EXPECT_EQ(fmt::detail::bit_cast<uint64_t>(s), 42ull);
-  s = fmt::detail::bit_cast<uint32_pair>(uint64_t(~0ull));
+  s = fmt::detail::bit_cast<uint32_pair>(~0ull);
```
vitaut, line 108: This looks unrelated and technically not correct.

test/std-format-test.cc
```diff
@@ -102,7 +102,7 @@ template <> struct std::formatter<S> {
   size_t width_arg_id = 0;
 
   // Parses a width argument id in the format { <digit> }.
-  constexpr auto parse(format_parse_context& ctx) {
+  auto parse(format_parse_context& ctx) {
```
vitaut, line 105: Please revert.

vitaut (changes requested):

test/std-format-test.cc
jgopel: Whoops, sorry about that - fixed

test/format-test.cc
jgopel: Whoops, sorry about that - fixed.

darklukee:
@jgopel thank you for doing this. It was annoying my team for some time now :)

vitaut:
Merged, thanks!

PR #1968, opened 2020-10-29
Remove accidental parenthesis

tohammer:
fails only when FMT_BUILTIN_CTZLL is not defined


vitaut:
Good catch, thanks!

PR #1974, opened 2020-11-03
Remove sizeof from unused variable silencer

darklukee:
Using sizeof causes some compilers to complain:
'operand of sizeof is not a type, variable, or dereferenced pointer'

static_cast itself should be enough to silence unused variable warning


vitaut:
Thank you

Issue #1973, opened 2020-11-02
visibility warnings with ARM GCC 7.3


alexezeder:
There is a weird warning related to the visibility of some classes of {fmt} on GCC 6.4 and 7.3 on ARM. These `-Wattributes` warnings produced with `-fvisibility=hidden` only.

For example, there is no warning for:
```cpp
#include "fmt/format.h"

int main(int argc, char** argv) {
    auto message = fmt::format(FMT_STRING("{}"), argv[0]);
}
```

But if I want to use literals then there is a warning telling that visibility of `FMT_COMPILE_STRING` struct (defined inside the `FMT_STRING`'s lambda) is greater than the visibility of its base `fmt::compile_string` struct:
https://godbolt.org/z/vfGanM

Same warning, simplified code:
https://godbolt.org/z/Pr8xKz
```cpp
#include "fmt/format.h"

template <typename... Args>
std::string func(Args&&... args) {
    return format(FMT_STRING("{}"), std::forward<Args>(args)...);
}

int main() {
    auto m = func(42);
}
```
ARM GCC 7.3 output:
```
In file included from <source>:1:0:
<source>: In instantiation of 'struct func(Args&& ...)::<lambda()> [with Args = {int}]::FMT_COMPILE_STRING':
<source>:5:19:   required from 'func(Args&& ...)::<lambda()> [with Args = {int}]'
<source>:5:19:   required from 'struct func(Args&& ...) [with Args = {int}; std::__cxx11::string = std::__cxx11::basic_string<char>]::<lambda()>'
<source>:5:19:   required from 'std::__cxx11::string func(Args&& ...) [with Args = {int}; std::__cxx11::string = std::__cxx11::basic_string<char>]'
<source>:9:21:   required from here
/opt/compiler-explorer/libs/fmt/trunk/include/fmt/format.h:3061:12: warning: 'func(Args&& ...)::<lambda()> [with Args = {int}]::FMT_COMPILE_STRING' declared with greater visibility than the type of its field 'func(Args&& ...)::<lambda()> [with Args = {int}]::FMT_COMPILE_STRING::<anonymous>' [-Wattributes]
     struct FMT_COMPILE_STRING : base {                            \
            ^
/opt/compiler-explorer/libs/fmt/trunk/include/fmt/format.h:3081:23: note: in expansion of macro 'FMT_STRING_IMPL'
 #define FMT_STRING(s) FMT_STRING_IMPL(s, fmt::compile_string)
[...]

PR #1975, opened 2020-11-03
-Wattributes visibility warning with some GCC versions

alexezeder:

More info about the problem that should be fixed by this PR: https://github.com/fmtlib/fmt/issues/1973

There are a few steps in this PR:
1. fail CI build on ARM by passing `-fvisibility=hidden` to the tests, looks like it was not specific to ARM, but x86_64 gcc-6 also produces this warning, so there was no need to create additional CI configuration, but I didn't know that
2. fix the problem, so CI should pass again
3. revert my commit which failed the CI build without this fix


include/fmt/format.h
```diff
@@ -3198,17 +3198,24 @@ FMT_CONSTEXPR basic_string_view<Char> compile_string_to_view(
...
-    /* Use a macro-like name to avoid shadowing warnings. */      \
-    struct FMT_COMPILE_STRING : base {                            \
-      using char_type = fmt::remove_cvref_t<decltype(s[0])>;      \
-      FMT_MAYBE_UNUSED FMT_CONSTEXPR                              \
-      operator fmt::basic_string_view<char_type>() const {        \
-        return fmt::detail::compile_string_to_view<char_type>(s); \
-      }                                                           \
-    };                                                            \
-    return FMT_COMPILE_STRING();                                  \
+#if defined(FMT_GCC_VERSION) && FMT_GCC_VERSION >= 400
```
phprus, line 3201: Bug present only in GCC 6.4 - 7.3:
#if FMT_GCC_VERSION >= 604 && FMT_GCC_VERSION < 704

.travis.yml
```diff
@@ -58,6 +58,20 @@ matrix:
           packages:
             - g++-8
 
+      # ARM g++ 6 on Linux with C++14
```
phprus, line 61: Specialized configuration is not needed.

include/fmt/format.h
vitaut: There is no need to overspecify. The attribute is harmless.

.travis.yml
alexezeder: yeah, I will revert this commit, was not sure that this warning can also be produced on x86_64 gcc-6 (tested it on Compile Explorer)

include/fmt/format.h
alexezeder: Ok, I leave this line as it is then, right?..

vitaut: Yes

vitaut: Actually you don't need to check the version at all because we don't support gcc < 4, so this should be just `#if FMT_GCC_VERSION`.

.travis.yml
vitaut: I don't think we need another CI config for this.

test/CMakeLists.txt
```diff
@@ -83,6 +83,9 @@ function(add_fmt_test name)
   if (FMT_WERROR)
     target_compile_options(${name} PRIVATE ${WERROR_FLAG})
   endif ()
+  if (CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+    target_compile_options(${name} PRIVATE -fvisibility=hidden -Werror=attributes)
```
vitaut, line 87: Please revert.

vitaut (changes requested):
Thanks for the PR. Mostly looks good but please address inline comments.

include/fmt/format.h
alexezeder: done

.travis.yml
alexezeder: reverted

include/fmt/format.h
```diff
@@ -3198,17 +3198,24 @@ FMT_CONSTEXPR basic_string_view<Char> compile_string_to_view(
...
-    struct FMT_COMPILE_STRING : base {                            \
-      using char_type = fmt::remove_cvref_t<decltype(s[0])>;      \
-      FMT_MAYBE_UNUSED FMT_CONSTEXPR                              \
-      operator fmt::basic_string_view<char_type>() const {        \
-        return fmt::detail::compile_string_to_view<char_type>(s); \
-      }                                                           \
-    };                                                            \
-    return FMT_COMPILE_STRING();                                  \
+#if FMT_GCC_VERSION
+#  define FMT_COMPILE_STRING_VISIBILITY_ATTRIBUTE \
```
vitaut, line 3202: I suggest renaming `FMT_COMPILE_STRING_VISIBILITY_ATTRIBUTE ` to `FMT_GCC_VISIBILITY_HIDDEN` to indicate that it's a gcc attribute and that it can be applied in other places (not specific to COMPILE_STRING).

alexezeder: Should it be placed in some other place then?

vitaut: Sure, somewhere next to https://github.com/fmtlib/fmt/blob/cba5970cd8a330b5f4d4a26281121fac90815a6b/include/fmt/format.h#L66-L70

include/fmt/format.h
```diff
@@ -3198,17 +3198,24 @@ FMT_CONSTEXPR basic_string_view<Char> compile_string_to_view(
...
+    __attribute__((visibility("hidden")))
+#else
+#  define FMT_COMPILE_STRING_VISIBILITY_ATTRIBUTE
+#endif
+
+#define FMT_STRING_IMPL(s, base)                                               \
+  [] {                                                                         \
+    /* Use a macro-like name to avoid shadowing warnings. */                   \
+    struct FMT_COMPILE_STRING_VISIBILITY_ATTRIBUTE FMT_COMPILE_STRING : base { \
```
vitaut, line 3211: Also please add a short comment explaining why we need visibility hidden here.

include/fmt/format.h
alexezeder: ok, and another question: since it is not only GCC-specific maybe the name should be `FMT_VISIBILITY_HIDDEN`?

alexezeder: don't mind, it's probably out of the scope of this PR, so I will create only GCC-specific `FMT_GCC_VISIBILITY_HIDDEN`
IMHO, if there will be a need for the same attribute but for all compilers then `FMT_VISIBILITY_HIDDEN` should be added with `FMT_GCC_VISIBILITY_HIDDEN` inside GCC-specific branch

alexezeder: renamed to `FMT_GCC_VISIBILITY_HIDDEN` and moved

include/fmt/format.h
alexezeder: done, hope this will be enough

vitaut:
LGTM but please rebase since there seem to be some conflicts.

alexezeder:
> LGTM but please rebase since there seem to be some conflicts.

hmm... @vitaut, are you sure, because I can see "_This branch has no conflicts with the base branch_" text here, also I'm able to merge in my local repo

vitaut:
Yes, but I guess I can squash the commits.

PR #1977, opened 2020-11-04
Fix test on OpenBSD

fagg:

Hello,

I am working on packaging fmt for OpenBSD. During this process I encountered a problem with one of the tests. It appears that for some reason the library is expecting particular error message wording in one of the tests. OpenBSD returns a different error message.

As far as I can tell, this is purely cosmetic (I may be wrong). If there's a better way to resolve this please let me know.

Btw with this change, the entire test suite passes on OpenBSD.

This PR just adds an ifdef to substitute the message when fmt is being built on OpenBSD.

### DIFF
--- a/test/gtest-extra-test.cc
+++ b/test/gtest-extra-test.cc
@@ -375,9 +375,15 @@ TEST(OutputRedirectTest, DupErrorInCtor) {
   file copy = file::dup(fd);
   FMT_POSIX(close(fd));
   std::unique_ptr<OutputRedirect> redir{nullptr};
+#ifdef __OpenBSD__
+  EXPECT_SYSTEM_ERROR_NOASSERT(
+      redir.reset(new OutputRedirect(f.get())), EBADF,
+      fmt::format("cannot flush stream"));
+#else
   EXPECT_SYSTEM_ERROR_NOASSERT(
       redir.reset(new OutputRedirect(f.get())), EBADF,
       fmt::format("cannot duplicate file descriptor {}", fd));
+#endif
   copy.dup2(fd);  // "undo" close or dtor will fail
 }
 


vitaut:
Thanks for the PR. Just changing the expected error doesn't look right. Could you capture the stack trace where the exception is being thrown?

vitaut:
I think a better fix would be to move `flush` https://github.com/fmtlib/fmt/blob/49544ea9432a13285846aca6d191f01e8ed00655/test/gtest-extra.cc#L32 after the call to `dup` here: https://github.com/fmtlib/fmt/blob/49544ea9432a13285846aca6d191f01e8ed00655/test/gtest-extra.cc#L35

This way the error should be consistent despite the differences in `flush`.

fagg:
> I think a better fix would be to move `flush` from here:
> 
> https://github.com/fmtlib/fmt/blob/49544ea9432a13285846aca6d191f01e8ed00655/test/gtest-extra.cc#L32
> 
> to after the call to `dup` here:
> 
> https://github.com/fmtlib/fmt/blob/49544ea9432a13285846aca6d191f01e8ed00655/test/gtest-extra.cc#L35
> 
> This way the error should be consistent despite the differences in `flush`.

vitaut,

Thanks for the review and the  advice. I'll make the change you suggested and test it out, and get back to you.

fagg:
G'day Victor,

I tried your suggestion - no dice.

I've attached ctest output. before.txt is with no modifications. after.txt is with my fix reverted and with your suggestion of moving the flush() call.

[before.txt](https://github.com/fmtlib/fmt/files/5489771/before.txt)
[after.txt](https://github.com/fmtlib/fmt/files/5489772/after.txt)

vitaut:
Could you capture the stack trace where the exception with the wrong message is being thrown?

fagg:
Victor Zverovich <notifications@github.com> writes:

> Could you capture the stack trace where the exception with the wrong
> message is being thrown?

Hmm...I'm trying but it seems the debug build has some disagreement with
OpenBSD's debugger. Let me try figure that out and I'll get back to you.

fagg:
Ok I have worked out  that in flush(), the fflush call throws up with "Bad file descriptor". I need to look into this but I am expecting there's a use-after-free somewhere here, and OpenBSD's mitigations are catching it.

I'll keep digging and see what I find, but I think for now the original PR isn't relevant.

PR #1978, opened 2020-11-04
update doc GitHub Actions workflow

alexezeder:
I agree that my contributions are licensed under the {fmt} license, and agree to future changes to the licensing.

Minor changes to build HTML documentation using GitHub Actions, since help is wanted https://github.com/fmtlib/fmt/issues/1969

PR #1980, opened 2020-11-04
update doc GitHub Actions workflow

alexezeder:

Same as https://github.com/fmtlib/fmt/pull/1978 but for `master` branch now


vitaut:
Thanks for the PR. Unfortunately it cannot be merged in because the purpose of the doc config is to update the online documentation, not just to verify that the documentation builds.

alexezeder:
This is a good approach 👍. However I know nothing about the update process of online documentation, so looks like I cannot help with that.

PR #1982, opened 2020-11-05
Printf get container

rimathia:
I agree that my contributions are licensed under the {fmt} license, and agree to future changes to the licensing.

This affects the performance of fmt::sprintf in the case where the format string is at least on the order of 100 characters long.

The benchmark:

```c++
#include <benchmark/benchmark.h>
#include <fmt/printf.h>

void long_printf_format_string(benchmark::State& state) {
  std::string format_string = std::string(state.range(0), 'A') + "%s";
  std::string message = "Hi";
  for (auto _ : state) {
    benchmark::DoNotOptimize(fmt::sprintf(format_string, message));
  };
}
BENCHMARK(long_printf_format_string)->Arg(10)->Arg(100)->Arg(1000)->Arg(10000);

void long_fmt_format_string(benchmark::State& state) {
  std::string format_string = std::string(state.range(0), 'A') + "{}";
  std::string message = "Hi";
  for (auto _ : state) {
    benchmark::DoNotOptimize(fmt::format(format_string, message));
  };
}
BENCHMARK(long_fmt_format_string)->Arg(10)->Arg(100)->Arg(1000)->Arg(10000);
```

The benchmark numbers change as follows:

master (60dc273513449dc959fc8212464058c54c7c36bb)
```
--------------------------------------------------------------------------
Benchmark                                Time             CPU   Iterations
--------------------------------------------------------------------------
long_printf_format_string/10          39.9 ns         39.9 ns     17364987
long_printf_format_string/100          188 ns          188 ns      3697522
long_printf_format_string/1000        1402 ns         1402 ns       494033
long_printf_format_string/10000      12409 ns        12409 ns        55024
long_fmt_format_string/10             28.0 ns         28.0 ns     24726944
long_fmt_format_string/100            88.2 ns         88.2 ns      7890169
long_fmt_format_string/1000            333 ns          333 ns      2119728
long_fmt_format_string/10000           873 ns          873 ns       779553
```
branch (104b3de77b637f94334dcfd726abe3d38c3d13c)
```
--------------------------------------------------------------------------
Benchmark                                Time             CPU   Iterations
--------------------------------------------------------------------------
long_printf_format_string/10          36.7 ns         36.7 ns     17982978
long_printf_format_string/100         95.8 ns         95.8 ns      7214487
long_printf_format_string/1000         337 ns          337 ns      2075802
long_printf_format_string/10000        763 ns          763 ns       910155
long_fmt_format_string/10             28.0 ns         28.0 ns     24945565
[...]

PR #1984, opened 2020-11-05
use memchr for searching for '%' in printf format string

rimathia:

This is the same change as in the closed pull request https://github.com/fmtlib/fmt/pull/1853 but this time with a simple microbenchmark to argue for it. I hope this isn't considered impolite.

The change starts to be noticeable at a format string length of about 100 characters.

The microbenchmark is the same as in https://github.com/fmtlib/fmt/pull/1982:

```c++
#include <benchmark/benchmark.h>
#include <fmt/printf.h>

void long_argument(benchmark::State& state) {
  std::string long_string(state.range(0), 'A');
  for (auto _ : state) {
    benchmark::DoNotOptimize(fmt::sprintf("%s", long_string));
  };
}
BENCHMARK(long_argument)->Arg(10)->Arg(100)->Arg(1000)->Arg(10000);

void long_format_string(benchmark::State& state) {
  std::string format_string = std::string(state.range(0), 'A') + "%s";
  std::string message = "Hi";
  for (auto _ : state) {
    benchmark::DoNotOptimize(fmt::sprintf(format_string, message));
  };
}
BENCHMARK(long_format_string)->Arg(10)->Arg(100)->Arg(1000)->Arg(10000);

BENCHMARK_MAIN();
```

The microbenchmark numbers:

master (038057eb3ef7ff083eb0c51802c3831473c614d0)
```
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
long_argument/10               36.6 ns         36.6 ns     17808714
long_argument/100               188 ns          188 ns      3716840
long_argument/1000             1284 ns         1284 ns       538288
long_argument/10000           11769 ns        11767 ns        62667
long_format_string/10          39.6 ns         39.6 ns     17552922
long_format_string/100          237 ns          237 ns      2934014
long_format_string/1000        1882 ns         1882 ns       383536
long_format_string/10000      18333 ns        18328 ns        39354
```
branch (e66299973b4f1d580ff1d94bdf920fbb60380ca5)
```
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
long_argument/10               39.7 ns         39.7 ns     16266435
long_argument/100               191 ns          191 ns      3609438
long_argument/1000             1304 ns         1304 ns       535373
long_argument/10000           11039 ns        11039 ns        62367
long_format_string/10          39.1 ns         39.1 ns     17829580
long_format_string/100          191 ns          191 ns      3675833
long_format_string/1000        1383 ns         1383 ns       493149
long_format_string/10000      12418 ns        12418 ns        55520
```

### DIFF
--- a/include/fmt/printf.h
+++ b/include/fmt/printf.h
@@ -474,13 +474,8 @@ OutputIt basic_printf_context<OutputIt, Char>::format() {
   const Char* end = parse_ctx_.end();
   auto it = start;
   while (it != end) {
-    if (end - it < 32) {
-      // Use a simple linear search instead of memchr for small strings.
-      it = std::find(it, end, '%');
-    } else {
-      if (!detail::find<false, Char>(it, end, '%', it)) {
-        it = end;
-      }
+    if (!detail::find<false, Char>(it, end, '%', it)) {
+      it = end;
     }
     if (it == end) {
       continue;
--- a/include/fmt/printf.h
+++ b/include/fmt/printf.h
@@ -474,8 +474,18 @@ OutputIt basic_printf_context<OutputIt, Char>::format() {
   const Char* end = parse_ctx_.end();
   auto it = start;
   while (it != end) {
+    if (end - it < 32) {
+      // Use a simple linear search instead of memchr for small strings.
+      it = std::find(it, end, '%');
+    } else {
+      if (!detail::find<false, Char>(it, end, '%', it)) {
+        it = end;
+      }
+    }
+    if (it == end) {
+      continue;
+    }
     char_type c = *it++;
-    if (c != '%') continue;
     if (it != end && *it == c) {
       out = std::copy(start, it, out);
       start = ++it;
--- a/include/fmt/printf.h
+++ b/include/fmt/printf.h
@@ -475,10 +475,8 @@ OutputIt basic_printf_context<OutputIt, Char>::format() {
   auto it = start;
   while (it != end) {
     if (!detail::find<false, Char>(it, end, '%', it)) {
-      it = end;
-    }
-    if (it == end) {
-      continue;
+      it = end;  // detail::find leaves it == nullptr if it doesn't find '%'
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR. Please fix formatting in the PR description and post benchmark results as text, not as image.

include/fmt/printf.h
```diff
@@ -474,8 +474,18 @@ OutputIt basic_printf_context<OutputIt, Char>::format() {
   const Char* end = parse_ctx_.end();
   auto it = start;
   while (it != end) {
+    if (end - it < 32) {
+      // Use a simple linear search instead of memchr for small strings.
+      it = std::find(it, end, '%');
+    } else {
```
vitaut, line 480: I don't think this branch is necessary. `std::find` and `defail::find` will likely provide an equivalent code. I suggest removing it.

vitaut (changes requested):

include/fmt/printf.h
rimathia: Happy to do so.

rimathia:
I have clicked on 'resolve comment' in the hope that this would make the red 'Changes requested' go away, I wasn't completely sure whether this is something I should do or you as the reviewer should do.

include/fmt/printf.h
```diff
@@ -474,8 +474,13 @@ OutputIt basic_printf_context<OutputIt, Char>::format() {
   const Char* end = parse_ctx_.end();
   auto it = start;
   while (it != end) {
+    if (!detail::find<false, Char>(it, end, '%', it)) {
+      it = end;
```
vitaut, line 478: I think this should be `break` and the if below should be removed because if we didn't find `%` there is no point in searching again.

rimathia: Thanks for the catch.

PR #1989, opened 2020-11-06
Make OpenBSD not use fcntl

fagg:

This is a more elegant workaround to the failing test I encountered the other day. I am still not 100% sure why there's a failure with fcntl, there's possibly a difference in behavior that's non-obvious between other systems and OpenBSD. For now, just opt OpenBSD out of using fcntl.


vitaut:
Thanks for the PR but I don't think it's a good idea to disable this functionality just because of a fairly minor test issue.

PR #1990, opened 2020-11-06
Fix build failure when not using fcntl with -Werror

fagg:

If not using fcntl, but building with -Werror enabled, there's a couple functions in the tests that aren't ifdef guarded and don't actually get called. This results in the tests failing to build (log attached) due to unused function warnings.

This PR just adds them to an ifdef guard so the tests build correctly.

[blows_up.txt](https://github.com/fmtlib/fmt/files/5503811/blows_up.txt)

### DIFF
--- a/test/gtest-extra-test.cc
+++ b/test/gtest-extra-test.cc
@@ -22,16 +22,6 @@
 
 namespace {
 
-#if FMT_USE_FCNTL
-// This is used to suppress coverity warnings about untrusted values.
-std::string sanitize(const std::string& s) {
-  std::string result;
-  for (std::string::const_iterator i = s.begin(), end = s.end(); i != end; ++i)
-    result.push_back(static_cast<char>(*i & 0xff));
-  return result;
-}
-#endif
-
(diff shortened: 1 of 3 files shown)


test/gtest-extra-test.cc
```diff
@@ -22,13 +22,15 @@
...
 
+#if FMT_USE_FCNTL
 // This is used to suppress coverity warnings about untrusted values.
 std::string sanitize(const std::string& s) {
   std::string result;
   for (std::string::const_iterator i = s.begin(), end = s.end(); i != end; ++i)
     result.push_back(static_cast<char>(*i & 0xff));
   return result;
 }
+#endif
```
vitaut, line 33: This is no longer used and can be removed completely.

vitaut (changes requested):

test/gtest-extra-test.cc
fagg: Even better, I'll nix it and update.

fagg:
Removed sanitize() as suggested

vitaut:
Thanks

PR #1991, opened 2020-11-07
:art: [ci] Specify the exact version of clang to use

jgopel:
Problem:
- The version of clang to use is specified only as `clang++`. This is
  inconsistent with the specifications for gcc and exposed to unexpected
  failure if the default changes.

Solution:
- Specify the exact version of clang to use. I chose `clang++-9` as that
  is the version that `clang++` is currently resolving to.


vitaut:
Good idea, thanks!

Issue #1987, opened 2020-11-06
Updated github CI only runs against C++11 standard


jgopel:
When switching from Travis to github CI (roughly cba5970c), it seems like the testing for standards other than C++11 was lost.

This issue is relevant because compilers behave differently at different standards. For example, with clang 11.0.0, a pedantic build of fmt succeeds in C++11 mode but fails with the following error in C++20 mode.

```
In file included from /Users/jonathan/work/fmt/test/compile-test.cc:16:
/Users/jonathan/work/fmt/include/fmt/compile.h:518:25: error: implicit conversion changes signedness: 'long' to 'unsigned long' [-Werror,-Wsign-conversion]
  return {f, pos + (end - str.data()) + 1, ctx.next_arg_id()};
                 ~  ~~~~^~~~~~~~~~~~
/Users/jonathan/work/fmt/include/fmt/compile.h:538:31: note: in instantiation of function template specialization 'fmt::v7::detail::parse_specs<int, char>' requested here
      constexpr auto result = parse_specs<id_type>(str, POS + 2, ID);
                              ^
/Users/jonathan/work/fmt/include/fmt/compile.h:569:17: note: in instantiation of function template specialization 'fmt::v7::detail::compile_format_string<fmt::v7::detail::type_list<int>, 0, 0, FMT_COMPILE_STRING>' requested here
        detail::compile_format_string<detail::type_list<Args...>, 0, 0>(
                ^
/Users/jonathan/work/fmt/include/fmt/compile.h:648:37: note: in instantiation of function template specialization 'fmt::v7::detail::compile<int, FMT_COMPILE_STRING, 0>' requested here
  constexpr auto compiled = detail::compile<Args...>(S());
                                    ^
/Users/jonathan/work/fmt/test/compile-test.cc:140:24: note: in instantiation of function template specialization 'fmt::v7::format<FMT_COMPILE_STRING, int, 0>' requested here
  EXPECT_EQ("42", fmt::format(FMT_COMPILE("{:x}"), 0x42));
                       ^
```

I think the correct fix here is to add C++ standards to the build matrix and I would be happy to submit a PR for that work, if there's interest. If that's not palatable then I would like to see the setup should return to the way it was previously with C++11 being tested in g++-4.8, C++14 being tested in gcc-6, and C++17 being tested in gcc-8.

vitaut:
#1969

jgopel:
[...]

PR #1992, opened 2020-11-08
Add C++ standards to CI config (Linux)

alexezeder:
For Linux workflow, I added several C++ standards with excluding unsupported/excess standard for each presented compiler.
This is related to the issue https://github.com/fmtlib/fmt/issues/1987.
Also, I fixed 1 compile warning (-Werror,-Wsign-conversion) for clang++ with C++17 standard, otherwise, CI build fails.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -8,6 +8,7 @@ jobs:
     strategy:
       matrix:
         cxx: [g++-4.8, g++-8, g++-10, clang++-9]
+        standard: [11, 14, 17, 20]
         build_type: [Debug, Release]
         include:
           - cxx: clang++-9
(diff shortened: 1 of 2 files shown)


.github/workflows/linux.yml
```diff
@@ -8,13 +8,49 @@ jobs:
...
+            standard: 14
+          # does not support c++17
+          - cxx: g++-4.8
+            standard: 17
+          # does not support c++20
+          - cxx: g++-4.8
+            standard: 20
+
+          ### gcc 8
+          # has c++14 as default mode
```
darklukee, line 32: Doesn't mean someone won't use c++11 on gcc-8.
If resources aren't a constrain I wouldn't exclude it from checks.

Same comment for other excludes like this one.

alexezeder: I was just impressed by the number of configurations without any filtering.
Maybe I over-filtered them. 
But I don't know if they make any sense, for example, is it possible that c++11 on gcc-8 fails while c++11 on gcc-4.8 doesn't? It can be possible of course, because of some bugs, but if we are hunting for bugs here then this config should have a huge number of configurations.

darklukee: I see your point.
However following this though, standard shouldn't be a matrix parameter.
Check default on each compiler and have separate matrix for all cpp standards tested on newest compilers only.

alexezeder: Ok, this is a good suggestion.

alexezeder: But I think it will be better to create a new PR with the described logic because the logic here and the logic you proposed differ. I can create it, or maybe you want to do this?

jgopel:
A quick note here in case this is preferred over #1999: You may want to consider disabling C++20 support on `gcc-8` builds. `gcc-8` only has partial support for 20 and it will never fully support it. The GCC folks have kindly documented this for us here: https://gcc.gnu.org/projects/cxx-status.html.

alexezeder:
@jgopel hmm... but C++20 on gcc-8 is disabled.

I'm closing this PR. https://github.com/fmtlib/fmt/pull/1999 is better because it has additional fixes.

I think the most important thing in this PR is the idea proposed by @darklukee that standard shouldn't be a matrix parameter

Following this idea I would make a configuration with several jobs inside, like:
```
jobs:
  cxx11:
    ...

  cxx14:
    ...

  cxx17_cxx20: # it's combined because no compilers have c++17 as default standard, better be separated
    ...
```
Where the `cxx11` job should run on ubuntu 16.04 to use the lowest possible compilers: gcc-4.8 (also available on 18.04) and clang-3.5. Maybe instead of this prepared image, this job should use a Docker container with gcc-4.8 and clang-3.0. Why these compilers? Because the [documentation](https://fmt.dev/latest/index.html#portability) says:

> These are available in GCC 4.8, Clang 3.0, MSVC 19.0 (2015) and more recent compiler version. For older compilers use {fmt} version 4.x which is maintained and only requires C++98.

_Off-top: what a shame that GitHub Action doesn't support YAML anchors._

jgopel:
> @jgopel hmm... but C++20 on gcc-8 is disabled.

I totally missed the additional blocks after the gcc-4.8 block - I stand corrected

PR #1999, opened 2020-11-10
Add C++ standards to the build matrix

jgopel:
Problem:
- A variety of compilers are tested in CI, but they are all tested on
  the C++11 standard. This prevents CI from catching bugs that exist
  only on a particular version of the C++ standard.

Solution:
- Parameterize CI on C++ standards 11, 14, 17, and 20.

This also includes fixes for compiler warnings discovered along the way. I've included a detailed writeup of each compilation error and the fix in the commit notes for each fix.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -9,12 +9,22 @@ jobs:
       matrix:
         cxx: [g++-4.8, g++-8, g++-10, clang++-9]
         build_type: [Debug, Release]
+        standard: [11, 14, 17, 20]
         include:
           - cxx: clang++-9
             build_type: Debug
             fuzz: -DFMT_FUZZ=ON -DFMT_FUZZ_LINKMAIN=ON
           - cxx: g++-4.8
             install: sudo apt install g++-4.8
+        exclude:
+          - cxx: g++-4.8
+            standard: 14
+          - cxx: g++-4.8
+            standard: 17
+          - cxx: g++-4.8
+            standard: 20
+          - cxx: g++-8
+            standard: 20
 
     steps:
     - uses: actions/checkout@v2
@@ -30,7 +40,8 @@ jobs:
         CXX: ${{matrix.cxx}}
       run: |
         cmake -DCMAKE_BUILD_TYPE=${{matrix.build_type}} ${{matrix.fuzz}} \
-              -DFMT_DOC=OFF -DFMT_PEDANTIC=ON -DFMT_WERROR=ON $GITHUB_WORKSPACE
+              -DFMT_DOC=OFF -DFMT_PEDANTIC=ON -DFMT_WERROR=ON \
+              -DCMAKE_CXX_STANDARD=${{matrix.standard}} $GITHUB_WORKSPACE
 
     - name: Build
       working-directory: ${{runner.workspace}}/build
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -7,6 +7,7 @@ jobs:
     runs-on: macos-10.15
     strategy:
       matrix:
+        standard: [11, 14, 17, 20]
         build_type: [Debug, Release]
 
     steps:
@@ -19,7 +20,8 @@ jobs:
       working-directory: ${{runner.workspace}}/build
       run: |
         cmake -DCMAKE_BUILD_TYPE=${{matrix.build_type}} \
-              -DFMT_DOC=OFF -DFMT_PEDANTIC=ON -DFMT_WERROR=ON $GITHUB_WORKSPACE
+              -DFMT_DOC=OFF -DFMT_PEDANTIC=ON -DFMT_WERROR=ON \
+              -DCMAKE_CXX_STANDARD=${{matrix.standard}} $GITHUB_WORKSPACE
 
     - name: Build
       working-directory: ${{runner.workspace}}/build


include/fmt/compile.h
```diff
@@ -515,7 +515,7 @@ constexpr parse_specs_result<T, Char> parse_specs(basic_string_view<Char> str,
   auto ctx = basic_format_parse_context<Char>(str, {}, arg_id + 1);
   auto f = formatter<T, Char>();
   auto end = f.parse(ctx);
-  return {f, pos + (end - str.data()) + 1, ctx.next_arg_id()};
+  return {f, pos + static_cast<decltype(pos)>(end - str.data()) + 1, ctx.next_arg_id()};
```
jgopel, line 518: @vitaut I assume this isn't the solution that you want here - I just went with this to have something to start the discussion with. Ultimately, the problem here is that `end` and `str.data()` are both pointer types and the result of subtraction of pointer types is `ptrdiff_t`, which is `signed long` on x86_64. The addition to `pos`, which is of an unsigned type, causes a sign conversion error.

I'm fairly confident that the `ptrdiff_t` would never be a negative type, but there's nothing binding that behavior.

Hopefully that explanation makes sense - in light of that how would you like to proceed here?

jgopel: Here's a minimal repro of the warning if you'd like to play around with it. https://godbolt.org/z/s7Th1f

jgopel:
I wrote a little script that generates a build directory for every possible compiler version & standard version combination. It's nothing fancy, but getting the details right took a few minutes. Would there be any interest in upstreaming that? Maybe something like `support/generate-all-builds.py`?

jgopel:
Looks like this partially duplicates #1992 - not sure how to proceed here.

.github/workflows/linux.yml
```diff
@@ -9,12 +9,22 @@ jobs:
       matrix:
         cxx: [g++-4.8, g++-8, g++-10, clang++-9]
         build_type: [Debug, Release]
+        standard: [11, 14, 17, 20]
```
vitaut, line 12: We should limit the number of configs to make this manageable. 14 is definitely not needed and the rest should be trimmed considerably.

alexezeder:
> Looks like this partially duplicates #1992 - not sure how to proceed here.

Done. You may want to read my final words there.

.github/workflows/linux.yml
jgopel: Can you help me understand what aspects of this you see as unmanageable so that I can arrive at an agreeable solution?

Depending on how thin you'd like to cut down the tested configurations, I'm considering something more whitelist-y than this approach. Roughly:

```
matrix:
  build_type: [Debug, Release]
  include:
    - cxx: gcc-4.8
      standard: 11
    - cxx: gcc-8
      standard: 14
    - cxx: gcc-10
      standard: 17
    - cxx: gcc-10
      standard: 20
    - cxx: clang-9
      standard: 20
```

jgopel: Per https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#example-running-with-more-than-one-operating-system that approach would allow this PR to integrate @alexezeder and @darklukee's suggestion of using varied OSs.

vitaut: > Can you help me understand what aspects of this you see as unmanageable so that I can arrive at an agreeable solution?

Too many configs. We should only add ones that are likely to provide high CI signal, ideally one config per std, not compiler-version * std.

>  I'm considering something more whitelist-y than this approach.

LGTM

include/fmt/compile.h
vitaut: Please use `to_unsigned` instead of a cast.

test/std-format-test.cc
```diff
@@ -102,7 +102,7 @@ template <> struct std::formatter<S> {
   size_t width_arg_id = 0;
 
   // Parses a width argument id in the format { <digit> }.
-  constexpr auto parse(format_parse_context& ctx) {
+  auto parse(format_parse_context& ctx) {
```
vitaut, line 105: What is this for?

Also could you move the fixes to a separate PR and post the warnings they are fixing for future reference?

vitaut (changes requested):
Thanks for the PR. Looks good in general, just a few minor comments.

test/std-format-test.cc
jgopel: Detailed writeup in the commit notes here: https://github.com/fmtlib/fmt/pull/1999/commits/34ea1472eb8dd2463819bdcfb4b81b6f64374cc3

I'll split these all out to individual PRs.

vitaut:
> Would there be any interest in upstreaming that? Maybe something like support/generate-all-builds.py?

Probably not but you can post it here in case anyone is interested.

jgopel:
> > Would there be any interest in upstreaming that? Maybe something like support/generate-all-builds.py?
> 
> Probably not but you can post it here in case anyone is interested.

Here's a gist https://gist.github.com/jgopel/1d9fcdaf1f82c6f0440487cba791ee68

include/fmt/compile.h
jgopel: Handled in #2009

test/std-format-test.cc
jgopel: Created #2008, #2009, and #2010 to resolve the warnings.

vitaut:
Closing since the new standards have been added in https://github.com/fmtlib/fmt/commit/f8640d4050504ea15096c3861925956db40d436a but thanks for the PR.

jgopel:
> Closing since the new standards have been added in [f8640d4](https://github.com/fmtlib/fmt/commit/f8640d4050504ea15096c3861925956db40d436a) but thanks for the PR.

I'm not sure that does what you want it to do, @vitaut. I had experimented with that approach and it didn't generated both debug and release jobs for those standards. See: https://github.com/jgopel/fmt/actions/runs/361646531 and https://github.com/jgopel/fmt/actions/runs/361658036. The only way I've found success so far is with an approach like this https://github.com/jgopel/fmt/actions/runs/363543436.

Note that there's an open issue with GitHub Actions about roughly this problem. You can find that here: https://github.community/t/reusing-sharing-inheriting-steps-between-jobs-declarations/16851

.github/workflows/linux.yml
jgopel: As explained here https://github.com/fmtlib/fmt/pull/1999#issuecomment-727259874, the approach I suggested does not work as expected. The current approach seems to be the only way to achieve the end goal of testing multiple compilers at 1 or more standards in both `Debug` and `Release` mode.

jgopel:
I've updated this to reflect my latest thinking about what this feature should look like - including support for using an older OS as discussed by @alexezeder and @darklukee on #1992. No pressure to merge this, just wanted to document what I've learned and put forth my best attempt at a solution that has all the desired attributes.

vitaut:
> it didn't generated both debug and release jobs for those standards

That's fine.

PR #2001, opened 2020-11-10
fix gcc warning of missing override

killerbot242:


vitaut:
Thank you

Issue #2002, opened 2020-11-10
FMT_ENFORCE_COMPILE_STRING broken in 7.1.2 with format-inl.h


yeswalrus:
Several lines in format-inl.h do not use FMT_STRING, which causes static assertion failures with FMT_ENFORCE_COMPILE_STRING.

```
In file included from external/fmt/src/format.cc:8:
In file included from external/fmt/include/fmt/format-inl.h:28:
In file included from external/fmt/include/fmt/format.h:44:
external/fmt/include/fmt/core.h:522:3: error: static_assert failed due to requirement 'is_compile_string<char [7]>::value' "FMT_ENFORCE_COMPILE_STRING requires all format strings to use FMT_STRING."
  static_assert(is_compile_string<S>::value,
  ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~
external/fmt/include/fmt/core.h:1625:11: note: in instantiation of function template specialization 'fmt::v7::detail::check_format_string<fmt::v7::basic_string_view<char> &, char *&, char [7], 0>' requested here
  detail::check_format_string<Args...>(format_str);
          ^
external/fmt/include/fmt/core.h:2004:28: note: in instantiation of function template specialization 'fmt::v7::make_args_checked<fmt::v7::basic_string_view<char> &, char *&, char [7], char>' requested here
  const auto& vargs = fmt::make_args_checked<Args...>(format_str, args...);
                           ^
external/fmt/include/fmt/format-inl.h:2723:9: note: in instantiation of function template specialization 'fmt::v7::format_to<fmt::v7::detail::buffer_appender<char>, char [7], fmt::v7::basic_string_view<char> &, char *&, true>' requested here
        format_to(detail::buffer_appender<char>(out), "{}: {}", message,

vitaut:
Could you submit a PR to fix those?

yeswalrus:
I'm working on it and have a patch for just those files, but I've discovered a number of other places with issues in chrono.h that are much trickier. Do you know why the format strings there are given in the form of constant char arrays? ie `const char format[] = {'{','}',0};`

vitaut:
> Do you know why the format strings there are given in the form of constant char arrays?

This is to handle the case when `Char` is not `char`.

yeswalrus:
Interesting. Is there any reason it would be bad to replace these with FMT_STRING("...") equivalents?

yeswalrus:
[...]

PR #2006, opened 2020-11-11
Always use FMT_STRING internally where possible [Issue #2002]

yeswalrus:
Update a few internal calls to fmt::format to use FMT_STRING, improving compatibility with the FMT_ENFORCE_COMPILE_STRING compile option.

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -763,29 +763,19 @@ inline std::chrono::duration<Rep, std::milli> get_milliseconds(
 }
 
 template <typename Char, typename Rep, typename OutputIt,
-          FMT_ENABLE_IF(std::is_integral<Rep>::value && std::is_same<Char, char>::value)>
+          FMT_ENABLE_IF(std::is_integral<Rep>::value)>
 OutputIt format_duration_value(OutputIt out, Rep val, int) {
-  return format_to(out, FMT_STRING("{}"), val);
+  static FMT_CONSTEXPR_DECL const Char format[] = {'{', '}', 0};
+  return format_to(out, compile_string_to_view(format), val);
 }
 
 template <typename Char, typename Rep, typename OutputIt,
-          FMT_ENABLE_IF(std::is_integral<Rep>::value && std::is_same<Char, wchar_t>::value)>
-OutputIt format_duration_value(OutputIt out, Rep val, int) {
-  return format_to(out, FMT_STRING(L"{}"), val);
-}
-
-template <typename Char, typename Rep, typename OutputIt,
-          FMT_ENABLE_IF(std::is_floating_point<Rep>::value && std::is_same<Char, char>::value)>
+          FMT_ENABLE_IF(std::is_floating_point<Rep>::value)>
 OutputIt format_duration_value(OutputIt out, Rep val, int precision) {
-  if (precision >= 0) return format_to(out, FMT_STRING("{:.{}f}"), val, precision);
-  return format_to(out, FMT_STRING("{:g}"), val);
-}
-
-template <typename Char, typename Rep, typename OutputIt,
-          FMT_ENABLE_IF(std::is_floating_point<Rep>::value && std::is_same<Char, wchar_t>::value)>
-OutputIt format_duration_value(OutputIt out, Rep val, int precision) {
-  if (precision >= 0) return format_to(out, FMT_STRING("L{:.{}f}"), val, precision);
-  return format_to(out, FMT_STRING("L{:g}"), val);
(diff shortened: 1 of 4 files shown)


include/fmt/color.h
```diff
@@ -578,7 +578,7 @@ inline std::basic_string<Char> vformat(
 */
 template <typename S, typename... Args, typename Char = char_t<S>>
 inline std::basic_string<Char> format(const text_style& ts, const S& format_str,
-                                      const Args&... args) {
+                                      Args&&... args) {
```
vitaut, line 581: This looks unrelated, please revert.

yeswalrus: Done, though this was done since I noticed it would bring the function more into alignment with the other examples of and documentation for functions using `make_args_checked`

include/fmt/chrono.h
```diff
@@ -764,13 +764,20 @@ inline std::chrono::duration<Rep, std::milli> get_milliseconds(
...
 template <typename Char, typename Rep, typename OutputIt>
 OutputIt format_duration_value(OutputIt out, Rep val, int precision) {
+  using context = FMT_BUFFER_CONTEXT(type_identity_t<Char>);
+
   const Char pr_f[] = {'{', ':', '.', '{', '}', 'f', '}', 0};
-  if (precision >= 0) return format_to(out, pr_f, val, precision);
+  if (precision >= 0) {
+    return vformat_to(out, to_string_view(pr_f),
+                      make_format_args<context>(val, precision));
+  }
```
vitaut, line 773: Please make `pr_f``static constexpr` and keep `format_to` (don't change it to `vformat_to`). `static constexpr` strings should work with `FMT_STRING`:

```c++
static constexpr const Char pr_f[] = {'{', ':', '.', '{', '}', 'f', '}', 0};
  if (precision >= 0) return format_to(out, FMT_STRING(pr_f), val, precision);
```

Same below.

vitaut (changes requested):
Thanks for the PR. Looks good in general but please address the inline comment.

include/fmt/chrono.h
yeswalrus: ah, good to know! Won't that break on platforms where FMT_CONSTEXPR_DECL evaluates to noop though?

vitaut: On that platform `FMT_STRING` should also evaluate to noop (at least in theory).

yeswalrus: Nm, tested on godbolt and realized that's a perfectly valid c++11 feature

yeswalrus: This apparently breaks on windows, even after correcting it to use SFINAE to avoid compiling the `{:.{}f}` format string with an integral type. I'm adding a test for using FMT_STRING with a static FMT_CONSTEXPR_DECL, and I'll see if I can get a workaround

yeswalrus:
I know you asked me not to modifiy tests, but I've added a demonstration of a failure to format-test.cc starting at line 1819, and filed a bug report against MSVC++ here: https://developercommunity.visualstudio.com/content/problem/1256077/internal-compiler-failure-with-macro-defining-lamb.html
I've modified chrono.h to use enable_if in a sadly combinatoric way to avoid having to use FMT_STRING on a constexpr char array and to avoid compiling a compile-time checked format string with a bad type.

yeswalrus:
I've simplified by directly calling the underlying function, detail::compile_string_to_view, to work around the MSVC crash. It simplifies things significantly, so LMK which version you'd prefer.

vitaut (changes requested):
Mostly looks good but please address inline comments.

test/format-test.cc
```diff
@@ -1807,11 +1807,27 @@ fmt::string_view to_string_view(string_like) { return "foo"; }
...
+  (void)static_with_null;
+  (void)static_with_null_wide;
+  (void)static_no_null;
+  (void)static_no_null_wide;
+#if !defined(_MSC_VER)
+  EXPECT_EQ("42", fmt::format(FMT_STRING(static_with_null), 42));
+  EXPECT_EQ(L"42", fmt::format(FMT_STRING(static_with_null_wide), 42));
+  EXPECT_EQ("42", fmt::format(FMT_STRING(static_no_null), 42));
+  EXPECT_EQ(L"42", fmt::format(FMT_STRING(static_no_null_wide), 42));
+#endif
```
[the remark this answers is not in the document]
yeswalrus: Not quite - you test with constexpr locals, which are a c++ 17 feature, but static constexpr char arrays are valid all the way back to c++11 and this exposes a bug in MSVC.

vitaut:
Thank you!

PR #2010, opened 2020-11-11
:bug: Cannot call non-constexpr function in constexpr context

jgopel:
Problem:
- gcc-8 gives the following error when compiling this function on all
  standards:
```
    test/std-format-test.cc: In member function 'constexpr auto std::formatter<S>::parse(std::format_parse_context&)':
    test/std-format-test.cc:112:17: error: call to non-'constexpr' function 'int isdigit(int)'
         ~~~~~~~^~~
```

Solution:
- Write a `constexpr` version of `isdigit` for use in this function.

### DIFF
--- a/test/std-format-test.cc
+++ b/test/std-format-test.cc
@@ -103,13 +103,17 @@ template <> struct std::formatter<S> {
 
   // Parses a width argument id in the format { <digit> }.
   constexpr auto parse(format_parse_context& ctx) {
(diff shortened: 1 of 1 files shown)


foonathan:
I think a better solution would be to use `'0' <= c <= '9'` instead, which can keep it `constexpr`.

vitaut:
Thanks

PR #2015, opened 2020-11-12
fix wrong order of expected and actual in truncated_printf_context te…

rimathia:
…st in PrintfDetermineOutputSize, make test of normal sprintf in a separate step

### DIFF
--- a/test/printf-test.cc
+++ b/test/printf-test.cc
@@ -617,12 +617,14 @@ TEST(PrintfTest, PrintfDetermineOutputSize) {
 
   const auto format_string = "%s";
   const auto format_arg = "Hello";
-  const auto expected_size = fmt::sprintf(format_string, format_arg).size();
+  const auto expected_size = 5;
 
-  EXPECT_EQ((truncated_printf_context(
+  EXPECT_EQ(expected_size, fmt::sprintf(format_string, format_arg).size());
+
+  EXPECT_EQ(expected_size,
+            (truncated_printf_context(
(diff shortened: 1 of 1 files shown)


rimathia:
This is admittedly a very pedantic change, but when the test fails "expected" and "actual" are labeled as such in the output.

test/printf-test.cc
```diff
@@ -617,12 +617,14 @@ TEST(PrintfTest, PrintfDetermineOutputSize) {
 
   const auto format_string = "%s";
   const auto format_arg = "Hello";
-  const auto expected_size = fmt::sprintf(format_string, format_arg).size();
+  const auto expected_size = 5;
 
-  EXPECT_EQ((truncated_printf_context(
+  EXPECT_EQ(expected_size, fmt::sprintf(format_string, format_arg).size());
```
vitaut, line 622: This can be removed since `sprintf` is tested elsewhere.

vitaut (changes requested):
Thanks for the PR. GTest has moved away from this order of expected/actual (https://github.com/google/googletest/blob/master/googletest/docs/primer.md) so there is no need to reorder arguments:

> Historical note: Before February 2016 *_EQ had a convention of calling it as ASSERT_EQ(expected, actual), so lots of existing code uses this order. Now *_EQ treats both parameters in the same way.

rimathia:
Because my assumption of one correct place for "actual" and "expected" in the test was wrong this is moot.

Issue #1874, opened 2020-09-16
Should a UDL version of `FMT_STRING` be added?


sth:
Usage of the `FMT_STRING()` macro could be replaced by a user defined literal operator that creates a `fmt::compile_string`.

The advantage would be a more concise syntax and avoiding the "ugly" use of a marco: `fmt::print(FMT_STRING("{}: {}"), a, b)` could be written as `fmt::print("{}: {}"_fmt, a, b)`. The disadvantage would be increased API complexity by having this additional option to create a format string and possible compatibility questions for compilers without sufficient UDL support.

I know there is a `""_format()` UDL, but as I understand it that replaces the whole `fmt::format()` function and can't be used together with other functions like `fmt::print()` or `fmt::format_to()`. A UDL that just creates a `fmt::compile_string` can be used more generically in all places where a format string is expected.

It can be implemented like this:

```c++
template <typename Char, Char... CHARS>
struct udl_compile_string : fmt::compile_string {
   using char_type = Char;

   static FMT_CONSTEXPR_DECL char_type s[] = {CHARS..., '\0'};

   FMT_MAYBE_UNUSED FMT_CONSTEXPR
   operator fmt::basic_string_view<char_type>() const {
      return fmt::detail::compile_string_to_view<char_type>(s);
   }
};

template <typename Char, Char... CHARS>
FMT_CONSTEXPR udl_compile_string<Char, CHARS...> operator""_fmt() {
   return {};
}
```

vitaut:
Do you mean `FMT_STRING`?

sth:
Yes that's what I mean... I edited it.

vitaut:
Yes, a UDL would be better than a macro. Unfortunately it will only be implementable in a portable and efficient(ish) way in C++20.

kitegi:
here's a c++20 implementation https://godbolt.org/z/zYfErb

though this is currently only supported on gcc 10+, as far as i know

vitaut:
Closing in favor of #2022. We can have compile-time checks without wrapping a format string in a UDL.

PR #2019, opened 2020-11-14
Basics of formatting at compile-time based on compile-time API

alexezeder:
I need to say that this PR does **not** introduce fully-functioning formatting via {fmt} at compile-time. By fully-functioning I mean that I handled only a few cases (bool, integer, string) of the entire functionality of this library. 

This PR based on compile-time API from `compile.h`. This forces end-users to use `FMT_COMPILE` macro in cases when formatting should happen at compile-time.

- new compile-time formatting tests added into `compile-test` test case, available only if C++ standard is greater or equal to C++20
- new `FMT_CONSTEXPR20` macro definition added
- new `is_constant_evaluated()` function in `fmt::details` namespace added, that invokes `std::is_constant_evaluated()` if std library has it, otherwise falls back to returning `false`

What functionality of the library is not available at compile-time with this PR:

- pointers formatting - `std::bit_cast` is not available yet, so it's not possible yet
- floating-point numbers formatting - looks like it's possible, but takes some time
- format specifications (padding, precision, etc.) - haven't tried, takes some time
- colors, locales, other functionality I'm not aware of - haven't tried, takes some time


alexezeder:
Since the current CI configuration lacks C++ standards other than C++11 and I was eager to know if it compiles with standards other than C++11, I prepared a special branch to check it on CI
https://github.com/alexezeder/fmt/actions/runs/363499597 (looks like g++-10 does not support C++20, so ignore it)

include/fmt/format.h
```diff
@@ -278,9 +278,879 @@ FMT_END_NAMESPACE
 #  define FMT_DEPRECATED_NUMERIC_ALIGN 0
 #endif
 
+#define FMT_POWERS_OF_10(factor)                                             \
+  factor * 10, (factor)*100, (factor)*1000, (factor)*10000, (factor)*100000, \
+      (factor)*1000000, (factor)*10000000, (factor)*100000000,               \
+      (factor)*1000000000
```
vitaut, line 284: Why turn this and other data into macros?

test/CMakeLists.txt
```diff
@@ -146,6 +146,15 @@ else ()
...
 endif ()
 
+if (${CMAKE_CXX_STANDARD} GREATER_EQUAL "20")
+  add_fmt_executable(compile-time-formatting-test
+      compile-time-formatting-test.cc test-main.cc)
+  target_link_libraries(compile-time-formatting-test gmock)
+  target_include_directories(compile-time-formatting-test SYSTEM PUBLIC gtest gmock)
+  target_link_libraries(compile-time-formatting-test fmt::fmt-header-only)
+  add_test(NAME compile-time-formatting-test COMMAND compile-time-formatting-test)
+endif ()
```
vitaut, line 156: This is not needed, just add new test cases to `compile-test`.

include/fmt/format.h
alexezeder: Maybe I overdid it, but I think that some more entities from there will be _exposed_  to use them at compile-time later. Like data for floating-point formatting, hex and maybe other.

vitaut: FP data won't need to move, so please revert those changes.

alexezeder: Done, but some of them will return to `format.h` later.

test/CMakeLists.txt
alexezeder: Done

alexezeder:
Looks like even gcc-10 works now. Not sure why didn't it work for the first time...

include/fmt/core.h
```diff
@@ -96,6 +96,20 @@
 #  define FMT_CONSTEXPR_DECL
 #endif
 
+#if __cplusplus >= 201402L
+#  define FMT_CONSTEXPR14 constexpr
+#else
+#  define FMT_CONSTEXPR14
+#endif
```
vitaut, line 103: `FMT_CONSTEXPR` is only enabled on C++14 so you don't need a separate `FMT_CONSTEXPR14`. Please remove.

include/fmt/core.h
```diff
@@ -96,6 +96,20 @@
...
+
+#if __cplusplus >= 202002L
+#  define FMT_CONSTEXPR20 constexpr
+#  define FMT_IS_CONSTANT_EVALUATED std::is_constant_evaluated()
```
vitaut, line 107: This should be a function rather than a macro.

include/fmt/format.h
```diff
@@ -278,9 +278,41 @@ FMT_END_NAMESPACE
...
+        {'6', '6'}, {'6', '7'}, {'6', '8'}, {'6', '9'}, {'7', '0'},         \
+        {'7', '1'}, {'7', '2'}, {'7', '3'}, {'7', '4'}, {'7', '5'},         \
+        {'7', '6'}, {'7', '7'}, {'7', '8'}, {'7', '9'}, {'8', '0'},         \
+        {'8', '1'}, {'8', '2'}, {'8', '3'}, {'8', '4'}, {'8', '5'},         \
+        {'8', '6'}, {'8', '7'}, {'8', '8'}, {'8', '9'}, {'9', '0'},         \
+        {'9', '1'}, {'9', '2'}, {'9', '3'}, {'9', '4'}, {'9', '5'},         \
+        {'9', '6'}, {'9', '7'}, {'9', '8'}, {                               \
+      '9', '9'                                                              \
+    }                                                                       \
+  }
```
vitaut, line 305: Can we avoid a macro?

include/fmt/format.h
```diff
@@ -278,9 +278,41 @@ FMT_END_NAMESPACE
...
+
 FMT_BEGIN_NAMESPACE
 namespace detail {
 
+using data_digit_pair = char[2];
+
+namespace compile_time_formatting_data {
+constexpr data_digit_pair digits[] = FMT_DATA_DIGITS;
+}
```
vitaut, line 314: Why move this out of `basic_data`?

include/fmt/format.h
```diff
@@ -390,7 +422,8 @@ inline buffer_appender<T> reserve(buffer_appender<T> it, size_t n) {
   return it;
 }
 
-template <typename Iterator> inline Iterator& reserve(Iterator& it, size_t) {
+template <typename Iterator>
+inline constexpr Iterator& reserve(Iterator& it, size_t) {
```
vitaut, line 426: Please revert (apply clang-format).

include/fmt/format.h
```diff
@@ -967,10 +994,27 @@ inline int count_digits(uint64_t n) {
     count += 4;
   }
 }
+
+#if defined(FMT_BUILTIN_CLZLL)
```
vitaut, line 998: Please avoid unnecessary changes.

vitaut (changes requested):

include/fmt/format.h
alexezeder: > Why move this out of basic_data?
> Can we avoid a macro?

I need this value at compile time as well as the entire library needs it at runtime, so to avoid having two separate definitions I added a macro definition for that. This allows me to use exactly the same value in `basic_data` and in-place. At least that was the idea, but I just don't know what is better - to create a global variable in some namespace (as right now) or to create a variable in every function that needed it.
But here is a direct answer to your question: basic data is not available at compile-time, because it is declared but not defined, only `format-inl.h` sets values in it, wherever these values defined in the static library or they defined later if {fmt} is header-only.

include/fmt/format.h
alexezeder: If you mean incorrect code-style here then it's just how clang-format works for this line, the line exceeded the column limit, that's all, I can revert this change, but it's strange... Otherwise, you could mind `constexpr` here which is necessary. Please explain what do you mean.

include/fmt/format.h
alexezeder: I'm sorry but this is necessary. Maybe there are not enough tests for that (like test with `uint32_t` and `uint64_t` values), but I can add them. It's just unable to work without `constexpr` `count_digits`.

alexezeder: Actually, I _will_ add them anyway.

include/fmt/core.h
alexezeder: Done

include/fmt/core.h
alexezeder: Done.
I think macro was better here because it's just way simpler than a function.

include/fmt/format.h
alexezeder: Test with uint32_t and uint64_t values added.

include/fmt/core.h
```diff
@@ -279,6 +290,14 @@ struct monostate {};
 
 namespace detail {
 
+constexpr bool is_constant_evaluated() FMT_DETECTED_NOEXCEPT {
+#if FMT_HAS_IS_CONSTANT_EVALUATED
```
vitaut, line 294: This macro is not needed, please check `__cpp_lib_is_constant_evaluated` instead.

include/fmt/format.h
vitaut: I think it should be enough to make `digits` `constexpr` and move the definition to `format.h`.

include/fmt/format.h
vitaut: Nevermind, I see you added `constexpr` which is why the formatting changed.

include/fmt/format.h
vitaut: Changing `#ifdef` to `#if defined(FMT_BUILTIN_CLZLL)` is unnecessary. Same elsewhere.

vitaut (changes requested):

include/fmt/format.h
alexezeder: Ok, got it

include/fmt/format.h
alexezeder: Resolved 🙂

include/fmt/format.h
alexezeder: This is surely out of the scope of this PR, but moving `digits` to `format.h` is only one small piece. When someone decides to format `int` as hex then looks like `hex_digits` is used (conditionally 😒). With output coloring - `foreground_color`, `background_color`, etc. With padding - `left_padding_shifts` and `right_padding_shifts`. Needless to say, the floating-point formatting uses this data.

Anyway, I will move `digits` to `format.h` and make it `constexpr`.

alexezeder: Also, this change will probably break ABI (I remember some issue about that recently), but it can be bypassed by leaving `digits` in data inside `format-inl.h`, but initializing it by `digits` inside `format.h`.

vitaut: > Also, this change will probably break ABI (I remember some issue about that recently), but it can be bypassed by leaving digits in data inside format-inl.h, but initializing it by newly created digits inside format.h.

Good point. We should definitely keep `basic_data::digits` as is.

Thinking more of it, I don't think we need `constexpr` version of `digits` at all because it's an optimization that doesn't make sense at compile time. Instead I suggest changing the implementation of `format_decimal` with a separate path guarded by `is_constant_evaluated` that does simple decimal to string conversion that doesn't require any data tables.

vitaut (changes requested):

include/fmt/format.h
```diff
@@ -943,17 +969,7 @@ FMT_EXTERN template struct basic_data<void>;
...
-// except for n == 0 in which case count_digits returns 1.
-inline int count_digits(uint64_t n) {
-  // https://github.com/fmtlib/format-benchmark/blob/master/digits10
-  auto t = bsr2log10(FMT_BUILTIN_CLZLL(n | 1) ^ 63);
-  return t - (n < data::zero_or_powers_of_10_64_new[t]);
-}
-#else
-// Fallback version of count_digits used when __builtin_clz is not available.
-inline int count_digits(uint64_t n) {
+template <typename T> FMT_CONSTEXPR int count_digits_trivial(T n) {
```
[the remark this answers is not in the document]
alexezeder: But it's used both for `uint32_t` and `uint64_t`, not for `uint128_t` yet only because I was stopped by upper-case literal `U` there... looks like it has the same meaning as `u` literal, so even `uint128_t` can use this function. That's why I have template there.

include/fmt/format.h
```diff
@@ -908,9 +937,6 @@ template <typename T = void> struct FMT_EXTERN_TEMPLATE_API basic_data {
   static const uint64_t powers_of_5_64[];
   static const uint32_t dragonbox_pow10_recovery_errors[];
 #endif
-  // GCC generates slightly better code for pairs than chars.
-  using digit_pair = char[2];
-  static const digit_pair digits[];
```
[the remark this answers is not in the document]
alexezeder: Looks like it's already outdated 🙂

include/fmt/core.h
```diff
@@ -96,6 +96,12 @@
 #  define FMT_CONSTEXPR_DECL
 #endif
 
+#if __cplusplus >= 202002L
+#  define FMT_CONSTEXPR20 constexpr
+#else
+#  define FMT_CONSTEXPR20
```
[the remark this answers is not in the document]
alexezeder: Done

include/fmt/format.h
```diff
@@ -390,7 +414,8 @@ inline buffer_appender<T> reserve(buffer_appender<T> it, size_t n) {
   return it;
 }
 
-template <typename Iterator> inline Iterator& reserve(Iterator& it, size_t) {
+template <typename Iterator>
+inline constexpr Iterator& reserve(Iterator& it, size_t) {
```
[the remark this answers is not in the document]
alexezeder: Done, also for `base_iterator()`

include/fmt/format.h
```diff
@@ -587,14 +612,18 @@ using needs_conversion = bool_constant<
...
-OutputIt copy_str(InputIt begin, InputIt end, OutputIt it) {
+FMT_CONSTEXPR OutputIt copy_str(InputIt begin, InputIt end, OutputIt it) {
   while (begin != end) *it++ = *begin++;
   return it;
 }
 
 template <typename OutChar, typename InputIt,
           FMT_ENABLE_IF(!needs_conversion<InputIt, OutChar>::value)>
-inline OutChar* copy_str(InputIt begin, InputIt end, OutChar* out) {
+inline FMT_CONSTEXPR20 OutChar* copy_str(InputIt begin, InputIt end,
```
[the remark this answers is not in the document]
alexezeder: Done, for all usages of `FMT_CONSTEXPR20`

alexezeder:
🤨 I don't know what exactly happened to `doc` build, but it's was not able to get some dependencies and it's the only one that failed.

test/compile-test.cc
```diff
@@ -171,3 +174,97 @@ TEST(CompileTest, TextAndArg) {
...
+    func(buffer);
+    return *this;
+  }
+
+  template <typename T> constexpr bool operator==(const T& rhs) const noexcept {
+    return (std::string_view(rhs).compare(buffer.data()) == 0);
+  }
+
+  std::array<char, max_string_length> buffer{};
+};
```
vitaut, line 191: Can we turn this into a consteval formatting function that returns an array to simplify tests?

Something like:

```c++
template <size_t max_string_length, typename... Args>
std::array<char, max_string_length> test_format(const char* format_str, const Args&... args) {
  ...
}
```

vitaut (commented):
Looks great, just one small suggestion to simplify tests.

include/fmt/format.h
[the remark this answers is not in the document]
alexezeder: What about `template` here, should I expand usage of `count_digits_fallback` to `count_digits` for `uint128_t`?

test/compile-test.cc
alexezeder: We could, but in that case, `FMT_COMPILE` should be outside of this `test_format` function, so every call would look like:
```cpp
constexpr auto result = test_format<5>(FMT_COMPILE("{}"), true);
```
Otherwise, I'm unable to create a compiled format from the `format_str` argument of `test_format`, it should have _baked_ string inside to be passed just as the type that can be created inside `compile.h` functions:
https://github.com/fmtlib/fmt/blob/b5dac0f0f8a112fba7b01165c792cd2aa8659c05/include/fmt/compile.h#L665-L668
But function argument is not enough because it's neither a string literal to use macro there nor a template parameter.

alexezeder: But my attempts to find any solution (before I realized why it's impossible) ended up with some sort of replacement for `FMT_COMPILE` which is only available with C++20 of course (only trunk of Clang also): https://godbolt.org/z/of9j3s

And yes, even with this replacement of `FMT_COMPILE` the usage in our situation would look like:
```cpp
constexpr auto result = test_format<5>("{}"_cf, true);
```

vitaut: > FMT_COMPILE should be outside of this test_format function

Sure.

alexezeder: > Can we turn this into a consteval formatting function that returns an array to simplify tests?

Not exactly `array` since a comparison operator is needed for GTest, but usage is simpler now

test/compile-test.cc
```diff
@@ -171,3 +174,71 @@ TEST(CompileTest, TextAndArg) {
...
+consteval auto test_format(auto format, const Args&... args) {
+  test_string<max_string_length> string{};
+  fmt::format_to(string.buffer.data(), format, args...);
+  return string;
+}
+
+TEST(CompileTimeFormattingTest, Bool) {
+  {
+    constexpr auto result = test_format<5>(FMT_COMPILE("{}"), true);
+    EXPECT_EQ(result, "true");
```
vitaut, line 197: Now that `test_format` is consteval we don't need to store the result in a constexpr variable. This can be just 

```c++
    EXPECT_EQ(test_format<5>(FMT_COMPILE("{}"), true), "true");
```

And the blocks can be removed.

alexezeder: Actually, it doesn't work because of `EXPECT_EQ` internals, for example, this code on Clang produces:
```
test/compile-test.cc:196:30: error: lambda expression in an unevaluated operand
    EXPECT_EQ(test_format<5>(FMT_COMPILE("{}"), true), "true");
                             ^
include/fmt/compile.h:38:24: note: expanded from macro 'FMT_COMPILE'
                       ^
include/fmt/format.h:3246:3: note: expanded from macro 'FMT_STRING_IMPL'
  [] {                                                                     \
```

alexezeder: But code with `_cf` literal probably will 😏

vitaut: OK, let's keep as is then.

test/compile-test.cc
```diff
@@ -171,3 +174,71 @@ TEST(CompileTest, TextAndArg) {
...
+  }
+  {
+    constexpr auto result =
+        test_format<17>(FMT_COMPILE("{} is {}"), "The answer", "42");
+    EXPECT_EQ(result, "The answer is 42");
+  }
+}
+
+TEST(CompileTimeFormattingTest, Combination) {
+  {
```
vitaut, line 238: This block is redundant.

alexezeder: You mean `Combination` subcase (or whatever it is)?

vitaut: The latter (`{}` is redundant).

alexezeder: Done

alexezeder:
@vitaut I also remind you that there is a question from me without an answer (https://github.com/fmtlib/fmt/pull/2019#discussion_r532106402), this is why I didn't resolve the thread - I fulfilled only one of your points. `count_digits_fallback` still has template argument.

test/compile-test.cc
alexezeder: Apparently, something like `_cf` was mentioned before https://github.com/fmtlib/fmt/issues/1874#issuecomment-716097084 and https://github.com/fmtlib/fmt/issues/613#issuecomment-491906397, and it works also on GCC 10. The `FMT_COMPILE` (not `FMT_STRING` since format can be checked with https://github.com/fmtlib/fmt/issues/2022) replacement will be my next PR probably if no one will do that 😏

include/fmt/format.h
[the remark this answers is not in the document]
vitaut: Let's keep it as a template since it's just a fallback.

vitaut:
Merged, thanks!

alexezeder:
Since we have the base now, I will try to cover the remaining functionality of the library by following PRs.

PR #2020, opened 2020-11-14
:new: [CI] Test with C++14 in Windows 2019

jgopel:
Problem:
- Both Windows builds test C++14

Solution:
- Use C++11 for `windows-2016` builds and C++14 for `windows-2019`
  builds.


vitaut:
Thank you!

Issue #1981, opened 2020-11-05
Add support for std::byte to fmt::join


hellow554:
There is the [`std::byte`](https://en.cppreference.com/w/cpp/types/byte) type since C++17.

It would be nice to have a built-in way to format them properly.

My current ad-hoc solution is

```cpp
template<> struct fmt::formatter<std::byte> : fmt::formatter<int> {
    template<class FormatContext> auto format(const std::byte &b, FormatContext &ctx) {
        return fmt::formatter<int>::format(std::to_integer<int>(b), ctx);
    }
};
```

vitaut:
`std::byte` is already supported:

```c++
#include <fmt/core.h>

int main() {
  fmt::print("{}", std::byte(42));
}
```
prints `42` (https://godbolt.org/z/MKccMT).

hellow554:
My bad:

https://godbolt.org/z/f3oTno

```cpp
#include <cstddef>
#include <vector>

#include <fmt/core.h>
#include <fmt/format.h>

int main() {
    std::vector<std::byte> vv;

    fmt::print("{}", fmt::join(vv, ","));
}
```

doesn't work

MikeGitb:
Could this be related to https://github.com/fmtlib/fmt/issues/2014?

Btw.: I'd find it much more reasonable, if std::byte got formatted as hex by default (ideally with leading zeros if necessary) - it is supposed to represent raw memory after all, not an integral number.

hellow554:
@MikeGitb it was an example to get it working for me. You can of course use `{:x}`, but having that as default? Hmm, not sure about it.

MikeGitb:
@hellow554 : That was more meant for @vitaut, referring to the result of `  fmt::print("{}", std::byte(42));`  not about your workaournd for `container<std::byte>`, which of course should be consistent with formatting of a single `std::byte`. but I guess thats off topic for the problem at hand.

vitaut:
Added a `formatter` specialization for `std::byte` in https://github.com/fmtlib/fmt/commit/4a6eadbde0619b249bde5e019d69e3e8afb02260, so it should work with `join` and other APIs. Thanks for reporting.

phprus:
Without a `#include <cstddef>`, there will be a compilation error in GCC-7:
```
include/fmt/format.h:3556:20: error: ‘byte’ is not a member of ‘std’
 FMT_FORMAT_AS(std::byte, unsigned);
```

PR #2034, opened 2020-11-25
Fix join for specific types (#1981)

kamibo:
Try to apply the same rules for iterator value type (join) as for
regular types.

Try to convert iterator value type to a formattable one.
For instance if the join operation refers to an enum class iterator
(such as std::byte) then try to use the enum class underlying type.

Also try to use a fallback formatter if available.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -3674,21 +3674,39 @@ struct arg_join : detail::view {
       : begin(b), end(e), sep(s) {}
 };
 
+template <typename Char>
+using iterator_arg_mapper = detail::arg_mapper<FMT_BUFFER_CONTEXT(Char)>;
+
+template <typename It, typename Char>
+using iterator_format_type =
+    decltype(std::declval<iterator_arg_mapper<Char>>().map(
+        std::declval<typename std::iterator_traits<It>::value_type>()));
+
+template <typename It, typename Char>
+using iterator_formatter = conditional_t<
+    has_formatter<iterator_format_type<It, Char>,
+                  FMT_BUFFER_CONTEXT(Char)>::value,
+    formatter<iterator_format_type<It, Char>, Char>,
+    detail::fallback_formatter<iterator_format_type<It, Char>, Char>>;
+
 template <typename It, typename Sentinel, typename Char>
 struct formatter<arg_join<It, Sentinel, Char>, Char>
-    : formatter<typename std::iterator_traits<It>::value_type, Char> {
+    : iterator_formatter<It, Char> {
   template <typename FormatContext>
   auto format(const arg_join<It, Sentinel, Char>& value, FormatContext& ctx)
       -> decltype(ctx.out()) {
-    using base = formatter<typename std::iterator_traits<It>::value_type, Char>;
+    using base = iterator_formatter<It, Char>;
+
+    iterator_arg_mapper<Char> mapper;
+
     auto it = value.begin;
     auto out = ctx.out();
     if (it != value.end) {
-      out = base::format(*it++, ctx);
+      out = base::format(mapper.map(*it++), ctx);
       while (it != value.end) {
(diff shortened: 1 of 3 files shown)


include/fmt/format.h
```diff
@@ -3674,21 +3674,39 @@ struct arg_join : detail::view {
...
+template <typename Char>
+using iterator_arg_mapper = detail::arg_mapper<FMT_BUFFER_CONTEXT(Char)>;
+
+template <typename It, typename Char>
+using iterator_format_type =
+    decltype(std::declval<iterator_arg_mapper<Char>>().map(
+        std::declval<typename std::iterator_traits<It>::value_type>()));
+
+template <typename It, typename Char>
+using iterator_formatter = conditional_t<
```
kamibo, line 3686: This is definitely not so nice.
Also TBH I'm not sure that using `FMT_BUFFER_CONTEXT` is correct here.

vitaut (changes requested):
Thanks for the PR.

You shouldn't be changing the `formatter` specialization for `arg_join`. Instead it's enough to add a formatter specialization for byte using `FORMAT_AS` here: https://github.com/fmtlib/fmt/blob/master/include/fmt/format.h#L3521 and add a cast in the implementation of `FORMAT_AS`.

kamibo:
I agree that it would fix the `std::byte` issue.
However it would not fix "similar issues".
Don't you think it would be nice to treat range value type the same way as "simple type"?
Ok let's take the `std::byte` issue from #1981, this is what I've understood so far:

`fmt::print("{}", std::byte(42));`

This works because of the conversion done here https://github.com/fmtlib/fmt/blob/3f4839ce3d72f197cca62a43df3e92080524cf31/include/fmt/core.h#L1229

However

```
std::vector<std::byte> vv;

fmt::print("{}", fmt::join(vv, ","));
```
This does not work because no conversion occurs.
Here is the type used https://github.com/fmtlib/fmt/blob/3f4839ce3d72f197cca62a43df3e92080524cf31/include/fmt/format.h#L3679

Am I missing something?

So I tried to apply the same conversion.
I agree that while it works, it is not satisfying but it solves "all the cases".

vitaut:
Conversion is just an implementation detail. The current way of detecting if type is formattable is to check `has_formatter` which in turn checks `formatter` specializations. I don't think it's necessary to change that. `formatter` specialization for `byte` is useful anyway and adding it makes `byte` work with `join` and any other adapter.

kamibo:
I'm sorry but I feel quiet confused.
The issue is beyond `std::byte`.
It does not work regarding conversion issue and also it is not work with any  "fallback formatter" (ostream).
For instance, in the code I work on, we have this:

```c++
struct RouteWithEvents
{
   ...
};

inline std::ostream& operator<<(std::ostream& os, const RouteWithEvents& route_with_events)
{
  return os << fmt::format("RouteWithEvents(Route: {}: {})", route_with_events.route, route_with_events.events_on_route);
}

using RoutesWithEvents = std::vector<RouteWithEvents>;

inline std::ostream& operator<<(std::ostream& os, const RoutesWithEvents& routes_with_events)
{
  return os << fmt::format("RoutesWithEvents({})", fmt::join(routes_with_events, ", "));
} 
```
Unless I missed something, this just doesn't work since 6.0.0 version while it works nicely with the previous version (5.3.0).
Do you consider this is not an issue?

vitaut:
> Do you consider this is not an issue?

I'm OK with adding support for fallback formatter to `join` but it should be done via `has_fallback_formatter`.

vitaut:
Closing this in favor of `std::byte`-specific change: https://github.com/fmtlib/fmt/commit/4a6eadbde0619b249bde5e019d69e3e8afb02260.

kamibo:
Ok Thanks for your  answer @vitaut!

PR #2037, opened 2020-11-30
Add print overload accepting locale to locale.h

yeswalrus:
Nothing complicated, just fills a coverage gap in the API by providing fmt::print overloads in locale.h. Feel free to reject if this feature is undesired.


vitaut:
Thanks for the PR but I don't think these are worth adding to the core library.

PR #2038, opened 2020-11-30
Add tests for FMT_ENFORCE_COMPILE_STRING, fix several errors

yeswalrus:
Though all format strings will be compile time checked by default in c++20, FMT_ENFORCE_COMPILE_STRING will still be useful for users stuck on older c++ versions. 

This PR adds a test file with API coverage taken from other existing tests, uncovering 4 places where FMT_ENFORCE_COMPILE_STRING caused unexpected breakages. It additionally resolves those errors, allowing the test to pass.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -165,13 +165,10 @@ functions in their ``formatter`` specializations.
 
 .. _udt:
 
-Enforcing Compile-time Format Checks
-------------------------------------
-To force the use of Compile-time checks, compile with 
-``FMT_ENFORCE_COMPILE_STRING`` defined. When this preprocessor definition is 
-given, functions accepting ``FMT_STRING`` will fail to compile with regular 
-strings. When this is set, runtime-checked formatting is still possible using
-``fmt::vformat``, ``fmt::vprint``, etc.
+To force the use of compile-time checks, define the preprocessor variable
+``FMT_ENFORCE_COMPILE_STRING``. When set, functions accepting ``FMT_STRING``
+will fail to compile with regular strings. Runtime-checked
+formatting is still possible using ``fmt::vformat``, ``fmt::vprint``, etc.
 
 Formatting User-defined Types
 -----------------------------
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -165,6 +165,14 @@ functions in their ``formatter`` specializations.
 
 .. _udt:
 
+Enforcing Compile-time Format Checks
+------------------------------------
+To force the use of Compile-time checks, compile with 
+``FMT_ENFORCE_COMPILE_STRING`` defined. When this preprocessor definition is 
+given, functions accepting ``FMT_STRING`` will fail to compile with regular 
+strings. When this is set, runtime-checked formatting is still possible using
+``fmt::vformat``, ``fmt::vprint``, etc.
+
 Formatting User-defined Types
 -----------------------------
 
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -769,7 +769,7 @@ template <typename Char, typename Rep, typename OutputIt,
           FMT_ENABLE_IF(std::is_integral<Rep>::value)>
 OutputIt format_duration_value(OutputIt out, Rep val, int) {
   static FMT_CONSTEXPR_DECL const Char format[] = {'{', '}', 0};
-  return format_to(out, compile_string_to_view(format), val);
+  return format_to(out, FMT_STRING(format), val);
 }
 
 template <typename Char, typename Rep, typename OutputIt,
@@ -778,9 +778,9 @@ OutputIt format_duration_value(OutputIt out, Rep val, int precision) {
   static FMT_CONSTEXPR_DECL const Char pr_f[] = {'{', ':', '.', '{',
                                                  '}', 'f', '}', 0};
   if (precision >= 0)
-    return format_to(out, compile_string_to_view(pr_f), val, precision);
+    return format_to(out, FMT_STRING(pr_f), val, precision);
   static FMT_CONSTEXPR_DECL const Char fp_f[] = {'{', ':', 'g', '}', 0};
-  return format_to(out, compile_string_to_view(fp_f), val);
+  return format_to(out, FMT_STRING(fp_f), val);
 }
 
 template <typename Char, typename OutputIt>
@@ -802,10 +802,10 @@ OutputIt format_duration_unit(OutputIt out) {
     return copy_unit(string_view(unit), out, Char());
   static FMT_CONSTEXPR_DECL const Char num_f[] = {'[', '{', '}', ']', 's', 0};
   if (const_check(Period::den == 1))
-    return format_to(out, compile_string_to_view(num_f), Period::num);
+    return format_to(out, FMT_STRING(num_f), Period::num);
   static FMT_CONSTEXPR_DECL const Char num_def_f[] = {'[', '{', '}', '/', '{',
                                                       '}', ']', 's', 0};
-  return format_to(out, compile_string_to_view(num_def_f), Period::num,
+  return format_to(out, FMT_STRING(num_def_f), Period::num,
                    Period::den);
 }
 
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -798,19 +798,17 @@ OutputIt format_duration_unit(OutputIt out) {
   if (const char* unit = get_units<Period>())
     return copy_unit(string_view(unit), out, Char());
 
-  basic_memory_buffer<Char> buffer;
-  auto bufOut = std::back_inserter(buffer);
-  *bufOut++ = '[';
-  bufOut = write<Char>(bufOut, Period::num);
+  *out++ = '[';
+  out = write<Char>(out, Period::num);
 
   if (const_check(Period::den != 1)) {
-    *bufOut++ = '/';
-    bufOut = write<Char>(bufOut, Period::den);
+    *out++ = '/';
+    out = write<Char>(out, Period::den);
   }
 
-  *bufOut++ = ']';
-  *bufOut++ = 's';
-  return write<Char>(out, {buffer.data(), buffer.size()});
+  *out++ = ']';
+  *out++ = 's';
+  return out;
 }
 
 template <typename FormatContext, typename OutputIt, typename Rep,
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -770,7 +770,8 @@ template <typename Char, typename Rep, typename OutputIt,
 OutputIt format_duration_value(OutputIt out, Rep val, int) {
   static FMT_CONSTEXPR_DECL const Char format[] = {'{', '}', 0};
 
-#if defined(FMT_MSC_VER) && FMT_MSC_VER < 1928
+// Workaround a compiler error in MSVC < 16.8.2
+#if defined(FMT_MSC_VER) && FMT_MSC_VER <= 1928
   return vformat_to(out, to_string_view(format),
                     make_format_args<buffer_context<Char>>(val));
 #else
@@ -784,7 +785,7 @@ OutputIt format_duration_value(OutputIt out, Rep val, int precision) {
   static FMT_CONSTEXPR_DECL const Char pr_f[] = {'{', ':', '.', '{',
                                                  '}', 'f', '}', 0};
   if (precision >= 0) {
-#if defined(FMT_MSC_VER) && FMT_MSC_VER < 1928
+#if defined(FMT_MSC_VER) && FMT_MSC_VER <= 1928
     return vformat_to(out, to_string_view(pr_f),
                       make_format_args<buffer_context<Char>>(val, precision));
 #else
@@ -793,7 +794,7 @@ OutputIt format_duration_value(OutputIt out, Rep val, int precision) {
   }
   static FMT_CONSTEXPR_DECL const Char fp_f[] = {'{', ':', 'g', '}', 0};
 
-#if defined(FMT_MSC_VER) && FMT_MSC_VER < 1928
+#if defined(FMT_MSC_VER) && FMT_MSC_VER <= 1928
   return vformat_to(out, to_string_view(fp_f),
                     make_format_args<buffer_context<Char>>(val));
 #else
@@ -820,7 +821,7 @@ OutputIt format_duration_unit(OutputIt out) {
     return copy_unit(string_view(unit), out, Char());
   static FMT_CONSTEXPR_DECL const Char num_f[] = {'[', '{', '}', ']', 's', 0};
   if (const_check(Period::den == 1)) {
-#if defined(FMT_MSC_VER) && FMT_MSC_VER < 1928
+#if defined(FMT_MSC_VER) && FMT_MSC_VER <= 1928
     return vformat_to(out, to_string_view(num_f),
                       make_format_args<buffer_context<Char>>(Period::num));
 #else
@@ -829,7 +830,7 @@ OutputIt format_duration_unit(OutputIt out) {
   }
   static FMT_CONSTEXPR_DECL const Char num_def_f[] = {'[', '{', '}', '/', '{',
                                                       '}', ']', 's', 0};
-#if defined(FMT_MSC_VER) && FMT_MSC_VER < 1928
+#if defined(FMT_MSC_VER) && FMT_MSC_VER <= 1928
   return vformat_to(
       out, to_string_view(num_def_f),
       make_format_args<buffer_context<Char>>(Period::num, Period::den));
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -777,8 +777,7 @@ template <typename Char, typename Rep, typename OutputIt,
 OutputIt format_duration_value(OutputIt out, Rep val, int precision) {
   static FMT_CONSTEXPR_DECL const Char pr_f[] = {'{', ':', '.', '{',
                                                  '}', 'f', '}', 0};
-  if (precision >= 0)
-    return format_to(out, FMT_STRING(pr_f), val, precision);
+  if (precision >= 0) return format_to(out, FMT_STRING(pr_f), val, precision);
   static FMT_CONSTEXPR_DECL const Char fp_f[] = {'{', ':', 'g', '}', 0};
   return format_to(out, FMT_STRING(fp_f), val);
 }
@@ -805,8 +804,7 @@ OutputIt format_duration_unit(OutputIt out) {
     return format_to(out, FMT_STRING(num_f), Period::num);
   static FMT_CONSTEXPR_DECL const Char num_def_f[] = {'[', '{', '}', '/', '{',
                                                       '}', ']', 's', 0};
-  return format_to(out, FMT_STRING(num_def_f), Period::num,
-                   Period::den);
+  return format_to(out, FMT_STRING(num_def_f), Period::num, Period::den);
 }
 
 template <typename FormatContext, typename OutputIt, typename Rep,
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -770,8 +770,8 @@ template <typename Char, typename Rep, typename OutputIt,
 OutputIt format_duration_value(OutputIt out, Rep val, int) {
   static FMT_CONSTEXPR_DECL const Char format[] = {'{', '}', 0};
 
-// Workaround a compiler error in MSVC < 16.8.2
-#if defined(FMT_MSC_VER) && FMT_MSC_VER <= 1928
+// Workaround a compiler error in MSVC <= 16.8.2
+#if FMT_MSC_VER && FMT_MSC_VER <= 1928
   return vformat_to(out, to_string_view(format),
                     make_format_args<buffer_context<Char>>(val));
 #else
@@ -785,7 +785,7 @@ OutputIt format_duration_value(OutputIt out, Rep val, int precision) {
   static FMT_CONSTEXPR_DECL const Char pr_f[] = {'{', ':', '.', '{',
                                                  '}', 'f', '}', 0};
   if (precision >= 0) {
-#if defined(FMT_MSC_VER) && FMT_MSC_VER <= 1928
+#if FMT_MSC_VER && FMT_MSC_VER <= 1928
     return vformat_to(out, to_string_view(pr_f),
                       make_format_args<buffer_context<Char>>(val, precision));
 #else
@@ -794,7 +794,7 @@ OutputIt format_duration_value(OutputIt out, Rep val, int precision) {
   }
   static FMT_CONSTEXPR_DECL const Char fp_f[] = {'{', ':', 'g', '}', 0};
 
-#if defined(FMT_MSC_VER) && FMT_MSC_VER <= 1928
+#if FMT_MSC_VER && FMT_MSC_VER <= 1928
   return vformat_to(out, to_string_view(fp_f),
                     make_format_args<buffer_context<Char>>(val));
 #else
@@ -821,7 +821,7 @@ OutputIt format_duration_unit(OutputIt out) {
     return copy_unit(string_view(unit), out, Char());
   static FMT_CONSTEXPR_DECL const Char num_f[] = {'[', '{', '}', ']', 's', 0};
   if (const_check(Period::den == 1)) {
-#if defined(FMT_MSC_VER) && FMT_MSC_VER <= 1928
+#if FMT_MSC_VER && FMT_MSC_VER <= 1928
     return vformat_to(out, to_string_view(num_f),
                       make_format_args<buffer_context<Char>>(Period::num));
 #else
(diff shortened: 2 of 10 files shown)


include/fmt/chrono.h
```diff
@@ -769,18 +769,37 @@ template <typename Char, typename Rep, typename OutputIt,
           FMT_ENABLE_IF(std::is_integral<Rep>::value)>
 OutputIt format_duration_value(OutputIt out, Rep val, int) {
   static FMT_CONSTEXPR_DECL const Char format[] = {'{', '}', 0};
-  return format_to(out, compile_string_to_view(format), val);
+
+// Workaround a compiler error in MSVC < 16.8.2
+#if defined(FMT_MSC_VER) && FMT_MSC_VER <= 1928
```
yeswalrus, line 774: My last fix was actually in error and did not hold up to CI once a test was in place

include/fmt/core.h
```diff
@@ -1841,6 +1841,15 @@ inline auto format_to_n(OutputIt out, size_t n, const S& format_str,
   Returns the number of characters in the output of
   ``format(format_str, args...)``.
  */
+template <typename S, typename... Args,
+          FMT_ENABLE_IF(is_compile_string<S>::value)>
+inline size_t formatted_size(const S& format_str, Args&&... args) {
```
yeswalrus, line 1846: This appeared to be a gap in the API

include/fmt/chrono.h
```diff
@@ -769,18 +769,37 @@ template <typename Char, typename Rep, typename OutputIt,
...
   static FMT_CONSTEXPR_DECL const Char format[] = {'{', '}', 0};
-  return format_to(out, compile_string_to_view(format), val);
+
+// Workaround a compiler error in MSVC <= 16.8.2
+#if FMT_MSC_VER && FMT_MSC_VER <= 1928
+  return vformat_to(out, to_string_view(format),
+                    make_format_args<buffer_context<Char>>(val));
+#else
+  return format_to(out, FMT_STRING(format), val);
+#endif
```
vitaut, line 779: Could you move this workaround to a separate PR and provide more details?

vitaut (changes requested):
Thanks for the PR.

Testing `FMT_ENFORCE_COMPILE_STRING` is a good idea but I think we should avoid duplicating tests. The new test should only check that the public headers compile with `FMT_ENFORCE_COMPILE_STRING` set with no or few checks of its own.

yeswalrus:
I agree we should avoid duplicating tests, but we kinda have to to a certain extent since without a file exercising the APIs nothing will actually get instantiated? One option might be making a separate file for all the tests of FMT_STRING specifically that could then be compiled with and without FMT_ENFORCE_COMPILE_STRING, would that work for you?

doc/api.rst
```diff
@@ -165,6 +165,14 @@ functions in their ``formatter`` specializations.
 
 .. _udt:
 
+Enforcing Compile-time Format Checks
+------------------------------------
```
vitaut, line 169: I think this doesn't need to be a separate section. Instead add a paragraph to the Compile-time Format String Checks section above.

doc/api.rst
```diff
@@ -165,6 +165,14 @@ functions in their ``formatter`` specializations.
...
+To force the use of Compile-time checks, compile with
```
vitaut, line 170: Compile -> compile

yeswalrus:
https://github.com/fmtlib/fmt/pull/2049 was closed, but the modifications to chrono.h are required. An alternative here might be to say that FMT_ENFORCE_COMPILE_STRING is simply not supported at all in MSVC, and disable the enforce_compile_string test for that compiler

include/fmt/core.h
yeswalrus: I could break it out into a separate followup PR, but it seemed relevant since without it there's no way to call formatted_size while FMT_ENFORCE_COMPILE_STRING is set

vitaut: Please move to a separate PR.

include/fmt/chrono.h
```diff
@@ -769,18 +769,38 @@ template <typename Char, typename Rep, typename OutputIt,
...
-  return format_to(out, compile_string_to_view(format), val);
+
+// Note(12/3/2020): Workaround an as-of-yet unfixed compiler error in MSVC.
+// See https://developercommunity.visualstudio.com/content/problem/1277597/internal-compiler-c0001-error-on-complex-nested-la.html
+#if FMT_MSC_VER
+  return vformat_to(out, to_string_view(format),
+                    make_format_args<buffer_context<Char>>(val));
+#else
+  return format_to(out, FMT_STRING(format), val);
+#endif
```
vitaut, line 780: As commented in #2049, I don't think it's worth doing.

test/CMakeLists.txt
```diff
@@ -106,6 +106,10 @@ add_fmt_test(printf-test)
 add_fmt_test(ranges-test)
 add_fmt_test(scan-test)
 
+
```
vitaut, line 109: nit: extra newline

test/CMakeLists.txt
```diff
@@ -106,6 +106,10 @@ add_fmt_test(printf-test)
...
+add_fmt_test(enforce-compiletime-test)
```
vitaut, line 110: enforce-compiletime-test -> enforce-compile-string-test for consistency with the macro name

test/CMakeLists.txt
```diff
@@ -106,6 +106,10 @@ add_fmt_test(printf-test)
...
+target_compile_definitions(enforce-compiletime-test PRIVATE "-DFMT_ENFORCE_COMPILE_STRING")
```
vitaut, line 111: Please break the long line.

test/enforce-compiletime-test.cc
```diff
@@ -0,0 +1,137 @@
...
+#include <list>
+#include <map>
+#include <sstream>
+#include <string>
+#include <utility>
+#include <vector>
+
+#ifdef WIN32
+#  define _CRT_SECURE_NO_WARNINGS
+#endif
```
vitaut, line 19: This is not needed, please remove.

test/enforce-compiletime-test.cc
```diff
@@ -0,0 +1,137 @@
...
+
+#include "fmt/chrono.h"
+#include "fmt/color.h"
+#include "fmt/format.h"
+#include "fmt/locale.h"
+#include "fmt/ostream.h"
+#include "fmt/ranges.h"
+
+// Exercise the API to verify that everything we expect to can compile.
+void TestFormatApi() {
```
vitaut, line 29: Here and elsewhere please follow naming conventions, e.g. TestFormatApi -> test_format_api

test/enforce-compiletime-test.cc
```diff
@@ -0,0 +1,137 @@
...
+  (void)fmt::format(FMT_STRING("{}"), 42);
+  (void)fmt::format(FMT_STRING(L"{}"), 42);
+#if !FMT_GCC_VERSION
```
vitaut, line 32: Why is it conditionally compiled?

test/enforce-compiletime-test.cc
```diff
@@ -0,0 +1,137 @@
...
+  (void)fmt::format(FMT_STRING("noop"));
+#endif
+
+  (void)fmt::to_string(42);
+  (void)fmt::to_wstring(42);
+
+  std::list<char> out;
+  fmt::format_to(std::back_inserter(out), FMT_STRING("{}"), 42);
+  std::stringstream s;
+  fmt::format_to(std::ostream_iterator<char>(s), FMT_STRING("{}"), 42);
```
vitaut, line 42: This looks redundant.

test/enforce-compiletime-test.cc
```diff
@@ -0,0 +1,137 @@
...
+  wchar_t wbuffer[4];
+  (void)fmt::format_to_n(wbuffer, 3, FMT_STRING(L"{}"), 12345);
+
+  (void)fmt::formatted_size(FMT_STRING("{}"), 42);
+}
+void TestLiteralsApi() {
+#if FMT_USE_UDL_TEMPLATE
+  // Passing user-defined literals directly to EXPECT_EQ causes problems
+  // with macro argument stringification (#) on some versions of GCC.
+  // Workaround: Assing the UDL result to a variable before the macro.
```
vitaut, line 56: There is no EXPECT_EQ here

test/enforce-compiletime-test.cc
```diff
@@ -0,0 +1,137 @@
...
+    ++data;
+    return tmp;
+  }
+  char& operator*() { return *data; }
+};
+
+void FormatToNOutputIteratorTest() {
+  char buf[10] = {};
+  fmt::format_to_n(test_output_iterator{buf}, 10, FMT_STRING("{}"), 42);
+}
```
vitaut, line 91: This is redundant, please remove.

test/enforce-compiletime-test.cc
```diff
@@ -0,0 +1,137 @@
...
+
+void TestChrono() {
+#ifndef FMT_STATIC_THOUSANDS_SEPARATOR
```
vitaut, line 94: Why conditionally compiled?

test/enforce-compiletime-test.cc
```diff
@@ -0,0 +1,137 @@
...
+  const char* p;
+  const char* begin() const { return p; }
+  zstring_sentinel end() const { return {}; }
+};
+
+void TestZString() {
+  zstring hello{"hello"};
+  (void)fmt::format(FMT_STRING("{}"), hello);
+  (void)fmt::format(FMT_STRING("{}"), fmt::join(hello, "_"));
+}
```
vitaut, line 126: This is redundant.

test/enforce-compiletime-test.cc
```diff
@@ -0,0 +1,137 @@
...
+
+int main(int, char**) {
```
vitaut, line 128: Arguments are unused, please remove.

test/enforce-compiletime-test.cc
```diff
@@ -0,0 +1,137 @@
...
+  TestFormatApi();
+  TestLiteralsApi();
+  FormatToNOutputIteratorTest();
+  TestChrono();
+  TestTextStyle();
+  TestZString();
+
+  return 0;
```
vitaut, line 136: This is not needed.

vitaut (changes requested):

test/enforce-compiletime-test.cc
yeswalrus: TBH, I'm not sure - all the tests covering formatting of chrono types in chrono_test.cc were too so I figured it was important? I'll try removing it

test/enforce-compiletime-test.cc
yeswalrus: I removed the 2nd overload but this actually exposed an issue with ranges.h:294

test/enforce-compiletime-test.cc
yeswalrus: Because it crashes in GCC if it's not - This is the issue I reported here: https://github.com/fmtlib/fmt/issues/2039. It seems to still be a problem: https://github.com/fmtlib/fmt/pull/2038/checks?check_run_id=1544063743

test/enforce-compiletime-test.cc
mwinterb: I lack context for this particular test, but: `FMT_STATIC_THOUSANDS_SEPARATOR` is "overloaded" to mean "I don't like `<locale>` and want nothing to do with it", and so the `locale_ref` functions are not implemented if `FMT_STATIC_THOUSANDS_SEPARATOR` is defined. However, `chrono_formatter::format_localized` always tries to use the `locale_ref` functions, so `chrono-test` won't link in that situation.

yeswalrus: Thank you for the explanation. It seems to work just fine without the ifdef in this situation so I'll leave them removed.

yeswalrus:
I've tried to solve the issue with GCC & calls to format with no arguments, but I'm not super familiar with the constexpr machinery at play & how to debug it.

include/fmt/chrono.h
```diff
@@ -769,18 +769,24 @@ template <typename Char, typename Rep, typename OutputIt,
           FMT_ENABLE_IF(std::is_integral<Rep>::value)>
 OutputIt format_duration_value(OutputIt out, Rep val, int) {
   static FMT_CONSTEXPR_DECL const Char format[] = {'{', '}', 0};
-  return format_to(out, compile_string_to_view(format), val);
+
```
vitaut, line 772: nit: extra newline

include/fmt/chrono.h
```diff
@@ -769,18 +769,24 @@ template <typename Char, typename Rep, typename OutputIt,
...
+  return vformat_to(out, to_string_view(format),
```
vitaut, line 773: Here and below: why not wrap `format` in `FMT_STRING` and use `format_to`?

include/fmt/core.h
```diff
@@ -1841,6 +1841,7 @@ inline auto format_to_n(OutputIt out, size_t n, const S& format_str,
   Returns the number of characters in the output of
   ``format(format_str, args...)``.
  */
+
```
vitaut, line 1844: Extra newline

include/fmt/ranges.h
```diff
@@ -291,10 +291,10 @@ struct formatter<TupleT, Char, enable_if_t<fmt::is_tuple_like<TupleT>::value>> {
...
         out = detail::copy(formatting.delimiter, out);
       }
-      out = format_to(out,
-                      detail::format_str_quoted(
-                          (formatting.add_delimiter_spaces && i > 0), v),
-                      v);
+      out = vformat_to(out,
+                       detail::format_str_quoted(
+                           (formatting.add_delimiter_spaces && i > 0), v),
+                       make_format_args<FormatContext>(v));
```
vitaut, line 297: Here and below: it would be better to refactor this to use compile string and `format_to`.

test/CMakeLists.txt
```diff
@@ -106,6 +106,14 @@ add_fmt_test(printf-test)
 add_fmt_test(ranges-test)
 add_fmt_test(scan-test)
 
+if(NOT MSVC)
```
vitaut, line 109: nit: missing space after `if`.

test/CMakeLists.txt
```diff
@@ -106,6 +106,14 @@ add_fmt_test(printf-test)
...
+  # FMT_ENFORCE_COMPILE_STRING not supported under MSVC
+  # See https://developercommunity.visualstudio.com/content/problem/1277597/internal-compiler-c0001-error-on-complex-nested-la.html
+  add_fmt_test(enforce-compile-string-test)
+  target_compile_definitions(enforce-compile-string-test PRIVATE
+    "-DFMT_ENFORCE_COMPILE_STRING")
+endif()
```
vitaut, line 115: ditto

test/enforce-compile-string-test.cc
```diff
@@ -0,0 +1,95 @@
...
+// For the license information refer to format.h.
+
+#include <array>
+#include <iterator>
+#include <list>
+#include <map>
+#include <sstream>
+#include <string>
+#include <utility>
+#include <vector>
```
vitaut, line 15: Some of these are unused.

test/enforce-compile-string-test.cc
```diff
@@ -0,0 +1,95 @@
...
+#include "fmt/chrono.h"
+#include "fmt/color.h"
+#include "fmt/format.h"
+#include "fmt/locale.h"
+#include "fmt/ostream.h"
+#include "fmt/ranges.h"
+
+// Exercise the API to verify that everything we expect to can compile.
+void test_format_api() {
+  (void)fmt::format(FMT_STRING("{}"), 42);
```
vitaut, line 26: I don't think cast is necessary.

test/enforce-compile-string-test.cc
```diff
@@ -0,0 +1,95 @@
...
+  (void)fmt::format(FMT_STRING(L"{}"), 42);
+#if !FMT_GCC_VERSION  // Currently will not compile: See
+                      // https://github.com/fmtlib/fmt/issues/2039
```
vitaut, line 29: This should compile now, please rebase and remove.

test/enforce-compile-string-test.cc
```diff
@@ -0,0 +1,95 @@
...
+}
+
+void test_literals_api() {
+#if FMT_USE_UDL_TEMPLATE
+  using namespace fmt::literals;
+
+  auto udl_format = "{}c{}"_format("ab", 1);
+  auto udl_format_w = L"{}c{}"_format(L"ab", 1);
+  (void)udl_format;
+  (void)udl_format_w;
```
vitaut, line 53: Variables are not needed, please remove.

test/enforce-compile-string-test.cc
```diff
@@ -0,0 +1,95 @@
...
+struct zstring {
+  const char* p;
+  const char* begin() const { return p; }
+  zstring_sentinel end() const { return {}; }
+};
+
+void test_zstring() {
+  zstring hello{"hello"};
+  (void)fmt::format(FMT_STRING("{}"), hello);
+}
```
vitaut, line 87: This looks unnecessary. If you want to check ranges, you can use a standard container.

vitaut (changes requested):

include/fmt/chrono.h
yeswalrus: Because then it wouldn't compile on windows because of the MSVC error?

vitaut: AFAICS the MSVC bug is specific to static constexpr which we don't need here. I suggest replacing with:

```c++
  return format_to(out, FMT_STRING("{}"), val);
```

which seems to work: https://godbolt.org/z/Teqnhq.

yeswalrus: I thought we needed the static constexpr array so that the format string type would match the type of `Char`?

include/fmt/ranges.h
yeswalrus: I considered that but given that the format string used appeared to be determined at run-time I wasn't sure it was possible. I'll give it a go.

include/fmt/chrono.h
vitaut: On a second thought, we don't need `format_to` at all. This should be just `write<Char>(out, val)`.

test/enforce-compile-string-test.cc
yeswalrus: IIRC apple clang complains about an unused return value if it's not present, I'll try it again though.

include/fmt/chrono.h
yeswalrus: Excellent! I had no idea about that API. I've redone the other calls that were using format_to in chrono as well.

include/fmt/ranges.h
yeswalrus: Looking at this more I'm reaching the same conclusion - `formatting` can't be used from within `FMT_STRING` - doing so causes 
```
/home/walter.gray/development/walrus/fmt/include/fmt/ranges.h:371:29: error: use of non-static data member 'formatting' of 'formatter<T, Char, enable_if_t<fmt::is_range<T, Char>::value && (has_formatter<detail::value_type<T>, format_context>::value || detail::has_fallback_formatter<detail::value_type<T>, format_context>::value)> >' from nested type 'FMT_COMPILE_STRING'
                           (formatting.add_delimiter_spaces && i > 0), *it)),```
                           
I couldn't have a function that returned one of several format strings, and I don't think I could have a function that returned a constexpr string that would work with FMT_STRING either - 
```
constexpr const char* getString() {
    return "{} bananna";
}
int main(int argc, char** argv) {
    fmt::print(FMT_STRING(getString()), "fried");
}
```
for example will not compile in godbolt. I also cant return the result of FMT_STRING since they'd be different types between branches. My best guess for how to do this would be to have a wrapping function that used template specialization on a bool for add_delimiter_spaces and runtime switching on i > 0 and called format_to with the appropriate FMT_STRING directly. Am I missing an option here or should I try that?

File: include/fmt/chrono.h [hunk 31]
```diff
@@ -800,13 +797,20 @@ template <typename Char, typename Period, typename OutputIt>
...
-  static FMT_CONSTEXPR_DECL const Char num_f[] = {'[', '{', '}', ']', 's', 0};
-  if (const_check(Period::den == 1))
-    return format_to(out, compile_string_to_view(num_f), Period::num);
-  static FMT_CONSTEXPR_DECL const Char num_def_f[] = {'[', '{', '}', '/', '{',
-                                                      '}', ']', 's', 0};
-  return format_to(out, compile_string_to_view(num_def_f), Period::num,
-                   Period::den);
+
+  basic_memory_buffer<Char> buffer;
+  auto bufOut = std::back_inserter(buffer);
```
vitaut, line 802: This is unnecessary. You can write to `out` directly.

include/fmt/ranges.h
vitaut: Let's use `write` instead of `format_to` here too.

test/enforce-compile-string-test.cc
```diff
@@ -0,0 +1,75 @@
...
+                    "rgb(255,20,30)");
+
+  fmt::text_style ts = fg(fmt::rgb(255, 20, 30));
+  std::string out;
+  fmt::format_to(std::back_inserter(out), ts,
+                 FMT_STRING("rgb(255,20,30){}{}{}"), 1, 2, 3);
+}
+
+void test_range() {
+  std::array<char, 5> hello = {'h','e','l','l','o'};
```
vitaut, line 65: Please apply clang-format.

vitaut (changes requested):
Mostly looks good, just a few minor comments.

yeswalrus:
Done! Good call on refactoring to use write - worked like a charm and IMO the code is easier to follow now as well.

include/fmt/locale.h
```diff
@@ -52,8 +52,8 @@ inline OutputIt vformat_to(
 
 template <typename OutputIt, typename S, typename... Args,
           bool enable = detail::is_output_iterator<OutputIt, char_t<S>>::value>
-inline auto format_to(OutputIt out, const std::locale& loc,
-                      const S& format_str, Args&&... args) ->
+inline auto format_to(OutputIt out, const std::locale& loc, const S& format_str,
+                      Args&&... args) ->
```
vitaut, line 56: Unrelated, please revert.

include/fmt/ranges.h
```diff
@@ -247,31 +247,48 @@ template <typename Range>
 using value_type =
     remove_cvref_t<decltype(*detail::range_begin(std::declval<Range>()))>;
 
-template <typename Arg, FMT_ENABLE_IF(!is_like_std_string<
-                                      typename std::decay<Arg>::type>::value)>
-FMT_CONSTEXPR const char* format_str_quoted(bool add_space, const Arg&) {
-  return add_space ? " {}" : "{}";
+template <typename OutputIt, typename Formatting>
+OutputIt add_range_formatting_spaces(OutputIt out,
+                                     const Formatting& formatting) {
```
vitaut, line 252: add_range_formatting_spaces -> write_delimiter

include/fmt/ranges.h
```diff
@@ -247,31 +247,48 @@ template <typename Range>
...
+  if (const_check(formatting.add_prepostfix_space)) {
+    *out++ = ' ';
+  }
```
vitaut, line 255: Let's just remove this since `add_prepostfix_space` is always false.

include/fmt/ranges.h
```diff
@@ -247,31 +247,48 @@ template <typename Range>
...
+  out = detail::copy(formatting.delimiter, out);
+  if (const_check(formatting.add_delimiter_spaces)) {
+    *out++ = ' ';
+  }
```
vitaut, line 259: Let's write space unconditionally since `add_delimiter_spaces` is always true and remove `formatting` argument.

include/fmt/ranges.h
```diff
@@ -247,31 +247,48 @@ template <typename Range>
...
+  return out;
 }
 
-template <typename Arg, FMT_ENABLE_IF(is_like_std_string<
-                                      typename std::decay<Arg>::type>::value)>
-FMT_CONSTEXPR const char* format_str_quoted(bool add_space, const Arg&) {
-  return add_space ? " \"{}\"" : "\"{}\"";
+template <
+    typename Char, typename OutputIt, typename Arg,
+    FMT_ENABLE_IF(std::is_same<Arg, const char*>::value ||
```
vitaut, line 265: What is `std::is_same<Arg, const char*>::value` for?

include/fmt/ranges.h
```diff
@@ -363,19 +375,19 @@ struct formatter<
...
       if (++i > formatting.range_length_limit) {
-        out = format_to(out, " ... <other elements>");
+        out = format_to(out, FMT_STRING("{}"), " ... <other elements>");
         break;
       }
     }
-    if (formatting.add_prepostfix_space) *out++ = ' ';
+    if (formatting.add_prepostfix_space) {
+      *out++ = ' ';
+    }
```
vitaut, line 390: Please revert (or remove).

vitaut (changes requested):

include/fmt/ranges.h
yeswalrus: My impression was that in user-declared overloads it could not be?

include/fmt/ranges.h
yeswalrus: I thought that this was required to make it so that for example a `const char * array_of_literals[]` would be formatted with `"`s, but I appear to be incorrect. I've added a test-case to cover that and you're correct, it appears unnecessary, though I couldn't say why.

vitaut:
Thank you!

PR #2042, opened 2020-12-01
Fix formatting with compile-time API when format string is empty

alexezeder:
Right now {fmt} cannot be used with empty string format with compile-time API, while it can be used with regular API:

```cpp
fmt::format(""); // ok
fmt::format(FMT_COMPILE("")); // compilation error
```
Proof on [Compiler Explorer](https://godbolt.org/z/6Yj38a)

With this tiny fix, compile-time API also get this useless feature.
Also, a test for `FMT_COMPILE` with an empty string added.


vitaut:
Thank you!

PR #2043, opened 2020-12-01
Add UDL as replacement for FMT_COMPILE

alexezeder:
Right now {fmt} has a special macro to invoke compile-time API functions - `FMT_COMPILE`. This macro can be replaced with UDL for modern compilers that support C++20. The chosen name in this PR for this feature is `_cf`.

```cpp
fmt::format(FMT_COMPILE("{}"), 42); // 🙁 not modern
fmt::format("{}"_cf, 42); // 🙂 modern as hell
```

Note that only GCC 9.3+ support that. Clang (trunk) is also able to compile this code, but it doesn't have `__cpp_nontype_template_parameter_class` defined, I hope it'll be fixed in Clang 12 release.
Proof on [Compiler Explorer](https://godbolt.org/z/boTTr9)

`constexpr` is used instead of `consteval` (even though it would make more sense) to lower GCC version.

Also, I should note that this PR is based on my own ideas and the following comments: https://github.com/fmtlib/fmt/issues/1874#issuecomment-716097084, https://github.com/fmtlib/fmt/issues/613#issuecomment-491906397.

### DIFF
--- a/include/fmt/compile.h
+++ b/include/fmt/compile.h
@@ -13,6 +13,12 @@
 
 #include "format.h"
 
+#if !defined(FMT_USE_NONTYPE_TEMPLATE_PARAMETERS) &&   \
+    defined(__cpp_nontype_template_parameter_class) && \
+    (!FMT_GCC_VERSION || FMT_GCC_VERSION >= 903)
+#  define FMT_USE_NONTYPE_TEMPLATE_PARAMETERS
+#endif
+
 FMT_BEGIN_NAMESPACE
 namespace detail {
 
@@ -37,9 +43,8 @@ struct is_compiled_string : std::is_base_of<compiled_string, S> {};
  */
 #define FMT_COMPILE(s) FMT_STRING_IMPL(s, fmt::detail::compiled_string)
 
-#if defined(__cpp_nontype_template_parameter_class) && \
-    (!FMT_GCC_VERSION || FMT_GCC_VERSION >= 903)
-template <typename Char, std::size_t N> struct fixed_string {
+#ifdef FMT_USE_NONTYPE_TEMPLATE_PARAMETERS
+template <typename Char, size_t N> struct fixed_string {
   constexpr fixed_string(const Char (&str)[N + 1]) {
     copy_str<Char, const Char*, Char*>(static_cast<const Char*>(str), str + N,
                                        data);
@@ -47,10 +52,10 @@ template <typename Char, std::size_t N> struct fixed_string {
   Char data[N]{};
 };
 
-template <typename Char, std::size_t N>
+template <typename Char, size_t N>
 fixed_string(const Char (&str)[N]) -> fixed_string<Char, N - 1>;
 
-template <typename Char, std::size_t N, fixed_string<Char, N> Str>
+template <typename Char, size_t N, fixed_string<Char, N> Str>
 struct udl_compiled_string : compiled_string {
   using char_type = Char;
   constexpr operator basic_string_view<char_type>() const {
@@ -720,8 +725,7 @@ size_t formatted_size(const CompiledFormat& cf, const Args&... args) {
   return format_to(detail::counting_iterator(), cf, args...).count();
 }
 
-#if defined(__cpp_nontype_template_parameter_class) && \
-    (!FMT_GCC_VERSION || FMT_GCC_VERSION >= 903)
+#ifdef FMT_USE_NONTYPE_TEMPLATE_PARAMETERS
 inline namespace literals {
 template <detail::fixed_string Str>
 constexpr detail::udl_compiled_string<remove_cvref_t<decltype(Str.data[0])>,
--- a/include/fmt/compile.h
+++ b/include/fmt/compile.h
@@ -55,9 +55,6 @@ template <typename Char, size_t N> struct fixed_string {
   Char data[N]{};
 };
 
-template <typename Char, size_t N>
-fixed_string(const Char (&str)[N]) -> fixed_string<Char, N>;
-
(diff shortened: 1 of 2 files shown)


include/fmt/compile.h
```diff
@@ -37,6 +37,28 @@ struct is_compiled_string : std::is_base_of<compiled_string, S> {};
...
 #define FMT_COMPILE(s) FMT_STRING_IMPL(s, fmt::detail::compiled_string)
 
+#if defined(__cpp_nontype_template_parameter_class) && \
+    (!FMT_GCC_VERSION || FMT_GCC_VERSION >= 903)
+template <typename Char, std::size_t N> struct fixed_string {
+  constexpr fixed_string(const Char (&str)[N + 1]) {
+    copy_str<Char, const Char*, Char*>(static_cast<const Char*>(str), str + N,
+                                       data);
+  }
+  Char data[N]{};
```
vitaut, line 47: `{}` seems unnecessary, please remove.

include/fmt/compile.h
```diff
@@ -37,6 +37,28 @@ struct is_compiled_string : std::is_base_of<compiled_string, S> {};
  */
 #define FMT_COMPILE(s) FMT_STRING_IMPL(s, fmt::detail::compiled_string)
 
+#if defined(__cpp_nontype_template_parameter_class) && \
+    (!FMT_GCC_VERSION || FMT_GCC_VERSION >= 903)
+template <typename Char, std::size_t N> struct fixed_string {
```
vitaut, line 42: std::size_t -> size_t for consistency with the rest of the library.

include/fmt/compile.h
```diff
@@ -37,6 +37,28 @@ struct is_compiled_string : std::is_base_of<compiled_string, S> {};
  */
 #define FMT_COMPILE(s) FMT_STRING_IMPL(s, fmt::detail::compiled_string)
 
+#if defined(__cpp_nontype_template_parameter_class) && \
+    (!FMT_GCC_VERSION || FMT_GCC_VERSION >= 903)
```
vitaut, line 41: I suggest moving this into a separate macro `FMT_USE_NONTYPE_TEMPLATE_PARAMETERS` since it is used more than once (and to make it controllable).

vitaut (changes requested):
Thanks for the PR! Mostly looks good, just a few nits inline.

include/fmt/compile.h
alexezeder: It's either `{}` here or `: data{}` initialization in constructor for GCC 9.3, otherwise:
```
<source>:8:13: error: member 'fmt::v7::detail::fixed_string<char, 2>::data' must be initialized by mem-initializer in 'constexpr' constructor
<source>:12:8: note: declared here
   12 |   Char data[N];
      |        ^~~~
```

include/fmt/compile.h
alexezeder: Done

include/fmt/compile.h
alexezeder: Done

yeswalrus:
I thought the plan was to just make compile-time checks the default in c++20?

alexezeder:
@yeswalrus I think you are confusing `FMT_STRING` and `FMT_COMPILE`. They are similar, they even use the same `FMT_STRING_IMPL` macro under the hood. But `FMT_STRING` is used to indicate that format string should be just checked, while `FMT_COMPILE` is used to indicate that format string should be baked in form of some compiled format that has prepared parts for formatting.
This PR is about the `FMT_COMPILE` which invokes compile-time API.

yeswalrus:
Correct, my mistake.

include/fmt/compile.h
```diff
@@ -13,6 +13,12 @@
 
 #include "format.h"
 
+#if !defined(FMT_USE_NONTYPE_TEMPLATE_PARAMETERS) &&   \
+    defined(__cpp_nontype_template_parameter_class) && \
+    (!FMT_GCC_VERSION || FMT_GCC_VERSION >= 903)
+#  define FMT_USE_NONTYPE_TEMPLATE_PARAMETERS
+#endif
```
vitaut, line 20: Please make `FMT_USE_NONTYPE_TEMPLATE_PARAMETERS` defined to 1/0 for consistency with other similar macros and to make it possible to disable the feature. See e.g. https://github.com/fmtlib/fmt/blob/33f9a6d3609c7001afc7019c8308f4af8942ce0a/include/fmt/core.h#L178-L185

include/fmt/compile.h
```diff
@@ -37,6 +43,27 @@ struct is_compiled_string : std::is_base_of<compiled_string, S> {};
  */
 #define FMT_COMPILE(s) FMT_STRING_IMPL(s, fmt::detail::compiled_string)
 
+#ifdef FMT_USE_NONTYPE_TEMPLATE_PARAMETERS
+template <typename Char, size_t N> struct fixed_string {
```
vitaut, line 47: Is `fixed_string` actually needed? Can we directly use an array instead?

vitaut (changes requested):

include/fmt/compile.h
alexezeder: Looks like yes, it's actually needed. The raw C-style array should be wrapped in some type, and this wrapper-type should also have a constructor with `const Char (&str)[N]` as an argument. Of course, I can be wrong here, but all my attempts ended up with nothing working.

**But** I realized that the deduction guide for `fixed_string` can be removed now because it holds a null-terminated string.

include/fmt/compile.h
alexezeder: Done

vitaut:
Merged, thanks.

PR #2044, opened 2020-12-02
Simplify CompileTimeFormattingTest tests by reordering arguments of EXPECT_EQ

alexezeder:
Extends #2019 by simplifying tests. This change was requested there https://github.com/fmtlib/fmt/pull/2019#discussion_r532121123 but I didn't manage to simplify it.

Turns out that the position in `EXPECT_EQ` macro is the key:
```cpp
EXPECT_EQ(test_format<5>(FMT_COMPILE("{}"), true), "true"); // compilation error
EXPECT_EQ("true", test_format<5>(FMT_COMPILE("{}"), true)); // ok
```

In this PR arguments are just swapped, so the result is the first argument. Just like all over the file:
https://github.com/fmtlib/fmt/blob/5a493560f59369e9fa664e8945b8e8a8ec4391b2/test/compile-test.cc#L125-L136

### DIFF
--- a/test/compile-test.cc
+++ b/test/compile-test.cc
@@ -180,7 +180,6 @@ template <size_t max_string_length> struct test_string {
   template <typename T> constexpr bool operator==(const T& rhs) const noexcept {
     return (std::string_view(rhs).compare(buffer.data()) == 0);
   }
-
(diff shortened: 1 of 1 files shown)


vitaut:
Nice!

PR #2053, opened 2020-12-07
Suppress gcc warning on privates-only class

Finkman:
Since gcc 9 it warns about is_streamable to have only private methods.


vitaut (changes requested):
The warning can be suppressed in a simpler way by adding a defaulted ctor.

Finkman:
oh yes... that's much less invasive

vitaut:
Thanks!

PR #2056, opened 2020-12-09
Add support of most format_specs for formatting at compile-time

alexezeder:
Extends https://github.com/fmtlib/fmt/pull/2019 by adding support of formatting specs for bool, integer, char and string:

- [fill](https://fmt.dev/latest/syntax.html#grammar-token-fill) `<a character other than '{' or '}'>`, like `{:*>4}`, `{:ж^4}`
- [align](https://fmt.dev/latest/syntax.html#grammar-token-align) `"<" | ">" | "^"`, like `{:^4}`
- [sign](https://fmt.dev/latest/syntax.html#grammar-token-sign) `"+" | "-" | " "`, like `{:+}`
- `"#"`, like `{:#b}`
- `"0"`, like `{:04}`
- [width](https://fmt.dev/latest/syntax.html#grammar-token-width) `integer | "{"  "}"`, like `{:4}`, `{:{}}`
- [int_type](https://fmt.dev/latest/syntax.html#grammar-token-int_type) `"b" | "B" | "d" | "o" | "x" | "X"`, like `{:b}`, `{:x}`, `{:X}`.

I think that bool, integer, char and string formatting should be fully supported at compile-time, of course, if there are no caveats I'm not aware of. Floating-point formatting is not addressed by this PR, so it's still unavailable at compile-time.

Also `test_string` in `compile-test` updated a bit to use not `std::` but `fmt:string_view`, this eliminates unnecessary include there

### DIFF
--- a/include/fmt/compile.h
+++ b/include/fmt/compile.h
@@ -469,7 +469,7 @@ template <typename Char, typename T, int N> struct spec_field {
   formatter<T, Char> fmt;
 
   template <typename OutputIt, typename... Args>
-  constexpr OutputIt format(OutputIt out, const Args&... args) {
+  constexpr OutputIt format(OutputIt out, const Args&... args) const {
     // This ensures that the argument type is convertile to `const T&`.
     const T& arg = get<N>(args...);
     const auto& vargs =
@@ -488,7 +488,7 @@ template <typename L, typename R> struct concat {
   using char_type = typename L::char_type;
 
   template <typename OutputIt, typename... Args>
-  constexpr OutputIt format(OutputIt out, const Args&... args) {
+  constexpr OutputIt format(OutputIt out, const Args&... args) const {
     out = lhs.format(out, args...);
     return rhs.format(out, args...);
   }
@@ -635,7 +635,7 @@ FMT_DEPRECATED auto compile(const Args&... args)
 template <typename CompiledFormat, typename... Args,
           typename Char = typename CompiledFormat::char_type,
           FMT_ENABLE_IF(detail::is_compiled_format<CompiledFormat>::value)>
-FMT_INLINE std::basic_string<Char> format(CompiledFormat& cf,
+FMT_INLINE std::basic_string<Char> format(const CompiledFormat& cf,
                                           const Args&... args) {
   basic_memory_buffer<Char> buffer;
   cf.format(detail::buffer_appender<Char>(buffer), args...);
@@ -644,7 +644,7 @@ FMT_INLINE std::basic_string<Char> format(CompiledFormat& cf,
 
 template <typename OutputIt, typename CompiledFormat, typename... Args,
           FMT_ENABLE_IF(detail::is_compiled_format<CompiledFormat>::value)>
-constexpr OutputIt format_to(OutputIt out, CompiledFormat& cf,
+constexpr OutputIt format_to(OutputIt out, const CompiledFormat& cf,
                              const Args&... args) {
   return cf.format(out, args...);
 }
@@ -674,14 +674,14 @@ FMT_INLINE std::basic_string<typename S::char_type> format(const S&,
       return fmt::to_string(detail::first(args...));
   }
 #endif
-  auto compiled = detail::compile<Args...>(S());
+  constexpr auto compiled = detail::compile<Args...>(S());
   return format(compiled, std::forward<Args>(args)...);
 }
 
 template <typename OutputIt, typename CompiledFormat, typename... Args,
           FMT_ENABLE_IF(std::is_base_of<detail::basic_compiled_format,
                                         CompiledFormat>::value)>
-constexpr OutputIt format_to(OutputIt out, CompiledFormat& cf,
+constexpr OutputIt format_to(OutputIt out, const CompiledFormat& cf,
                              const Args&... args) {
   using char_type = typename CompiledFormat::char_type;
   using context = format_context_t<OutputIt, char_type>;
@@ -692,7 +692,7 @@ constexpr OutputIt format_to(OutputIt out, CompiledFormat& cf,
 template <typename OutputIt, typename S, typename... Args,
           FMT_ENABLE_IF(detail::is_compiled_string<S>::value)>
 FMT_CONSTEXPR OutputIt format_to(OutputIt out, const S&, const Args&... args) {
-  auto compiled = detail::compile<Args...>(S());
+  constexpr auto compiled = detail::compile<Args...>(S());
   return format_to(out, compiled, args...);
 }
 
@@ -714,7 +714,7 @@ template <typename OutputIt, typename S, typename... Args,
           FMT_ENABLE_IF(detail::is_compiled_string<S>::value)>
 format_to_n_result<OutputIt> format_to_n(OutputIt out, size_t n, const S&,
                                          const Args&... args) {
-  auto compiled = detail::compile<Args...>(S());
+  constexpr auto compiled = detail::compile<Args...>(S());
   auto it = format_to(detail::truncating_iterator<OutputIt>(out, n), compiled,
                       args...);
   return {it.base(), it.count()};
--- a/include/fmt/compile.h
+++ b/include/fmt/compile.h
@@ -466,10 +466,10 @@ struct is_compiled_format<field<Char, T, N>> : std::true_type {};
 // A replacement field that refers to argument N and has format specifiers.
 template <typename Char, typename T, int N> struct spec_field {
   using char_type = Char;
-  mutable formatter<T, Char> fmt;
+  formatter<T, Char> fmt;
 
   template <typename OutputIt, typename... Args>
-  OutputIt format(OutputIt out, const Args&... args) const {
+  constexpr OutputIt format(OutputIt out, const Args&... args) {
     // This ensures that the argument type is convertile to `const T&`.
     const T& arg = get<N>(args...);
     const auto& vargs =
@@ -488,7 +488,7 @@ template <typename L, typename R> struct concat {
   using char_type = typename L::char_type;
 
   template <typename OutputIt, typename... Args>
-  constexpr OutputIt format(OutputIt out, const Args&... args) const {
+  constexpr OutputIt format(OutputIt out, const Args&... args) {
     out = lhs.format(out, args...);
     return rhs.format(out, args...);
   }
@@ -635,7 +635,7 @@ FMT_DEPRECATED auto compile(const Args&... args)
 template <typename CompiledFormat, typename... Args,
           typename Char = typename CompiledFormat::char_type,
           FMT_ENABLE_IF(detail::is_compiled_format<CompiledFormat>::value)>
-FMT_INLINE std::basic_string<Char> format(const CompiledFormat& cf,
+FMT_INLINE std::basic_string<Char> format(CompiledFormat& cf,
                                           const Args&... args) {
   basic_memory_buffer<Char> buffer;
   cf.format(detail::buffer_appender<Char>(buffer), args...);
@@ -644,7 +644,7 @@ FMT_INLINE std::basic_string<Char> format(const CompiledFormat& cf,
 
 template <typename OutputIt, typename CompiledFormat, typename... Args,
           FMT_ENABLE_IF(detail::is_compiled_format<CompiledFormat>::value)>
-constexpr OutputIt format_to(OutputIt out, const CompiledFormat& cf,
+constexpr OutputIt format_to(OutputIt out, CompiledFormat& cf,
                              const Args&... args) {
   return cf.format(out, args...);
 }
@@ -674,14 +674,14 @@ FMT_INLINE std::basic_string<typename S::char_type> format(const S&,
       return fmt::to_string(detail::first(args...));
   }
 #endif
-  constexpr auto compiled = detail::compile<Args...>(S());
+  auto compiled = detail::compile<Args...>(S());
   return format(compiled, std::forward<Args>(args)...);
 }
 
 template <typename OutputIt, typename CompiledFormat, typename... Args,
           FMT_ENABLE_IF(std::is_base_of<detail::basic_compiled_format,
                                         CompiledFormat>::value)>
-constexpr OutputIt format_to(OutputIt out, const CompiledFormat& cf,
+constexpr OutputIt format_to(OutputIt out, CompiledFormat& cf,
                              const Args&... args) {
   using char_type = typename CompiledFormat::char_type;
   using context = format_context_t<OutputIt, char_type>;
@@ -692,7 +692,7 @@ constexpr OutputIt format_to(OutputIt out, const CompiledFormat& cf,
 template <typename OutputIt, typename S, typename... Args,
           FMT_ENABLE_IF(detail::is_compiled_string<S>::value)>
 FMT_CONSTEXPR OutputIt format_to(OutputIt out, const S&, const Args&... args) {
-  constexpr auto compiled = detail::compile<Args...>(S());
+  auto compiled = detail::compile<Args...>(S());
   return format_to(out, compiled, args...);
 }
 
@@ -714,7 +714,7 @@ template <typename OutputIt, typename S, typename... Args,
           FMT_ENABLE_IF(detail::is_compiled_string<S>::value)>
 format_to_n_result<OutputIt> format_to_n(OutputIt out, size_t n, const S&,
                                          const Args&... args) {
-  constexpr auto compiled = detail::compile<Args...>(S());
+  auto compiled = detail::compile<Args...>(S());
   auto it = format_to(detail::truncating_iterator<OutputIt>(out, n), compiled,
                       args...);
   return {it.base(), it.count()};
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -283,7 +283,7 @@ struct monostate {};
 
 namespace detail {
 
-constexpr bool is_constant_evaluated() FMT_DETECTED_NOEXCEPT {
+constexpr bool is_constant_evaluated() FMT_NOEXCEPT {
 #ifdef __cpp_lib_is_constant_evaluated
   return std::is_constant_evaluated();
 #else
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -481,7 +481,7 @@ inline basic_string_view<Char> to_string_view(
 }
 
 template <typename Char>
-inline basic_string_view<Char> to_string_view(basic_string_view<Char> s) {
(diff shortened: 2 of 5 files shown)


alexezeder:
Also, looks like tests for formatting at compile-time (`CompileTimeFormattingTest` in `compile-test.cc`) do not run because the GCC 10.2 defines `__cplusplus` less than `202002L`, at least on my PC 🙂

include/fmt/core.h
```diff
@@ -1480,28 +1480,30 @@ template <typename OutputIt, typename Char> class basic_format_context {
...
-                       detail::locale_ref loc = detail::locale_ref())
+  constexpr basic_format_context(
+      OutputIt out, basic_format_args<basic_format_context> ctx_args,
+      detail::locale_ref loc = detail::locale_ref())
       : out_(out), args_(ctx_args), loc_(loc) {}
 
-  format_arg arg(int id) const { return args_.get(id); }
-  format_arg arg(basic_string_view<char_type> name) { return args_.get(name); }
+  constexpr format_arg arg(int id) const { return args_.get(id); }
+  FMT_CONSTEXPR format_arg arg(basic_string_view<char_type> name) {
```
vitaut, line 1489: Here and below: can we use `constexpr` instead of `FMT_CONSTEXPR`?

include/fmt/core.h
```diff
@@ -1644,33 +1646,35 @@ template <typename Context> class basic_format_args {
...
+  constexpr basic_format_args() : desc_(0), args_(nullptr) {}
 
   /**
    \rst
    Constructs a `basic_format_args` object from `~fmt::format_arg_store`.
    \endrst
    */
   template <typename... Args>
-  FMT_INLINE basic_format_args(const format_arg_store<Context, Args...>& store)
+  constexpr basic_format_args(const format_arg_store<Context, Args...>& store)
```
vitaut, line 1677: Here and elsewhere: please keep `FMT_INLINE`.

vitaut (changes requested):
Thanks for the PR!

include/fmt/compile.h
```diff
@@ -466,10 +466,10 @@ struct is_compiled_format<field<Char, T, N>> : std::true_type {};
 // A replacement field that refers to argument N and has format specifiers.
 template <typename Char, typename T, int N> struct spec_field {
   using char_type = Char;
-  mutable formatter<T, Char> fmt;
+  formatter<T, Char> fmt;
 
   template <typename OutputIt, typename... Args>
-  OutputIt format(OutputIt out, const Args&... args) const {
+  constexpr OutputIt format(OutputIt out, const Args&... args) {
```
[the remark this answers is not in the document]
alexezeder: All these problems started with `mutable`, GCC thinks that `mutable formatter<T, Char> fmt` in `spec_field` structure cannot be used at compile-time. As I can see Clang <7.0 also thinks the same way, this is why the side effect of provided PR is that Clang <7.0 with C++17 support can also use `FMT_COMPILE` with specs.

I failed to get a small code example which raises this error, I tried to create a failing example by recreating all significant parts of the error stack trace, but strangely [it works](https://godbolt.org/z/asa8K6). But I can present you the huge example with this error, this example is created by combining all needed headers of {fmt} together (version from this PR, except `compile.h`) and by making everything needed as `constexpr` in `compile.h` without `mutable`-related changes. Here it is on Compiler Explorer: https://godbolt.org/z/n1Kz38

Since the behaviour of GCC 10.2 and Clang <7.0 is kind of similar, I also tried to recreate the error with Clang 6.0 with my small code example that IMHO should fail - [no luck](https://godbolt.org/z/xvdMaf), so looks like I'm missing something here. But of course, it fails to compile with a bit modified huge example where all {fmt} headers are combined: https://godbolt.org/z/P17TGG

So, this became non-const to eliminate `mutable`, `const CompiledFormat& cf` became non-const references because method `format` is no more `const`, `constexpr auto compiled` became just `auto compiled` because const object cannot be passed via non-const reference. This is my explanation for why I made these changes, of course, I can miss some more elegant solution, especially considering that I failed to recreate this error in a small example. In case it's true and more elegant solution exist I'll be glad to implement it.

include/fmt/core.h
```diff
@@ -938,9 +938,9 @@ struct arg_data<T, Char, NUM_ARGS, 0> {
   T args_[NUM_ARGS != 0 ? NUM_ARGS : +1];
 
   template <typename... U>
-  FMT_INLINE arg_data(const U&... init) : args_{init...} {}
-  FMT_INLINE const T* args() const { return args_; }
-  FMT_INLINE std::nullptr_t named_args() { return nullptr; }
+  FMT_CONSTEXPR arg_data(const U&... init) : args_{init...} {}
+  FMT_CONSTEXPR const T* args() const { return args_; }
+  FMT_CONSTEXPR std::nullptr_t named_args() { return nullptr; }
```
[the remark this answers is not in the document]
alexezeder: Replacing `FMT_CONSTEXPR` with `constexpr` here (as here https://github.com/alexezeder/fmt/commit/3c9c7c798c789d029876383b459c0206859ab18c) leads to these failed actions:

- Linux - https://github.com/alexezeder/fmt/actions/runs/422137428
- MacOS - https://github.com/alexezeder/fmt/actions/runs/422137426

TLDR:
```
error: 'constexpr' non-static member function will not be implicitly 'const' in C++14; add 'const' to avoid a change in behavior [-Werror,-Wconstexpr-not-const]
```

include/fmt/core.h
alexezeder: Replacing `FMT_CONSTEXPR` with `constexpr` here (as here https://github.com/alexezeder/fmt/commit/1901008dd977eb29f90f6e1fbb01abf4a47f6d0b) leads to these failed actions:

- Linux - https://github.com/alexezeder/fmt/actions/runs/422139689
- MacOS - https://github.com/alexezeder/fmt/actions/runs/422139686

TLDR:
```
/home/runner/work/fmt/fmt/include/fmt/core.h:1489:24: error: 'constexpr' non-static member function will not be implicitly 'const' in C++14; add 'const' to avoid a change in behavior [-Werror,-Wconstexpr-not-const]
  constexpr format_arg arg(basic_string_view<char_type> name) {
                       ^
                                                              const
/home/runner/work/fmt/fmt/include/fmt/core.h:1495:35: error: 'constexpr' non-static member function will not be implicitly 'const' in C++14; add 'const' to avoid a change in behavior [-Werror,-Wconstexpr-not-const]
  constexpr detail::error_handler error_handler() { return {}; }
                                  ^
                                                  const
/home/runner/work/fmt/fmt/include/fmt/core.h:1499:22: error: 'constexpr' non-static member function will not be implicitly 'const' in C++14; add 'const' to avoid a change in behavior [-Werror,-Wconstexpr-not-const]
  constexpr iterator out() { return out_; }
                     ^
                           const
/home/runner/work/fmt/fmt/include/fmt/core.h:1506:32: error: 'constexpr' non-static member function will not be implicitly 'const' in C++14; add 'const' to avoid a change in behavior [-Werror,-Wconstexpr-not-const]
  constexpr detail::locale_ref locale() { return loc_; }
                               ^
                                        const
```

include/fmt/core.h
```diff
@@ -1073,17 +1073,17 @@ template <typename Context> class value {
...
   FMT_INLINE value(float val) : float_value(val) {}
   FMT_INLINE value(double val) : double_value(val) {}
   FMT_INLINE value(long double val) : long_double_value(val) {}
-  FMT_INLINE value(bool val) : bool_value(val) {}
-  FMT_INLINE value(char_type val) : char_value(val) {}
-  FMT_INLINE value(const char_type* val) { string.data = val; }
-  FMT_INLINE value(basic_string_view<char_type> val) {
+  constexpr FMT_INLINE value(bool val) : bool_value(val) {}
+  constexpr FMT_INLINE value(char_type val) : char_value(val) {}
+  FMT_CONSTEXPR value(const char_type* val) : string() { string.data = val; }
```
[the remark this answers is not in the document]
alexezeder: Done with `is_constant_evaluated()`

include/fmt/core.h
alexezeder: Yes, I missed that `FMT_INLINE` has an attribute inside, and this line can use `FMT_INLINE` **and** `constexpr`. But those lines where `FMT_INLINE` was replaced by `FMT_CONSTEXPR(20)` will have double `inline` specifier, which makes some compilers [unhappy](https://godbolt.org/z/8Exfo8) and some [very unhappy](https://godbolt.org/z/v8xq8e).

vitaut:
> I suggest adding define for this whole formatting at compile-time feature in the same manner as with other features, so if it's not predefined then set it by the expression we already have __cplusplus >= 202002L. Is it a good idea, should I do that in this PR or not?

We shouldn't be adding a macro for this. Instead `__cplusplus >= 202002L` should be replaced with a more specific check.

include/fmt/compile.h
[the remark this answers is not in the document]
vitaut: I think the correct solution is to make `formatter::format` const and keep this `format` function const.

include/fmt/core.h
```diff
@@ -283,6 +283,14 @@ struct monostate {};
 
 namespace detail {
 
+constexpr bool is_constant_evaluated() FMT_DETECTED_NOEXCEPT {
```
vitaut, line 286: This should be `FMT_NOEXCEPT`, not `FMT_DETECTED_NOEXCEPT`.

include/fmt/core.h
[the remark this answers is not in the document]
vitaut: OK, not a big deal.

include/fmt/core.h
```diff
@@ -1073,17 +1081,22 @@ template <typename Context> class value {
...
-  FMT_INLINE value(char_type val) : char_value(val) {}
-  FMT_INLINE value(const char_type* val) { string.data = val; }
-  FMT_INLINE value(basic_string_view<char_type> val) {
+  constexpr FMT_INLINE value(bool val) : bool_value(val) {}
+  constexpr FMT_INLINE value(char_type val) : char_value(val) {}
+  FMT_CONSTEXPR value(const char_type* val) {
+    string.data = val;
+    if (is_constant_evaluated()) {
+      string.size = {};
+    }
```
vitaut, line 1097: Remove braces for consistency with the rest of the code.

vitaut (changes requested):

include/fmt/compile.h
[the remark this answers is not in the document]
alexezeder: Ok, by in my opinion, there is a potential problem with this solution, with making `format()` function in formatters const - maybe if a some of formatters have `format()` function non-const on purpose then it cannot be used at compile-time. I just don't know:

1. why all or almost all formatters currently defined in {fmt} has non-const `format()` function?
2. how rare it is that `format()` function in custom formatters is non-const on purpose?

I mean, I can definitely try to make this function const for formatters which are used in test, but since you have better perspective here it will be helpful if you agree or disagree with me for this potential problem.

Actually, I don't even consider to go this way because some already created formatters would be unavailable at compile-time. On the other hand the code I have in this PR right now does not change formatters at all.

alexezeder:
> We shouldn't be adding a macro for this. Instead __cplusplus >= 202002L should be replaced with a more specific check.

Is it okay that this check will be used twice?

- https://github.com/fmtlib/fmt/blob/de42f13712cbd6476a252a0cb9e25feb8122ba4d/include/fmt/format.h#L285
- https://github.com/fmtlib/fmt/blob/de42f13712cbd6476a252a0cb9e25feb8122ba4d/test/compile-test.cc#L187

include/fmt/compile.h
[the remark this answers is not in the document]
vitaut: > why all or almost all formatters currently defined in {fmt} has non-const format() function?

No good reason, just less typing =).

> how rare it is that format() function in custom formatters is non-const on purpose?

I don't think it should necessary at all. `format` shouldn't mutate `formatter`, only apply format specifiers stored in it to the input.

> already created formatters would be unavailable at compile-time

We can fix this for all {fmt} formatters and later provide a workaround for user-defined ones.

include/fmt/core.h
alexezeder: Done

include/fmt/core.h
alexezeder: Done

include/fmt/compile.h
```diff
@@ -674,14 +674,14 @@ FMT_INLINE std::basic_string<typename S::char_type> format(const S&,
       return fmt::to_string(detail::first(args...));
   }
 #endif
-  constexpr auto compiled = detail::compile<Args...>(S());
+  auto compiled = detail::compile<Args...>(S());
```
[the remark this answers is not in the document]
alexezeder: Reverted back

include/fmt/compile.h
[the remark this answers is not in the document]
alexezeder: Ok, I reverted back most of the changes in `compile.h` and mark all of `formatter`s as const, at least those I found.

> and later provide a workaround for user-defined ones.

But there is no such thing as workarounds at compile-time 🙂, that why it's so cool to check your code just by making it execute at compile-time. Of course, I'm talking about workarounds that are about removing the constness of objects, not interface workarounds.

include/fmt/core.h
alexezeder: So looks like one of these macros shouldn't set `inline` and that would be really strange if this macro is going to be `FMT_INLINE`.

I can give an example of [Hedley](https://github.com/nemequ/hedley), it does not fallback to `inline` in the case where `constexpr` is not available - https://github.com/nemequ/hedley/blob/5e50f6b735aeb4e09e3b2ad3d1717db3c0613bfa/hedley.h#L1269-L1276
and I think it's right because I can always mark a function as `inline` with or without `constexpr`, no matter that `constexpr` implies `inline`.

In case you agree with me here I can create a separate PR to remove `inline` from `FMT_CONSTEXPR(20)` macro. Otherwise I just don't know how to fix these problems without introducing `FMT_CONSTEXPR(20)_NO_INLINE` macro, which probably would be also bad naming because there is `FMT_NOINLINE` macro that expands to additional attribute...

vitaut: > In case you agree with me here I can create a separate PR to remove inline from FMT_CONSTEXPR(20) macro.

Sure.

include/fmt/format.h
```diff
@@ -1143,12 +1138,14 @@ inline format_decimal_result<Iterator> format_decimal(Iterator out, UInt value,
...
-                         bool upper = false) {
+FMT_CONSTEXPR20 Char* format_uint(Char* buffer, UInt value, int num_digits,
+                                  bool upper = false) {
   buffer += num_digits;
   Char* end = buffer;
   do {
-    const char* digits = upper ? "0123456789ABCDEF" : data::hex_digits;
+    const char* digits = upper ? "0123456789ABCDEF"
+                               : (is_constant_evaluated() ? "0123456789abcdef"
+                                                          : data::hex_digits);
```
vitaut, line 1148: Can we make data::hex_digits constexpr instead?

include/fmt/format.h
```diff
@@ -1557,15 +1555,24 @@ FMT_NOINLINE OutputIt fill(OutputIt it, size_t n, const fill_t<Char>& fill) {
...
+  size_t left_padding = 0;
+  if (is_constant_evaluated()) {
+    const char left_padding_shifts[] = {31, 31, 0, 1, 0};
+    const char right_padding_shifts[] = {0, 31, 0, 1, 0};
+    auto* shifts =
+        align == align::left ? left_padding_shifts : right_padding_shifts;
+    left_padding = padding >> shifts[specs.align];
+  } else {
+    auto* shifts = align == align::left ? data::left_padding_shifts
+                                        : data::right_padding_shifts;
```
vitaut, line 1573: Same with shifts.

include/fmt/format.h
```diff
@@ -3527,7 +3540,22 @@ struct formatter<T, Char,
...
+    detail::handle_dynamic_spec<detail::precision_checker>(
+        specs.precision, specs.precision_ref, ctx);
+    using af = detail::arg_formatter<typename FormatContext::iterator,
+                                     typename FormatContext::char_type>;
+    return visit_format_arg(af(ctx, nullptr, &specs),
+                            detail::make_arg<FormatContext>(val));
+  }
+
+  template <typename FormatContext>
+  FMT_CONSTEXPR auto format(const T& val, FormatContext& ctx)
```
vitaut, line 3557: non-const overload is not needed, please remove.

vitaut (changes requested):

include/fmt/format.h
alexezeder: But I [tried](https://github.com/fmtlib/fmt/pull/2019#discussion_r527027972) to make them available both at compile-time and at runtime via macro definitions. Right now they are embedding into the static library, so I don't even touch them in this PR to not break the ABI. Maybe such variables should be declared as `constexpr` and be copied to that static data structure for runtime uses and still be chosen for compile-time uses.

include/fmt/format.h
alexezeder: The `const` one copies the` specs_` member into a temporary variable to be able to modify it, while the non-const one just uses this member field. So this change can introduce performance loss if this copy is heavy enough.

include/fmt/format.h
alexezeder: Also, we can go the same way as with `digits` - by implementing cast specially for compile-time. It's going to be more complex, but it's possible.

vitaut: I mean defining it as

```c++
static constexpr char hex_digits[] = "0123456789abcdef";
```
in the `basic_data` struct. This is a tiny piece of data so we don't need to worry about having it in the header.

include/fmt/format.h
vitaut: That's fine, we can optimize this later.

include/fmt/core.h
alexezeder: Done https://github.com/fmtlib/fmt/pull/2075

include/fmt/compile.h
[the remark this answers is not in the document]
vitaut: The workaround is to copy the formatter when needed.

vitaut:
Thank you!

alexezeder:
But there is still the problem with `FMT_INLINE`...

vitaut:
> But there is still the problem with FMT_INLINE.

I thought you are addressing this in a separate PR.

alexezeder:
Yes, but in this PR some of them is missing due to a need for `FMT_CONSTEXPR`, but I will probably revert them back in that separate PR.
I just thought that that PR would be merged before this 🙂.

PR #2065, opened 2020-12-15
Add test for formatting with no args

yeswalrus:
Demonstrates a bug reported in https://github.com/fmtlib/fmt/issues/2039.

The code involved here is a bit beyond me & the error message produced is rather inscrutable

### DIFF
--- a/test/format-test.cc
+++ b/test/format-test.cc
@@ -1822,6 +1822,7 @@ static FMT_CONSTEXPR_DECL const char static_no_null[2] = {'{', '}'};
 static FMT_CONSTEXPR_DECL const wchar_t static_no_null_wide[2] = {'{', '}'};
 
 TEST(FormatTest, CompileTimeString) {
+  EXPECT_EQ("42", fmt::format(FMT_STRING("42")));
   EXPECT_EQ("42", fmt::format(FMT_STRING("{}"), 42));
   EXPECT_EQ(L"42", fmt::format(FMT_STRING(L"{}"), 42));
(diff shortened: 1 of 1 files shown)


alexezeder:
Looks like it's a GCC bug to me, because it's reproducible only on GCC with `-O2`+ optimization level and `-Warray-bounds` diagnostic enabled. I got this warning also on [Compiler Explorer](https://godbolt.org/z/jT9TzY) but only for ARM GCC, on the other hand, x86_64 GCC does not output any warning on Compiler Explorer, but of course it does that on GitHub Actions environment and also on my PC.
Somehow GCC manages to get a possibility of a negative value in `id` in this function during compilation:
https://github.com/fmtlib/fmt/blob/e737672614dfad3a6df23ffe3f2348fcfa4d3944/include/fmt/format.h#L3221-L3225
This happens probably because of some weird optimization as this function is not even invoked for the provided format string.
I managed to get rid of this warning by explicitly checking if the `id` value is negative then call `std::abort()` to _guard_ that array subscript in `on_format_specs` function for negative values:
```cpp
if (id < 0) std::abort();
return id < num_args ? parse_funcs_[id](context_) : begin;
```
But it can introduce a side-effect in runtime because of the additional check.

vitaut:
Good catch, thanks! Worked around in https://github.com/fmtlib/fmt/commit/3551f5d11876bf7167267df364d000eff47f416c.

alexezeder:
@vitaut, just for information, it's not specific for gcc **_10_** only, but looks like for all gcc versions, at least for 9.2, 7.3

vitaut:
Thanks for the info, @alexezeder.

PR #2067, opened 2020-12-16
Disable warning about format string

HazardyKnusperkeks:
Reported by MinGW/GCC 10

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -383,12 +383,21 @@ inline std::tm gmtime(
 namespace detail {
 inline size_t strftime(char* str, size_t count, const char* format,
                        const std::tm* time) {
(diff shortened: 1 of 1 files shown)


vitaut (changes requested):
Thanks for the PR.

It's better to suppress the warning in the same way it is done for printf https://github.com/fmtlib/fmt/blob/e737672614dfad3a6df23ffe3f2348fcfa4d3944/include/fmt/format-inl.h#L2529 instead of using compiler-specific `#pragma warning`.

HazardyKnusperkeks:
Updated.

And for the record, the internal compiler error (which forbids auto) seems to be still in place with GCC/MinGW 10.2.

vitaut:
Thanks

PR #2068, opened 2020-12-18
Forward arguments to work with views

dkavolis:
Fix compile error when using `fmt::join(...)` as an argument to `fmt::ostream::print(...)`.


vitaut:
Thank you!

PR #2072, opened 2020-12-22
Fix std::byte formatting with compile-time API

alexezeder:
Right now formatting of `std::byte` using compile-time API like so:
```cpp
fmt::format(FMT_COMPILE("{}"), std::byte{42})
```
ends up with compilation error, [proof on Compiler Explorer](https://godbolt.org/z/5xqrvr). Because there are 2 `write` functions which satisfy for the passed type `std::byte`:
https://github.com/fmtlib/fmt/blob/5a37e182deee09a873777ff3f8e901d117e5f39b/include/fmt/format.h#L2114-L2120
https://github.com/fmtlib/fmt/blob/5a37e182deee09a873777ff3f8e901d117e5f39b/include/fmt/format.h#L2150-L2162

Also, I found `__cpp_lib_byte` feature macro, so all new checks are based on this macro instead of `__cplusplus >= 201703L` and I changed old checks accordingly.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -2112,8 +2112,14 @@ FMT_CONSTEXPR OutputIt write(OutputIt out, T value) {
 }
 
 template <typename Char, typename OutputIt, typename T,
+#ifdef __cpp_lib_byte
+          FMT_ENABLE_IF(std::is_enum<T>::value &&
+                        !std::is_same<T, Char>::value &&
+                        !std::is_same<T, std::byte>::value)>
+#else
           FMT_ENABLE_IF(std::is_enum<T>::value &&
                         !std::is_same<T, Char>::value)>
+#endif
 FMT_CONSTEXPR OutputIt write(OutputIt out, T value) {
   return write<Char>(
       out, static_cast<typename std::underlying_type<T>::type>(value));
@@ -3562,7 +3568,7 @@ FMT_FORMAT_AS(Char*, const Char*);
 FMT_FORMAT_AS(std::basic_string<Char>, basic_string_view<Char>);
 FMT_FORMAT_AS(std::nullptr_t, const void*);
 FMT_FORMAT_AS(detail::std_string_view<Char>, basic_string_view<Char>);
-#if __cplusplus >= 201703L
+#ifdef __cpp_lib_byte
 FMT_FORMAT_AS(std::byte, unsigned);
 #endif
 
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -2111,12 +2111,14 @@ FMT_CONSTEXPR OutputIt write(OutputIt out, T value) {
   return base_iterator(out, it);
 }
 
+// FMT_ENABLE_IF() condition separated to workaround MSVC bug
 template <
     typename Char, typename OutputIt, typename T,
-    FMT_ENABLE_IF(
+    bool check =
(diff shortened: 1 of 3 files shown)


include/fmt/format.h
```diff
@@ -2112,8 +2112,14 @@ FMT_CONSTEXPR OutputIt write(OutputIt out, T value) {
...
 
 template <typename Char, typename OutputIt, typename T,
+#ifdef __cpp_lib_byte
+          FMT_ENABLE_IF(std::is_enum<T>::value &&
+                        !std::is_same<T, Char>::value &&
+                        !std::is_same<T, std::byte>::value)>
+#else
           FMT_ENABLE_IF(std::is_enum<T>::value &&
                         !std::is_same<T, Char>::value)>
+#endif
```
vitaut, line 2122: This should have an additional SFINAE check `mapped_type_constant<T, basic_format_context<OutputIt, Char>>::value != type::custom_type` instead of the check if T is byte.

vitaut (changes requested):
Thanks for the PR.

alexezeder:
Using some tricks I was able to get a precise location of the line that fails format-test on MSVC, here it is:
https://github.com/fmtlib/fmt/blob/044661470666ec07f85ddc894a849a7cdfc501e6/test/format-test.cc#L1992-L1993

This check fails with these values:
```
Value of: fmt::to_string(test_value)
Actual: "false"
Expected: "0"
```

I don't have MSVC to check what is going on there - why a value of `enum` converted as `bool`. Also, I was able to get on Github Actions that this function is actually called 😐:
https://github.com/fmtlib/fmt/blob/044661470666ec07f85ddc894a849a7cdfc501e6/include/fmt/format.h#L2125-L2128
from here:
https://github.com/fmtlib/fmt/blob/044661470666ec07f85ddc894a849a7cdfc501e6/include/fmt/format.h#L3813-L3818
while all other compilers invoke this function from `to_string` above:
https://github.com/fmtlib/fmt/blob/044661470666ec07f85ddc894a849a7cdfc501e6/include/fmt/format.h#L2114-L2123

I'm pretty confused about that so I think it's better if someone who has MSVC would check that because I just don't see any reason why this should happen. So probably this PR can be closed because I'm unable to fix that.

vitaut:
> So probably this PR can be closed because I'm unable to fix that.

OK, closing then until someone comes up with an msvc workaround.

alexezeder:
I found 2 workarounds for this problem, both of them you can see here - https://godbolt.org/z/nMYj69. I even tested one of them here https://github.com/alexezeder/fmt/commit/a3b1c0756abd0edf1796dd3c2c11587670351a24 and it [works](https://github.com/alexezeder/fmt/actions/runs/442932924).
IMHO it's definitely an MSVC bug, so I should probably report this bug to them, but I won't because I didn't find a way to do that without signing in using a Microsoft account.

alexezeder:
In case one of the proposed workarounds is okay, then this PR can be reopened, and then I can add the selected workaround here.
And probably an issue can be closed in this case.

vitaut:
> In case one of the proposed workarounds is okay, then this PR can be reopened, and then I can add the selected workaround here.

Sure, thanks for looking into this. I suggest going with the first workaround to avoid introducing unnecessary symbols.

vitaut:
Thank you!

PR #2073, opened 2020-12-23
Add cwchar to format.h for std::fputws

concatime:
- [x] I agree that my contributions are licensed under the {fmt} license, and agree to future changes to the licensing.

While trying to build fmtlib with EASTL, I realized that `cwchar` was missing:
```
include/fmt/format.h:3964:7: error: no member named 'fputws' in namespace 'std'; did you mean simply 'fputws'?
  if (std::fputws(buffer.data(), f) == -1)
```

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -37,6 +37,7 @@
 #include <cmath>
 #include <cstddef>  // std::byte
 #include <cstdint>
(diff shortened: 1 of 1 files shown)


vitaut:
Thank you!

concatime:
Always a pleasure :)

PR #2075, opened 2020-12-24
Remove fallback to inline specifier from FMT_CONSTEXPR(20) macro

alexezeder:
As it was discussed here - https://github.com/fmtlib/fmt/pull/2056#discussion_r542988874, an fallback to the `inline` inside `FMT_CONSTEXPR(20)` macro definitions is not always a good solution. Especially when `FMT_CONSTEXPR(20)` macro should be used together with `FMT_INLINE` macro, in C++11 they both expand to `inline` specifier.

In this PR I removed that fallback to `inline` and placed `inline` specifiers where it's needed (for non-template, not member or friend defined in class/struct/union functions), so everything should stay the same in C++11. When `FMT_CONSTEXPR(20)` is expanded to `constexpr` all these manually-tuned functions will have `constexpr inline` specifiers, which is okay.

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -456,27 +456,61 @@ namespace detail {
 template <typename Period> FMT_CONSTEXPR const char* get_units() {
   return nullptr;
 }
-template <> FMT_CONSTEXPR const char* get_units<std::atto>() { return "as"; }
-template <> FMT_CONSTEXPR const char* get_units<std::femto>() { return "fs"; }
-template <> FMT_CONSTEXPR const char* get_units<std::pico>() { return "ps"; }
-template <> FMT_CONSTEXPR const char* get_units<std::nano>() { return "ns"; }
-template <> FMT_CONSTEXPR const char* get_units<std::micro>() { return "µs"; }
-template <> FMT_CONSTEXPR const char* get_units<std::milli>() { return "ms"; }
-template <> FMT_CONSTEXPR const char* get_units<std::centi>() { return "cs"; }
-template <> FMT_CONSTEXPR const char* get_units<std::deci>() { return "ds"; }
-template <> FMT_CONSTEXPR const char* get_units<std::ratio<1>>() { return "s"; }
-template <> FMT_CONSTEXPR const char* get_units<std::deca>() { return "das"; }
-template <> FMT_CONSTEXPR const char* get_units<std::hecto>() { return "hs"; }
-template <> FMT_CONSTEXPR const char* get_units<std::kilo>() { return "ks"; }
-template <> FMT_CONSTEXPR const char* get_units<std::mega>() { return "Ms"; }
-template <> FMT_CONSTEXPR const char* get_units<std::giga>() { return "Gs"; }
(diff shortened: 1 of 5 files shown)


include/fmt/chrono.h
```diff
@@ -456,27 +456,61 @@ namespace detail {
...
-template <> FMT_CONSTEXPR const char* get_units<std::deca>() { return "das"; }
-template <> FMT_CONSTEXPR const char* get_units<std::hecto>() { return "hs"; }
-template <> FMT_CONSTEXPR const char* get_units<std::kilo>() { return "ks"; }
-template <> FMT_CONSTEXPR const char* get_units<std::mega>() { return "Ms"; }
-template <> FMT_CONSTEXPR const char* get_units<std::giga>() { return "Gs"; }
-template <> FMT_CONSTEXPR const char* get_units<std::tera>() { return "Ts"; }
-template <> FMT_CONSTEXPR const char* get_units<std::peta>() { return "Ps"; }
-template <> FMT_CONSTEXPR const char* get_units<std::exa>() { return "Es"; }
-template <> FMT_CONSTEXPR const char* get_units<std::ratio<60>>() {
+template <> FMT_CONSTEXPR inline const char* get_units<std::atto>() {
```
vitaut, line 459: Can we remove `inline` from all these functions? Ideally all of this should be rewritten without template specializations but this is out of scope of the current PR.

test/format-test.cc
```diff
@@ -2367,13 +2367,13 @@ struct test_error_handler {
   }
 };
 
-FMT_CONSTEXPR size_t len(const char* s) {
+FMT_CONSTEXPR inline size_t len(const char* s) {
```
vitaut, line 2370: inline is not needed here.

test/format-test.cc
```diff
@@ -2367,13 +2367,13 @@ struct test_error_handler {
...
   size_t len = 0;
   while (*s++) ++len;
   return len;
 }
 
-FMT_CONSTEXPR bool equal(const char* s1, const char* s2) {
+FMT_CONSTEXPR inline bool equal(const char* s1, const char* s2) {
```
vitaut, line 2376: same

vitaut (changes requested):
Thanks for the PR. Mostly looks good, just a few minor comments inline =).

include/fmt/chrono.h
alexezeder: It causes multiple definitions of these functions.
Maybe a test for multiple definitions can be added, which would include all headers in two object files and link them together. Because now I have to include `chrono.h` into `util.h` in the `test` directory to get this error.

test/format-test.cc
alexezeder: Reverted back

test/format-test.cc
alexezeder: Reverted back

include/fmt/chrono.h
vitaut: Makes sense. I don't think the multiple definition test is necessary.

PR #2081, opened 2020-12-31
Correct a typo on syntax.rst (documentation)

0x416c69:

Thank you for this awesome library, I was just reading the documentation until I've noticed this little typo.

tm_mon is supposed to be 8 here.

### DIFF
--- a/doc/syntax.rst
+++ b/doc/syntax.rst
@@ -444,7 +444,7 @@ Using type-specific formatting::
 
    auto t = tm();
    t.tm_year = 2010 - 1900;
-   t.tm_mon = 6;
+   t.tm_mon = 8;
    t.tm_mday = 4;
    t.tm_hour = 12;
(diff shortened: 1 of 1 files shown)


doc/syntax.rst
```diff
@@ -444,7 +444,7 @@ Using type-specific formatting::
 
    auto t = tm();
    t.tm_year = 2010 - 1900;
-   t.tm_mon = 6;
+   t.tm_mon = 8;
```
vitaut, line 447: I think it should be 7 because tm_mon starts from 0: https://godbolt.org/z/b9fv81

vitaut (changes requested):
Thanks for the PR!

doc/syntax.rst
0x416c69: Yeah, you're right.

Issue #1381, opened 2019-10-30
Extend locale-specific numeric formatting


vitaut:
As discussed in #1372 and elsewhere, consider adding a locale modifier to floating-point format specs, e.g.

```c++
auto s = fmt::format("{:.2Lf}\n", 7975.3569);
// s == 7975.36 - decimal point taken from locale
```

9737551526:
> @vitaut change the name and address @calumr 

> 

> Gave me `a```

peter-b:
Please implement [P1892R1 Extended locale-specific presentation specifiers for `std::format`](https://isocpp.org/files/papers/P1892R1.pdf), since it was voted into the C++ working draft yesterday.

peter-b:
@vitaut If fmt implements the `L` modifier, am I correct in thinking that it will need to continue to implement `n` conversions for backwards compatibility?

vitaut:
Yes, but it can be phased out in stages: first add a macro to disable it, then announce that it's being deprecated, then switch the default. Will take multiple release cycles.

jtorresfabra:
What would be the correct way to workaround it in the meantime? Meaning I would like to use both Locale and f presentation specifier. Any advice?

vitaut:
You can format with `f` and then replace `.` with a decimal separator of your choice.

vitaut:
Implemented in https://github.com/fmtlib/fmt/commit/e4f2cf455e4ff3a065ff931e40b50dae182b4d6f.

PR #2095, opened 2021-01-15
Move default decimal point constant to inline function

a212:

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -1807,6 +1807,11 @@ struct big_decimal_fp {
   int exponent;
 };
 
+template <typename Char>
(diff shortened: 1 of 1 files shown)


vitaut:
`.` is self-descriptive and is not going to change so the only benefit that I see is avoiding one cast. I don't think it's worth adding a new function.

a212:
It'll allow to change this constant in only one place.
For example:
1) I want to use only comma as decimal point in my project.
2) I wish to change the default decimal point without using locale (currently I can't change decimal point for 'f' modifier)

vitaut:
There are no plans to change the default. The `f` specifier will be supported as part of #1381.

Issue #2098, opened 2021-01-16
Build issue: let user decide the MASTER_PROJECT variable


randomizedthinking:
This is a small question on building the fmt library. In my use case, fmt is included as a git submodule, and I would like to build and install fmt automatically along with the main project. Yet, the current fmt CMakefile would not allow me to do it because the `MASTER_PROJECT` variable in CMakeLists.txt is determined in the build time.

The solution would be straightforward: define a `FMT_MASTER_PROJECT` and check whether it is defined before setting it up. If you think it is worthwhile to have, I can submit a PR.

```cmake
if(NOT DEFINED FMT_MASTER_PROJECT)
    if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
        set(FMT_MASTER_PROJECT ON)
    else()
        set(FMT_MASTER_PROJECT OFF)
    endif()
endif()
```

vitaut:
Sounds reasonable. Please submit a PR to make `MASTER_PROJECT` configurable.

PR #2100, opened 2021-01-17
Add build variable: FMT_MASTER_PROJECT

randomizedthinking:
The build variable FMT_MASTER_PROJECT is added so user can build and install fmt library even if the project is included as a submodule. See discussion: #2098.

The PR is slightly updated to address two change requests.


vitaut:
Thank you!

PR #2107, opened 2021-01-19
Add missing detail namespace

yeswalrus:
This is a minor fix for an issue I discovered in one of my own projects that for some reason I cannot seem to reproduce it in ranges-test.cc. I'm not sure *why* exactly this isn't having any troubles internally, but it does definitely appear to be invalid syntax.


vitaut:
Thank you!

Issue #2078, opened 2020-12-28
Feedback needed: compile-time API improvements


alexezeder:
What is it about?
----

I have a portion of ideas about compiling format string at compile-time (not formatting at compile-time since I kinda stuck there), maybe some of these ideas were proposed before.

Since there is still no support for `arg_id` I was thinking that maybe I can provide full support of parsing formatting string. And by starting with that I ended up with some not so bad ideas for {fmt} in general that I want to share to get some feedback.

Ideas
---

1. First of all, I was thinking about the idea to fallback on runtime formatting in general and I think it's not a great idea honestly. Because I, as a user of {fmt}, expect to get some formatters structure prepared at compile-time by using `FMT_COMPILE` or `_cf`, so all my calls to {fmt} can be just optimized-out or, let's say, be well-prepared. But with this kind of fallback, I would get runtime calls to the static library (w/o `FMT_HEADER_ONLY`) or somehow optimized inlined calls (with `FMT_HEADER_ONLY`). And all of this without any indication, which is not intuitive at all.

    But if formatting string compilation fully supports parsing then there is not any need for that fallback I guess.

2. Currently, `FMT_COMPILE` and `_cf` just mark formatting string so it can be compiled later inside of `fmt::format*` functions, so this code:

    ```cpp
    constexpr auto compiled_string = FMT_COMPILE("{} {}") /* or "{}"_cf */; // not actually compiled
    fmt::format(compiled_string, true, 42.0f);
    fmt::format(compiled_string, nullptr, 42);
    fmt::format(compiled_string, "foo", std::byte{42});
    ```

    parses formatting string 3 times, but it can parse it only once.

    In the case of parsing it only once, `compiled_string` would not have prepared formatters because there are no arguments passed, so no type information, but it would have compiled replacement fields and of course it would validate the passed formatting string. So the main task of `fmt::format*` functions here is to match passed arguments to the compiled ones and to prepare formatters for the compiled replacement fields.

    Of course, there are additional template instantiations with this approach, but string parsing is not the easiest operation too.

3. Support for the named arguments with compiled formatting string can be introduced with no runtime cost because we can get all argument names at compile-time. But it requires us to change `fmt::arg` to be able to hold a name as template parameter - `fmt::arg<"name">(42)`, it's possible since C++20, also `"name"_a` can produce templated version too in C++20 (like `_cf`).

[...]

PR #2111, opened 2021-01-22
Add full support of format string parsing in compile-time API

alexezeder:
Some time ago, I wrote that I would try to add support of named arguments in compile-time API [here](https://github.com/fmtlib/fmt/issues/2078#issuecomment-753662719). This took some time... especially when I realize that I can replace most of the functions written by me with functions that already exist in {fmt}.

Some points for the changes:

- [works with C++17](https://github.com/alexezeder/fmt/actions/runs/509533441) as [it did before](https://github.com/alexezeder/fmt/actions/runs/506035706), I tried to make additional implementation for C++20 with non-type template parameters to eliminate unnecessary template instantiations, but I ended up recreating every used type to make them structural and facing with [this fancy problem](https://godbolt.org/z/7hrzGb) I was not even thinking about, so there is no additional C++20 implementation for now (and maybe forever)
- custom (for compile-time API's `FMT_COMPILE` and `_cf`) parsing functions (`compile_format_string()` + `parse_tail()`) replaced with the usage of `format_string_compiler`, this gives full support of format string parsing, it's already written and tested code, and that's nice
- one-step algorithm with recursion of `compile_format_string()` + `parse_tail()` functions replaced by two-steps algorithm (kinda):
  - first step is to prepare an array of `format_part` using `format_string_compiler`
  - second step is to create the same compiled format object as it was before (with recursion of `concat<L, R>`)
- manual indexing with `{0}` and automatic indexing with `{name}` usage works exactly as it works in runtime API (thanks to `format_string_compiler`)
- named replacement fields are compiled into their own `compiled_format` types (`runtime_named_field` and `runtime_named_spec_field`), but these types just hold name or name+specs to be able to use `write()` function or to create `formatter` for specified specs
- single `*index` argument kind divided into `index_auto` and `index_manual` to have meta info about how the argument was created, maybe it's not the best solution, but this meta info needed for proper `spec_field` creation
- tests added

I was afraid of making compilation time significantly longer by using the current approach, so I made a very straightforward test by compiling `compile-test.cc` with the same tests (`master` version of this file) several times, here are average times (mean ± σ, big thanks to [hyperfine](https://github.com/sharkdp/hyperfine)):
Compiler | master | this PR
------------ | --------- | --------------
clang++-11 | 2.831 s ±  0.016 s | 2.894 s ±  0.019 s
clang++-11 `-g` | 3.373 s ±  0.026 s | 3.425 s ±  0.022 s
clang++-11 `-O3 -DNDEBUG` | 10.632 s ±  0.016 s | 10.601 s ±  0.042 s
g++-10 | 4.728 s ±  0.024 s | 4.819 s ±  0.023 s
g++-10 `-g` | 5.299 s ±  0.024 s | 5.403 s ±  0.032 s
g++-10 `-O3 -DNDEBUG` | 11.787 s ±  0.028 s | 11.978 s ±  0.033 s

So, there is some compilation time increase (in most cases), but I'm not sure how bad it is.

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -1090,7 +1090,7 @@ struct formatter<std::chrono::duration<Rep, Period>, Char> {
 
     template <typename Id> FMT_CONSTEXPR arg_ref_type make_arg_ref(Id arg_id) {
       context.check_arg_id(arg_id);
-      return arg_ref_type(arg_id);
+      return arg_ref_type(arg_id, true);
     }
 
     FMT_CONSTEXPR arg_ref_type make_arg_ref(basic_string_view<Char> arg_id) {
@@ -1099,7 +1099,7 @@ struct formatter<std::chrono::duration<Rep, Period>, Char> {
     }
 
     FMT_CONSTEXPR arg_ref_type make_arg_ref(detail::auto_id) {
-      return arg_ref_type(context.next_arg_id());
+      return arg_ref_type(context.next_arg_id(), false);
     }
 
     void on_error(const char* msg) { FMT_THROW(format_error(msg)); }
--- a/include/fmt/compile.h
+++ b/include/fmt/compile.h
@@ -649,14 +649,11 @@ template <typename CompiledString> constexpr auto get_parts_array() {
   return result;
 }
 
-template <typename CompiledString>
-constexpr static auto compiled_parts = get_parts_array<CompiledString>();
-
 template <typename Args, typename CompiledString, size_t part_index>
 constexpr auto make_compiled_part() {
   using char_type = typename CompiledString::char_type;
 
-  constexpr format_part part = compiled_parts<CompiledString>[part_index];
+  constexpr auto part = get_parts_array<CompiledString>()[part_index];
   if constexpr (part.part_kind == decltype(part)::kind::arg_index_auto ||
                 part.part_kind == decltype(part)::kind::arg_index_manual) {
     using id_type = get_type<part.val.arg_index, Args>;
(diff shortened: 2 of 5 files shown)


alexezeder:
Hmm... looks like there is a problem with parsing specs for custom types because the default specs parser (`parse_format_specs()` function) is invoked even for custom types in `format_string_compiler::on_format_specs()`. For example:
```cpp
struct test_struct {
  int value;
};

template <> struct formatter<test_struct> {
  enum class output_type { two, four } type{};

  constexpr auto parse(format_parse_context& ctx) {
    auto it = ctx.begin(), end = ctx.end();
    while (it != end && *it != '}') {
      ++it;
    }
    auto spec = string_view(ctx.begin(), it - ctx.begin());
    if (compare(spec, "t")) { // custom compare() cause fmt::string_view comp operators are not constexpr
      type = output_type::two;
    } else if (compare(spec, "four symbols, please")) {
      type = output_type::four;
    }
    return it;
  }

  auto format(const test_struct& object, FormatContext& ctx) const {
    return format_to(ctx.out(), type == output_type::two ? "{:>2}" : "{:>4}", object.value);
  }
};

int main() {
  format("{:t}", test_struct{42}); // OK - "42"
  format("{:four symbols, please}", test_struct{42}); // OK - "  42"
  format(FMT_COMPILE("{:t}"), test_struct{42}); // OK - "42", cause 't' works as type in default specs parser
  format(FMT_COMPILE("{:four symbols, please}"), test_struct{42}); // compilation error
  return 0;
}
```

I will create a new variant of `format_string_compiler` then, keeping the old one untouched.

But for now, I'm converting this PR to a draft.

alexezeder:
Despite the fact that I found a workaround for the problem described above, it's so bad that I would probably take the initial approach with keeping that recursion of `compile_format_string()` + `parse_tail()` functions to produce compiled format structure.

The workaround is pretty straightforward - pass a tuple-like object which will have optional formatters for each passed argument.
Something like `tuple<optional<formatter<int>>, optional<formatter<float>>>` for `int, float` arguments. This tuple-like object will be filled for each spec'ed argument in `format_string_compiler::on_format_specs()` to be later used in combining of `concat`-based compiled format.

Should I say that it looks really bad? 😏 

Since _I think_ it can be done better (with a custom string pass, as it is right now) I am closing this PR.

PR #2114, opened 2021-01-23
Make ranges-test available with C++11

alexezeder:
I noticed here this strange check:
https://github.com/fmtlib/fmt/blob/acef0bb51ade288bf7930d1b74d1338f49f9f900/test/ranges-test.cc#L16-L18
So I decided to do it right - by using `__cpp_if_constexpr` feature macro. But later I realized that since there is no mention in docs that ranges are supported only with C++14 (or C++17 because of `constexpr if`) it's strange to me that test doesn't check C++11, especially when this check is not necessary at all with only few changes in the test.
With this PR `ranges-test` compiles with C++11 on probably most ancient supported compilers - gcc-4.8 and clang-3.5 (couldn't test 3.4).

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -381,7 +381,7 @@ template <typename T> inline T* make_checked(T* p, size_t) { return p; }
 #endif
 
 template <typename Container, FMT_ENABLE_IF(is_contiguous<Container>::value)>
-#if FMT_CLANG_VERSION
+#if FMT_CLANG_VERSION >= 307
 __attribute__((no_sanitize("undefined")))
 #endif
 inline checked_ptr<typename Container::value_type>
@@ -940,8 +940,8 @@ template <typename T = void> struct FMT_EXTERN_TEMPLATE_API basic_data {
   static const char reset_color[5];
   static const wchar_t wreset_color[5];
   static const char signs[];
-  static constexpr const char left_padding_shifts[] = {31, 31, 0, 1, 0};
-  static constexpr const char right_padding_shifts[] = {0, 31, 0, 1, 0};
+  static constexpr const char left_padding_shifts[5] = {31, 31, 0, 1, 0};
+  static constexpr const char right_padding_shifts[5] = {0, 31, 0, 1, 0};
 
   // DEPRECATED! These are for ABI compatibility.
   static const uint32_t zero_or_powers_of_10_32[];
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -368,7 +368,7 @@ template <typename Char, typename... T> struct tuple_arg_join : detail::view {
   basic_string_view<Char> sep;
 
   tuple_arg_join(const std::tuple<T...>& t, basic_string_view<Char> s)
-      : tuple{t}, sep{s} {}
+      : tuple(t), sep{s} {}
 };
 
 template <typename Char, typename... T>
--- a/test/ranges-test.cc
+++ b/test/ranges-test.cc
@@ -18,11 +18,16 @@
 
 #include "gtest.h"
 
-#if FMT_GCC_VERSION == 0 || FMT_GCC_VERSION >= 601
-#  define FMT_ENABLE_C_STYLE_ARRAY_TEST
+#if !FMT_GCC_VERSION || FMT_GCC_VERSION >= 601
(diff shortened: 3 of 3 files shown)


include/fmt/format.h
```diff
@@ -381,7 +381,7 @@ template <typename T> inline T* make_checked(T* p, size_t) { return p; }
 #endif
 
 template <typename Container, FMT_ENABLE_IF(is_contiguous<Container>::value)>
-#if FMT_CLANG_VERSION
+#if FMT_CLANG_VERSION >= 307
 __attribute__((no_sanitize("undefined")))
```
alexezeder, line 385: Looks like there is no `no_sanitize` attribute prior to Clang 3.7

include/fmt/format.h
```diff
@@ -940,8 +940,8 @@ template <typename T = void> struct FMT_EXTERN_TEMPLATE_API basic_data {
   static const char reset_color[5];
   static const wchar_t wreset_color[5];
   static const char signs[];
-  static constexpr const char left_padding_shifts[] = {31, 31, 0, 1, 0};
-  static constexpr const char right_padding_shifts[] = {0, 31, 0, 1, 0};
+  static constexpr const char left_padding_shifts[5] = {31, 31, 0, 1, 0};
+  static constexpr const char right_padding_shifts[5] = {0, 31, 0, 1, 0};
```
alexezeder, line 944: Some GCC versions fail to determine the size of this array - https://godbolt.org/z/f4s98c

include/fmt/ranges.h
```diff
@@ -368,7 +368,7 @@ template <typename Char, typename... T> struct tuple_arg_join : detail::view {
   basic_string_view<Char> sep;
 
   tuple_arg_join(const std::tuple<T...>& t, basic_string_view<Char> s)
-      : tuple{t}, sep{s} {}
+      : tuple(t), sep{s} {}
```
alexezeder, line 371: GCC 4.8 fails to initialize tuple ref by braces - https://godbolt.org/z/r7WMT1

test/ranges-test.cc
```diff
@@ -11,17 +11,18 @@
...
-    (defined(_MSVC_LANG) && _MSVC_LANG > 201402L && _MSC_VER >= 1910)
+#include "gtest.h"
 
-#  include <array>
-#  include <map>
-#  include <string>
-#  include <vector>
+#if FMT_GCC_VERSION == 0 || FMT_GCC_VERSION >= 601
+#  define FMT_ENABLE_C_STYLE_ARRAY_TEST
+#endif
```
alexezeder, line 23: It's strange but looks like GCC <6 unable to pass C-style array correctly, so {fmt} is unable right now to format them - https://godbolt.org/z/3jv87x

test/ranges-test.cc
```diff
@@ -96,15 +100,17 @@ TEST(RangesTest, JoinInitializerList) {
...
-    if constexpr (N == 0)
-      return i;
-    else if constexpr (N == 1)
-      return fmt::string_view{str};
+  template <size_t N> fmt::enable_if_t<N == 0, int32_t> get() const noexcept {
+    return i;
+  }
+  template <size_t N>
+  fmt::enable_if_t<N == 1, fmt::string_view> get() const noexcept {
+    return {str};
```
alexezeder, line 108: No C++14 `decltype(auto)` (probably was unneeded anyway) and C++17 `constexpr if`, only C++11 code

test/ranges-test.cc
```diff
@@ -125,7 +131,7 @@ TEST(RangesTest, FormatStruct) {
 
 TEST(RangesTest, FormatTo) {
   char buf[10];
-  auto end = fmt::format_to(buf, "{}", std::vector{1, 2, 3});
+  auto end = fmt::format_to(buf, "{}", std::vector<int>{1, 2, 3});
```
alexezeder, line 134: No C++17 template deduction

alexezeder (commented):
Some points for the PR.

include/fmt/format.h
vitaut: This is a gcc bug https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66921.

vitaut:
Looks great, thanks!

PR #2119, opened 2021-01-28
fix #2118: FMT_COMPILE did not work with tm formatter

### LINKED ISSUE #2118: `FMT_COMPILE` does not work with `tm` (fmt/chrono.h)
https://godbolt.org/z/b8PYK3

*Compiler*: x86-64 gcc (any version can trigger)
Flags: -std=gnu++2a -O2

Different from #2116, this one does not seem to be ESP32-specific.

The solution seems trivial. I will send a pull request shortly.

summivox:
Fixes #2118.


vitaut:
Thanks!

PR #2122, opened 2021-01-29
Fix ordering of install commands for CMake

ibaned:
the library itself needs to be installed before
the fmt-targets.cmake file is installed,
otherwise the installed targets file doesn't
actually point to the library using
IMPORTED_LOCATION


vitaut:
Thank you!

PR #2128, opened 2021-02-07
Fix fmt::localtime formatting not working in wide-char string contexts

yNeo:
Issue was caused by incorrectly using an explicit type for `parse_context` parameter instead of a templated type, which meant that the `parse` function would only work in regular string contexts.


vitaut:
Thanks. LGTM but could you add a test case in https://github.com/fmtlib/fmt/blob/master/test/chrono-test.cc?

vitaut:
Added a test and merged in https://github.com/fmtlib/fmt/commit/c5979d564e1ec40c383eaca610c8aa7cca1733a4.

PR #2129, opened 2021-02-07
Add full support of format string parsing in compile-time API

alexezeder:
Compile-time API functionality extended to support manual ordering and named arguments. Unlike my first attempt to do this in https://github.com/fmtlib/fmt/pull/2111, where I tried to use a format part array, here I'm just reusing that recursion of functions `compile_format_string()` and `parse_tail()`.

Some points for the changes:
* works with C++17 as it did before
* unlike https://github.com/fmtlib/fmt/pull/2111, this PR fully supports custom parsing in formatters
* manual indexing with `{0}` and automatic indexing with `{name}` work exactly as they work in the runtime API
* a switch of the argument indexing between automatic to manual or manual to automatic is detected at compile-time, and `static_assert`s fail with corresponding messages
* usage of named arguments with specs is unsupported because these arguments cannot provide type information unless their names are available at compile-time (I wrote about that with more details [here](https://github.com/fmtlib/fmt/issues/2078#issuecomment-769167144)), in case if named argument is used with specs in format string then `unknown_format()` is returned from string compilation procedure, thus we fallback to the runtime API for this string (but this fallback is currently broken)
* tests added

### DIFF
--- a/include/fmt/compile.h
+++ b/include/fmt/compile.h
@@ -469,28 +469,25 @@ template <typename Char> struct runtime_named_field {
   basic_string_view<Char> name;
 
   template <typename OutputIt, typename T>
-  constexpr static bool try_format_argument(OutputIt& end, OutputIt out,
+  constexpr static bool try_format_argument(OutputIt& out,
                                             basic_string_view<Char> arg_name,
                                             const T& arg) {
-    if constexpr (!is_named_arg<typename std::remove_cv<T>::type>::value) {
-      return false;
-    } else {
+    if constexpr (is_named_arg<typename std::remove_cv<T>::type>::value) {
       if (arg_name == arg.name) {
-        end = write<Char>(out, arg.value);
+        out = write<Char>(out, arg.value);
         return true;
       }
-      return false;
     }
+    return false;
   }
 
   template <typename OutputIt, typename... Args>
   constexpr OutputIt format(OutputIt out, const Args&... args) const {
-    auto end = out;
-    bool found = (try_format_argument(end, out, name, args) || ...);
+    bool found = (try_format_argument(out, name, args) || ...);
     if (!found) {
       throw format_error("argument with specified name is not found");
     }
-    return end;
+    return out;
   }
 };
 
@@ -570,7 +567,7 @@ template <typename T, typename Char> struct parse_specs_result {
   int next_arg_id;
 };
 
-constexpr int manual_indexing_id() { return -1; }
+constexpr int manual_indexing_id = -1;
 
 template <typename T, typename Char>
 constexpr parse_specs_result<T, Char> parse_specs(basic_string_view<Char> str,
@@ -580,7 +577,7 @@ constexpr parse_specs_result<T, Char> parse_specs(basic_string_view<Char> str,
   auto f = formatter<T, Char>();
   auto end = f.parse(ctx);
   return {f, pos + fmt::detail::to_unsigned(end - str.data()) + 1,
-          next_arg_id == 0 ? manual_indexing_id() : ctx.next_arg_id()};
+          next_arg_id == 0 ? manual_indexing_id : ctx.next_arg_id()};
 }
 
 template <typename Char> struct arg_id_handler {
@@ -629,12 +626,12 @@ constexpr auto compile_format_string(S format_str) {
     if constexpr (str[POS + 1] == '{') {
       return parse_tail<Args, POS + 2, ID>(make_text(str, POS, 1), format_str);
     } else if constexpr (str[POS + 1] == '}' || str[POS + 1] == ':') {
-      static_assert(ID != manual_indexing_id(),
+      static_assert(ID != manual_indexing_id,
                     "cannot switch from manual to automatic argument indexing");
       using id_type = get_type<ID, Args>;
       if constexpr (str[POS + 1] == '}') {
         constexpr auto next_id =
-            ID != manual_indexing_id() ? ID + 1 : manual_indexing_id();
+            ID != manual_indexing_id ? ID + 1 : manual_indexing_id;
         return parse_tail<Args, POS + 2, next_id>(
             field<char_type, id_type, ID>(), format_str);
       } else {
@@ -651,12 +648,12 @@ constexpr auto compile_format_string(S format_str) {
       static_assert(c == '}' || c == ':', "missing '}' in format string");
       if constexpr (arg_id_result.arg_id.kind == arg_id_kind::index) {
         static_assert(
-            ID == manual_indexing_id() || ID == 0,
+            ID == manual_indexing_id || ID == 0,
             "cannot switch from automatic to manual argument indexing");
         constexpr auto arg_index = arg_id_result.arg_id.val.index;
         using id_type = get_type<arg_index, Args>;
         if constexpr (c == '}') {
-          return parse_tail<Args, arg_id_end_pos + 1, manual_indexing_id()>(
+          return parse_tail<Args, arg_id_end_pos + 1, manual_indexing_id>(
               field<char_type, id_type, arg_index>(), format_str);
         } else if constexpr (c == ':') {
           constexpr auto result =
@@ -667,7 +664,7 @@ constexpr auto compile_format_string(S format_str) {
         }
       } else if constexpr (arg_id_result.arg_id.kind == arg_id_kind::name) {
         static_assert(
-            ID != manual_indexing_id(),
+            ID != manual_indexing_id,
             "cannot switch from manual to automatic argument indexing");
         if constexpr (c == '}') {
           return parse_tail<Args, arg_id_end_pos + 1, ID + 1>(
--- a/include/fmt/compile.h
+++ b/include/fmt/compile.h
@@ -454,9 +454,14 @@ template <typename Char, typename T, int N> struct field {
 
   template <typename OutputIt, typename... Args>
   constexpr OutputIt format(OutputIt out, const Args&... args) const {
-    // This ensures that the argument type is convertile to `const T&`.
-    const T& arg = get<N>(args...);
-    return write<Char>(out, arg);
+    if constexpr (is_named_arg<typename std::remove_cv<T>::type>::value) {
+      decltype(T::value) arg = get<N>(args...).value;
+      return write<Char>(out, arg);
+    } else {
+      // This ensures that the argument type is convertile to `const T&`.
+      const T& arg = get<N>(args...);
+      return write<Char>(out, arg);
+    }
   }
 };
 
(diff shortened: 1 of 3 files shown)


include/fmt/compile.h
```diff
@@ -463,6 +463,40 @@ template <typename Char, typename T, int N> struct field {
...
+                                            const T& arg) {
+    if constexpr (!is_named_arg<typename std::remove_cv<T>::type>::value) {
+      return false;
+    } else {
+      if (arg_name == arg.name) {
+        end = write<Char>(out, arg.value);
+        return true;
+      }
+      return false;
+    }
```
vitaut, line 483: Let's simplify this a bit:

```c++
    if constexpr (is_named_arg<typename std::remove_cv<T>::type>::value) {
      if (arg_name == arg.name) {
        end = write<Char>(out, arg.value);
        return true;
      }
    }
    return false;
```

include/fmt/compile.h
```diff
@@ -463,6 +463,40 @@ template <typename Char, typename T, int N> struct field {
 template <typename Char, typename T, int N>
 struct is_compiled_format<field<Char, T, N>> : std::true_type {};
 
+// A replacement field that refers to argument with name.
+template <typename Char> struct runtime_named_field {
+  using char_type = Char;
+  basic_string_view<Char> name;
+
+  template <typename OutputIt, typename T>
+  constexpr static bool try_format_argument(OutputIt& end, OutputIt out,
```
vitaut, line 472: I don't think we need both `end` and `out`. It's enough to pass `out` by reference.

include/fmt/compile.h
```diff
@@ -463,6 +463,40 @@ template <typename Char, typename T, int N> struct field {
...
+                                            basic_string_view<Char> arg_name,
+                                            const T& arg) {
+    if constexpr (!is_named_arg<typename std::remove_cv<T>::type>::value) {
+      return false;
+    } else {
+      if (arg_name == arg.name) {
```
vitaut, line 478: What if there multiple arguments with the same name?

include/fmt/compile.h
```diff
@@ -536,15 +570,51 @@ template <typename T, typename Char> struct parse_specs_result {
   int next_arg_id;
 };
 
+constexpr int manual_indexing_id() { return -1; }
```
vitaut, line 573: Why is it a function?

include/fmt/compile.h
```diff
@@ -536,15 +570,51 @@ template <typename T, typename Char> struct parse_specs_result {
...
   return {f, pos + fmt::detail::to_unsigned(end - str.data()) + 1,
-          ctx.next_arg_id()};
+          next_arg_id == 0 ? manual_indexing_id() : ctx.next_arg_id()};
+}
+
+template <typename Char> struct arg_id_handler {
+  constexpr void on_error(const char* message) { throw format_error(message); }
+
+  constexpr int on_arg_id() {
+    throw format_error("handler cannot be used for empty arg_id");
```
vitaut, line 590: "for empty arg_id" -> "with automatic indexing"

Also can this be an assert?

include/fmt/compile.h
```diff
@@ -536,15 +570,51 @@ template <typename T, typename Char> struct parse_specs_result {
...
+    return 0;
+  }
+
+  arg_ref<Char> arg_id;
+};
+
+template <typename Char> struct parse_arg_id_result {
+  arg_ref<Char> arg_id;
+  const Char* arg_id_end;
+};
```
vitaut, line 610: Can we pass `begin` by reference in `parse_arg_id` and avoid introducing this struct?

test/compile-test.cc
```diff
@@ -139,13 +139,136 @@ TEST(CompileTest, FormatWideString) {
...
+  }
+
+  template <typename FormatContext>
+  auto format(const test_custom_formattable&, FormatContext& ctx) const
+      -> decltype(ctx.out()) {
+    return format_to(ctx.out(), type == output_type::two ? "{:>2}" : "{:>4}",
+                     42);
+  }
+};
+FMT_END_NAMESPACE
```
vitaut, line 177: I suggest using one of the existing formatters such as duration formatter instead of introducing a new one here.

test/compile-test.cc
```diff
@@ -139,13 +139,136 @@ TEST(CompileTest, FormatWideString) {
...
+      return format_to(ctx.out(), "{:{}}{:{}}", 41, widths[0], 42, widths[1]);
+    } else if (amount == 3) {
+      return format_to(ctx.out(), "{:{}}{:{}}{:{}}", 41, widths[0], 42,
+                       widths[1], 43, widths[2]);
+    } else {
+      throw format_error("formatting error");
+    }
+  }
+};
+FMT_END_NAMESPACE
```
vitaut, line 227: Same here. duration formatter has dynamic field support.

vitaut (changes requested):
Looks great, thanks for another high quality PR!

include/fmt/compile.h
alexezeder: Nice! Applied this simplification.

include/fmt/compile.h
alexezeder: And this one too.

include/fmt/compile.h
alexezeder: Well... I guess the first one will be chosen 🙂. Probably I need to investigate how does runtime API handles that, but right now, it's just implemented in the easiest way possible.

include/fmt/compile.h
alexezeder: Made it a variable.
I was just confused by all that global variables linkage complexity that remains the same for constexpr variables. But it looks like all compilers would just optimize out this variable anyway.

include/fmt/compile.h
alexezeder: But it actually can be used for automatic indexing with named identifiers. Both runtime and (now) compile-time APIs keep automatic indexing when a named argument identifier is used. So it just cannot be used for an unnamed argument identifier in the automatic indexing mode, which this message is trying to say. By the way, this function cannot be used in normal conditions because the code that invokes this handler actually controls that this handler is used only for numeric or named arguments. As long as it's true, no one would see this message, but when someone breaks the parsing code, they will get this message.

> Also can this be an assert?

As I said, it just indicates an internal error, so the cause of this compile-time error can be everything not compile-time friendly. I saw several usages of `throw format_error(...)` and use it too.

include/fmt/compile.h
alexezeder: Hmm... it would be a reference to the pointer, or (_IMHO_ better) a pointer to the pointer, is it ok?

test/compile-test.cc
alexezeder: One problem here is that the `chrono::duration` formatter is not ready to be used with compile-time API because of that `format()` constness requirement. Should I update it in this PR or the separate one?

test/compile-test.cc
alexezeder: Actually, the previous one (about custom formatter) and this are not the same.
Yes, it has dynamic field support. But as far as I can see, it supports the same set of nested replacement fields as the default formatter, `{:{}.{}}`. So handling 2 dynamic fields for the default formatter would probably be enough to pass the test with `chrono::duration` formatter.
While this custom formatter has a custom syntax for nested replacement fields (non `{:{}.{}}`), and it has 3 of them. So handling default dynamic fields wouldn't be enough to pass the test.

include/fmt/compile.h
alexezeder: It looks like runtime API also chooses the first argument with the specified name - https://godbolt.org/z/8PWEvo.

include/fmt/compile.h
vitaut: I'm not entirely sure what you mean by "unnamed argument identifier". Both "{}" and "{:...}" denote automatic indexing which is why I'm suggesting this minor wording change. It doesn't matter much since it's an internal error but a bit more consistent with the wording elsewhere.

> it just indicates an internal error

Right and this is exactly why I'm suggesting to use an assert if possible. This will distinguish an internal error from a user error even though they both result in a compilation error. If assert doesn't work for some reason, then throw is OK.

include/fmt/compile.h
vitaut: Sure, I think reference is better unless it can be null.

test/compile-test.cc
vitaut: > Should I update it in this PR or the separate one?

This PR is OK since it should be a small change.

test/compile-test.cc
vitaut: I don't think we need to test the implementation of exotic formatter specializations here.

vitaut (changes requested):

alexezeder:
Actually, I'm going to make this PR a draft (yep, again 😄).
For some reason, I assumed that named arguments should be used with automatic indexing only, which is false for runtime API - https://godbolt.org/z/oTPqKK. This ended up with the wrong implementation here. I don't know why I just haven't checked all those cases before.
I need to make few changes to make the same behavior in compile-time API as runtime API has. Anyway, I think all comments will be valid even after applying these changes.

include/fmt/compile.h
alexezeder: Yes, "unnamed argument identifier" sounds a bit strange. 🙂
But the problem is probably in my wrong understanding of how named arguments work. After updating this PR (as I wrote [here](https://github.com/fmtlib/fmt/pull/2129#issuecomment-779475748)), this wording problem would be probably eliminated.

alexezeder: Done

include/fmt/compile.h
alexezeder: Actually, it's probably impossible because there is a need to have `arg_id_end` as a constexpr variable or, more importantly, begin has to be a non-constexpr variable in that case, but it should be used in constexpr context.

test/compile-test.cc
alexezeder: Done with the weirdest looking format string from chrono-test

test/compile-test.cc
alexezeder: Done with format string from chrono-test that uses dynamic specs

include/fmt/compile.h
```diff
@@ -670,8 +777,15 @@ FMT_INLINE std::basic_string<typename S::char_type> format(const S&,
 #ifdef __cpp_if_constexpr
   if constexpr (std::is_same<typename S::char_type, char>::value) {
     constexpr basic_string_view<typename S::char_type> str = S();
-    if constexpr (str.size() == 2 && str[0] == '{' && str[1] == '}')
-      return fmt::to_string(detail::first(args...));
+    if constexpr (str.size() == 2 && str[0] == '{' && str[1] == '}') {
+      auto first = detail::first(args...);
```
vitaut, line 781: This may introduce an extra copy. Please use `const auto&` instead of `auto&`.

vitaut (changes requested):
Two minor comments, otherwise looks good.

include/fmt/compile.h
```diff
@@ -454,15 +454,51 @@ template <typename Char, typename T, int N> struct field {
 
   template <typename OutputIt, typename... Args>
   constexpr OutputIt format(OutputIt out, const Args&... args) const {
-    // This ensures that the argument type is convertile to `const T&`.
-    const T& arg = get<N>(args...);
-    return write<Char>(out, arg);
+    if constexpr (is_named_arg<typename std::remove_cv<T>::type>::value) {
+      decltype(T::value) arg = get<N>(args...).value;
```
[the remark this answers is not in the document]
alexezeder: Done

include/fmt/compile.h
alexezeder: Done

vitaut:
Merged, thanks!

PR #2131, opened 2021-02-09
fmt::ptr: Support function pointers

mikecrowe:
Passing a function pointer to fmt::ptr results in:

 In file included from /home/mac/git/fmt/test/gmock/gmock.h:238,
                  from /home/mac/git/fmt/test/format-test.cc:31:
 .../fmt/test/format-test.cc: In member function ‘virtual void FormatterTest_FormatPointer_Test::TestBody()’:
 .../fmt/test/format-test.cc:1486:56: error: no matching function for call to ‘ptr(void (&)(int, double, std::__cxx11::string))’
              format("{}", fmt::ptr(function_pointer_test)));

Let's add an overload to support that usage.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -314,6 +314,7 @@ Utilities
 .. doxygenfunction:: fmt::ptr(const T *p)
 .. doxygenfunction:: fmt::ptr(const std::unique_ptr<T> &p)
 .. doxygenfunction:: fmt::ptr(const std::shared_ptr<T> &p)
+.. doxygenfunction:: fmt::ptr(T (*fn)(Args...))
 
 .. doxygenfunction:: fmt::to_string(const T &value)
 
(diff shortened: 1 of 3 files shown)


mikecrowe:
Using `detail::bit_cast` rather than `reinterpret_cast` appears to have been sufficient to make g++4.8 happy. However, I'm not sure what planet MSVC is on with:
```
D:\a\fmt\fmt\test\format-test.cc(1485,1): error C2668: 'fmt::v7::ptr': ambiguous call to overloaded function [D:\a\fmt\build\test\format-test.vcxproj]
D:\a\fmt\fmt\include\fmt/format.h(3742,60): message : could be 'const void *fmt::v7::ptr<void,int,double,std::string>(T (__cdecl *)(int,double,std::string))' [D:\a\fmt\build\test\format-test.vcxproj]
          with
          [
              T=void
          ]
D:\a\fmt\fmt\include\fmt/format.h(3735,42): message : or       'const void *fmt::v7::ptr<void(int,double,std::string)>(T (__cdecl *))' [D:\a\fmt\build\test\format-test.vcxproj]
          with
          [
              T=void (int,double,std::string)
          ]
D:\a\fmt\fmt\test\format-test.cc(1486,1): message : while trying to match the argument list '(overloaded-function)' [D:\a\fmt\build\test\format-test.vcxproj]
```
I suppose this means that the overload isn't even necessary for MSVC.

vitaut:
Thank you!

PR #2143, opened 2021-02-21
Fix fallback to runtime API from compile-time API

alexezeder:
Right now, when {fmt} fails to prepare compiled format for format string (using `FMT_COMPILE` or `_cf`) it also fails to report about that clearly. Here is an example - https://godbolt.org/z/G1ETdc.

There [were](https://github.com/fmtlib/fmt/pull/2061) some attempts to make the error a bit clearer, but as it [turned out](https://github.com/fmtlib/fmt/pull/2061#issuecomment-743775046) {fmt} should fall back to the runtime API in those cases. This PR adds that fallback.

Some points for the PR:
* test added
* there is a bit of code duplication introduced, because I decided that this code is clearer:
  ```cpp
  #ifdef __cpp_if_constexpr
    if constexpr (std::is_same<remove_cvref_t<decltype(compiled)>,
                               detail::unknown_format>()) {
      return format(static_cast<basic_string_view<typename S::char_type>>(S()),
                    std::forward<Args>(args)...);
    } else {
      return format(compiled, std::forward<Args>(args)...);
    }
  #else
    return format(compiled, std::forward<Args>(args)...);
  #endif
  ```
  then this one:
  ```cpp
  #ifdef __cpp_if_constexpr
    if constexpr (std::is_same<remove_cvref_t<decltype(compiled)>,
                               detail::unknown_format>()) {
      return format(static_cast<basic_string_view<typename S::char_type>>(S()),
                    std::forward<Args>(args)...);
    } else {
  #endif
      return format(compiled, std::forward<Args>(args)...);
  #ifdef __cpp_if_constexpr
    }
  #endif
  ```

### DIFF
--- a/include/fmt/compile.h
+++ b/include/fmt/compile.h
@@ -22,10 +22,6 @@
 #  endif
 #endif
 
-#ifndef FMT_ENABLE_FALLBACK_TO_RUNTIME_API
-#  define FMT_ENABLE_FALLBACK_TO_RUNTIME_API 1
-#endif
-
 FMT_BEGIN_NAMESPACE
 namespace detail {
 
@@ -708,12 +704,6 @@ constexpr auto compile(S format_str) {
     constexpr auto result =
         detail::compile_format_string<detail::type_list<Args...>, 0, 0>(
             format_str);
-#  if !FMT_ENABLE_FALLBACK_TO_RUNTIME_API
-    static_assert(!std::is_same<remove_cvref_t<decltype(result)>,
-                                detail::unknown_format>(),
-                  "format string is invalid for compile-time API, "
-                  "and fallback to runtime API is disabled");
-#  endif
     return result;
   }
 }
--- a/include/fmt/compile.h
+++ b/include/fmt/compile.h
@@ -844,25 +844,9 @@ template <typename OutputIt, typename S, typename... Args,
           FMT_ENABLE_IF(detail::is_compiled_string<S>::value)>
 format_to_n_result<OutputIt> format_to_n(OutputIt out, size_t n, const S&,
                                          Args&&... args) {
-  constexpr auto compiled = detail::compile<Args...>(S());
-#ifdef __cpp_if_constexpr
-  if constexpr (std::is_same<remove_cvref_t<decltype(compiled)>,
-                             detail::unknown_format>()) {
-    auto it =
-        format_to(detail::truncating_iterator<OutputIt>(out, n),
-                  static_cast<basic_string_view<typename S::char_type>>(S()),
-                  std::forward<Args>(args)...);
-    return {it.base(), it.count()};
-  } else {
-    auto it = format_to(detail::truncating_iterator<OutputIt>(out, n), compiled,
-                        std::forward<Args>(args)...);
-    return {it.base(), it.count()};
-  }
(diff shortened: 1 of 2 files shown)


include/fmt/compile.h
```diff
@@ -789,7 +794,17 @@ FMT_INLINE std::basic_string<typename S::char_type> format(const S&,
   }
 #endif
   constexpr auto compiled = detail::compile<Args...>(S());
+#ifdef __cpp_if_constexpr
+  if constexpr (std::is_same<remove_cvref_t<decltype(compiled)>,
```
vitaut, line 798: This doesn't need to be a constexpr if. The compiler will optimize the unneeded branch away.

include/fmt/compile.h
```diff
@@ -22,6 +22,10 @@
 #  endif
 #endif
 
+#ifndef FMT_ENABLE_FALLBACK_TO_RUNTIME_API
+#  define FMT_ENABLE_FALLBACK_TO_RUNTIME_API 1
+#endif
```
vitaut, line 27: I don't think we need this to be configurable considering that format string compilation is just an optimization. We don't provide any performance guarantees with or without compilation.

vitaut (changes requested):
Thanks for the PR. Looks good but I don't think we should provide a way to disable the fallback.

include/fmt/compile.h
alexezeder: The problem without using `constexpr if` is that else branch becomes invalid (before the optimization), for the same reason the current implementation fails to fall back to runtime API: https://github.com/alexezeder/fmt/runs/2007151954?check_suite_focus=true#step:5:70, https://github.com/alexezeder/fmt/runs/2007151989?check_suite_focus=true#step:5:70.
So it has to be `constexpr if` to omit else branch completely.
Of course, another option is to fix `char_t` for `unknown_format` (and maybe more fixes), but it seems unnecessary with `constexpr if`.

include/fmt/compile.h
alexezeder: Looks like I misunderstood you here - https://github.com/fmtlib/fmt/issues/2078#issuecomment-753644522:
> It can be sometimes useful to enforce the use of a specific API but it shouldn't be the default.

include/fmt/compile.h
```diff
@@ -827,11 +843,26 @@ auto format_to_n(OutputIt out, size_t n, const CompiledFormat& cf,
...
+        format_to(detail::truncating_iterator<OutputIt>(out, n),
+                  static_cast<basic_string_view<typename S::char_type>>(S()),
+                  std::forward<Args>(args)...);
+    return {it.base(), it.count()};
+  } else {
+    auto it = format_to(detail::truncating_iterator<OutputIt>(out, n), compiled,
+                        std::forward<Args>(args)...);
+    return {it.base(), it.count()};
+  }
+#else
```
vitaut, line 861: Can we replace all of these by forwarding `S` to `format_to` call below which will handle fallback?

alexezeder: Yes, we can. It will create one more template instantiation, but it shouldn't be tough.
So, it's done.

vitaut:
Thank you!

Issue #2140, opened 2021-02-21
Make error


tanim87:
Dear Concern, 
While making the fmt, I am getting the following error:
Building CXX object test/CMakeFiles/format-test.dir/format-test.cc.o
/usr/local/src/fmt/test/format-test.cc:1487:20: error: call to 'ptr' is ambiguous

Would you please help me with it? 
I am using Ubuntu 16.04

Thanks.

alexezeder:
Could you please provide compiler info?

tanim87:
gcc 8.10

alexezeder:
Looks like it should just work with the `master` branch of {fmt} on [Compiler Explorer](https://godbolt.org/z/8W1zne).
Do you have full output for this error, like function candidates, since the call is ambiguous?

tanim87:
![Screenshot_2021-02-21_01-06-44](https://user-images.githubusercontent.com/42817244/108618158-5f619a80-73e1-11eb-8f9d-80c8b629982d.png)
I have attached the complete screenshot. I could not find log file that would give more insight. Let me know if this will work. Thanks.

alexezeder:
Thanks, this definitely helped. For some reason, this specific version of GCC is able to decay function pointers to void pointers, just like MSVC, so this branch should be disabled for this GCC version too:
https://github.com/fmtlib/fmt/blob/7e72673d87cc44340f4b808d0b536e4e191342b5/include/fmt/format.h#L3837-L3844
You can check it locally just by commenting out the code above.

But since I don't know how to make it in a more versatile way, I think we should just wait for more experienced people for proper fix.

vitaut:
I think we could SFINAE this overload away on function to `const void*` pointer convertibility.

alexezeder:
Actually there is another problem with `fmt::ptr` and old Clang versions (<7.0) - https://godbolt.org/z/j57Kzx. Actually, this problem is even more interesting than I thought - https://godbolt.org/z/sEsTE7.

But as far as I can see, following workarounds are able to satisfy all compilers - https://godbolt.org/z/Y8sadG. I'll create a PR with this code soon.

tanim87:
Your suggestion worked! Thanks.

PR #2144, opened 2021-02-22
Fix `fmt::get` with function pointers for some GCC versions and legacy Clang

alexezeder:
Fixes https://github.com/fmtlib/fmt/issues/2140.

Some points:
- legacy Clang (prior to 7.0) treats function pointers also as `const T*` pointers but unable to convert them
- some GCC versions decay function pointers to `const void*`, exactly like MSVC does
- `inline` specifiers removed because all functions are templates

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -3833,21 +3833,16 @@ FMT_CONSTEXPR void advance_to(
     auto s = fmt::format("{}", fmt::ptr(p));
   \endrst
  */
-template <typename T> inline const void* ptr(const T* p) { return p; }
-template <typename T> inline const void* ptr(const std::unique_ptr<T>& p) {
+template <typename T> const void* ptr(T p) {
+  static_assert(std::is_pointer<T>::value, "");
+  return detail::bit_cast<const void*>(p);
+}
+template <typename T> const void* ptr(const std::unique_ptr<T>& p) {
   return p.get();
 }
-template <typename T> inline const void* ptr(const std::shared_ptr<T>& p) {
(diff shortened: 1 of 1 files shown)


include/fmt/format.h
```diff
@@ -3833,18 +3833,28 @@ FMT_CONSTEXPR void advance_to(
...
-template <typename T> inline const void* ptr(const std::unique_ptr<T>& p) {
+template <typename T>
+#if FMT_CLANG_VERSION && FMT_CLANG_VERSION < 700
+enable_if_t<!std::is_function<T>::value, const void*>
+#else
+const void*
+#endif
+ptr(const T* p) {
+  return p;
+}
```
vitaut, line 3844: Can we change this to

```c++
template <typename T>
const void* ptr(T p) {
  static_assert(std::is_pointer<T>::value, "");
  return detail::bit_cast<const void*>(p);
}
```
and remove the `ptr(T (*fn)(Args...))` overload below?

vitaut (changes requested):
Thank you!

include/fmt/format.h
alexezeder: Yes, we can. Another great code simplification.

alexezeder:
This is unrelated to this PR, but maybe this overload should also be added later:
```cpp
inline const void* ptr(std::nullptr_t) {
  return nullptr;
}
```

vitaut:
> This is unrelated to this PR, but maybe this overload should also be added later

I don't think it's necessary because `nullptr` can be passed directly. `ptr` is there for typed pointers.

PR #2145, opened 2021-02-22
Print error message in FMT_THROW when exception is disabled

NobodyXu:
Currently, if exception is disabled, an error in `fmt` will result a termination via `FMT_ASSERT`, but there is no error message printed on `stderr`, so it can be a little bit confusing for the user on why the program failed.

This PR modifies `FMT_THROW` to pass `exception.what()` to `FMT_ASSERT` as the error message.

Although it doesn't provide more information which part of format string caused the error, it is better than printing no error message at all.

Signed-off-by: Jiahao XU <Jiahao_XU@outlook.com>

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -123,8 +123,7 @@ FMT_END_NAMESPACE
 #  else
 #    define FMT_THROW(x)       \
       do {                     \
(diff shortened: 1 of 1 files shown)


vitaut:
Thank you!

PR #2151, opened 2021-02-25
Add error_reporter_t for customizable error handling when '-fno-exceptions'

NobodyXu:
`fmt::error_reporter_t` is a template class that enables user to install their own error report handler.

This is especially useful when the exception is disabled since there is no exception available.

To use `fmt::error_reporter_t`, one would use `FMT_ERROR`, which supersedes `FMT_THROW` in that inside macro `FMT_ERROR`, no exception is created. Instead, an exception is thrown from `fmt::error_reporter_t` only if exception is enabled.

Abstraction of change:

```
// format.h
enum class Error_type {
  format_error,
  system_error,
  runtime_error,
# ifdef _WIN3
  windows_error,
# endif
};

/**
 * Use error_reporter_t as a customization point.
 */

template <Error_type error_type, typename = void> struct error_reporter_t;

template <typename void_type> struct error_reporter_t<Error_type::format_error, void_type> {
    template <typename Char = char>
    FMT_NORETURN void operator () (int line, const char *func, const char *file, const char *msg,
                                   basic_string_view<Char> fmt = basic_string_view<Char>());
};

template <typename void_type> struct error_reporter_t<Error_type::system_error, void_type> {
    template <typename ...Args>
    FMT_NORETURN void operator () (int line, const char *func, const char *file,
                                   int error_code, const char *msg,
                                   const Args&... args);
};

template <typename void_type> struct error_reporter_t<Error_type::runtime_error, void_type> {
    FMT_NORETURN void operator () (int line, const char *func, const char *file, const char *msg);
}; 

# ifdef __GNUC__
#  define FMT_ERROR(error_type, ...)                                          \
    do {                                                                      \
        ::fmt::error_reporter_t<(error_type)> error_reporter;                 \
        error_reporter(__LINE__, __PRETTY_FUNCTION__, __FILE__, __VA_ARGS__); \
    } while (0)
# else
#  define FMT_ERROR(error_type, ...)                                           \
    do {                                                                      \
        ::fmt::error_reporter_t<(error_type)> error_reporter;                 \
        error_reporter(__LINE__, __FUNCTION__, __FILE__, __VA_ARGS__); \
    } while (0)
# endif
```

`fmt::error_reporter_t` also has a partial specialization for `windows_error`:

```
// os.h
template <typename void_type> struct error_reporter_t<Error_type::windows_error, void_type> {
  template <typename... Args>
  FMT_NORETURN void operator () (int line, const char *func, const char *file, 
                                 int error_code, const char *message,
                                 const Args&... args) {
#ifdef FMT_EXCEPTIONS
    static_cast<void>(line);
    static_cast<void>(func);
    static_cast<void>(file);

    throw windows_error(error_code, message, args...);
#else
    report_windows_error(error_code, format(message, args...));
    std::terminate();
#endif
  }
};
```

The rest of the PR simply changes all use of `FMT_THROW` to `FMT_ERROR`.

Signed-off-by: Jiahao XU <Jiahao_XU@outlook.com>

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -326,7 +326,7 @@ inline std::tm localtime(std::time_t time) {
   };
   dispatcher lt(time);
   // Too big time values may be unsupported.
-  if (!lt.run()) FMT_THROW(format_error("time_t value out of range"));
+  if (!lt.run()) FMT_ERROR(Error_type::format_error, "time_t value out of range");
   return lt.tm_;
 }
 
@@ -371,7 +371,7 @@ inline std::tm gmtime(std::time_t time) {
   };
   dispatcher gt(time);
   // Too big time values may be unsupported.
-  if (!gt.run()) FMT_THROW(format_error("time_t value out of range"));
+  if (!gt.run()) FMT_ERROR(Error_type::format_error, "time_t value out of range");
   return gt.tm_;
 }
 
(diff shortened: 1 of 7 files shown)


vitaut:
Thanks for the PR but it's already possible to customize error handling via `FMT_THROW`. I don't think we should provide more facilities around it considering that disabling exceptions is not recommended.

Issue #2148, opened 2021-02-24
Header-only mode still includes cassert, causing assert macro to be defined.


toojays:
Consider the following example:
```c++
#define FMT_HEADER_ONLY
#include <fmt/format.h>
//#undef assert

struct Test
{
    void assert (bool pass, const char *test_name)
    {
        fmt::print("{}: {}", test_name, pass ? "PASS" : "FAIL");
    }
};

int main()
{
    Test test;
    test.assert(true, "mumble");
    return 0;
}
```
It fails compilation ([godbolt](https://godbolt.org/z/Y4zfWq)) with:
```
<source>:7:50: error: macro "assert" passed 2 arguments, but takes just 1
    7 |     void assert (bool pass, const char *test_name)
      |                                                  ^
In file included from /opt/compiler-explorer/gcc-10.2.0/include/c++/10.2.0/cassert:44,
                 from /opt/compiler-explorer/libs/fmt/trunk/include/fmt/format-inl.h:12,
                 from /opt/compiler-explorer/libs/fmt/trunk/include/fmt/format.h:4230,
                 from <source>:2:
/usr/include/assert.h:92: note: macro "assert" defined here
   92 | #  define assert(expr)       \
      | 
<source>:16:31: error: macro "assert" passed 2 arguments, but takes just 1
   16 |     test.assert(true, "mumble");
      |                               ^
In file included from /opt/compiler-explorer/gcc-10.2.0/include/c++/10.2.0/cassert:44,
                 from /opt/compiler-explorer/libs/fmt/trunk/include/fmt/format-inl.h:12,
                 from /opt/compiler-explorer/libs/fmt/trunk/include/fmt/format.h:4230,
                 from <source>:2:
/usr/include/assert.h:92: note: macro "assert" defined here
   92 | #  define assert(expr)       \
      | 
<source>:7:10: error: variable or field 'assert' declared void
    7 |     void assert (bool pass, const char *test_name)
      |          ^~~~~~
<source>:10:5: error: expected ';' at end of member declaration
   10 |     }
      |     ^
      |      ;
<source>: In function 'int main()':
<source>:16:10: error: 'struct Test' has no member named 'assert'
   16 |     test.assert(true, "mumble");
      |          ^~~~~~
```
As a workaround we can ``#undef assert`` immediately after including ``format.h``.

It looks like some work was done to remove the dependency on cassert in 2f9acd18, but this did not extend to the header only mode?

[...]

PR #2152, opened 2021-02-25
Don't include <cassert>. (#2148)

toojays:
This commit replaces use of the assert() macro in format-inl.h with
FMT_ASSERT(). This allows us to drop the cassert include.

Note that the change to ``FMT_ASSERT`` in test-assert.h is to avoid the following warning:
```
In file included from /home/jscott/src/fmt/test/format-impl-test.cc:10:
/home/jscott/src/fmt/include/fmt/format-inl.h: In destructor ‘fmt::v7::detail::bigint::~bigint()’:
/home/jscott/src/fmt/test/test-assert.h:26:52: warning: throw will always call terminate() [-Wterminate]
   26 |   if (!(condition)) throw assertion_failure(message);
      |                                                    ^
/home/jscott/src/fmt/include/fmt/format-inl.h:1300:15: note: in expansion of macro ‘FMT_ASSERT’
 1300 |   ~bigint() { FMT_ASSERT(bigits_.capacity() <= bigits_capacity, ""); }
      |               ^~~~~~~~~~
/home/jscott/src/fmt/test/test-assert.h:26:52: note: in C++11 destructors default to noexcept
   26 |   if (!(condition)) throw assertion_failure(message);
      |                                                    ^
/home/jscott/src/fmt/include/fmt/format-inl.h:1300:15: note: in expansion of macro ‘FMT_ASSERT’
 1300 |   ~bigint() { FMT_ASSERT(bigits_.capacity() <= bigits_capacity, ""); }
      |               ^~~~~~~~~~
```
Actually it seems that just moving the ``throw`` into a separate function (even inline) is enough to squelch the warning, with GCC 9.3 at least. In the end I went with this more explicit version.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -147,7 +147,7 @@ FMT_FUNC void format_error_code(detail::buffer<char>& out, int error_code,
   if (message.size() <= inline_buffer_size - error_code_size)
     format_to(it, FMT_STRING("{}{}"), message, SEP);
   format_to(it, FMT_STRING("{}{}"), ERROR_STR, error_code);
-  FMT_ASSERT(out.size() <= inline_buffer_size, "must not require dynamic alloc");
+  FMT_ASSERT(out.size() <= inline_buffer_size, "");
 }
 
 FMT_FUNC void report_error(format_func func, int error_code,
@@ -1218,7 +1218,7 @@ struct accumulator {
     if (lower < n) ++upper;
   }
   void operator>>=(int shift) {
-    FMT_ASSERT(shift == 32, "only 32-bit shift is supported");
+    FMT_ASSERT(shift == 32, "");
     (void)shift;
     lower = (upper << 32) | (lower >> 32);
     upper >>= 32;
@@ -1323,7 +1323,7 @@ class bigint {
   int num_bigits() const { return static_cast<int>(bigits_.size()) + exp_; }
 
   FMT_NOINLINE bigint& operator<<=(int shift) {
-    FMT_ASSERT(shift >= 0, "shift must be non-negative");
+    FMT_ASSERT(shift >= 0, "");
     exp_ += shift / bigit_bits;
     shift %= bigit_bits;
     if (shift == 0) return *this;
@@ -1385,7 +1385,7 @@ class bigint {
 
   // Assigns pow(10, exp) to this bigint.
   void assign_pow10(int exp) {
-    FMT_ASSERT(exp >= 0, "exponent must be non-negative");
+    FMT_ASSERT(exp >= 0, "");
     if (exp == 0) return assign(1);
     // Find the top bit.
     int bitmask = 1;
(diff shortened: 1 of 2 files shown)


include/fmt/format-inl.h
```diff
@@ -148,7 +147,7 @@ FMT_FUNC void format_error_code(detail::buffer<char>& out, int error_code,
   if (message.size() <= inline_buffer_size - error_code_size)
     format_to(it, FMT_STRING("{}{}"), message, SEP);
   format_to(it, FMT_STRING("{}{}"), ERROR_STR, error_code);
-  assert(out.size() <= inline_buffer_size);
+  FMT_ASSERT(out.size() <= inline_buffer_size, "must not require dynamic alloc");
```
vitaut, line 150: I don't think we need assertion messages in any of these cases, I suggest keeping them empty (we should really make the message optional in `FMT_ASSERT` but not in this PR).

test/test-assert.h
```diff
@@ -22,8 +22,24 @@ class assertion_failure : public std::logic_error {
...
+#    pragma GCC diagnostic push
+#    pragma GCC diagnostic ignored "-Wterminate"
+#  endif
+
+  throw assertion_failure(message);
+
+#  if defined(__GNUC__) && __GNUC__ >= 6
+#    pragma GCC diagnostic pop
+#  endif
+}
```
vitaut, line 38: Do you need both the function and suppression? gcc doesn't seem to warn in this case: https://godbolt.org/z/PYzx9h

vitaut (changes requested):
Thanks for the PR.

test/test-assert.h
toojays: Thanks for your review. I've pushed a couple of commits which address your comments.

test/test-assert.h
```diff
@@ -22,8 +22,16 @@ class assertion_failure : public std::logic_error {
 
 void assertion_failure::avoid_weak_vtable() {}
 
+// We use a separate function (rather than throw directly from FMT_ASSERT) to
+// avoid GCC's -Wterminate warning when FMT_ASSERT is used in a destructor.
+inline void throw_assertion_failure (const char *message)
+{
```
vitaut, line 28: Please follow coding conventions (https://github.com/fmtlib/fmt/blob/master/CONTRIBUTING.md):

```
inline void throw_assertion_failure (const char* message) {
```

test/test-assert.h
```diff
@@ -22,8 +22,16 @@ class assertion_failure : public std::logic_error {
...
+  throw assertion_failure(message);
+}
+
+
```
vitaut, line 32: Please remove extra empty line.

vitaut (changes requested):

test/test-assert.h
toojays: Ooops. Thanks for your patience. Fixed with clang-format (which made the exact same edits you suggested).

Issue #2149, opened 2021-02-24
vformat_to_n(char *) fails to compile with GCC 8.


toojays:
Consider the following example of formatting to a fixed size buffer before writing to stdout:
```c++
#include <fmt/format.h>
#include <unistd.h>

void vlog(fmt::string_view format, fmt::format_args args) {
  char line[1024];
  auto [_, size] = fmt::vformat_to_n(line, sizeof(line), format, args);
  write(STDOUT_FILENO, line, size);
}

template <typename S, typename... Args>
void log(const S& format, Args&&... args) {
  vlog(format, fmt::make_args_checked<Args...>(format, args...));
}

int main() {
  log("invalid squishiness: {}", 42);
}
```
With GCC 9 and 10 this works. But with GCC 8 this fails to compile ([godbolt](https://godbolt.org/z/bhq7d3)) with:
```
<source>: In function 'void vlog(fmt::v7::string_view, fmt::v7::format_args)':
<source>:6:70: error: no matching function for call to 'vformat_to_n(char [1024], long unsigned int, fmt::v7::string_view&, fmt::v7::format_args&)'
   auto [_, size] = fmt::vformat_to_n(line, sizeof(line), format, args);
                                                                      ^
In file included from /opt/compiler-explorer/libs/fmt/trunk/include/fmt/format.h:46,
                 from <source>:1:
/opt/compiler-explorer/libs/fmt/trunk/include/fmt/core.h:1830:37: note: candidate: 'template<class OutputIt, class Char, class ... Args, typename std::enable_if<fmt::v7::detail::is_output_iterator<OutputIt, Char>::value, int>::type <anonymous> > fmt::v7::format_to_n_result<OutputIt> fmt::v7::vformat_to_n(OutputIt, size_t, fmt::v7::basic_string_view<Char>, fmt::v7::basic_format_args<fmt::v7::basic_format_context<fmt::v7::detail::buffer_appender<typename fmt::v7::type_identity<Char>::type>, typename fmt::v7::type_identity<Char>::type> >)'
 inline format_to_n_result<OutputIt> vformat_to_n(
                                     ^~~~~~~~~~~~
/opt/compiler-explorer/libs/fmt/trunk/include/fmt/core.h:1830:37: note:   template argument deduction/substitution failed:
/opt/compiler-explorer/libs/fmt/trunk/include/fmt/core.h:1829:11: error: no type named 'type' in 'struct std::enable_if<false, int>'
[...]

PR #2160, opened 2021-03-01
Use simplified void_t for all compilers except GCC 4.x

alexezeder:
Fixes https://github.com/fmtlib/fmt/issues/2149.

This PR moves a special `void_t` workaround for GCC 4.x:
https://github.com/fmtlib/fmt/blob/640acba8509d7458a9c749b243355d223b1cfa25/include/fmt/core.h#L1382-L1385
under the GCC < 5.0 macro check, adding a straightforward `void_t` implementation:
```cpp
template <typename...> using void_t = void;
```
for other compilers.

This solves the problem with GCC 8.x that fails with the current implementation of `void_t`.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -1379,10 +1379,14 @@ template <typename T> struct formattable : std::false_type {};
 
 namespace detail {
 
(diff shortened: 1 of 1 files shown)


vitaut:
Thank you!

Issue #2141, opened 2021-02-21
formatted_size() is broken for modern compile-time API


alexezeder:
There is a problem using `formatted_size()` with format strings for compile-time API - `FMT_COMPILE` and `_cf`. Here Is an example - https://godbolt.org/z/zz3c1E, and a simplified version of it - https://godbolt.org/z/drW1rG.

This problem can be solved in at least two ways:

1. By adding SFINAE check to https://github.com/fmtlib/fmt/blob/ab0f7d7fdcb46f35719197f7cae012fe843ec6a1/include/fmt/compile.h#L837-L840 so it would not be available for `compiled_string`-based format strings. Thus the version from `core.h` will be chosen.
2. By adding a new `formatted_size()` implementation specifically for `compiled_string`-based format strings. This also requires marking a conversion operator in `compiled_string`-based format strings explicit, which results in changing few lines in `compile.h` for conversion (with using "custom" `FMT_STRING_IMPL` macro).

There is no difference between these 2 ways in performance because of compilers optimizations. But there is a difference between runtime API and compile-time API (lack of support for named spec'ed fields, for example). So which way should we choose to fix broken `formatted_size()`?

Maybe it's not even a problem because, maybe, a format string from compile-time API shouldn't be used in `formatted_size()`.

vitaut:
I think we should go with the second option except that instead of making the conversion operator explicit the `formatted_size` overload should be templatized like other formatting functions.

PR #2161, opened 2021-03-01
Fix fmt::formatted_size() with compiled format (FMT_COMPILE) as argument

alexezeder:
Resolves https://github.com/fmtlib/fmt/issues/2141.

All useful info can be found in the issue, including code examples.

One small detail for this PR - `fmt::formatted_size()` in `compile.h` becomes templatized too to eliminate one unclear case - https://godbolt.org/z/Yxnccj. In few words, the implementation of `fmt::formatted_size()` for runtime API (non-compiled format string) changes depending on whether we include `fmt/compile.h` or not. You can see it in assembly for the `formatted_size`, it uses `counting_iterator`, which is used in the overload from `compile.h`:
https://github.com/fmtlib/fmt/blob/835b910e7d758efdfdce9f23df1b190deb3373db/include/fmt/compile.h#L837-L840
but not in the overload from `core.h`:
https://github.com/fmtlib/fmt/blob/835b910e7d758efdfdce9f23df1b190deb3373db/include/fmt/core.h#L1864-L1870

Maybe it's okay, but IMHO in this case, that behavior should be documented, right?

### DIFF
--- a/include/fmt/compile.h
+++ b/include/fmt/compile.h
@@ -834,7 +834,10 @@ format_to_n_result<OutputIt> format_to_n(OutputIt out, size_t n, const S&,
   return {it.base(), it.count()};
 }
 
-template <typename CompiledFormat, typename... Args>
+template <typename CompiledFormat, typename... Args,
+          FMT_ENABLE_IF(std::is_base_of<detail::basic_compiled_format,
+                                        CompiledFormat>::value ||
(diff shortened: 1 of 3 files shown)


vitaut:
> the implementation of fmt::formatted_size() for runtime API (non-compiled format string) changes depending on whether we include fmt/compile.h or not.

Good catch, `formatted_size` in `fmt/compile.h` was previously underconstrained.

PR #2179, opened 2021-03-14
Fix MSVC /clr builds

### LINKED ISSUE #2176: Missing identifier 'ctz'/'ctzll' when compiling with MSVC and /clr
If I try to use fmt in a MSVC project with /clr enabled, I get the following errors:

> 1>D:\Temp\ConsoleApplication1\fmt-7.1.3\include\fmt\format-inl.h(2017,11): error C3861: 'ctz': identifier not found
> 1>D:\Temp\ConsoleApplication1\fmt-7.1.3\include\fmt\format-inl.h(2045,11): error C3861: 'ctzll': identifier not found

Here is a simple example to reproduce these errors:

```C++
#define FMT_HEADER_ONLY
#include <fmt/format.h>

int main()
{
}
```

It seems due to _MANAGED beeing defined in /clr builds, the replacement funtions for 'clz', 'clzll', 'ctz' and 'ctzll' are never defined and missing.

https://github.com/fmtlib/fmt/blob/af567538a031557e16b83c3a730246f079f2ca11/include/fmt/format.h#L206-L207

I think the better approach here would be to remove the _MANAGED check and instead force this section to native mode, to be able to use intrinsics without warnings:

```C++
// Switch to native
#pragma managed(push,off)
// Do native stuff here
#pragma managed(pop)
// Switch back to previous mode
```

Brainy0207:
Resolves #2176

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -204,17 +204,18 @@ FMT_END_NAMESPACE
 // __builtin_clz and __builtin_clzll, so only define FMT_BUILTIN_CLZ using the
 // MSVC intrinsics if the clz and clzll builtins are not available.
 #if FMT_MSC_VER && !defined(FMT_BUILTIN_CLZLL) && \
-    !defined(FMT_BUILTIN_CTZLL) && !defined(_MANAGED)
+    !defined(FMT_BUILTIN_CTZLL)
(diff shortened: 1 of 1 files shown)


vitaut:
Thank you!

PR #2183, opened 2021-03-15
Fix harmless MSVS warning about using undefined _MANAGED symbol

vadz:
Since the changes of 1305cbeb (Fix MSVC2019 error C2049 when compiling
with /clr (#1897), 2020-09-23), compiling fmt with MSVS 2019 resulted in

fmt\include\fmt\core.h(180,32): warning C4668: '_MANAGED' is not defined
as a preprocessor macro, replacing with '0' for '#if/#elif'.

when the (disabled by default) warning C4668 was enabled.

Fix this simply by checking if _MANAGED is defined before testing it.

---


vitaut:
Thank you!

PR #2186, opened 2021-03-19
add default cases

imsherlock:
Adding default case for switch statements where the compilation
flag -Wswitch-default is present on the command line when spdlog
is included in external projects.

Signed-off-by: Ryan Sherlock <ryan.m.sherlock@gmail.com>


vitaut:
Thanks

PR #2187, opened 2021-03-20
"Use" `fwrite` result (workaround for `warn_unused_result`)

### LINKED ISSUE #2185: (void)fwrite triggers -Wunused-result again
Regression of #1098 introduced here https://github.com/fmtlib/fmt/commit/211d31240698703bbada7826dd3d0c6dd27c1133

https://github.com/fmtlib/fmt/blob/master/include/fmt/format-inl.h#L158

Unfortunately the `(void)` cast doesn't actually convince GCC that we've paid attention to the `std::fwrite` result.

Maybe `report_error` can return a bool indicating whether the error was reported successfullly, calculated from the fwrite result. We don't have to actually *look* at that bool, but GCC may shut up if we look like we care.

It's all kind of silly. The `std::fputc` on the next line has no such annotation but it's just as capable of failing.

```
 [2021/03/17 13:37:16.683] /opt/mongodbtoolchain/v3/bin/g++ -o build/cached/third_party/fmt/dist/src/format.o -c -std=c++17 -Wall  -Werror  src/third_party/fmt/dist/src/format.cc

 [2021/03/17 13:37:18.244] In file included from src/third_party/fmt/dist/src/format.cc:8:
 [2021/03/17 13:37:18.244] src/third_party/fmt/dist/include/fmt/format-inl.h: In function 'void fmt::v7::detail::report_error(fmt::v7::detail::format_func, int, fmt::v7::string_view)':
 [2021/03/17 13:37:18.244] src/third_party/fmt/dist/include/fmt/format-inl.h:158:20: error: ignoring return value of 'size_t fwrite(const void*, size_t, size_t, FILE*)', declared with attribute warn_unused_result [-Werror=unused-result]
 [2021/03/17 13:37:18.244]    (void)std::fwrite(full_message.data(), full_message.size(), 1, stderr);
 [2021/03/17 13:37:18.244]          ~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

BillyDonahue:
Fixes #2185

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -155,7 +155,9 @@ FMT_FUNC void report_error(format_func func, int error_code,
   memory_buffer full_message;
   func(full_message, error_code, message);
   // Don't use fwrite_fully because the latter may throw.
-  (void)std::fwrite(full_message.data(), full_message.size(), 1, stderr);
+  if (std::fwrite(full_message.data(), full_message.size(), 1, stderr)) {
+      ;
+  }
   std::fputc('\n', stderr);
 }
 


include/fmt/format-inl.h
```diff
@@ -155,7 +155,9 @@ FMT_FUNC void report_error(format_func func, int error_code,
   memory_buffer full_message;
   func(full_message, error_code, message);
   // Don't use fwrite_fully because the latter may throw.
-  (void)std::fwrite(full_message.data(), full_message.size(), 1, stderr);
+  if (std::fwrite(full_message.data(), full_message.size(), 1, stderr)) {
+      ;
+  }
   std::fputc('\n', stderr);
```
vitaut, line 161: I think we shouldn't be trying to write `\n` if an error occurs which will also solve the warning: 

```c++
  if (std::fwrite(full_message.data(), full_message.size(), 1, stderr) > 0)
    std::fputc('\n', stderr);
```

vitaut (changes requested):
Thanks for the PR

vitaut:
Merged with a small tweak.

PR #2188, opened 2021-03-20
add default case for core

imsherlock:
Adding default case for switch statements where the compilation
flag -Wswitch-default is present on the command line.

Signed-off-by: Ryan Sherlock <ryan.m.sherlock@gmail.com>


vitaut:
Thanks for the PR but this is intentional.

imsherlock:
Why is that? The default case has no effect other than suppressing an error when using -Wswitch-default.

vitaut:
Cases are currently exhaustive so default may obscure the missing case.

imsherlock:
I understand the concern there. Is there an alternative approach? Would including a comment explaining its reasoning be suitable for inclusion?

vitaut:
I am not sure if it's worth adding a comment because it applies to all code, not just this case: an enum switch should either be exhaustive or have a default.

PR #2192, opened 2021-03-25
Fix C++17 build

phprus:
- Fix C++17 Visual Studio builds.
- Use C++17 for `windows-2019` builds.
- Removed options for unsupported compiler (MSVC11).
- Removed options, not needed after enabling C++11+ features in gmock/gtest.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -16,7 +16,7 @@ jobs:
           - os: windows-2016
             standard: 11
           - os: windows-2019
-            standard: 14
+            standard: 17
           - os: windows-2016
             platform: Win32
             build_type: Debug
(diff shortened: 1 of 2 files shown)


.github/workflows/windows.yml
```diff
@@ -12,18 +12,22 @@ jobs:
         os: [windows-2016, windows-2019]
         platform: [Win32, x64]
         build_type: [Debug, Release]
+        standard: [11, 14, 17]
         include:
-          - os: windows-2016
-            standard: 11
-          - os: windows-2019
-            standard: 14
```
vitaut, line 19: We don't need this many configurations, it's enough to change 14 to 17 here.

vitaut (changes requested):
Thanks for the PR. In general looks good but let's keep the number of configurations manageable.

vitaut:
Thank you!

PR #2194, opened 2021-03-27
Intel

yafshar:
This PR includes: 
- Extra flag to prevent Intel compiler (with Clang front-end) warning 
- A user-defined constructor for a const declaration of `buffer_size`
 

vitaut:
Thank you!

Issue #2197, opened 2021-03-29
fmt::ostream cannot be moved while holding buffered data


vtta:
The following code will throw.
[https://godbolt.org/z/58hM1h94e](https://godbolt.org/z/58hM1h94e)

```c++
#include <fmt/os.h>

int main() {
  fmt::ostream out = fmt::output_file("test-file");
  out.print("fail!");
  fmt::ostream moved(std::move(out));
  moved.print("hello");
  return 0;
}
```

vtta:
related to #1844

vtta:
PR: #2198

PR #2198, opened 2021-03-29
Fix: fmt::ostream cannot be moved while holding buffered data #2197

vtta:
#2197

### DIFF
--- a/include/fmt/os.h
+++ b/include/fmt/os.h
@@ -396,8 +396,10 @@ class ostream final : private detail::buffer<char> {
   ostream(ostream&& other)
       : detail::buffer<char>(other.data(), other.size(), other.capacity()),
         file_(std::move(other.file_)) {
+    other.clear();
     other.set(nullptr, 0);
   }
+
   ~ostream() {
     flush();
     delete[] data();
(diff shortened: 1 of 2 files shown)


include/fmt/os.h
```diff
@@ -396,8 +396,10 @@ class ostream final : private detail::buffer<char> {
   ostream(ostream&& other)
       : detail::buffer<char>(other.data(), other.size(), other.capacity()),
         file_(std::move(other.file_)) {
+    other.clear();
     other.set(nullptr, 0);
```
vitaut, line 400: I wonder if this can be replaced with

```
other = detail::buffer<char>();
```

vitaut (commented):
Thanks for the PR!

alexezeder:
Hmm.... @vitaut, is a line in PR description:

> I agree that my contributions are licensed under the {fmt} license, and agree to future changes to the licensing.

still necessary?

include/fmt/os.h
vtta: It seems reasonable, but I prefer the consistency between `set` and `clear`. 
Another issue is that creating another object in ctors seems confusing.

PR #2204, opened 2021-03-31
Make buffer_appender default-constructible when back_insert_iterator is

### LINKED ISSUE #2195: `buffer_appender` does not conform to `std::output_iterator`
I've run into this before with `truncating_iterator` (see #2156), apparently, the same applies for `buffer_appender`.

CC @randomnetcat :-)

randomnetcat:
Fixes #2195.


vitaut:
Thank you!

PR #2206, opened 2021-04-01
Add speech synthesis support

vitaut:
Add speech synthesis support to {fmt}.

For example:

```c++
#include <fmt/os.h>

int main() {
  fmt::say(
    R"SHANTY(
    There [[emph +]] once was a pen that [[emph +]] put to [[emph +]] paper,
    It [[emph -]] was [[emph +]] gonna be written [[emph +]] sooner or later,
    Ctrl-S went down, the proposal sent out,
    Vote [[emph +]] down my committee vote [[emph +]] down.

    [[emph +]] Soon, may the Unicode come,
    So [[emph +]] we can write aleph and [[emph +]] xĭ and [[emph +]] từng,
    One day when '23 is done,
    We'll take our bom and go.)SHANTY");
}
```
Output:

https://user-images.githubusercontent.com/576385/113313097-c9623f00-92bf-11eb-8903-f7acaa7c982f.mov

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        cxx: [g++-4.8, g++-8, g++-10, clang++-9]
+        cxx: [g++-4.8, g++-10, clang++-9]
         build_type: [Debug, Release]
         std: [11]
         os: [ubuntu-18.04]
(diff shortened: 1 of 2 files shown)


tigerw:
The best kind of feature creep. 3D text render output when

vitaut:
> 3D text render output when

PRs are welcome (not).

rcombs:
In all seriousness, I don't think this is a great idea to have in the library as-is, since it might not be obvious to users that this isn't safe to pass arbitrary input to, nor that it's highly non-portable (`say` only exists on macOS).

It'd be safer to at least use `FILE* f = popen("say", "w");` and `fmt::print(f, format_str, args…);`.

vitaut:
The function has been removed so that it is not accidentally used in an unsafe way. It was broken with C++20 and more recently with fmt 12 anyway.

PR #2211, opened 2021-04-05
fix #2210

### LINKED ISSUE #2210: warning C4127: conditional expression is constant
building the latest commit (95da4847274c42c110fd5bf1bf61d3f94be4a08e) with `/W4` with `msvc 19.27.29111` yields the following diagnostic

```
fmt\core.h(399): warning C4127: conditional expression is constant
fmt\core.h(399): note: consider using 'if constexpr' statement instead
fmt\core.h(398): note: while compiling class template member function 'fmt::v7::basic_string_view<wchar_t>::basic_string_view(const Char *)'
        with
        [
            Char=wchar_t
        ]
fmt\src\os.cc(126): note: see reference to function template instantiation 'fmt::v7::basic_string_view<wchar_t>::basic_string_view(const Char *)' being compiled
        with
        [
            Char=wchar_t
        ]
fmt\format.h(1222): note: see reference to class template instantiation 'fmt::v7::basic_string_view<wchar_t>' being compiled
```

Apache-HB:

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -98,13 +98,6 @@
 #  define FMT_CONSTEXPR_DECL
 #endif
 
-// C++17's char_traits::length() is constexpr.
-#if __cplusplus >= 201703L
-#  define FMT_CONSTEXPR_CHAR_TRAITS_LENGTH FMT_CONSTEXPR
-#else
-#  define FMT_CONSTEXPR_CHAR_TRAITS_LENGTH
-#endif
-
 #ifndef FMT_OVERRIDE
 #  if FMT_HAS_FEATURE(cxx_override_control) || \
       (FMT_GCC_VERSION >= 408 && FMT_HAS_GXX_CXX11) || FMT_MSC_VER >= 1900
@@ -399,8 +392,12 @@ template <typename Char> class basic_string_view {
     the size with ``std::char_traits<Char>::length``.
     \endrst
    */
-  FMT_CONSTEXPR_CHAR_TRAITS_LENGTH FMT_INLINE basic_string_view(const Char* s) : data_(s) {
-    if FMT_CONSTEXPR_CHAR_TRAITS_LENGTH (std::is_same<Char, char>::value && !detail::is_constant_evaluated())
+// C++17's char_traits::length() is constexpr.
+#if __cplusplus >= 201703L
+  constexpr
+#endif
+  FMT_INLINE basic_string_view(const Char* s) : data_(s) {
+    if (detail::const_check(std::is_same<Char, char>::value && !detail::is_constant_evaluated()))
       size_ = std::strlen(reinterpret_cast<const char*>(s));
     else
       size_ = std::char_traits<Char>::length(s);
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -392,7 +392,7 @@ template <typename Char> class basic_string_view {
     the size with ``std::char_traits<Char>::length``.
     \endrst
    */
-#if __cplusplus >= 201703L
+#if __cplusplus >= 201703L // C++17's char_traits::length() is constexpr.
   constexpr
 #endif
       FMT_INLINE
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -392,12 +392,14 @@ template <typename Char> class basic_string_view {
     the size with ``std::char_traits<Char>::length``.
     \endrst
    */
-// C++17's char_traits::length() is constexpr.
 #if __cplusplus >= 201703L
   constexpr
 #endif
-  FMT_INLINE basic_string_view(const Char* s) : data_(s) {
-    if (detail::const_check(std::is_same<Char, char>::value && !detail::is_constant_evaluated()))
+      FMT_INLINE
+      basic_string_view(const Char* s)
+      : data_(s) {
+    if (detail::const_check(std::is_same<Char, char>::value &&
+                            !detail::is_constant_evaluated()))
       size_ = std::strlen(reinterpret_cast<const char*>(s));
     else
       size_ = std::char_traits<Char>::length(s);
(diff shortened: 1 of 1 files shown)


include/fmt/core.h
```diff
@@ -392,11 +399,8 @@ template <typename Char> class basic_string_view {
     the size with ``std::char_traits<Char>::length``.
     \endrst
    */
-#if __cplusplus >= 201703L  // C++17's char_traits::length() is constexpr.
-  FMT_CONSTEXPR
-#endif
-  FMT_INLINE basic_string_view(const Char* s) : data_(s) {
-    if (std::is_same<Char, char>::value && !detail::is_constant_evaluated())
+  FMT_CONSTEXPR_CHAR_TRAITS_LENGTH FMT_INLINE basic_string_view(const Char* s) : data_(s) {
+    if FMT_CONSTEXPR_CHAR_TRAITS_LENGTH (std::is_same<Char, char>::value && !detail::is_constant_evaluated())
```
alexezeder, line 403: This line becomes wrong with `FMT_CONSTEXPR_CHAR_TRAITS_LENGTH` replaced as `FMT_CONSTEXPR`, i.e. `constexpr`. From [cppreference](https://en.cppreference.com/w/cpp/types/is_constant_evaluated) about `std::is_constant_evaluated()` (which is used inside `detail::is_constant_evaluated()` when supported):

> When directly used as the condition of static_assert declaration or **_constexpr if_** statement, std::is_constant_evaluated() always returns true.

So it shouldn't be `if constexpr`, just `if`.

include/fmt/core.h
```diff
@@ -98,6 +98,13 @@
 #  define FMT_CONSTEXPR_DECL
 #endif
 
+// C++17's char_traits::length() is constexpr.
+#if __cplusplus >= 201703L
+#  define FMT_CONSTEXPR_CHAR_TRAITS_LENGTH FMT_CONSTEXPR
+#else
+#  define FMT_CONSTEXPR_CHAR_TRAITS_LENGTH
+#endif
+
```
alexezeder, line 107: Also I'm not sure if it's worth introducing this special definition to use it only once.

include/fmt/core.h
```diff
@@ -392,11 +399,8 @@ template <typename Char> class basic_string_view {
     the size with ``std::char_traits<Char>::length``.
     \endrst
    */
-#if __cplusplus >= 201703L  // C++17's char_traits::length() is constexpr.
-  FMT_CONSTEXPR
```
alexezeder, line 396: And another observation, it's not about this PR, just these lines in general: probably `constexpr` can be used here, not `FMT_CONSTEXPR`, since it's already under the `__cplusplus >= 201703L` macro check.

alexezeder: Looks like this actually can be applied to this PR too, since it's also used in newly introduced macro definition.

alexezeder:
Anyway, feel free to ignore my review, and just wait for comments from maintainers 🙂

include/fmt/core.h
alexezeder: The correct way to fix this warning I can imagine (probably it's also very unclear): combine `__cplusplus` check together with `__cpp_lib_is_constant_evaluated` check to use `if constexpr` only if `detail::is_constant_evaluated()` unconditionally false.

vitaut: Just replace `FMT_CONSTEXPR_CHAR_TRAITS_LENGTH` with `detail::const_check` and revert other changes.

vitaut (changes requested):
Thanks for the PR.

include/fmt/core.h
```diff
@@ -392,11 +392,12 @@ template <typename Char> class basic_string_view {
...
+// C++17's char_traits::length() is constexpr.
```
vitaut, line 395: Looks good but please move this comment back to the next line to avoid interfering with an apidoc comment above and apply clang-format.

include/fmt/core.h
```diff
@@ -392,11 +392,14 @@ template <typename Char> class basic_string_view {
     the size with ``std::char_traits<Char>::length``.
     \endrst
    */
-#if __cplusplus >= 201703L  // C++17's char_traits::length() is constexpr.
-  FMT_CONSTEXPR
+#if __cplusplus >= 201703L
```
alexezeder, line 395: Please revert back `  // C++17's char_traits::length() is constexpr.` comment to this line.

include/fmt/core.h
```diff
@@ -392,11 +392,14 @@ template <typename Char> class basic_string_view {
     the size with ``std::char_traits<Char>::length``.
     \endrst
    */
-#if __cplusplus >= 201703L  // C++17's char_traits::length() is constexpr.
-  FMT_CONSTEXPR
+#if __cplusplus >= 201703L // C++17's char_traits::length() is constexpr.
```
alexezeder, line 395: Apply clang-format here, please

include/fmt/core.h
```diff
@@ -393,10 +393,13 @@ template <typename Char> class basic_string_view {
...
    */
 #if __cplusplus >= 201703L  // C++17's char_traits::length() is constexpr.
-  FMT_CONSTEXPR
+  constexpr
 #endif
-  FMT_INLINE basic_string_view(const Char* s) : data_(s) {
-    if (std::is_same<Char, char>::value && !detail::is_constant_evaluated())
+      FMT_INLINE
+      basic_string_view(const Char* s)
+      : data_(s) {
```
vitaut, line 400: Did clang-format do this?

alexezeder: Yeah, it's:
```cpp
#if __cplusplus >= 201703L  // C++17's char_traits::length() is constexpr.
  FMT_CONSTEXPR
#endif
  FMT_INLINE
  basic_string_view(const Char* s) : data_(s) {
```
or:
```cpp
#if __cplusplus >= 201703L  // C++17's char_traits::length() is constexpr.
  constexpr
#endif
      FMT_INLINE
      basic_string_view(const Char* s)
      : data_(s) {
```

vitaut: LOL

vitaut:
Merged, thanks.

PR #2213, opened 2021-04-05
Fixing VS2019 warning C4127: conditional expression is constant

chronoxor:


vitaut:
This should be covered by #2211 but thanks anyway.

vitaut:
Fixed in #2211.

PR #2215, opened 2021-04-06
constexpr uint128_wrapper

jstaahl:
I noticed a static initializer was introduced by the fmt library in our binary. It was attributed to:
[include/fmt/format-inl.h#L387](https://github.com/fmtlib/fmt/blob/95da4847274c42c110fd5bf1bf61d3f94be4a08e/include/fmt/format-inl.h#L387).

This can be avoided by making the uint128_wrapper constexpr.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -901,14 +901,14 @@ struct FMT_EXTERN_TEMPLATE_API uint128_wrapper {
 #if FMT_USE_INT128
   uint128_t internal_;
 
-  uint128_wrapper(uint64_t high, uint64_t low) FMT_NOEXCEPT
+  FMT_CONSTEXPR uint128_wrapper(uint64_t high, uint64_t low) FMT_NOEXCEPT
       : internal_{static_cast<uint128_t>(low) |
                   (static_cast<uint128_t>(high) << 64)} {}
 
-  uint128_wrapper(uint128_t u) : internal_{u} {}
+  FMT_CONSTEXPR uint128_wrapper(uint128_t u) : internal_{u} {}
 
-  uint64_t high() const FMT_NOEXCEPT { return uint64_t(internal_ >> 64); }
-  uint64_t low() const FMT_NOEXCEPT { return uint64_t(internal_); }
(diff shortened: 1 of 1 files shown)


include/fmt/format.h
```diff
@@ -901,14 +901,14 @@ struct FMT_EXTERN_TEMPLATE_API uint128_wrapper {
 #if FMT_USE_INT128
   uint128_t internal_;
 
-  uint128_wrapper(uint64_t high, uint64_t low) FMT_NOEXCEPT
+  FMT_CONSTEXPR uint128_wrapper(uint64_t high, uint64_t low) FMT_NOEXCEPT
```
vitaut, line 904: Please use `constexpr` instead of `FMT_CONSTEXPR` here and below since this looks compatible with C++11 constexpr.

vitaut (changes requested):
Thanks for the PR.

include/fmt/format.h
jstaahl: Ahh thanks yup it is compatible with C++11 constexpr. Done.

include/fmt/format.h
```diff
@@ -901,14 +901,14 @@ struct FMT_EXTERN_TEMPLATE_API uint128_wrapper {
...
+  constexpr uint128_wrapper(uint64_t high, uint64_t low) FMT_NOEXCEPT
       : internal_{static_cast<uint128_t>(low) |
                   (static_cast<uint128_t>(high) << 64)} {}
 
-  uint128_wrapper(uint128_t u) : internal_{u} {}
+  constexpr uint128_wrapper(uint128_t u) : internal_{u} {}
 
-  uint64_t high() const FMT_NOEXCEPT { return uint64_t(internal_ >> 64); }
-  uint64_t low() const FMT_NOEXCEPT { return uint64_t(internal_); }
+  constexpr uint64_t high() const FMT_NOEXCEPT { return uint64_t(internal_ >> 64); }
```
alexezeder, line 910: Could you apply clang-format, please.

jstaahl: done

PR #2216, opened 2021-04-06
Specify size for `prefixes` static data

### LINKED ISSUE #2205: test/compile-test.cc:339:21: error: consteval function 'test_format not a constant expression.
In test-enabled cmake, I got this error
```
fmt-master/test/compile-test.cc:339:21: error: call to consteval function 'test_format<5, char, bool, FMT_COMPILE_STRING>' is not a constant expression
   EXPECT_EQ("true", test_format<5>(FMT_COMPILE("{}"), true));
```

compiler:  clang 12 from https://llvm.org/builds/ in `--target=x86_64-w64-windows-gnu` mode

Any ideas?

alexezeder:
Resolves #2205, at least compile-time part of it.

A newer version of Clang should probably be used on CI to eliminate potential problems with compile-time tests in the future.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -971,8 +971,8 @@ template <typename T = void> struct FMT_EXTERN_TEMPLATE_API basic_data {
   static const char reset_color[5];
   static const wchar_t wreset_color[5];
   static const char signs[];
-  static constexpr const unsigned prefixes[] = {0, 0, 0x1000000u | '+',
-                                                0x1000000u | ' '};
+  static constexpr const unsigned prefixes[4] = {0, 0, 0x1000000u | '+',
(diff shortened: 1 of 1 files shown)


vitaut:
Thank you!

alexezeder:
Hmm... but why are there no CI checks?.. 🤔

vitaut:
Because it's perfect =)

denchat:
May I ask why `static` is not redundant with `constexpr` presenting?
Doesn't just `constexpr` automatically make it `static`, without saying `static`?

vitaut:
> May I ask why static is not redundant with constexpr presenting?

These are data members so `static` is required.

PR #2217, opened 2021-04-08
Linux CI config improvements

alexezeder:
Some points for this PR:

* Clang 9 C++11 build here:
  https://github.com/fmtlib/fmt/blob/5a8bf1f6a3b5f2248bbe657aae1cdfb850802504/.github/workflows/linux.yml#L29-L31
  was implicitly covered by matrix. I removed those lines, but that change does nothing for the Clang 9 C++11 build, it [works](https://github.com/fmtlib/fmt/runs/2292750007?check_suite_focus=true) as [before](https://github.com/fmtlib/fmt/runs/2290330649?check_suite_focus=true).
* Build type was not specified for some builds, like:
  https://github.com/fmtlib/fmt/blob/5a8bf1f6a3b5f2248bbe657aae1cdfb850802504/.github/workflows/linux.yml#L22-L24
  so it's not passed to CMake correctly, this led to [messages](https://github.com/fmtlib/fmt/runs/2290330696?check_suite_focus=true#step:5:4) [like](https://github.com/fmtlib/fmt/runs/2290330716?check_suite_focus=true#step:5:4) [this](https://github.com/fmtlib/fmt/runs/2290330748?check_suite_focus=true#step:5:4):
  ```
  CMake Error: Invalid value used with --config
  ```
* Clang 11 C++20 build added since it was welcomed [here](https://github.com/fmtlib/fmt/issues/2205#issuecomment-814510274).
* Clang 11 C++20 with LLVM libc++ build added. Thanks to this additional check, I found 2 problems, which are fixed in this PR:
  * lack of CXXFLAGS passing in `find-package-test`, thus `-stdlib=libc++` flag was ignored, and an incorrect standard library was used, [error log](https://github.com/alexezeder/fmt/runs/2292091058?check_suite_focus=true#step:6:1547);
  * lack of `<array>` header include in `compile-test`, I fixed it by using a C-style array instead of `std::array`, [error log](https://github.com/alexezeder/fmt/runs/2291966185?check_suite_focus=true#step:5:64).


vitaut:
Looks great, thanks!

PR #2218, opened 2021-04-08
Suppress redef warning of _CRT_SECURE_NO_WARNINGS if any.

denchat:
add a pre-check of _CRT_SECURE_NO_WARNINGS before the def

### DIFF
--- a/test/chrono-test.cc
+++ b/test/chrono-test.cc
@@ -6,7 +6,9 @@
 // For the license information refer to format.h.
 
 #ifdef WIN32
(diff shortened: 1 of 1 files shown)


test/chrono-test.cc
```diff
@@ -6,7 +6,9 @@
 // For the license information refer to format.h.
 
 #ifdef WIN32
```
vitaut, line 8: Let's get rid of WIN32 for simplicity. There is no harm in having the macro always defined.

vitaut:
Thank you

PR #2222, opened 2021-04-09
Make string_view comparison as constexpr

denchat:
So we can do something like [this](https://compiler-explorer.com/z/dc3vaPTrx) as string_view could.
```
  constexpr auto r = ( "1" == test_format<2>(FMT_COMPILE("{:d}"), true) );
```

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -434,7 +434,7 @@ template <typename Char> class basic_string_view {
   }
 
   // Lexicographically compare this string reference to other.
-  int compare(basic_string_view other) const {
+  FMT_CONSTEXPR int compare(basic_string_view other) const {
     size_t str_size = size_ < other.size_ ? size_ : other.size_;
     int result = std::char_traits<Char>::compare(data_, other.data_, str_size);
     if (result == 0)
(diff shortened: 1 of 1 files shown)


include/fmt/core.h
```diff
@@ -434,30 +434,30 @@ template <typename Char> class basic_string_view {
   }
 
   // Lexicographically compare this string reference to other.
-  int compare(basic_string_view other) const {
+  FMT_CONSTEXPR int compare(basic_string_view other) const {
```
alexezeder, line 437: This function uses `std::char_traits::compare()` which is [available at compile time starting from C++17](https://en.cppreference.com/w/cpp/string/char_traits/compare), while `FMT_CONSTEXPR` is replaced with `constexpr` starting from C++14 (with some exceptions). So for C++14, this code is invalid at compile-time, but no one uses it with this standard anyway, so there is no compilation error.

vitaut:
Thanks for the PR but unfortunately it can't be merged because the comparison should be compatible with C++14.

alexezeder:
But we can make it compatible. 😉

But I, personally, don't see good reasoning for this PR. Of course, it would make `fmt::string_view` comparable at compile-time, but use-cases for this are more interesting for me. 
Because the first use-case that came to my mind was to use `critical_assert` instead of GTest macro for compile-time formatting checks. But that's not cool because GTest at least outputs something.

vitaut:
Also `fmt::string_view` is just a fallback and in most cases its standard counterpart should be used instead.

include/fmt/core.h
denchat: Thanks for clarification.

denchat: Let's do this then
https://github.com/llvm-mirror/libcxx/blob/master/include/__string#L114

Issue #2228, opened 2021-04-12
warning C4910: '__declspec(dllexport)' and 'extern' are incompatible on an explicit instantiation


moiwi:
Compiling format.h with FMT_EXPORT defined gives the warning

`format.h(998,71): warning C4910: 'fmt::v7::detail::basic_data<void>': '__declspec(dllexport)' and 'extern' are incompatible on an explicit instantiation (compiling source file os.cc)`

with Visual Studio 2019.
Why was

```
#ifndef FMT_EXPORTED

#endif
```

removed?

vitaut:
It was removed in #2220 because of clang warnings. It seems that MSVC and clang handle dllexport in incompatible ways.

vitaut:
We might need both `FMT_INSTANTIATION_DECL_API` and `FMT_INSTANTIATION_DEF_API` defined based on the compiler (sigh).

denchat:
I'm trying to fix this at https://github.com/fmtlib/fmt/pull/2229 .
However I'm not sure that...

#ifndef FMT_EXPORTED
  extern template struct <name>
#endif

is also required by msvc.

I don't have msvc, please report back.

moiwi:
It seems

```
#ifndef FMT_EXPORTED

#endif
```

is needed for MSVC. Without I still get ehe same warning

`'__declspec(dllexport)' and 'extern' are incompatible on an explicit instantiation`

denchat:
Okey, I wrapped it up.

However, I've been thinking that msvc should never has seen `extern template` together with `dllexport` in the 1st place.
The preprocessor is supposed to behave like the previous one on msvc!?
That's odd or maybe I might have missed out something.

Please check again.

moiwi:
#2229 fixes my problem.
Without "#ifndef FMT_EXPORTED" the preprocessor output is
```
template <typename T = void> struct __declspec(dllexport) basic_data {
  static const uint64_t powers_of_10_64[];
  //...
  static const uint64_t zero_or_powers_of_10_64[];
};
//...
extern template struct  basic_data<void>;
```
which results in the mentioned compiler warning.

vitaut:
Thinking more of it, `basic_data` doesn't seem to need `dllexport` at all. At least I was able to build fmt dll with it removed (https://github.com/fmtlib/fmt/commit/8bc45503596581cde93007016e0d57acba67a908) and the tests work.

vitaut:
[...]

PR #2229, opened 2021-04-13
Create separate dllexport marking points for clang and msvc.

denchat:
This should fix https://github.com/fmtlib/fmt/issues/2228

The issue is caused by changes in clang below.

[MinGW] Fix dllexport of explicit template instantiation
by mstorsjo on Apr 25 2019
https://reviews.llvm.org/D61118

Have to keep seperated `dllexport` marking points, until `msvc` follows `GNU` and `LLVM`'s policy or vice versa.


vitaut:
Thanks!

PR #2230, opened 2021-04-13
Try to suppress MVSC warn of narrowing

denchat:
Try to suppress MVSC warn of narrowing at

> fmt\test\format-test.cc(774): warning C4838: conversion from 'int' to 'wchar_t' requires a narrowing conversion

### DIFF
--- a/test/format-test.cc
+++ b/test/format-test.cc
@@ -771,7 +771,10 @@ TEST(FormatterTest, SpaceSign) {
 
 TEST(FormatterTest, SignNotTruncated) {
   wchar_t format_str[] = {L'{', L':',
-                          '+' | static_case<wchar_t>(1 << fmt::detail::num_bits<char>()), L'}', 0};
+                          '+' | static_cast<wchar_t>(
+                            1 << fmt::detail::num_bits<char>() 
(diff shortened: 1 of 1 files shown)


test/format-test.cc
```diff
@@ -771,7 +771,10 @@ TEST(FormatterTest, SpaceSign) {
 
 TEST(FormatterTest, SignNotTruncated) {
   wchar_t format_str[] = {L'{', L':',
-                          '+' | (1 << fmt::detail::num_bits<char>()), L'}', 0};
+                          '+' | static_cast<wchar_t>(
+                            1 << fmt::detail::num_bits<char>() 
+                            & static_cast<typename std::make_unsigned<int>::type >(-1) ),
```
vitaut, line 776: The second static cast and & are not needed, please remove.

vitaut (changes requested):

vitaut:
Thanks

test/format-test.cc
denchat: May I ask if it's unneccessary always or just in this case?

vitaut: I don't see why it was added in the first place here. It has no effect.

PR #2232, opened 2021-04-13
Support char*_t in fmt::join

patstew:
Makes fmt::join templated on the string argument so you can use unicode strings with it.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -3753,14 +3753,9 @@ struct formatter<arg_join<It, Sentinel, Char>, Char> {
   Returns an object that formats the iterator range `[begin, end)` with elements
   separated by `sep`.
  */
-template <typename It, typename Sentinel>
-arg_join<It, Sentinel, char> join(It begin, Sentinel end, string_view sep) {
-  return {begin, end, sep};
-}
-
-template <typename It, typename Sentinel>
-arg_join<It, Sentinel, wchar_t> join(It begin, Sentinel end, wstring_view sep) {
-  return {begin, end, sep};
+template <typename It, typename Sentinel, typename S>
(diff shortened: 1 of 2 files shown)


denchat (commented):
This is brilliant!

Never thought chat_t<S> has such powerful features.
Its usage is kinda indirectly tricky and very powerful.

vitaut:
Thanks for the PR but `chat_t<S>` will likely go away and the use of `char*_t` strings is strongly discouraged. I recommend using UTF-8 char strings or adding an `join` overload to your codebase.

alexezeder:
Just to be sure, @vitaut, do you mean here:

> the use of char*_t strings is strongly discouraged. I recommend using UTF-8 char strings

that `std::u8string` shouldn't be used in favor of something like this - https://github.com/DuffsDevice/tiny-utf8?

vitaut:
It doesn't make any sense to use `std::u8string`, `std::string` or something better should be used instead. `std::string` can already work with UTF-8 and is compatible with native APIs unlike its `char*_t` counterparts.

patstew:
While I'd like to use UTF-8 everywhere, we also need to use char16_t for strings that get consumed by our GUI library (Qt). We already overload in our own codebase, but I was hoping to get it upstream after having to change it when updating fmtlib. Would a char_t-less version be acceptable?

denchat:
>   and is compatible with native APIs

Please give some examples.

vitaut:
> Would a char_t-less version be acceptable?

Not until the next major release when the revamp of the code unit type handling is planned (including the likely removal of `char_t`). It would be afterwards.

> Please give some examples.

Linux and OS X normally assume UTF-8 `char` and don't have any APIs that work with `char8_t`.

PR #2235, opened 2021-04-15
Support compiling {fmt} as named module

DanielaE:
Start by tagging official API for module export

### DIFF
--- a/include/fmt/color.h
+++ b/include/fmt/color.h
@@ -18,6 +18,7 @@
 #endif
 
 FMT_BEGIN_NAMESPACE
+FMT_MODULE_EXPORT_BEGIN
 
 enum class color : uint32_t {
   alice_blue = 0xF0F8FF,               // rgb(240,248,255)
@@ -205,6 +206,7 @@ struct rgb {
   uint8_t b;
 };
 
+FMT_MODULE_EXPORT_END
 namespace detail {
 
 // color is a struct of either a rgb color or a terminal color.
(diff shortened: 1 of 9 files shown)


include/fmt/color.h
```diff
@@ -231,6 +235,7 @@ struct color_type {
 }  // namespace detail
 
 /** A text style consisting of foreground and background colors and emphasis. */
+FMT_MODULE_EXPORT
```
vitaut, line 238: Why not include this class in the `FMT_MODULE_EXPORT_BEGIN/FMT_MODULE_EXPORT_END` block below and remove `FMT_MODULE_EXPORT`?

include/fmt/printf.h
```diff
@@ -194,10 +194,12 @@ FMT_DEPRECATED void printf(detail::buffer<Char>& buf,
 }
 using detail::vprintf;
 
+// already exported through "ostream.h" above
 template <typename Char>
 class basic_printf_parse_context : public basic_format_parse_context<Char> {
   using basic_format_parse_context<Char>::basic_format_parse_context;
 };
+// already exported through "ostream.h" above
```
vitaut, line 202: Let's remove this comment, the one above can apply to both contexts.

vitaut (commented):
Thank you for the PR. LGTM, just two minor comments inline.

include/fmt/color.h
DanielaE: Yes, why not? I'll tell ya: staring at code for extended periods of time didn't prevent me from being silly.

PR #2239, opened 2021-04-16
Use qualified name lookup rather than ADL.

DanielaE:
Name lookup within exported templates cannot find non-exported entities by ADL when instantiation takes place outside the module.


vitaut:
Thank you!

PR #2240, opened 2021-04-17
Provide a module interface unit

DanielaE:
To compile {fmt} as named module, there must be at least the source provided with the (primary) module interface.

I chose the most elegant and probably the most desirable implementation strategy: a single-file module with a private module partition. This has these benefits:

- easiest to use
- the existing headers require no changes other than providing macro hooks for the exported entities (if implemented cleanly)
- all macros are available during compilation and require no further attention
- the existing source parts all go into the non-exported, unreachable private module partition, which is basically a module implementation unit that's directly fused to end of the module interface unit
- there are only 2 artefacts created:
  1. the object file
  2. the built module interface (BMI)

The macro-based interfaces FMT_STRING and FMT_COMPILE are not yet implemented. Macros cannot get exported from modules.

The inclusion of the 'os' part is optional an can be configured from the command line at build time.
The module name can be configured from the command line at build time. By default it is 'fmt'.
The module may be configured and provided as a static or dynamic-load lib as before.

This requires a sufficiently complete and bug-free implementation of C++20 modules to compile!

At present, msvc 16.10-pre2 requires a workaround for successful compilation (which is not included). 16.10-pre3 is reportedly fine according the the compiler devs.
I have no experience how gcc and clang fare. Let's give them a serious drill!

@vitaut : this is how I do it n my own repo. Let's discuss this.

### DIFF
--- a/src/fmt.cc
+++ b/src/fmt.cc
@@ -1,0 +1,87 @@
+module;
+// put all implementation-provided headers into the global module fragment
+// to prevent attachment to this module
+#if !defined(_CRT_SECURE_NO_WARNINGS) && defined(_MSC_VER)
+#  define _CRT_SECURE_NO_WARNINGS
+#endif
+#if !defined(WIN32_LEAN_AND_MEAN) && defined(_WIN32)
+#  define WIN32_LEAN_AND_MEAN
+#endif
+
+#include <cctype>
+#include <cerrno>
+#include <climits>
+#include <clocale>
+#include <cmath>
+#include <cstdarg>
+#include <cstddef>
+#include <cstdint>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+#include <ctime>
+#include <cwchar>
+
+#include <algorithm>
+#include <chrono>
+#include <exception>
+#include <functional>
+#include <iterator>
+#include <limits>
+#include <locale>
+#include <memory>
+#include <ostream>
+#include <sstream>
+#include <stdexcept>
+#include <string>
+#include <string_view>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+#if _MSC_VER
+#  include <intrin.h>
+#endif
+#if defined __APPLE__ || defined(__FreeBSD__)
+#  include <xlocale.h>
+#endif
+#if __has_include(<winapifamily.h>)
+#  include <winapifamily.h>
+#endif
+#if (__has_include(<fcntl.h>) || defined(__APPLE__) || \
+     defined(__linux__)) &&                            \
+    (!defined(WINAPI_FAMILY) || (WINAPI_FAMILY == WINAPI_FAMILY_DESKTOP_APP))
+#  include <fcntl.h>
+#  include <sys/stat.h>
+#  include <sys/types.h>
+#  ifndef _WIN32
+#    include <unistd.h>
+#  else
+#    include <io.h>
+#  endif
+#endif
+#ifdef _WIN32
+#  include <windows.h>
+#endif
+
+export module fmt;
+
+#define FMT_MODULE_EXPORT export
+#define FMT_MODULE_EXPORT_BEGIN export {
+#define FMT_MODULE_EXPORT_END }
+
+// all library-provided declarations and definitions
+// must be in the module purview to be exported
+#include "fmt/format.h"
+#include "fmt/args.h"
+#include "fmt/color.h"
+#include "fmt/compile.h"
+#include "fmt/locale.h"
+#include "fmt/chrono.h"
+#include "fmt/printf.h"
+#include "fmt/os.h"
+
+module : private;
+
+#include "format.cc"
+#include "os.cc"
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -270,6 +270,7 @@ FMT_GCC_PRAGMA("GCC optimize(\"Og\")")
 #endif
 
 FMT_BEGIN_NAMESPACE
+FMT_MODULE_EXPORT_BEGIN
 
 // Implementations of enable_if_t and other metafunctions for older systems.
 template <bool B, class T = void>
@@ -287,6 +288,8 @@ template <typename T> struct type_identity { using type = T; };
 template <typename T> using type_identity_t = typename type_identity<T>::type;
 
 struct monostate {};
+
+FMT_MODULE_EXPORT_END
 
 // An enable_if helper to be used in template parameters which results in much
 // shorter symbols: https://godbolt.org/z/sWw4vP. Extra parentheses are needed
--- a/include/fmt/ostream.h
+++ b/include/fmt/ostream.h
@@ -14,12 +14,8 @@
 
 FMT_BEGIN_NAMESPACE
 
-FMT_MODULE_EXPORT_BEGIN
-
 template <typename Char> class basic_printf_parse_context;
 template <typename OutputIt, typename Char> class basic_printf_context;
-
-FMT_MODULE_EXPORT_END
 
 namespace detail {
 
--- a/include/fmt/printf.h
+++ b/include/fmt/printf.h
@@ -195,12 +195,15 @@ FMT_DEPRECATED void printf(detail::buffer<Char>& buf,
 }
 using detail::vprintf;
 
-// Exported from ostream.h.
+FMT_MODULE_EXPORT_BEGIN
+
 template <typename Char>
 class basic_printf_parse_context : public basic_format_parse_context<Char> {
   using basic_format_parse_context<Char>::basic_format_parse_context;
 };
 template <typename OutputIt, typename Char> class basic_printf_context;
+
+FMT_MODULE_EXPORT_END
 
 /**
   \rst


src/fmt.cc
```diff
@@ -0,0 +1,85 @@
...
+#include <initializer_list>
+#include <iterator>
+#include <limits>
+#include <locale>
+#include <memory>
+#include <ostream>
+#include <sstream>
+#include <stdexcept>
+#include <string>
+#include <string_view>
```
vitaut, line 32: I don't think `<string_view>` is included anywhere in fmt.

src/fmt.cc
```diff
@@ -0,0 +1,85 @@
...
+  #include <clocale>
+  #if __has_include(<winapifamily.h>)
+    #include <winapifamily.h>
+  #endif
+  #if __has_include(<fcntl.h>)
+    #include <fcntl.h>
+  #endif
+#endif
+
+export module FMT_MODULE_NAME;
```
vitaut, line 56: I don't think the module name should be configurable. Let's use fmt here.

src/fmt.cc
```diff
@@ -0,0 +1,85 @@
...
+
+#define FMT_MODULE_EXPORT export
+#define FMT_MODULE_EXPORT_BEGIN export {
+#define FMT_MODULE_EXPORT_END }
+
+#define FMT_USE_NONTYPE_TEMPLATE_PARAMETERS 1
```
vitaut, line 62: What is this for?

src/fmt.cc
```diff
@@ -0,0 +1,85 @@
...
+
+// all library-provided declarations and definitions
+// must be in the module purview to be exported
+#include "fmt/format.h"
+#include "fmt/args.h"
+#include "fmt/color.h"
+#include "fmt/compile.h"
+#include "fmt/locale.h"
+#include "fmt/ostream.h"
+#include "fmt/ranges.h"
```
vitaut, line 72: These two should be moved to separate modules because in many cases they should be disabled / explicitly opted in.

vitaut (changes requested):
Thanks a lot for the PR! Looks good, just a few minor comments inline.

src/fmt.cc
DanielaE: `core.h` line 263
All implementation-provided headers were grepped out of the sources.

src/fmt.cc
DanielaE: Wonderful, we can agree on a name.

src/fmt.cc
vitaut: My bad, it's conditionally included from core.h

src/fmt.cc
DanielaE: C++20 modules unsurprisingly require C++20 which includes NTTPs. `compile.h` enables the use of NTTPs in a non-standard way that excludes compilers like msvc that support this feature. My current code to  support FMT_STRING and FMT_COMPILE relies on this feature. So I enable it by overriding the macro until `compile.h` gets updated to the reality.

src/fmt.cc
DanielaE: `ostream.h` is included through `printf.h` anyway. 
`ranges.h` can be implemented in a different module if it neither requires any macro from other parts of {fmt} nor any non-exported identifier. Not to mention that teensy tiny modules are not recommended afaik.
What do you think about conditionally including these parts of {fmt} by macro definitions from the command line like it is already proposed for `os.h`? So users of the module can decide on their own if they want this or not.

DanielaE: The currently existing dependency levels are as such:

1
core.h

2
args.h -> core.h
format.h -> core.h

3
compile.h -> format.h
ostream.h -> format.h
ranges.h -> format.h
color.h -> format.h
locale.h -> format.h
os.h -> format.h
format-inl.h -> format.h

4
printf.h -> ostream.h
chrono.h -> format.h locale.h

foonathan: Not that I have a say in this, but why split it up and/or have conditional inclusion (this also applies to OS)? In principle, it shouldn't matter how much stuff is included in the module and it's not like fmt has no/few standard library dependencies without them.

src/fmt.cc
alexezeder: Looks like we should use this as a fix - https://github.com/foonathan/lexy/issues/15

alexezeder: It should be fixed with https://github.com/fmtlib/fmt/pull/2247, but I need help with this PR, some strange MSVC errors...

src/fmt.cc
DanielaE: I'm not sure if I get your point here. You are right about your comment on the amount of stuff included in a module, but why do you mention the standard library? Do you propose not to carve out anything from inclusion into the module interface? If so we are on the same page. So far I've excluded `os.h` for the simple fact that it doesn't matter: nothing is exported from there because there doesn't seem to exist a documented interface that hints on which declarations to export. Additional exports can be added in a subsequent PR by tagging them accordingly. IIUC, Victor's concerns are about template specializations they bring into the TUs that import {fmt}. I have hardly any clue about `ranges.h` in this regard but I cannot really see a problem with `ostream.h`.

src/fmt.cc
DanielaE: This FMT_STRING thing is bothering me in a different context, too. At present `fmt::to_wstring()` cannot be invoked from client code because the identifier `FMT_STRING` cannot be looked-up from the body of instantiations of `to_wstring`. Neither unqualified lookup nor ADL works 😢. But that should be solved in a different PR.

src/fmt.cc
vitaut: It must be split because ostream and ranges provide formatter specialization that should not be available by default.

DanielaE: Which then rules out `printf.h`, too. Oh well...

DanielaE:
So I will proceed this way:

- make the module name fixed `fmt`
- keep FMT_USE_NONTYPE_TEMPLATE_PARAMETERS for the near future
- remove `os.h`, `ostream.h`, `printf.h`, and `ranges.h`.
- remove system headers that are no longer nominated by the remaining parts of {fmt}

Additional changes like already mentioned need to be subject of PRs against regular {fmt}.

src/fmt.cc
vitaut: `printf.h` dependency on `ostream.h` is accidental and will be removed by https://github.com/fmtlib/fmt/commit/b9ab5c8836bbffbe0a877f64d6faef8fbf4fd394.

src/fmt.cc
vitaut: > C++20 modules unsurprisingly require C++20 which includes NTTPs.

While this is true, modules and NTTP are independent features so presence of one doesn't imply presence of the other. I suggest removing this define from the current PR. It can be set when building {fmt}.

vitaut:
> remove os.h

I think it's OK to keep `os.h`. It's mostly experimental but not harmful.

src/fmt.cc
DanielaE: Ok with me.

DanielaE:
Looks like a plan to me.

vitaut:
Merged, thank you!

vitaut:
I am getting the following error when building the module with gcc:

```
$ g++-11 src/fmt.cc -I include -std=c++20 -fmodules-ts
In file included from src/fmt.cc:75:
include/fmt/format.h:3898:66: error: conflicting exporting declaration ‘template<class Char, typename std::enable_if<(! std::is_same<Char, char>::value), int>::type <anonymous> > std::__cxx11::basic_string<_CharT> fmt::v7::detail::vformat(fmt::v7::basic_string_view<Char>, fmt::v7::basic_format_args<fmt::v7::basic_format_context<fmt::v7::detail::buffer_appender<typename fmt::v7::type_identity<T>::type>, typename fmt::v7::type_identity<T>::type> >)’
 3898 |     basic_format_args<buffer_context<type_identity_t<Char>>> args) {
      |                                                                  ^
In file included from include/fmt/format.h:46,
                 from src/fmt.cc:75:
include/fmt/core.h:1800:25: note: previous declaration ‘template<class Char, typename std::enable_if<(! std::is_same<Char, char>::value), int>::type <anonymous> > std::__cxx11::basic_string<_CharT> fmt::v7::detail::vformat(fmt::v7::basic_string_view<Char>, fmt::v7::basic_format_args<fmt::v7::basic_format_context<fmt::v7::detail::buffer_appender<typename fmt::v7::type_identity<T>::type>, typename fmt::v7::type_identity<T>::type> >)’ here
 1800 | std::basic_string<Char> vformat(
      |                         ^~~~~~~
```

Any ideas?

vitaut:
I think the problem is that the declaration is exported twice.

DanielaE:
Ok, if so, the problem should be easy to resolve by exporting it only from the location where the name is intoduced first. That's most likely in core.h. MSVC didn't complain and I don't have any other compiler available to test with.

DanielaE:
Strange. The declaration in core.h is in namespace detail, the one in format.h is in namespace fmt. I hope neither file is changed too much, my clone isn't up-to-date.

PR #2242, opened 2021-04-18
Fix udl_compiled_string with non-byte chars (e.g. wchar)

alexezeder:
Without this fix, `udl_compiled_string` works only for char types that have a 1-byte size.

### DIFF
--- a/include/fmt/compile.h
+++ b/include/fmt/compile.h
@@ -947,8 +947,9 @@ size_t formatted_size(const CompiledFormat& cf, const Args&... args) {
 #if FMT_USE_NONTYPE_TEMPLATE_PARAMETERS
 inline namespace literals {
 template <detail::fixed_string Str>
-constexpr detail::udl_compiled_string<remove_cvref_t<decltype(Str.data[0])>,
-                                      sizeof(Str.data), Str>
+constexpr detail::udl_compiled_string<
+    remove_cvref_t<decltype(Str.data[0])>,
+    sizeof(Str.data) / sizeof(decltype(Str.data[0])), Str>
 operator""_cf() {
   return {};
 }
--- a/test/compile-test.cc
+++ b/test/compile-test.cc
@@ -316,6 +316,7 @@ TEST(CompileTest, CompileFormatStringLiteral) {
   using namespace fmt::literals;
   EXPECT_EQ("", fmt::format(""_cf));
   EXPECT_EQ("42", fmt::format("{}"_cf, 42));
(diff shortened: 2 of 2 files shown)


DanielaE (approved):
This is totally fine. I did exactly the same in a different context outside of UDLs.

include/fmt/compile.h
```diff
@@ -947,8 +947,9 @@ size_t formatted_size(const CompiledFormat& cf, const Args&... args) {
 #if FMT_USE_NONTYPE_TEMPLATE_PARAMETERS
 inline namespace literals {
 template <detail::fixed_string Str>
-constexpr detail::udl_compiled_string<remove_cvref_t<decltype(Str.data[0])>,
-                                      sizeof(Str.data), Str>
+constexpr detail::udl_compiled_string<
+    remove_cvref_t<decltype(Str.data[0])>,
```
vitaut, line 951: Can we move `remove_cvref_t<decltype(Str.data[0])>` to a default template parameter to avoid repetition? i.e.

```c++
template <detail::fixed_string Str, Char = remove_cvref_t<decltype(Str.data[0])>>
```

alexezeder: Looks like it's not possible because compilers expect to see certain template arguments - https://godbolt.org/z/6odY6E78h

vitaut: Can we at least do `typename Str::char_type`?

denchat: How about `value_type` as in `basic_string<>::value_type` ?

vitaut: Sure, `value_type` is a more appropriate.

alexezeder: Don't forget that `Str` is not a type, it's an object since it's passed as a non-type template parameter, so we have these options here:
```cpp
<remove_cvref_t<decltype(Str.data[0])>, sizeof(Str.data) / sizeof(decltype(Str.data[0])), Str>
vs.
<typename decltype(Str)::value_type, sizeof(Str.data) / sizeof(typename  decltype(Str)::value_type), Str>
```
and they both are not great.

But the size of `fixed_string` can be exposed by a static member or function, [like so](https://godbolt.org/z/G8ezYj8c9):
```cpp
<typename decltype(Str)::value_type, decltype(Str)::size, Str>
```
Let me know if this approach is preferable.

vitaut: I think it's fine as is then. Thanks for the fix.

Issue #2238, opened 2021-04-16
strftime() zero size check


yumeyao:
When I was browsing the code I came across the following comment about #367:
```
      if (size >= tm_format.size() * 256) {
        // If the buffer is 256 times larger than the format string, assume
        // that `strftime` gives an empty result. There doesn't seem to be a
        // better way to distinguish the two cases:
        // https://github.com/fmtlib/fmt/issues/367
        break;
      }
```
a possibly simple way to fix this is adding an extra space ` ` in `parse()`:
```
    tm_format.reserve(detail::to_unsigned(end - it + 1));
    tm_format.append(it, end);;
 +  tm_format.push_back(' ');
    tm_format.push_back('\0');
    return end;
```
and remove the space after formatting in `format()`:
```
 -  return std::copy(buf.begin(), buf.end(), ctx.out());
 +  return std::copy(buf.begin(), buf.end() - 1, ctx.out());
```

vitaut:
Good idea, thanks. Could you submit a PR?

PR #2244, opened 2021-04-19
Appending a space to guarantee non-empty strftime() result.

yumeyao:
As per discussed and requested on #2238.

Tested on VS 2019 + WSL Ubuntu 20.04.


vitaut:
Thank you!

PR #2243, opened 2021-04-18
Compile-time named arguments

alexezeder:
This PR makes the name of named arguments created via `_a` literal be available at compile-time. So named arguments can be checked at compile-time now exactly like non-named ones. 
I haven't implemented these checks for `FMT_STRING`, but I did it for compile-time API. Thus these lines:
```cpp
fmt::format(FMT_COMPILE("{0}"), 42);
fmt::format(FMT_COMPILE("{arg}"), "arg"_a = 42);
```
produce the same compiled format, so it's zero-overhead for arguments naming. Of course, `fmt::arg(name, value)`-based named arguments still have this overhead because they cannot provide their names at compile-time. Here are [the results](https://github.com/alexezeder/fmt_bnchmrk/tree/feature/named_args):

| Benchmark                         |    Time     |       CPU | Iterations |
| ----------------------------------|-------------|-----------|------------|
| FMTNonNamed/42                    | 14.3 ns     |   14.3 ns |   47750054 |
| FMTRuntimeNamed/42                | 28.0 ns     |   28.0 ns |   25015214 |
| FMTCompileNonNamed/42             | 2.28 ns     |   2.28 ns |  307524148 |
| FMTCompileRuntimeNamed/42         | 3.05 ns     |   3.05 ns |  229333563 |
| FMTCompileCompileTimeNamed/42     | 2.23 ns     |   2.23 ns |  316913616 |

Besides that performance improvement, a fallback to runtime API is no longer needed for named arguments with specs because type information is available at compile-time. Here are [the results](https://github.com/alexezeder/fmt_bnchmrk/tree/feature/named_args_with_specs):

| Benchmark                         |    Time     |       CPU | Iterations |
| ----------------------------------|-------------|-----------|------------|
| FMTNonNamed/42                    | 46.0 ns     |   46.0 ns |   15185138 |
| FMTRuntimeNamed/42                | 54.5 ns     |   54.5 ns |   12507523 |
| FMTCompileNonNamed/42             | 30.2 ns     |   30.2 ns |   23434189 |
| FMTCompileRuntimeNamed/42         | 54.3 ns     |   54.3 ns |   12678573 |
| FMTCompileCompileTimeNamed/42     | 29.8 ns     |   29.8 ns |   23493430 |

Some points for this PR:

* `_a` literal provides statically named argument (i.e. its name available at compile-time) in form of `statically_named_arg` type
[...]

PR #2246, opened 2021-04-20
C++17: std::char_traits<>::{compare,length} is constexpr.

phprus:
Discussion: https://github.com/fmtlib/fmt/pull/2243#discussion_r616803322

### DIFF
--- a/include/fmt/color.h
+++ b/include/fmt/color.h
@@ -432,7 +432,7 @@ template <typename Char> struct ansi_color_escape {
   FMT_CONSTEXPR operator const Char*() const FMT_NOEXCEPT { return buffer; }
 
   FMT_CONSTEXPR const Char* begin() const FMT_NOEXCEPT { return buffer; }
-  FMT_CONSTEXPR const Char* end() const FMT_NOEXCEPT {
+  FMT_CONSTEXPR_CHAR_TRAITS const Char* end() const FMT_NOEXCEPT {
     return buffer + std::char_traits<Char>::length(buffer);
   }
 
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -43,6 +43,22 @@
 #  define FMT_HAS_GXX_CXX11 0
 #endif
 
+// Check if constexpr std::char_traits<>::compare,length is supported.
+// libstdc++: present on GCC 7 and newer and __cplusplus >= 201703L
+// MSVC, libc++: always if __cplusplus >= 201703L
+// NOTE: FMT_GCC_VERSION  - is not libstdc++ version.
+//       _GLIBCXX_RELEASE - is present in GCC 7 libstdc++ and newer.
+#if __cplusplus >= 201703L
+#  ifndef __GLIBCXX__
+#    define FMT_CONSTEXPR_CHAR_TRAITS constexpr
+#  elif defined(_GLIBCXX_RELEASE) && _GLIBCXX_RELEASE >= 7
+#    define FMT_CONSTEXPR_CHAR_TRAITS constexpr
+#  endif
+#endif
+#ifndef FMT_CONSTEXPR_CHAR_TRAITS
+#  define FMT_CONSTEXPR_CHAR_TRAITS
+#endif
+
 #ifdef __NVCC__
 #  define FMT_NVCC __NVCC__
 #else
@@ -415,9 +431,7 @@ template <typename Char> class basic_string_view {
     the size with ``std::char_traits<Char>::length``.
     \endrst
    */
-#if __cplusplus >= 201703L  // C++17's char_traits::length() is constexpr.
-  constexpr
-#endif
+  FMT_CONSTEXPR_CHAR_TRAITS
       FMT_INLINE
       basic_string_view(const Char* s)
       : data_(s) {
@@ -457,7 +471,7 @@ template <typename Char> class basic_string_view {
   }
 
   // Lexicographically compare this string reference to other.
-  int compare(basic_string_view other) const {
+  FMT_CONSTEXPR_CHAR_TRAITS int compare(basic_string_view other) const {
     size_t str_size = size_ < other.size_ ? size_ : other.size_;
     int result = std::char_traits<Char>::compare(data_, other.data_, str_size);
     if (result == 0)
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -2149,7 +2149,7 @@ FMT_CONSTEXPR OutputIt write(OutputIt out, Char value) {
 }
 
 template <typename Char, typename OutputIt>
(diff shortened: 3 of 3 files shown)


include/fmt/format.h
```diff
@@ -2149,7 +2149,10 @@ FMT_CONSTEXPR OutputIt write(OutputIt out, Char value) {
 }
 
 template <typename Char, typename OutputIt>
-FMT_CONSTEXPR OutputIt write(OutputIt out, const Char* value) {
+#if FMT_HAS_CONSTEXPR_CHAR_TRAITS  // C++17's char_traits::length() is constexpr.
+FMT_CONSTEXPR
```
alexezeder, line 2153: No need to use `FMT_CONSTEXPR` macro here, just `constexpr` would be enough.

include/fmt/core.h
```diff
@@ -457,6 +468,9 @@ template <typename Char> class basic_string_view {
   }
 
   // Lexicographically compare this string reference to other.
+#if FMT_HAS_CONSTEXPR_CHAR_TRAITS  // C++17's char_traits::compare() is constexpr.
+  constexpr
+#endif
```
alexezeder, line 473: No need to make this function `constexpr` in this PR, I think.
This can be done as needed in the following PRs.

include/fmt/color.h
```diff
@@ -432,7 +432,10 @@ template <typename Char> struct ansi_color_escape {
   FMT_CONSTEXPR operator const Char*() const FMT_NOEXCEPT { return buffer; }
 
   FMT_CONSTEXPR const Char* begin() const FMT_NOEXCEPT { return buffer; }
-  FMT_CONSTEXPR const Char* end() const FMT_NOEXCEPT {
+#if FMT_HAS_CONSTEXPR_CHAR_TRAITS  // C++17's char_traits::length() is constexpr.
+  FMT_CONSTEXPR
```
alexezeder, line 436: No need to use `FMT_CONSTEXPR` macro here, just `constexpr` would be enough.

include/fmt/core.h
```diff
@@ -43,6 +43,17 @@
...
+#if __cplusplus >= 201703L
+#  ifndef __GLIBCXX__
+#    define FMT_HAS_CONSTEXPR_CHAR_TRAITS 1
+#  elif defined(_GLIBCXX_RELEASE) && _GLIBCXX_RELEASE >= 7
+#    define FMT_HAS_CONSTEXPR_CHAR_TRAITS 1
+#  endif
+#endif
+#ifndef FMT_HAS_CONSTEXPR_CHAR_TRAITS
+#  define FMT_HAS_CONSTEXPR_CHAR_TRAITS 0
+#endif
```
alexezeder, line 55: I suggest you comment on this check because it's not a really straightforward one 🙂

alexezeder (changes requested):
Just a few suggestions 😉

include/fmt/core.h
phprus: This change is for consistency across all uses of ``constexpr std::char_traits``.

alexezeder: All I'm just saying is that this function is non-`constexpr` currently, unlike `basic_string_view` constructor or `write(const Char* value)`, so there is no need to make it `constexpr` in this PR.

vitaut (changes requested):
Thanks for the PR. In general looks good but let's replace `FMT_HAS_CONSTEXPR_CHAR_TRAITS` with `FMT_CONSTEXPR_CHAR_TRAITS` that expands to `constexpr` or nothing depending on availability of constexpr char traits. This will make usage cleaner which is important because we plan to use this macro in multiple places.

vitaut:
Actually looking at the latest diff, you already made the suggested change =). GitHub confusingly shows the old code in Conversation.

vitaut (approved):

alexezeder:
It looks like there is a problem here that no one can ever expect: https://godbolt.org/z/Ps7q8dGME
TLDR: `__cplusplus` is `030c1fH = 199711` on MSVC even with `/std:c++17` option, because [it's MSVC](https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/) I guess...
🤦 🤦 🤦

phprus:
Oh!!!
My big mistake...

I can make pull request tonight.

Fix:
```
#if __cplusplus >= 201703L
#  ifndef __GLIBCXX__
#    define FMT_CONSTEXPR_CHAR_TRAITS constexpr
#  elif defined(_GLIBCXX_RELEASE) && _GLIBCXX_RELEASE >= 7
#    define FMT_CONSTEXPR_CHAR_TRAITS constexpr
#  endif
#endif
```
to:
```
#if defined(_MSC_VER)
// VS 2017 15.7+ (https://docs.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-150)
#  if _MSVC_LANG >= 201703L && _MSC_VER >= 1914
#    define FMT_CONSTEXPR_CHAR_TRAITS constexpr
#  endif
#elif __cplusplus >= 201703L
#  if defined(_LIBCPP_VERSION) && _LIBCPP_VERSION >= 4000 
// libc++4.0+ (https://libcxx.llvm.org/docs/Cxx1zStatus.html)
#    define FMT_CONSTEXPR_CHAR_TRAITS constexpr
#  elif defined(__GLIBCXX__) && defined(_GLIBCXX_RELEASE) && _GLIBCXX_RELEASE >= 7
#    define FMT_CONSTEXPR_CHAR_TRAITS constexpr
#  endif
#endif
```

New comment:
```
// Check if constexpr std::char_traits<>::compare,length is supported.
// libstdc++: GCC 7 and newer and __cplusplus >= 201703L
// MSVC: VS 2017 15.7 and newer and /std:c++17
// libc++: 4.0 and newer and if __cplusplus >= 201703L
// NOTE: FMT_GCC_VERSION  - is not libstdc++ version.
//       _GLIBCXX_RELEASE - is present in GCC 7 libstdc++ and newer.
```

phprus:
@alexezeder, Can you make a pull request?

alexezeder:
Actually I already somehow fixed this with https://github.com/fmtlib/fmt/pull/2247/commits/8a03ec0a17aa746b06a2e13f563be05b3443dcea:
```diff
- #if __cplusplus >= 201703L
+ #if __cplusplus >= 201703L || (defined(_MSVC_LANG) && _MSVC_LANG >= 201703L)
```
Looking at your proposal, probably it's not enough since you have an additional `_MSC_VER >= 1914` check there.

> Can you make a pull request?

Of course I can, but it's not needed right now because, as I said, I make it work for my PR. Feel free to make PR at your convenience.

phprus:
```_MSVC_LANG >= 201703L``` - VS2017 version 15.3+
```constexpr char_traits``` - VS 2017 version 15.7+

PR #2253, opened 2021-04-25
__THROW warning fix for e2k

ilyakurdyukov:


vitaut (changes requested):
Thanks for the PR. Looks good but please apply clang-format.

PR #2255, opened 2021-04-26
Keep defaulted destructors inline

DanielaE:
applies to exception classes in case of msvc only

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -77,6 +77,12 @@
 #  define FMT_GCC_VISIBILITY_HIDDEN
 #endif
 
+#if FMT_MSC_VER
+#  define FMT_MSC_DEFAULT = default
+#else
+#  define FMT_MSC_DEFAULT
+#endif
+
 #if __cplusplus == 201103L || __cplusplus == 201402L
 #  if defined(__INTEL_COMPILER) || defined(__PGI)
 #    define FMT_FALLTHROUGH
@@ -865,7 +871,7 @@ class FMT_API format_error : public std::runtime_error {
   format_error& operator=(const format_error&) = default;
   format_error(format_error&&) = default;
   format_error& operator=(format_error&&) = default;
-  ~format_error() FMT_NOEXCEPT FMT_OVERRIDE;
+  ~format_error() FMT_NOEXCEPT FMT_OVERRIDE FMT_MSC_DEFAULT;
 };
 
 FMT_MODULE_EXPORT_END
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR but could you elaborate why do this change?

DanielaE:
Otherwise I see this
```
fmt-m-pp.lib(fmt-m-pp.obj) : error LNK2005: "public: virtual __cdecl fmt::v7::format_error::~format_error(void)" (??1format_error@v7@fmt@@UEAA@XZ::<!fmt>) already defined in check.obj
check-m-pp.exe : fatal error LNK1169: one or more multiply defined symbols found

```
in my module compile check as soon as I add an example with a local custom formatter like this:
```
enum class check_color {red, green, blue};

namespace fmt {
template<> struct formatter<check_color> : formatter<string_view> {
// parse is inherited
    auto format(check_color c, FormatContext& ctx) {
        string_view name = "?";
        switch (c) {
            case check_color::red:   name = "red"; break;
            case check_color::green: name = "green"; break;
            case check_color::blue:  name = "blue"; break;
        }
        return formatter<string_view>::format(name, ctx);
    }
};    
}
result = fmt::format("{}", check_color::blue);
```
This is probably a defect in msvc, but the change looks beneficial to me for msvc also in the non-modular case. I can see no reason, not to do it. There are no related warnings emitted if I crank the warning level up to maximum and/or enable DLL export. But I can't tell for sure in every usecase, therefor the WIP status.

include/fmt/format.h
```diff
@@ -862,7 +862,11 @@ class FMT_API format_error : public std::runtime_error {
   format_error& operator=(const format_error&) = default;
   format_error(format_error&&) = default;
   format_error& operator=(format_error&&) = default;
-  ~format_error() FMT_NOEXCEPT FMT_OVERRIDE;
+  ~format_error() FMT_NOEXCEPT FMT_OVERRIDE
+#if FMT_MSC_VER
+    = default
+#endif
```
vitaut, line 868: I suggest moving this into a separate macro `FMT_MSC_DEFAULT` that expands to `= default` if `FMT_MSC_VER` is set.

vitaut (changes requested):

include/fmt/format.h
DanielaE: Yeah, just my thought. I wasn't sure if a single occurence warrants another macro. But now that it's blessed I will change it as suggested tomorrow.

PR #2256, opened 2021-04-26
Update GTest/GMock structure, apply some fixes

alexezeder:
* all GTest/GMock files moved to `test/gtest` directory
* `CMakeLists.txt` created in `test/gtest` from `CMakeLists.txt` in `test`
* GTest/GMock target in CMake renamed to `gtest` (was `gmock`)
* CMake `gtest` target updated to export includes as "gtest/gtest.h" or "gmock/gmock.h" only
* includes in tests updated: "gtest.h" -> "gtest/gtest.h", "gmock.h" -> "gmock/gmock.h"
* removed duplications of `target_include_directories` for GTest/GMock directories (CMake manages them)
* `.clang-format` file added into `test/gtest` directory to prevent formatting there
* obsolete `GTEST_LANG_CXX11` compile definition setting removed
* `std::is_trivially_copy_constructible` for GCC 4.8 & 4.9 fixed properly


vitaut (changes requested):
Thanks for the PR. In general looks good but let's use the directory structure generated by the fuse script (not introduce include directories) and use gtest instead of gmock as the top-level directory since the gmock component is way less important:

```
gtest/
  gtest/gtest.h
  gmock/gmock.h
  gmock-gtest-all.cc
```

alexezeder:
> use gtest instead of gmock as the top-level directory since the gmock component is way less important

Maybe I should also rename CMake target then?

vitaut:
> Maybe I should also rename CMake target then?

Good idea.

alexezeder:
> let's use the directory structure generated by the fuse script (not introduce include directories)
> ```
> gtest/
>   gtest/gtest.h
>   gmock/gmock.h
>   gmock-gtest-all.cc
> ```

Sorry for breaking your message into pieces, but I just realized that `include` directory was introduced on purpose. As we saw  [here](https://github.com/fmtlib/fmt/pull/2247#issuecomment-827157616), this structure makes compilers use GTest/GMock includes as a relative path, but not a system, while building `gmock-gtest-all.cc`, because of `""`, not `<>` includes. At the same time, only the system includes allows us to get rid of potential compiler warnings. Of course, there was a problem with `std::pod` only, but who knows what else suspicious can GTest library has by itself.

alexezeder:
On the other hand, maybe warnings (not errors, if no other `CXX_FLAGS` passed to CMake) from the GTest library, while building (currently) `gmock` CMake target, are not so bad to change the structure...

vitaut:
> maybe warnings (not errors, if no other CXX_FLAGS passed to CMake) from the GTest library, while building (currently) gmock CMake target, are not so bad to change the structure...

Yeah, I think warnings are OK.

alexezeder:
I updated this PR with few additional fixes like a fix for `std::is_pod` usage.

Also, there are 2 places I want to discuss:

1. https://github.com/fmtlib/fmt/blob/342973b3498e978d322586b3f6a536d7a7287046/test/CMakeLists.txt#L45-L50
   maybe it should be moved to the new `CMakeLists.txt` file? Because there are some fixes for MSVC already.
2. https://github.com/fmtlib/fmt/blob/342973b3498e978d322586b3f6a536d7a7287046/test/CMakeLists.txt#L52-L56
   maybe it should be removed? 😄

vitaut:
Awesome, thanks!

alexezeder:
Great, but what do you think about https://github.com/fmtlib/fmt/pull/2256#issuecomment-828842031? 🙂

vitaut:
> what do you think about #2256 (comment)?

Both suggestions look reasonable to me.

PR #2257, opened 2021-04-27
C++17: std::char_traits<>::{compare,length} is constexpr - v2

phprus:
Fixed my mistake in the previous patch.

Fix for Microsoft Visual Studio and clang/libc++ < 4.0.

Discussion: https://github.com/fmtlib/fmt/pull/2246#issuecomment-827184981

Links:
    https://docs.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-150
    https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-150
    https://libcxx.llvm.org/docs/Cxx1zStatus.html

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -44,14 +44,27 @@
 #endif
 
 // Check if constexpr std::char_traits<>::compare,length is supported.
-// libstdc++: present on GCC 7 and newer and __cplusplus >= 201703L
-// MSVC, libc++: always if __cplusplus >= 201703L
+//
+// libstdc++: GCC 7 and newer and __cplusplus >= 201703L
+// MSVC     : VS 2017 15.7 and newer and /std:c++17
+// https://docs.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance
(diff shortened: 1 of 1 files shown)


vitaut:
Thank you!

vitaut:
BTW Jonathan Wakely suggested using `__cpp_lib_constexpr_char_traits` to detect this in libstdc++: https://twitter.com/wakomeup/status/1387422935590260738.

phprus:
https://github.com/gcc-mirror/gcc/blob/releases/gcc-7.1.0/libstdc++-v3/include/bits/char_traits.h:
```
// #define __cpp_lib_constexpr_char_traits 201611
```
Macro ```__cpp_lib_constexpr_char_traits``` defined only in gcc-8 and newer, but constexpr char_traits is presented in gcc-7.

Tweet not found ("Hmm...this page doesn’t exist. Try searching for something else.") :(

vitaut:
Although on further investigation it turned out that `__cpp_lib_constexpr_char_traits` is not defined consistently.

vitaut:
> Tweet not found ("Hmm...this page doesn’t exist

It was retracted =)

Issue #2022, opened 2020-11-15
Compile-time format string checks by default


vitaut:
In C++20 we can have format string checks with the normal API, no macro or UDL is required (https://godbolt.org/z/qrh3ee):

```c++
#include <fmt/format.h>

template <typename... Args>
struct format_str {
  fmt::string_view str;

  template <size_t N>
  consteval format_str(const char (&s)[N]) : str(s) {
    using checker = fmt::detail::format_string_checker<
      char, fmt::detail::error_handler, Args...>;
    fmt::detail::parse_format_string<true>(
      fmt::string_view(s, N), checker(s, {}));
  }
};

template <class... Args>
std::string format(
    format_str<std::type_identity_t<Args>...> fmt,
    const Args&... args) {
  return fmt::format(fmt.str, args...);
}

int main() {
  auto s = format("{:d}", 42);
  fmt::print("{}\n", s);
}
```

alexezeder:
One small concern here is that compile-time checks work only for format strings that don't have named identifiers since, currently, named arguments are not able to provide their names at compile-time.
I'm not so concern about the same behavior of compile-time checks invoked by `FMT_STRING`, because it's kinda extension, which just fails to work with named identifiers by failing a compilation process.
But it feels strange, at least for me, to disable compile-time checks (that will be enabled by default) only for format strings with named identifiers.

vitaut:
Compile-time checks can potentially be made to work with a UDL form of named arguments.

alexezeder:
> Compile-time checks can potentially be made to work with a UDL form of named arguments.

But that's only one way of creating named arguments. The problem with `fmt::arg(name, value)` would still be relevant.

vitaut:
> The problem with fmt::arg(name, value) would still be relevant.

Right. Maybe some day we'll have parametric expressions or constexpr arguments that will make it possible.

vitaut:
Compile-time checks are enabled now in compilers that have working `consteval` (gcc 10+, clang 11+).

PR #2259, opened 2021-04-27
Support 'to_wstring(T)' in module.

DanielaE:
FMT_STRING(x) doesn't work in modules yet because of FMT_STRING_IMPL. In this particular case, this alternative implementation might even be superior?

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -3741,6 +3741,21 @@ arg_join<detail::iterator_t<Range>, detail::sentinel_t<Range>, wchar_t> join(
   return join(std::begin(range), std::end(range), sep);
 }
 
+FMT_MODULE_EXPORT_END
+namespace detail {
+
+template <typename Char, typename T>
+inline std::basic_string<Char> integral_to_string(T value) {
+  // The buffer should be large enough to store the number including the sign or
+  // "false" for bool.
+  constexpr int max_size = detail::digits10<T>() + 2;
+  Char buffer[max_size > 5 ? static_cast<unsigned>(max_size) : 5];
+  Char* begin = buffer;
+  return std::basic_string<Char>(begin, detail::write<Char>(begin, value));
+}
+
+}
+FMT_MODULE_EXPORT_BEGIN
 /**
   \rst
   Converts *value* to ``std::string`` using the default format for type *T*.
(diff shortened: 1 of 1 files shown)


alexezeder:
Is it the only place where `FMT_STRING` cannot be used with modules? For example, `FMT_STRING` also used in `ranges.h`:
https://github.com/fmtlib/fmt/blob/84a36b99bfe0898daf194b666575fcfd5e23673c/include/fmt/ranges.h#L364

> FMT_STRING(x) doesn't work in modules yet because of FMT_STRING_IMPL

Does it mean that you want to make it work in the future? 🙂 
IMHO, it's not worth it, as well as `FMT_COMPILE`, since you also mentioned that macro [here](https://github.com/fmtlib/fmt/pull/2240#issue-617277763):

> The macro-based interfaces FMT_STRING and FMT_COMPILE are not yet implemented. Macros cannot get exported from modules.

As far as I can see, there is no need to export `FMT_COMPILE` macro from {fmt} module because `_cf` literal makes it obsolete.

It's a little more complicated with `FMT_STRING` - there is no literal for that. Still, there are (or will be, IDK, @vitaut knows it better) [Compile-time format string checks by default](https://github.com/fmtlib/fmt/issues/2022) that has one flaw (as I see it) - no other option but to have implicit fallback without checks for non-statically named arguments. `FMT_STRING` doesn't have this flaw because it handles this explicitly.

Aside from the concern with `FMT_STRING`, looks like {fmt} module can be fully functional without these macros. So probably macro-based interface can just be omitted. What do you think?

DanielaE:
At present, `ranges.h` is not part of modularized {fmt}.

I do have an implementation of FMT_STRING and FMT_COMPILE which is based on udl-based compile-time literal arguments. I still haven't looked into possible other solutions like immediate functions. Unfortunately, the current implementation is not yet supported in msvc's modules implementation (may be 16.10-pre3 will do). Even more unfortunately, no other compiler is even able to survive attempts to compile `fmt.cc`. Without the modules-related lines in there, every compiler should be fine and compile {fmt} into a single object file, with them an advanced compiler chould create a BMI, too.

vitaut:
Thanks for the suggestion but don't think it's necessary since this is an internal use of `FMT_STRING` that doesn't require this macro to be exported.

PR #2268, opened 2021-05-07
WIP: async format_arg_store buffer storage support

yumeyao:
## WIP: async format_arg_store buffer storage support

The idea of **async** storage is similar to `dynamic_format_arg_store`, by doing so, the **arg_store** and **format_str** can be queued up and let a consumer (it can be another thread) do the actual formatting. A typical scenario of it is a genuine **async** logger. I'd call it **async formatting logger**, rather than the common idea **async logger** which usually refers to **async flushing logger**.

But `dynamic_format_arg_store` has its drawbacks. It breaks triviality and aggregate construction, allocates memory by using a `std::vector`, disqualifies *is_packed* optimization...

To get rid from the overhead of `dynamic_format_arg_store`, here comes the idea of **async**. Initially it was called `fmtlog` because as explained above, the main scenario is **async formatting logger**. It's has been used in production in house for several months.  
When I was considering contributing this to public, I came up with the name **async**, because obviously async is not anything to be formatted by `libfmt` as other extensions, so there should not be a name confliction.  

## PR status
The logger part is not part of the contribution.  
This is quite immature on some parts of the design see **to be implemented/fixed/discussed** section.  
Also currently the code works on c++17 (or relaxed g++ that allows you using `if constexpr` without `c++17`) and un*x.  
code-style will be formatted whe it gets mature.

## features
* async - producing the `stored_entry` and cosuming it can happen on different threads.
* As fast as pure libfmt - if you just use it on the same thread (limited to non-copy types, explained below, see **layout**).
  * and even nearly identical assembly code, despite you must allocate some buffer large enough to hold things on the stack.  
    (say if you know the size in advance, which is possible right now if there is no copied buffers, you can almost make the identical code)
  * any optimization effort done by `libfmt` (like *is_packed*) is preserved.
  * despited copied buffers (see **layout**), any other offset/size of an arg is `constexpr`.
* No extra overhead on both producer side and consumer side, on both runtime-wise and template-bloat-wise.
  * producer: the constructor is as trivial as `format_arg_store`. This guarentees any possible optimizations from the compilers don't get disqualified.
  * consumer: the constructed entry itself already contains a `arg_store` that is layout-and-value-idential to `format_arg_store`. This guarantees the `fmt::format()` and likewise function can use it directly without extra processing.
* custom types supported, with **custom storage** support (as an extension to custom storage)

## modifications on libfmt
A private constructor of `basic_format_args` and a `static constexpr` member field of `format_arg_store` must be accessible. Currently this is done by making them public. Of course we can use `friend` declaration instead.

## layout
```
// A stored entry looks like:
// -----------------------------------------------------------------------
// | basic_async_entry | arg_store | stored_objs... | stored_buffers... |
// -----------------------------------------------------------------------
```
* basic_async_entry contains `format_str`, `desc_` (the `static constexpr` field mentioned above), and `dtor` (in case object copy happens and the destructor is non-trivial).
* arg_store is `format_arg_store` itself. layout-wise identical. aggregate construction.
* stored_objs stores any temproray objects (custom types), including move-constructed `std::string` from `std::string&&`.
* stored_buffers stores any buffer, like a **string** copied by buffer directly from `std::string_view|const std::string&|const char*`. Also custom storage is supported. For example you can let **hexdump** only copies the buffer, and let the consumer do the bin-to-hex conversion.
 
## public APIs
* `auto make_async_entry(format_str, args...)` creates an async entry. Only scalar types supported. It's maybe just used to demonstrate the **nealy identicaly assembly code to libfmt** feature.
* `size_t store_async_entry(void* buf, format_str, args...)` creates an async entry on **buf** (placement-new) with **transformed args** (*copy/move-construct args according to some rule*).
  * returns total size of the entry. Summing up all 4 sections mentioned above.
  * The basic idea is just (not the actual code, only for demonstration purpose):
    ```
    template <typename S, typename... Args, typename Char = char_t<S>>
    size_t store_async_entry(void* buf, const S& format_str, Args&&... args) {
        using async_entry_type = ...;
        using TRANSFORMER = ...;
        TRANSFORMER x(buf);
        new (buf) async_entry_type(   x.transform(std::forward<Args>(args))...   );
        // sizeof(async_entry_type) == sizeof(basic_async_entry) + sizeof(arg_store)
        // x.total_size() == sizeof(async_entry_type) + sizeof(stored_objs... and stored_buffers...)
        return x.total_size();
    }
    ```
  * Like memtioned above in **layout**, strings of different types will be handled distinguishedly. Eventually they will be transformed to `string_view`.
* `format|format_to|print(entry&, ...)` same usage as `libfmt` APIs.
* `entry.format|format_to|print(...)` don't destruct the `entry`.
  * use `entry::dtor_sentry(entry)` to destruct the `entry` upon the sentry destruction.

### custom buffer storage
An example of **hexdump**.
```
template <typename Iterator>
struct formatter<fmtlog::detail::hexdump_range<Iterator>> {
    fmtlog::detail::hexdump_format parsed_format;

    template<typename ParseContext>
    auto parse(ParseContext& ctx) -> decltype(ctx.begin()) { return parsed_format.parse(ctx); }

    template<typename FormatContext>
    auto format(const fmtlog::detail::hexdump_range<Iterator>& range, FormatContext& ctx) -> decltype(ctx.out()) {
        return range.format(ctx, parsed_format);
    }

//////////////////////////////////////
//// 1. note the return type. Then hexdump_stored_buffer& will take place of the original arg, and will be forwarded into entry construction.
//// 2. must be static
//// 3. the first argument is reference.
//// 4. you can even add another overload of T&&.
//// 5. of course this design is immature, but it wfm.
    static fmtlog::detail::hexdump_stored_buffer& store(char*& dest, const fmtlog::detail::hexdump_range<Iterator>& range) {
        return fmtlog::detail::hexdump_stored_buffer::store(dest, range);
    }
//////////////////////////////////////
};
```

## limitation
* no alignment specification support on stored buffer
* the entry must be aligned to 16 bytes(libfmt limitation). currently the public APIs don't do this for you. you may consider write a wrapper function before you use.
* `format_str` must be compile-time string. It's never copied.

## to be implemented/fixed/discussed (help from community wanted)
* named_arg not working on `store_async_entry()` yet. I just don't have the time to do it so far(because we don't use this feature in house). Sould be easy to implement.
* `std::ref()` support to capture the reference rather than copy the object.
* Compile-time string-literals in args don't need a whole copy of the string. However it's impossible to distinguish them so they're still copied.
  * solution1: `std::ref("my compile-time string")`
  * solution2: `c++11` custom string literals, like `"my compile-time string"cts` (cts: compile-time-string)
    * I'm seeing some discussion to enforce compile-time string check on `format_string`, if the UDL solution is chosen, we can use the same suffix.
* Currently there is no way to check if the buffer size is efficient. This is limited by the **custom storage** feature. Maybe we can change the prototype of custom storage function from  
  `static ret_type store(char*&, const T&/T&&)`  
  to  
  `static std::pair<ret_type, size_t> store(BUFFER_ITERATOR&, const T&/T&&)`  
  (where **BUFFER_ITERATOR** is a template argument, which can be something like `decltype(detail::counting_buffer<>().begin())`).
* Non-un*x OS support lacking because of the use of `stpcpy()|wcpcpy()`. `C23` has `memccpy()` and Windows has [`_memccpy()`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/memccpy). But I haven't tested the performance of `memccpy()` vs `strlen(); memcpy()`.
* And... of course, the test is pending enriched... Though there are already some production code in house using different args and placing the buffer on different memory addresses running for months so the basic functionality should be ok... Any suggestion like adding a test case is welcomed.
* code clean-up is welcomed too.
  * If you found a more intuitive way to write some part of the code.
  * compatibility/portable issues.
    * currently there are 4 `if constexpr` that needs to be converted to `c++11/14` friendly style.
* **code-styling** not open for now (except for CamelCase/snake_case fix). When this gets ready to merge, I'll format the code according to `libfmt` code style.

Should you feel needed, please feel free to open an issue/pr ticket on my fork (for example you think a new thread is a better way to track some specific discussion topic).

### P.S.  
In case you're curious: I do use *reddit*. I'm mainly a shit-poster on /r/dota2(afk for a long time since reddit got blocked in China). So let's discuss things here on github.  
Something happens on *reddit* and *another repo* just copies my idea of string buffer storage according to my **interpretation**, yes, I know that guy in person, and I believe he doesn't understand my code (obviously I showed him the experimental code `essential fmtlog done` which later became the motivation of his work) and I did interprete the detail to him.  
I don't want any drama happen on github, let's discuss and **improve the design and code**.

### DIFF
--- a/include/fmt/async.h
+++ b/include/fmt/async.h
@@ -1,0 +1,323 @@
+#ifndef FMT_ASYNC_H_
+#define FMT_ASYNC_H_
+
+#include "format.h"
+#include <tuple>
+
+namespace fmt {
+
+template <typename Context> struct basic_async_entry {
+protected:
+    using char_type = typename Context::char_type;
+    using format_arg = typename basic_format_args<Context>::format_arg;
+    using arg_destructor = void (*)(void *p);
+
+    basic_string_view<char_type> format_;
+    unsigned long long desc_;
+    arg_destructor dtor_;
+
+    FMT_CONSTEXPR basic_async_entry(basic_string_view<char_type> format) : format_(format), desc_(0), dtor_(0) {}
+    const format_arg* get_format_args() const;
+
+    template <typename OutIt, typename T = OutIt>
+    using enable_out = enable_if_t<detail::is_output_iterator<OutIt, char_type>::value, T>;
+
+    void destruct() { if (dtor_) dtor_(this); }
+public:
+    struct dtor_sentry {
+        dtor_sentry(basic_async_entry& entry) : entry_(entry) {}
+        ~dtor_sentry() { entry_.destruct(); }
+        basic_async_entry& entry_;
+    };
+
+    // libfmt public APIs
+    std::basic_string<char_type> format() const { return vformat(format_, {desc_, get_format_args()}); }
+
+    template <typename OutIt>
+    auto format_to(OutIt out) const -> enable_out<OutIt> {
+        return vformat_to(out, format_, {desc_, get_format_args()});
+    }
+
+    // template <typename OutIt>
+    // auto format_to(OutIt out, size_t n) const -> enable_out<OutIt, format_to_n_result<OutIt>> {
+    //     return vformat_to(detail::truncating_iterator<OutIt>(out, n), format_, {desc_, get_format_args()});
+    // }
+
+    size_t formatted_size() const {
+        detail::counting_buffer<> buf;
+        format_to(buf);
+        return buf.count();
+    }
+
+    void print(std::FILE* file = stdout) const { return vprint(file, format_, {desc_, get_format_args()}); }
+
+};
+
+template <typename Context, typename... Args> struct async_entry : basic_async_entry<Context> {
+    using entry = basic_async_entry<Context>;
+
+    format_arg_store<Context, Args...> arg_store_;
+
+    template <typename S>
+    FMT_CONSTEXPR async_entry(const S& format_str, const Args&... args) : entry(to_string_view(format_str)), arg_store_(args...) {
+        entry::desc_ = arg_store_.desc;
+    }
+    FMT_CONSTEXPR void set_dtor(typename entry::arg_destructor dtor) { this->dtor_ = dtor; }
+};
+
+template <typename Context>
+inline const typename basic_async_entry<Context>::format_arg* basic_async_entry<Context>::get_format_args() const {
+    union obfuscated_args {
+        const detail::value<Context>* values_;
+        const format_arg* args_;
+        intptr_t pointer_; // more efficient to add integer with size, as the compiler is able to avoid emitting branch
+    } args;
+    auto& entry = static_cast<const async_entry<Context>&>(*this);
+    args.values_ = entry.arg_store_.data_.args_;
+    if (entry.desc_ & detail::has_named_args_bit) {
+        args.pointer_ += (desc_ & detail::is_unpacked_bit) ? sizeof(*args.args_) : sizeof(*args.values_);
+    }
+    return args.args_;
+}
+
+//
+// A stored entry looks like:
+// -----------------------------------------------------------------------
+// | basic_async_entry | arg_store | stored_objs... | stored_buffers... |
+// -----------------------------------------------------------------------
+//
+namespace async {
+namespace detail {
+namespace detail = fmt::detail;
+template <typename T> using decay_t = typename std::decay<T>::type;
+template <typename T> using add_const_t = typename std::add_const<T>::type;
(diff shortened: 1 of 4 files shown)


yumeyao:
out of my expectation, MSVC has `stpcpy()` and supports relaxed `if constexpr` std requirement too.

anyway these still need to be fixed.

`stpcpy()` can be fixed using the fallback pattern as `localtime_s()|localtime_r()|localtime()` in `chrono.h`.

Issue #2269, opened 2021-05-07
Proposal: Add formatter for std::error_code.


phprus:
The ```std::error_code``` has are ```operator<<``` (https://en.cppreference.com/w/cpp/error/error_code/operator_ltlt), which is equivalent ```os << ec.category().name() << ':' << ec.value() ```.

Formatting with ostream is 4-5 (C++11) or 11-24 (C++17) times slower than the custom formatter class.

Simpe and possibly incorrect implementation: https://github.com/phprus/fmt-bench/blob/format-error_code/test_formatter.cpp (I agree that this code are licensed under the {fmt} license, and agree to future changes to the licensing.)
Benchmark (https://github.com/phprus/fmt-bench/tree/format-error_code):

### AppleClang 11.0.0.11000033 C++11

```
Running ./fmt_test
Run on (8 X 2700 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x4)
  L1 Instruction 32 KiB (x4)
  L2 Unified 256 KiB (x4)
  L3 Unified 6144 KiB (x1)
Load Average: 1.79, 1.73, 1.69
-----------------------------------------------------------------
Benchmark                       Time             CPU   Iterations
-----------------------------------------------------------------
FMTIostream/22                369 ns          368 ns      1953098
FMTIostreamCompile/22         385 ns          384 ns      1863938
FMTFormatter/22              75.9 ns         75.8 ns      9392947
FMTFormatterCompile/22       79.6 ns         79.5 ns      8838830
```

### AppleClang 11.0.0.11000033 C++17

```
Running ./fmt_test
Run on (8 X 2700 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x4)
  L1 Instruction 32 KiB (x4)
  L2 Unified 256 KiB (x4)
  L3 Unified 6144 KiB (x1)
Load Average: 1.68, 1.71, 1.69
-----------------------------------------------------------------
Benchmark                       Time             CPU   Iterations
-----------------------------------------------------------------
FMTIostream/22                389 ns          388 ns      1854806
FMTIostreamCompile/22         380 ns          378 ns      1794761
FMTFormatter/22              33.7 ns         33.7 ns     20944226
FMTFormatterCompile/22       16.4 ns         16.4 ns     42452029
```

vitaut:
Seems like a good idea, could you submit a PR?

phprus:
Yes, but I'm not sure if my implementation is correct.

[...]

PR #2270, opened 2021-05-08
Add a formatter specialization for std::error_code.

phprus:
Discussion: https://github.com/fmtlib/fmt/issues/2269

### DIFF
--- a/include/fmt/os.h
+++ b/include/fmt/os.h
@@ -129,6 +129,23 @@ class error_code {
   int get() const FMT_NOEXCEPT { return value_; }
 };
 
+template <typename Char> struct formatter<std::error_code, Char> {
+  template <typename ParseContext>
+  FMT_CONSTEXPR auto parse(ParseContext& ctx) -> decltype(ctx.begin()) {
+    return ctx.begin();
+  }
+
+  template <typename FormatContext>
+  FMT_CONSTEXPR auto format(const std::error_code& ec, FormatContext& ctx) const
+      -> decltype(ctx.out()) {
+    auto out = ctx.out();
+    out = detail::write<Char>(out, to_string_view(ec.category().name()));
+    out = detail::write<Char>(out, Char(':'));
+    out = detail::write<Char>(out, ec.value());
+    return out;
+  }
+};
+
 #ifdef _WIN32
 namespace detail {
 // A converter from UTF-16 to UTF-8.
--- a/test/os-test.cc
+++ b/test/os-test.cc
@@ -74,6 +74,30 @@ TEST(os_test, error_code) {
   EXPECT_EQ(error_code(42).get(), 42);
 }
 
+TEST(os_test, format_std_error_code) {
+  EXPECT_EQ("generic:42",
+            fmt::format(FMT_STRING("{0}"),
+                        std::error_code(42, std::generic_category())));
+  EXPECT_EQ("system:42",
+            fmt::format(FMT_STRING("{0}"),
+                        std::error_code(42, std::system_category())));
+  EXPECT_EQ("system:-42",
+            fmt::format(FMT_STRING("{0}"),
+                        std::error_code(-42, std::system_category())));
+}
+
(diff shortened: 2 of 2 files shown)


include/fmt/format.h
```diff
@@ -3061,6 +3061,24 @@ template <> struct formatter<bytes> {
...
+  FMT_CONSTEXPR auto format(const std::error_code& ec, FormatContext& ctx) const
+      -> decltype(ctx.out()) {
+    auto out = ctx.out();
+    out = detail::write<Char>(out, to_string_view(ec.category().name()));
+    out = detail::write<Char>(out, Char(':'));
+    out = detail::write<Char>(out, ec.value());
+    return out;
+  }
+};
+
```
vitaut, line 3081: This should go to `fmt/os.h` next to the `fmt::error_code` type.

test/compile-test.cc
```diff
@@ -192,6 +192,30 @@ TEST(compile_test, format_to_n) {
...
+            fmt::format(FMT_COMPILE(L"{0}"),
+                        std::error_code(42, std::generic_category())));
+  EXPECT_EQ(L"system:42",
+            fmt::format(FMT_COMPILE(L"{0}"),
+                        std::error_code(42, std::system_category())));
+  EXPECT_EQ(L"system:-42",
+            fmt::format(FMT_COMPILE(L"{0}"),
+                        std::error_code(-42, std::system_category())));
+}
+
```
vitaut, line 218: No need to replicate the other formatting tests. Please remove these and add one check that parse is constexpr to the other test.

test/format-test.cc
```diff
@@ -2494,6 +2494,24 @@ TEST(format_test, test_formatters_enabled) {
...
+}
+
+TEST(format_test, format_std_error_code_wide) {
+  EXPECT_EQ(L"generic:42",
+            fmt::format(L"{0}", std::error_code(42, std::generic_category())));
+  EXPECT_EQ(L"system:42",
+            fmt::format(L"{0}", std::error_code(42, std::system_category())));
+  EXPECT_EQ(L"system:-42",
+            fmt::format(L"{0}", std::error_code(-42, std::system_category())));
+}
```
vitaut, line 2513: This should go to `os-test.cc`.

vitaut (changes requested):
Thanks for the PR. Looks good in general, a few comments inline.

include/fmt/os.h
```diff
@@ -129,6 +129,24 @@ class error_code {
   int get() const FMT_NOEXCEPT { return value_; }
 };
 
+// A formatter specialization for std::error_code.
```
vitaut, line 132: Let's remove this comment since it's obvious from the next line that this is a formatter specialization for `std::error_code`.

test/compile-test.cc
```diff
@@ -84,6 +85,9 @@ TEST(compile_test, format_default) {
 #  ifdef __cpp_lib_byte
   EXPECT_EQ("42", fmt::format(FMT_COMPILE("{}"), std::byte{42}));
 #  endif
+  EXPECT_EQ("generic:42",
+            fmt::format(FMT_COMPILE("{}"),
+                        std::error_code(42, std::generic_category())));
```
vitaut, line 90: Let's remove this too as well as the fmt/os.h include above.

test/os-test.cc
```diff
@@ -74,6 +74,24 @@ TEST(os_test, error_code) {
   EXPECT_EQ(error_code(42).get(), 42);
 }
 
+TEST(os_test, format_std_error_code) {
+  EXPECT_EQ("generic:42",
+            fmt::format("{0}", std::error_code(42, std::generic_category())));
```
vitaut, line 79: Please wrap the format string in FMT_STRING - this will guarantee that parsing is constexpr and will ensure that the formatter works with format string compilation.

vitaut (changes requested):

vitaut:
Merged, thanks!

Issue #2266, opened 2021-05-05
Question about using custom error_code/fmt::system_error in C++11


phprus:
Tell me please why {fmt} uses its own implementation of ```int error_code``` and ```fmt::system_error```/```fmt::windows_error``` instead of ```std::error_code``` and ```std::system_error``` from C++11 standard?

vitaut:
`fmt::system_error` and `fmt::windows_error` predate C++11 requirement. We could probably migrate to `std::system_error` now.

vitaut:
Custom exception classes are gone now. Thanks for the suggestion.

phprus:
Maybe need a replace ```fmt::error_code``` to ```std::error_code```?

https://github.com/fmtlib/fmt/blob/master/include/fmt/os.h#L122
https://github.com/fmtlib/fmt/compare/cleanup#diff-9fb4e28ec2b9c93ff34b557433378a7f22ebdef90a692bf372896f18347afa0bR72

vitaut:
Maybe

PR #2273, opened 2021-05-09
Replace fmt::error_code to std::error_code

phprus:
Discussion: https://github.com/fmtlib/fmt/issues/2266#issuecomment-835353462


vitaut:
Thank you!

PR #2282, opened 2021-05-13
add missing header

DanielaE:
The exception-related changes added this new dependency that must be synchronized to the primary module interface unit.


vitaut:
Thank you!

Issue #2284, opened 2021-05-13
Warning: msvc 16.10-pre3 clashes with {fmt}


DanielaE:
For some not yet investigated reason, `<format>` from the STL gets pulled into `{fmt}`'s test for `compile.h`. This might affect user code as well. Interesting times ...

```
Microsoft (R) C/C++ Optimizing Compiler Version 19.29.30035 for x64
compile-test.cc
E:\fmt-module\fmtlib\include\fmt/compile.h(325): error C2668: 'std::make_format_args': ambiguous call to overloaded function
C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview-IDE\VC\Tools\MSVC\14.29.30035\include\format(2848): note: could be 'auto std::make_format_args<fmt::v7::basic_format_context<std::back_insert_iterator<std::string>,char>,_Ty>(const _Ty &)' [found using argument-dependent lookup]
        with
        [
            _Ty=std::chrono::duration<__std_tzdb_epoch_milli,std::milli>
        ]
E:\fmt-module\fmtlib\include\fmt\core.h(1640): note: or       'fmt::v7::format_arg_store<fmt::v7::basic_format_context<std::back_insert_iterator<std::string>,char>,_Ty> fmt::v7::make_format_args<fmt::v7::basic_format_context<std::back_insert_iterator<std::string>,char>,_Ty>(const _Ty &)'
        with
        [
            _Ty=std::chrono::duration<__std_tzdb_epoch_milli,std::milli>
        ]
E:\fmt-module\fmtlib\include\fmt/compile.h(323): note: while trying to match the argument list '(const _Ty)'
        with
        [
            _Ty=std::chrono::duration<__std_tzdb_epoch_milli,std::milli>
        ]
E:\fmt-module\fmtlib\include\fmt/compile.h(559): note: see reference to function template instantiation 'OutputIt fmt::v7::detail::spec_field<char_type,std::chrono::duration<__std_tzdb_epoch_milli,std::milli>,0>::format<std::back_insert_iterator<std::string>,_Ty>(OutputIt,const _Ty &) const' being compiled
        with
        [
            OutputIt=std::back_insert_iterator<std::string>,
            _Ty=std::chrono::duration<__std_tzdb_epoch_milli,std::milli>
        ]
[...]

PR #2287, opened 2021-05-14
Prevent ambiguity in name lookup

DanielaE:
Unqualified calls to `make_format_args` may find the same name by ADL in `<format>` if this C++20 header happens to be directly or indirectly included in a translation unit. Do a qualified lookup instead.


DanielaE:
This addresses issue #2284

vitaut:
Thank you!

PR #2288, opened 2021-05-14
Do *not* export namespace `detail`

DanielaE:


include/fmt/chrono.h
```diff
@@ -458,6 +456,7 @@ template <typename Char> struct formatter<std::tm, Char> {
   basic_string_view<Char> specs;
 };
 
+FMT_MODULE_EXPORT_END
 namespace detail {
```
vitaut, line 460: I suggest adding `FMT_BEGIN_DETAIL_NAMESPACE` / `FMT_END_DETAIL_NAMESPACE` macros to avoid repeating this sequence throughout the code.

vitaut (changes requested):

include/fmt/chrono.h
DanielaE: Makes sense. I will apply this suggestion in all of `fmt/include` where applicable.

DanielaE:
All headers included in the module have at most one single `export { ... }` group now which may have embedded not-exported `namespace detail { ... }` sections. Only those are affected by the newly introduced macros.

vitaut:
Thank you!

PR #2289, opened 2021-05-15
Silence msvc warning about an unused named parameter

DanielaE:
Warning C4100 may cause compile failures under strict warning regimes.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -2446,6 +2446,7 @@ constexpr int get_arg_index_by_name(basic_string_view<Char> name) {
   }
   if constexpr (sizeof...(Args) > 0)
     return get_arg_index_by_name<N + 1, Args...>(name);
+  (void)name;
   return invalid_arg_index;
 }
 #endif


include/fmt/core.h
```diff
@@ -2446,6 +2446,7 @@ constexpr int get_arg_index_by_name(basic_string_view<Char> name) {
   }
   if constexpr (sizeof...(Args) > 0)
     return get_arg_index_by_name<N + 1, Args...>(name);
+  (void)name;
```
vitaut, line 2449: Looks like a bug in MSVC because `name` is used in the line above. I recommend reporting to Microsoft.

DanielaE: I don't think so: if the control flow in an instantiation of 
```
template <int N, typename T, typename... Args, typename Char>
constexpr int get_arg_index_by_name(basic_string_view<Char> name)
```
reaches line 2449, then both `if constexpr` statements before were evaluated as false and their first (in this case: only) substatements become so called _discarded statements_ [stmt.if]/2 which leave no use of `name` in that template specialization.

alexezeder: Looks like it was also discussed on [Microsoft Developer Community](https://developercommunity2.visualstudio.com/t/constexpr-if-and-unused-parameter/276272), and resulted in joining the other compilers ([GCC](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81676) and Clang, at least) for these cases.

DanielaE: Thanks for both links!

Regarding the supposed GCC bug: the original poster gave a wrong description of the nature of the problem. He was missing the difference between a runtime `if` and a compile-time `if constexpr` as layed out in the standard. And "This is a valid program which should not produce a warning" misses the point of `-Wunused-but-set-parameter` explicitly requesting a warning.

This is exactly what the last poster on MS DevCon is rightfully complaining about.

DanielaE: I've referenced this PR on DevCon. According to my knowledge of the release schedule this will certainly not make it for 16.10 which will probably be around for a long time.

stevenhoving: I'm a bit unsure, but shouldn't uou use [[maybe_unused]] here? As far as I know it is what it is designed for.

DanielaE: Sure, whatever is preferred here. It's just the same as a few lines later.

vitaut: Thanks for investigating and I'm glad that this nonsense has already been reported. Merged this as a workaround for now.

vitaut: > "This is a valid program which should not produce a warning" misses the point of -Wunused-but-set-parameter explicitly requesting a warning.

I disagree, using a parameter in one branch of `if constexpr` is not conceptually different from using it in one branch of `if` (that may be optimized away). It's great that it was fixed in GCC.

DanielaE: The "misses" is referring to the blanket statement. A valid program may have many good reasons to issue warnings.

vitaut: Ah, that is true.

PR #2290, opened 2021-05-15
Fix Visual Studio warning

ilmai:
Visual Studio was complaining about unreachable code, presumably because vwindows_error was throwing the std::system_error instead of returning. Not 100% confident this is the correct fix, but quickly taking a look at how the function is used, this seems like the intended behavior.


vitaut:
Thank you

PR #2291, opened 2021-05-18
fix clang 8.0.x compilation error

sergiud:
With a clang 8.0.x based toolchain, the compiler insists that the `constexpr` value must be default initialized.

### DIFF
--- a/include/fmt/os.h
+++ b/include/fmt/os.h
@@ -391,7 +391,7 @@ struct ostream_params {
 };
 }  // namespace detail
 
-static constexpr detail::buffer_size buffer_size;
+static constexpr detail::buffer_size buffer_size{};
 
 /** A fast output stream which is not thread-safe. */
 class ostream final : private detail::buffer<char> {


include/fmt/os.h
```diff
@@ -391,7 +391,7 @@ struct ostream_params {
 };
 }  // namespace detail
 
-static constexpr detail::buffer_size buffer_size;
+static constexpr detail::buffer_size buffer_size{};
```
vitaut, line 394: Please add a short comment clarifying that {} is a workaround for gcc 8, preferably with a link to the bug report.

vitaut:
I wasn't able to repro the issue (https://godbolt.org/z/P99s1q6vf) so closing for now. Feel free to reopen with more details.

sergiud:
That's fine. It may be our specific combination of clang and musl responsible for this.

mathstuf:
I'm not seeing it with *clang* 8.x on Godbolt either, but it was observed with some older Xcode deployments at least. Godbolt doesn't provide those (and the Xcode toolchain versions don't match to any specific upstream clang releases either).

seanm:
You can see the error here for example: https://open.cdash.org/viewBuildError.php?buildid=7385287

```
/Users/builder/external/VTK/ThirdParty/fmt/vtkfmt/vtkfmt/os.h:397:31: error: default initialization of an object of const type 'const detail::buffer_size' without a user-provided default constructor
constexpr detail::buffer_size buffer_size;
                              ^
                                         {}
```

It reproduces with both macOS 10.11 + Xcode 8.2.1 (8C1002) and macOS 10.10 + Xcode 7.2.1 (7C1002).  Those are the newest Xcodes possible for those versions of macOS.

I've confirmed adding the `{}` fixes the issue. And there are no other compiler errors after that.

PR #2293, opened 2021-05-18
export missed symbol

sergiud:
When compiling with `-fvisibility=hidden`, linker errors occur because a symbol was not marked as being exported. The `final` keyword is not necessary for a class that is already marked as `final`.

How to reproduce:
```bash
CXXFLAGS=-fvisibility=hidden cmake -S . -B build/ -DBUILD_SHARED_LIBS:BOOL=ON -GNinja
cmake --build build/
```

[Log output](https://github.com/fmtlib/fmt/files/6504551/log.txt)

### DIFF
--- a/include/fmt/os.h
+++ b/include/fmt/os.h
@@ -394,7 +394,7 @@ struct ostream_params {
 static constexpr detail::buffer_size buffer_size;
 
 /** A fast output stream which is not thread-safe. */
-class ostream final : private detail::buffer<char> {
+class FMT_API ostream final : private detail::buffer<char> {
  private:
   file file_;
 
@@ -404,7 +404,7 @@ class ostream final : private detail::buffer<char> {
     clear();
   }
 
-  FMT_API void grow(size_t) override final;
+  void grow(size_t) override;
 
   ostream(cstring_view path, const detail::ostream_params& params)
       : file_(path, params.oflag) {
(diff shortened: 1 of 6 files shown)


include/fmt/os.h
```diff
@@ -394,7 +394,7 @@ struct ostream_params {
 static constexpr detail::buffer_size buffer_size;
 
 /** A fast output stream which is not thread-safe. */
-class ostream final : private detail::buffer<char> {
+class FMT_API ostream final : private detail::buffer<char> {
```
vitaut, line 397: Why move `FMT_API` to class? `grow` is the only non-inline member.

sergiud: The vtable also needs to be exported. Failing to do so, causes linker errors.

After enabling the tests, there seem to be more problems with `-fvisibility=hidden` in shared library builds. It may be good idea to enable the flag in CI builds.

sergiud:
@vitaut I've updated the error description and marked other missed symbols for export.

It might be a good idea to set the visibility to hidden by default because this leaves more room for compiler optimizations.

In CMake, the preferred way for generating export macros is to use the [GenerateExportHeader](https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html) module which creates a header with toolchain specific export defines. On a downside, this introduces an additional header. Is this something `fmt` would still consider?

vitaut:
> Is this something fmt would still consider?

No, we try to keep the number of headers small and especially avoid any generated headers.

vitaut:
There are some build failures.

sergiud:
I could reproduce the problem with `-flto` enabled (clang 11.0 and gcc 11.1).

vitaut:
Thank you!

Issue #2295, opened 2021-05-19
Warning: msvc 16.10-pre4 clashes with {fmt} 


JessyDL:
As a continuation of #2284, in `preview 4` more clashes are happening.

Unlike previously, you are not required to `#include <format>` as the latest STL has started to use its own implementation. Including `<chrono>` will now trigger the clash.

Example error:
```
fmt-src\include\fmt/core.h(2784,23): error C2668: 'fmt::v7::make_format_args': ambiguous call to overloaded function 
fmt-src\include\fmt/core.h(1625,46): message : could be 'fmt::v7::format_arg_store<fmt::v7::format_context,spdlog::filename_t,size_t,spdlog::filename_t> fmt::v7::make_format_args<fmt::v7::format_context,spdlog::filename_t,size_t,spdlog::filename_t>(const spdlog::filename_t &,const size_t &,const spdlog::filename_t &)' 
Microsoft Visual Studio\2019\Preview\VC\Tools\MSVC\14.29.30036\include\format(2848,17): message : or       'auto std::make_format_args<std::format_context,spdlog::filename_t,size_t,spdlog::filename_t>(const spdlog::filename_t &,const size_t &,const spdlog::filename_t &)' [found using argument-dependent lookup] 
fmt-src\include\fmt/core.h(2783,1): message : while trying to match the argument list '(spdlog::filename_t, size_t, spdlog::filename_t)' 
spdlog-src\include\spdlog/sinks/rotating_file_sink-inl.h(53): message : see reference to function template instantiation 'std::string fmt::v7::format<spdlog::filename_t&,size_t&,spdlog::filename_t&>(fmt::v7::basic_format_string<char,spdlog::filename_t &,size_t &,spdlog::filename_t &>,spdlog::filename_t &,size_t &,spdlog::filename_t &)' being compiled 
spdlog-src\include\spdlog/sinks/rotating_file_sink-inl.h(45): message : while compiling class template member function 'spdlog::filename_t spdlog::sinks::rotating_file_sink<std::mutex>::calc_filename(const spdlog::filename_t &,size_t)' 
spdlog-src\include\spdlog/sinks/rotating_file_sink-inl.h(33): message : see reference to function template instantiation 'spdlog::filename_t spdlog::sinks::rotating_file_sink<std::mutex>::calc_filename(const spdlog::filename_t &,size_t)' being compiled 
[...]

PR #2296, opened 2021-05-19
fix compile error on msvc preview 4 (16.10) involving lookup clash

JessyDL:
This PR resolves the issue mentioned in #2295 caused by a clash in the lookup between the `STL` shipped in vs2019 preview 4 (16.10) and `fmt`.

There's several other unqualified lookups to that function in particular, but as I didn't notice any causing issues I left them alone. If that changes I'll make a new PR to deal with them.


vitaut:
Thank you

PR #2297, opened 2021-05-19
Avoid unwanted sign extensions from MSVC in is_utf8.

mwinterb:
Microsoft's constexpr evaluator treats the type of micro[0] and micro[1] as
plain char, and so sign extends before comparing them to ints.
The normal compiler, including the optimizer, does not fail in this way,
so this is merely a "future proof" change in case someone uses is_utf8()
in a constant expression.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -391,11 +391,11 @@ FMT_CONSTEXPR typename std::make_unsigned<Int>::type to_unsigned(Int value) {
 FMT_MSC_WARNING(suppress : 4566) constexpr unsigned char micro[] = "\u00B5";
 
 constexpr bool is_utf8() {
-  // avoid buggy sign extensions in MSVC's constant evaluation mode
+  // Avoid buggy sign extensions in MSVC's constant evaluation mode.
   // https://developercommunity.visualstudio.com/t/C-difference-in-behavior-for-unsigned/1233612
   using uchar = unsigned char;
-  return FMT_UNICODE || (sizeof(micro) == 3 && uchar{micro[0]} == 0xC2 &&
-                         uchar{micro[1]} == 0xB5);
+  return FMT_UNICODE || (sizeof(micro) == 3 && uchar(micro[0]) == 0xC2 &&
+                         uchar(micro[1]) == 0xB5);
 }
 FMT_END_DETAIL_NAMESPACE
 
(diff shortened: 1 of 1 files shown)


mwinterb:
The smallest repro pointing out the issue is this: https://gcc.godbolt.org/z/xTxPfzqz9. `/J` is how Microsoft spells `-funsigned-char`. All compiles should return 1 from `main()` to indicate agreement between constant evaluation and "runtime" evaluation, but mostly default options do not do that.
I couldn't find any decent places to forcibly test this as none of the tests specify `/utf-8`, and it seems ODR-violating to just compile the tests with that and rude to add the option unconditionally when building the static library.
Another fix is just to make is_utf8 non-constexpr, but keeping the comment that it's explicitly non-constexpr to avoid Microsoft's bug.

include/fmt/core.h
```diff
@@ -391,8 +391,11 @@ FMT_CONSTEXPR typename std::make_unsigned<Int>::type to_unsigned(Int value) {
 FMT_MSC_WARNING(suppress : 4566) constexpr unsigned char micro[] = "\u00B5";
 
 constexpr bool is_utf8() {
-  return FMT_UNICODE ||
-         (sizeof(micro) == 3 && micro[0] == 0xC2 && micro[1] == 0xB5);
+  // avoid buggy sign extensions in MSVC's constant evaluation mode
+  // https://developercommunity.visualstudio.com/t/C-difference-in-behavior-for-unsigned/1233612
```
vitaut, line 395: nit: Please make it a proper sentence:
```
  // Avoid buggy sign extensions in MSVC's constant evaluation mode
  // https://developercommunity.visualstudio.com/t/C-difference-in-behavior-for-unsigned/1233612.
```

include/fmt/core.h
```diff
@@ -391,8 +391,11 @@ FMT_CONSTEXPR typename std::make_unsigned<Int>::type to_unsigned(Int value) {
...
+  using uchar = unsigned char;
+  return FMT_UNICODE || (sizeof(micro) == 3 && uchar{micro[0]} == 0xC2 &&
+                         uchar{micro[1]} == 0xB5);
```
vitaut, line 398: I suggest using a function-style cast (`uchar(...)`) for consistency with the rest of the library which doesn't use braced initialization for primitive types.

vitaut (changes requested):
Thanks for the PR. Looks good, just two nits.

vitaut:
Thanks!

PR #2300, opened 2021-05-21
Add missing Allocator template argument for basic_memory_buffer in fo…

voxmea:
…rmat_to

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -2699,10 +2699,10 @@ inline void vformat_to(
   return detail::vformat_to(buf, to_string_view(format_str), args);
 }
 
-template <typename S, typename... Args, size_t SIZE = inline_buffer_size,
-          typename Char = enable_if_t<detail::is_string<S>::value, char_t<S>>>
-inline auto format_to(basic_memory_buffer<Char, SIZE>& buf, const S& format_str,
-                      Args&&... args) ->
+template <typename S, typename... Args, typename Char, size_t SIZE,
+          typename Allocator, FMT_ENABLE_IF(detail::is_string<S>::value)>
+inline auto format_to(basic_memory_buffer<Char, SIZE, Allocator>& buf,
+                      const S& format_str, Args&&... args) ->
     typename buffer_context<Char>::iterator {
   const auto& vargs = fmt::make_args_checked<Args...>(format_str, args...);
   detail::vformat_to(buf, to_string_view(format_str), vargs);


include/fmt/format.h
```diff
@@ -2691,8 +2691,10 @@ inline void vformat_to(
 }
 
 template <typename S, typename... Args, size_t SIZE = inline_buffer_size,
-          typename Char = enable_if_t<detail::is_string<S>::value, char_t<S>>>
-inline auto format_to(basic_memory_buffer<Char, SIZE>& buf, const S& format_str,
+          typename Char = enable_if_t<detail::is_string<S>::value, char_t<S>>,
+          typename Allocator = std::allocator<Char>>
```
vitaut, line 2695: I think that `Allocator` shouldn't be defaulted because it's deduced from the argument. Otherwise looks good.

vitaut (changes requested):
Thanks for the PR.

include/fmt/format.h
voxmea: Good point, isn't that true for all of basic_memory_buffer's template arguments here?

voxmea: Never mind, the defaults means I can pass {} and get the defaulted values. I think that means I can remove the default SIZE then.

vitaut: It's true for `SIZE` (could you remove the default while at it?) but not for `Char` which is used for SFINAE.

voxmea: Oh, but it must be an lvalue. Sorry for the stream of thought here.

voxmea: Fixed this version so that SIZE and Allocator are deduced, and Char is still used for SFINAE.

voxmea: I'm not sure SFINAE actually works here since the template argument is being deduced: https://godbolt.org/z/WxqceTGef. Am I missing something?

alexezeder: SFINAE works here, as well as in yours example with small change - https://godbolt.org/z/M4brcnrTn.
In the original version the `U` template argument is deduced as `int`, so it's not defaulted, so SFINAE doesn't work there.

voxmea: I think you need to something like an extra unused template parameter:
typename = enable_if_t<detail::is_string<S>::value>>

alexezeder: Hmm, looks like I'm wrong, because your example is closer to the original code 🙂

voxmea: I think your case is different, it's more like an unused template argument and is different than defaulting a deducible parameter. I think here it needs to become an unused parameter:
typename = std::enable_if_t<std::is_string<S>::value>

voxmea: Pushed another version, this time with the SFINAE check to a non-deduced template parameter.

alexezeder: Then maybe `FMT_ENABLE_IF` can be used here as well as in:
https://github.com/fmtlib/fmt/blob/00149c0b6a0a0ec2ba9c7eefc342d13d21397219/include/fmt/format.h#L2694-L2695

alexezeder: Or return value can be used for SFINAE

vitaut: `FMT_ENABLE_IF` is preferable for consistency and occasionally smaller symbols sizes (maybe not in this case).

voxmea: Fixed. However, I don't know if the default argument for Char was intentional. I don't know how you can take advantage of a default argument there given that you must supply an lvalue basic_memory_buffer. How do you see the default argument to Char being used? Is there a way to skip specifying in the creation of the lvalue basic_memory_buffer?

voxmea: Oh, sorry, that was just an example of another spot where FMT_ENABLE_IF was being used.

voxmea: I pushed another version that removes the default template argument for vformat_to. Let me know if you want me to limit the change to just format_to.

vitaut: Please limit the changes to `format_to`.

alexezeder: > How do you see the default argument to Char being used?

I guess something like:
```cpp
format_to<std::basic_string<wchar_t>>(...)
```

alexezeder: But it's still deduced from `buf`, right? 🤔

voxmea: Updated to limit the change to `format_to`.

include/fmt/format.h
```diff
@@ -2699,9 +2699,10 @@ inline void vformat_to(
...
 }
 
-template <typename S, typename... Args, size_t SIZE = inline_buffer_size,
-          typename Char = enable_if_t<detail::is_string<S>::value, char_t<S>>>
-inline auto format_to(basic_memory_buffer<Char, SIZE>& buf, const S& format_str,
+template <typename S, typename... Args, typename Char, size_t SIZE,
+          typename Allocator, FMT_ENABLE_IF(detail::is_string<S>::value)>
+inline auto format_to(basic_memory_buffer<Char, SIZE, Allocator>& buf,
+                      const S& format_str,
                       Args&&... args) ->
```
alexezeder, line 2706: Does clang-format still break this onto the next line?

voxmea: Oops, thanks, this time run through clang-format.

vitaut:
Thank you

Issue #2299, opened 2021-05-20
Reduce library size


vitaut:
Need to revisit library size optimizations (https://www.zverovich.net/2020/05/21/reducing-library-size.html) since there was a bit of a regression compared to 7.0:

```
% clang++ -std=c++17  -I include -O2 -DNDEBUG test.cc src/format.cc
% strip a.out
% ls -l a.out
-rwxr-xr-x  1 viz  staff  193960 May 20 11:50 a.out
```

For comparison 7.0 generates
```
-rwxr-xr-x  1 viz  staff  127472 May 20 11:53 a.out
```
with the latest clang and libc++. Looks like some of the regression is due to newer clang and standard library.

vitaut:
7.1 is ~20k bigger probably because of adding Dragonbox.

vitaut:
Most of the regression happened in 7.1 so should be easy to track down.

vitaut:
The regression between 7.1.0 and 7.1.1 was to preserve ABI compatibility (https://github.com/fmtlib/fmt/commit/4081b2fe948aec01a49cc6e231bee4e02df3d61e) and is fixed now.

vitaut:
~30k regression was caused by https://github.com/fmtlib/fmt/commit/b268f8815d28659f65f73d0119c4152265266fd4.

sergiud:
Following #2293, I would suggest enabling `-fvisibility=hidden` and `-fvisibility-inlines-hidden` by default in `CMakeLists.txt`:
```cmake
set (CMAKE_CXX_VISIBILITY_PRESET hidden)
set (CMAKE_VISIBILITY_INLINES_HIDDEN ON)
```

```bash
$ du -h build-before/libfmt.so.7.1.4
244K
$ du -h build-after/libfmt.so.7.1.4
212K
```

I can provide a PR.

vitaut:
> I would suggest enabling -fvisibility=hidden and -fvisibility-inlines-hidden by default

Good idea but I would suggest using target properties instead of global CMake flags.

> I can provide a PR.

Please do.

vitaut:
With https://github.com/fmtlib/fmt/commit/b268f8815d28659f65f73d0119c4152265266fd4 regression fixed:

```
% clang++ -std=c++17  -I include -O2 -DNDEBUG test.cc src/format.cc && strip a.out && ls -l a.out
-rwxr-xr-x  1 viz  staff  161168 May 21 06:27 a.out
```

[...]

PR #2301, opened 2021-05-21
cmake: hide private symbols by default

sergiud:
Enable  `-fvisibility=hidden` and `-fvisibility-inlines-hidden` by default in `CMakeLists.txt`. Reduces the size of the shared library from 244K to 212K. Allow users to override the default choice.

Related to #2299.


vitaut:
Thank you!

PR #2303, opened 2021-05-21
Add more emphases

data-man:
Closes #2302.

### DIFF
--- a/include/fmt/color.h
+++ b/include/fmt/color.h
@@ -185,9 +185,17 @@ enum class terminal_color : uint8_t {
 
 enum class emphasis : uint8_t {
   bold = 1,
-  italic = 1 << 1,
-  underline = 1 << 2,
-  strikethrough = 1 << 3
+  faint = 1 << 1,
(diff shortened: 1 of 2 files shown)


include/fmt/color.h
```diff
@@ -399,7 +407,7 @@ template <typename Char> struct ansi_color_escape {
       return;
     }
 
-    for (int i = 0; i < 7; i++) {
+    for (int i = 0; i < 8; i++) {
```
foonathan, line 410: I don't understand the buffer logic completely, but you've increased the buffer from a `3 * 4` to a `3 * 8` yet change this loop index from 7 to 8? It might be a good idea to introduce some constants...

data-man: My bad, thanks!

data-man: I'll continue tomorrow. Need to sleep. :)

data-man:
Sorry, my family tested positive for COVID so I have other concerns now.

Issue #2280, opened 2021-05-12
FMT_COMPILE(s) does not work with chrono example


toughengineer:
`chrono` example from here  
https://fmt.dev/latest/syntax.html  
works on its own
<details><summary>(click to expand the initial example)</summary>

```c++
#include <fmt/chrono.h>

auto t = tm();
t.tm_year = 2010 - 1900;
t.tm_mon = 6;
t.tm_mday = 4;
t.tm_hour = 12;
t.tm_min = 15;
t.tm_sec = 58;
fmt::print("{:%Y-%m-%d %H:%M:%S}", t);
```
</details>

but fails to compile with latest master with all latest major compilers when used with `FMT_COMPILE(s)` as explained here  
https://fmt.dev/latest/api.html#compile-api  
like this:

```c++
fmt::print(FMT_COMPILE("{:%Y-%m-%dT%H:%M:%S}"), t);
```

E.g. see  
https://godbolt.org/z/TzzG3cvGe

vitaut:
Same as #1938.

alexezeder:
But it works with this code: https://godbolt.org/z/3KG8P4zzz
So the code from the first message doesn't work because `fmt::print` doesn't have `FMT_COMPILE`-friendly overload.

PR #2304, opened 2021-05-21
Add `fmt::print()` overload to support compiled format (`FMT_COMPILE`)

alexezeder:
Resolves https://github.com/fmtlib/fmt/issues/2280.

By adding a new exported function - `detail::print_buffer(std::FILE*, const memory_buffer&)`. This function is used in both `vprint()` from `format-inl.h` and a new `print()` from `compile.h`.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -2594,13 +2594,12 @@ extern "C" __declspec(dllimport) int __stdcall WriteConsoleW(  //
 }  // namespace detail
 #endif
 
-FMT_FUNC void vprint(std::FILE* f, string_view format_str, format_args args) {
-  memory_buffer buffer;
-  detail::vformat_to(buffer, format_str, args);
+namespace detail {
+FMT_FUNC void print(std::FILE* f, string_view text) {
 #ifdef _WIN32
   auto fd = _fileno(f);
   if (_isatty(fd)) {
-    detail::utf8_to_utf16 u16(string_view(buffer.data(), buffer.size()));
+    detail::utf8_to_utf16 u16(string_view(text.data(), text.size()));
     auto written = detail::dword();
     if (detail::WriteConsoleW(reinterpret_cast<void*>(_get_osfhandle(fd)),
                               u16.c_str(), static_cast<uint32_t>(u16.size()),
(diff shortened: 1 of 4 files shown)


vitaut:
Thanks for the PR. In general looks good but I wonder if it's worth doing considering that format string processing time is likely small compared to I/O. Could you provide some benchmarks?

alexezeder:
Okay, I have [the results](https://github.com/alexezeder/fmt_bnchmrk/tree/print), and there are some improvements. It worth noting though that these results may be too specific for my PC.

Using static library:

| Benchmark             |      Time    |
| ----------------------|--------------|
| JustWrite42           |   12.2 ns    |
| PrintTrivial          |   25.2 ns    |
| PrintTrivialCompile   |   20.3 ns    |
| PrintComplex          |    216 ns    |
| PrintComplexCompile   |    135 ns    |

Using header-only library:

| Benchmark             |      Time    |
| ----------------------|--------------|
| JustWrite42           |   12.1 ns    |
| PrintTrivial          |   25.2 ns    |
| PrintTrivialCompile   |   16.6 ns    |
| PrintComplex          |    216 ns    |
| PrintComplexCompile   |    141 ns    |

Here a short notes about each benchmark:

* JustWrite42 - just write `"42"` to file using `std::fwrite()` (just to measure base I/O speed)
* PrintTrivial - use `fmt::print("{}", 42)` to write `42`
* PrintTrivialCompile - same as PrintTrivial, but using compile-time API
* PrintComplex - same as trivial, but with `"{:08x} {:0.2f} {:>7}"` as format string
* PrintComplexCompile - same as PrintComplex, but using compile-time API

include/fmt/format-inl.h
```diff
@@ -2594,9 +2594,8 @@ extern "C" __declspec(dllimport) int __stdcall WriteConsoleW(  //
 }  // namespace detail
 #endif
 
-FMT_FUNC void vprint(std::FILE* f, string_view format_str, format_args args) {
-  memory_buffer buffer;
-  detail::vformat_to(buffer, format_str, args);
+namespace detail {
+FMT_FUNC void print_buffer(std::FILE* f, const memory_buffer& buffer) {
```
vitaut, line 2598: I think it would be slightly better to pass the text as a `string_view` instead of a buffer. This also means that the name would have to change. Maybe `detail::print`?

vitaut:
> Okay, I have the results, and there are some improvements. 

Thanks for the benchmarks, they look convincing.

include/fmt/format-inl.h
alexezeder: Done, the most challenging part of this change was to name the new argument correctly 😄, but since you mentioned that it's the text, I used `text` as a name.

PR #2309, opened 2021-05-23
More module tests

DanielaE:
Most of the API of core.h, format.h, and args.h

Just narrow strings. Tests for wide strings will come later.

### DIFF
--- a/test/module-test.cc
+++ b/test/module-test.cc
@@ -9,6 +9,27 @@
 // All Rights Reserved
 // {fmt} module.
 
+#include <bit>
+#include <chrono>
+#include <exception>
+#include <iterator>
+#include <locale>
+#include <memory>
+#include <ostream>
+#include <string>
+#include <string_view>
+#include <system_error>
+
+#if (__has_include(<fcntl.h>) || defined(__APPLE__) || \
+     defined(__linux__)) &&                              \
+    (!defined(WINAPI_FAMILY) || (WINAPI_FAMILY == WINAPI_FAMILY_DESKTOP_APP))
+#  include <fcntl.h>
+#  define FMT_USE_FCNTL 1
+#else
+#  define FMT_USE_FCNTL 0
+#endif
+#define FMT_NOEXCEPT noexcept
+
 import fmt;
 
 // check for macros leaking from BMI
@@ -19,11 +40,12 @@ static bool macro_leaked =
   false;
 #endif
 
-#include "gtest/gtest.h"
+#include "gtest-extra.h"
 
 // an implicitly exported namespace must be visible [module.interface]/2.2
 TEST(module_test, namespace) {
   using namespace fmt;
+  using namespace fmt::literals;
   ASSERT_TRUE(true);
 }
 
@@ -34,8 +56,8 @@ bool oops_detail_namespace_is_visible;
 namespace fmt {
 bool namespace_detail_invisible() {
 #if defined(FMT_HIDE_MODULE_BUGS) && \
-  defined(_MSC_FULL_VER) && _MSC_FULL_VER <= 192930036
-  // bug in msvc 16.10-pre4:
+  defined(_MSC_FULL_VER) && _MSC_FULL_VER <= 192930129
+  // bug in msvc up to 16.11-pre1:
   // the namespace is visible even when it is neither
   // implicitly nor explicitly exported
   return true;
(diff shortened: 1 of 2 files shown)


vitaut (changes requested):
Thanks for the PR. Having a module test seems like a good idea but I am concerned about duplication. I think we don't need to replicate so many test cases here, it's enough to exercise every user-facing public function like `format` or `format_to` once. In particular, we shouldn't be testing multiple built-in argument types because they all go through the same path.

Also there are some json files that I don't think should be here.

DanielaE:
Right, these two additional files sneaked in by accident. Thanks for pointing this out.

Regarding duplication: it's not just the name like e.g. `format` but **all** of the overloads, and as these are templates which may cause different specializations within the library, as many of them should be instantiated to check not only _visibility_ but also _reachability_. This aspect is a non-issue in traditional libraries, because everything visible is also reachable, and there are possibly more declarations and semantic properties reachable than visible. In the modules world things are much different: visibility and name-lookup in instantiations must be put to a test. I've already found a few places by these tests where lookup is an issue that need to be audited and addressed.

If you happen to have sections to point me to which concern you I will look into them and remove superfluous duplicates. I'd love to have a tool to aide me but I'm not aware of one.

test/module-test.cc
```diff
@@ -63,3 +73,507 @@ TEST(module_test, macros) {
...
+  EXPECT_EQ("-INF-", fmt::format("{:-^5F}", inf));
+  EXPECT_EQ("-INF ", fmt::format("{: 5G}", -inf));
+  EXPECT_EQ("42", fmt::format("{:s}", "42"));
+  EXPECT_EQ("a", fmt::format("{:c}", 'a'));
+  EXPECT_EQ("a", fmt::format("{:c}", unsigned char{'a'}));
+  EXPECT_EQ("a", fmt::format("{:c}", signed char{'a'}));
+  EXPECT_EQ("false", fmt::format("{}", false));
+  EXPECT_EQ("0B1", fmt::format("{:#B}", true));
+  EXPECT_EQ("0x2a", fmt::format("{:p}", (void *)42));
+  EXPECT_EQ("1234", fmt::format("{:L}", 1234));
```
vitaut, line 129: For example here we don't need to test all argument types. Since they all go through the same code path, it is very unlikely that one of them works and another doesn't. I suggest picking one type (e.g. int) and testing public methods with it. In the unlikely case that we find some problem in the future, we can add a regression test.

test/module-test.cc
```diff
@@ -63,3 +73,507 @@ TEST(module_test, macros) {
...
+    EXPECT_EQ(5u, result.size);
+    EXPECT_EQ(buffer + 3, result.out);
+    EXPECT_EQ("123", std::string_view(buffer));
+  }
+}
+
+template <typename... Args>
+auto create_args(Args&&... args) {
+  return fmt::make_format_args<fmt::format_context>(args...);
+}
```
vitaut, line 165: I suggest using `make_format_args` directly instead of introducing a wrapper function.

DanielaE:
Your suggestions are in. This is also updated to msvc 16.11-pre1

test/module-test.cc
```diff
@@ -55,11 +76,551 @@ TEST(module_test, detail_namespace) {
...
+  char buffer[4] = {0};
+  fmt::format_to_n(buffer, 3, "{}", 12345);
+}
+
+TEST(module_test, format_args) {
+  auto no_args = fmt::format_args();
+  EXPECT_FALSE(no_args.get(1));
+
+  using ctx = fmt::format_context;
+  fmt::basic_format_args<ctx> args = fmt::make_format_args<ctx>(42);
```
vitaut, line 136: `ctx` can be dropped here because it is the default. Same in multiple places below.

test/module-test.cc
```diff
@@ -55,11 +76,551 @@ TEST(module_test, detail_namespace) {
...
+  EXPECT_TRUE(fmt::is_char<char>());
+  EXPECT_TRUE(fmt::is_char<wchar_t>());
+  EXPECT_TRUE(fmt::is_char<char8_t>());
+  EXPECT_TRUE(fmt::is_char<char16_t>());
+  EXPECT_TRUE(fmt::is_char<char32_t>());
+  EXPECT_FALSE(fmt::is_char<signed char>());
+}
+
+TEST(module_test, ptr) {
+  auto p = (int*)42;
```
vitaut, line 274: Please use more narrow bit_cast or reinterpret_cast instead of a C-style cast.

test/module-test.cc
```diff
@@ -55,11 +76,551 @@ TEST(module_test, detail_namespace) {
...
+TEST(module_test, time_duration) {
+  EXPECT_EQ("42s", fmt::format("{}", std::chrono::seconds{42}));
+  using us = std::chrono::duration<double, std::micro>;
+  EXPECT_EQ("4.2µs", fmt::format("{:3.1}", us{4.234}));
+  const auto de = std::locale("de_DE");
+  EXPECT_EQ("4.2µs", fmt::format(de, "{:L}", us{4.2}));
+}
+
+TEST(module_test, weekday) {
+  const auto de = std::locale("de_DE");
```
vitaut, line 371: This may throw if the locale is not available.

test/module-test.cc
```diff
@@ -55,11 +76,551 @@ TEST(module_test, detail_namespace) {
...
+  EXPECT_EQ("fmt", nsv.c_str());
+  fmt::wcstring_view wsv(L"fmt");
+  EXPECT_EQ(L"fmt", wsv.c_str());
+}
+
+TEST(module_test, buffered_file) {
+  EXPECT_TRUE(fmt::buffered_file{}.get() == nullptr);
+}
+
+TEST(module_test, move) {
```
vitaut, line 434: This test should probably be called `output_file`, not `move`.

test/module-test.cc
```diff
@@ -55,11 +76,551 @@ TEST(module_test, detail_namespace) {
...
+                "");
+}
+
+TEST(core_test, is_formattable) {
+  static_assert(fmt::is_formattable<enabled_formatter>::value, "");
+  static_assert(!fmt::is_formattable<disabled_formatter>::value, "");
+  static_assert(fmt::is_formattable<disabled_formatter_convertible>::value, "");
+}
+
+struct convertible_to_int {
```
vitaut, line 517: Everything starting from here to the end of the file falls into the category of "functional testing", please remove. It is very unlikely that this functionality behaves differently between modular and non-modular build because it goes through the same argument handling paths as everything else.

vitaut (changes requested):
> Your suggestions are in. This is also updated to msvc 16.11-pre1

Thank you. A few more comments, otherwise looks great.

test/module-test.cc
DanielaE: Yeah ...
Luckily we don't care here.

test/module-test.cc
DanielaE: Copy & waste strikes again.

test/module-test.cc
DanielaE: I've condensed it as much as possible to name-lookup.

test/module-test.cc
vitaut: Why don't we care? Since this is a non-functional test I suggest using the `std::locale::classic()` which is always available.

DanielaE: We don't care for the same reason why you suggest using the classic locale: it just has to look up the correct overloads. If you prefer `std::locale::classic()` then I'll change it accordingly.

vitaut: > it just has to look up the correct overloads.

It would enough if we didn't actually run this code, only compile. But my understanding is that we eventually want to run this code.

> If you prefer std::locale::classic() then I'll change it accordingly.

Please do.

vitaut:
Thank you!

PR #2312, opened 2021-05-24
Update integer presentation types documentation.

mordante:
Documents that the 'c' type is a valid type for integer types. Since
boolean uses the same types as integer its documentation is
automatically updated.


vitaut:
Thank you!

PR #2314, opened 2021-05-25
Fixes warning about uint type

macedot:
uint is already a type defined by other libraries (ie, opencv)


vitaut:
Thanks for the PR but this is a local alias. We don't guarantee global uniqueness of all identifiers.

PR #2315, opened 2021-05-26
qualify make_format_args

jmelas:
Latest MSVC 16.10.0 introduced a function std::make_format_args which clashes with fmt::v7::make_format_args.
This means that all calls to make_format_args need to by qualified to avoid ADL clash


vitaut:
Thank you

PR #2326, opened 2021-05-30
Fix `FMT_COMPILE()` with custom types

alexezeder:
By reading the changelog for upcoming release, I realized that it does not mention one of the major breaking change of custom formatters for compile-time formatting - `const`-qualified `format` method, so I decided to check how do custom formatters work at compile-time and realized that they actually don't work.

There are 3 commits, by the complexity of changes from the commit that just fixes the library, to the commit that adds dynamic width handling for the `test_formattable` class in compile-test.cc. The latter two can be removed or squashed into the first one, on demand. This time I cannot [use custom formatters](https://github.com/fmtlib/fmt/pull/2129#discussion_r575826176) from `chrono.h` because they use `memory_buffer` and it's not ready to be used at compile-time yet.

### DIFF
--- a/test/compile-test.cc
+++ b/test/compile-test.cc
@@ -53,17 +53,6 @@ TEST(iterator_test, truncating_back_inserter) {
   EXPECT_EQ(buffer, "42");
 }
 
-struct test_formattable {};
-
-FMT_BEGIN_NAMESPACE
-template <> struct formatter<test_formattable> : formatter<const char*> {
-  template <typename FormatContext>
-  auto format(test_formattable, FormatContext& ctx) -> decltype(ctx.out()) {
-    return formatter<const char*>::format("foo", ctx);
-  }
-};
-FMT_END_NAMESPACE
-
 TEST(compile_test, compile_fallback) {
   // FMT_COMPILE should fallback on runtime formatting when `if constexpr` is
   // not available.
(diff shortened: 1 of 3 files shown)


test/compile-test.cc
```diff
@@ -53,24 +53,76 @@ TEST(iterator_test, truncating_back_inserter) {
...
+      return detail::arg_ref<char>(arg_id);
+    }
+    FMT_CONSTEXPR detail::arg_ref<char> make_arg_ref(detail::auto_id) {
+      return detail::arg_ref<char>(context.next_arg_id());
+    }
+    template <typename Id> FMT_CONSTEXPR void on_dynamic_width(Id arg_id) {
+      f.width_ref = make_arg_ref(arg_id);
+    }
+    template <typename Id> FMT_CONSTEXPR void on_dynamic_precision(Id) {}
+  };
```
vitaut, line 100: This is a bit of an overkill for the test. Let's do some basic parsing instead of relying on internal API.

vitaut (changes requested):
Thanks for the PR.

vitaut:
> it does not mention one of the major breaking change of custom formatters for compile-time formatting - const-qualified format method

Good catch, added to changelog in https://github.com/fmtlib/fmt/commit/9bb406d78d209641de8deb6429c168ac9f05b461.

test/compile-test.cc
alexezeder: So, commits 1+2 then, dropping the 3

vitaut: Yeah, could you update the PR?

alexezeder: Of course, I will do so, but maybe an hour later

alexezeder: Done

vitaut:
Thank you!

PR #2332, opened 2021-05-31
Re-enable module testing

DanielaE:
Prepare for compilation with gcc (modules branch).

### DIFF
--- a/test/module-test.cc
+++ b/test/module-test.cc
@@ -9,6 +9,12 @@
 // All Rights Reserved
 // {fmt} module.
 
+#ifdef _MSC_FULL_VER
+// hide some implementation bugs in msvc
+// that are not essential to users of the module.
+#  define FMT_HIDE_MODULE_BUGS
+#endif
+
 #include <bit>
 #include <chrono>
 #include <exception>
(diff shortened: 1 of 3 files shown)


DanielaE:
This depends on #2324

DanielaE:
I've force-pushed my local work-in-progress branch with a lot of additional tests to see if module testing works. It does.

vitaut:
CI is still failing.

DanielaE:
Interesting. Never seen this. Is this a GTest feature to compare C-strings by _identity_ rather than by _equality_?

test/module-test.cc
```diff
@@ -405,9 +411,9 @@ TEST(module_test, color) {
 
 TEST(module_test, cstring_view) {
   fmt::cstring_view nsv("fmt");
-  EXPECT_EQ("fmt", nsv.c_str());
+  EXPECT_EQ("fmt", std::string_view(nsv.c_str()));
```
vitaut, line 414: I suggest using `EXPECT_STREQ` instead.

vitaut: Actually no, this checks the pointer comparison so the correct fix is to move "fmt" into a separate variable.

DanielaE: My new code cannot fall into the identity trap anymore.

vitaut: Not sure what you mean by identity trap but this must compare pointers, not strings, something like:
```c++
auto s = "fmt";
EXPECT_EQ(s, fmt::cstring_view(s).c_str());
```

vitaut:
Thank you!

PR #2333, opened 2021-05-31
Fix "undefined reference to `fmt::v7::detail::basic_data<void>::digits'"

phprus:
If static fmt library is used in shared library.
Error in commit 13e652939b222249073281e507848e3db9fa3e65.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -253,9 +253,10 @@
 #  endif
 #else
 #  define FMT_CLASS_API
-#  if defined(__GNUC__) || defined(__clang__)
-#    define FMT_API __attribute__((visibility("default")))
-#    define FMT_EXTERN_TEMPLATE_API FMT_API
+#  if defined(FMT_EXPORT) || defined(FMT_SHARED)
+#    if defined(__GNUC__) || defined(__clang__)
+#      define FMT_API __attribute__((visibility("default")))
+#    endif
 #  endif
 #endif
 #ifndef FMT_API
--- a/test/static-export-test/library.cc
+++ b/test/static-export-test/library.cc
@@ -1,0 +1,5 @@
+#include <fmt/compile.h>
+
+__attribute__((visibility("default"))) std::string foo() {
+  return fmt::format(FMT_COMPILE("foo bar {}"), 4242);
+}
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -903,7 +903,7 @@ template <typename T = void> struct basic_data {
   FMT_API static constexpr const char right_padding_shifts[] = {0, 31, 0, 1, 0};
 };
 
-#ifndef FMT_HEADER_ONLY
+#ifdef FMT_SHARED
 // Required for -flto, -fivisibility=hidden and -shared to work
 extern template struct basic_data<void>;
 #endif
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -213,6 +213,21 @@ if (FMT_PEDANTIC AND NOT WIN32)
     "-DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}")
 endif ()
 
+# This test are disabled on Windows because it is only *NIX issue.
+if (FMT_PEDANTIC AND NOT WIN32)
+  add_test(static-export-test ${CMAKE_CTEST_COMMAND}
+    -C ${CMAKE_BUILD_TYPE}
+    --build-and-test
+    "${CMAKE_CURRENT_SOURCE_DIR}/static-export-test"
+    "${CMAKE_CURRENT_BINARY_DIR}/static-export-test"
+    --build-generator ${CMAKE_GENERATOR}
+    --build-makeprogram ${CMAKE_MAKE_PROGRAM}
+    --build-options
+    "-DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}"
+    "-DCMAKE_CXX_STANDARD=${CMAKE_CXX_STANDARD}"
+    "-DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}")
+endif ()
+
 # Activate optional CUDA tests if CUDA is found. For version selection see
 # https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#cpp14-language-features
 if (FMT_CUDA_TEST)
--- a/test/static-export-test/CMakeLists.txt
+++ b/test/static-export-test/CMakeLists.txt
@@ -1,0 +1,30 @@
+cmake_minimum_required(VERSION 3.1...3.18)
+
+project(fmt-link CXX)
+
+set(BUILD_SHARED_LIBS OFF)
+set(CMAKE_VISIBILITY_INLINES_HIDDEN TRUE)
+set(CMAKE_CXX_VISIBILITY_PRESET "hidden")
+
+# Broken LTO on GCC 4
+if (CMAKE_COMPILER_IS_GNUCXX AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 5)
+  set(BROKEN_LTO ON)
+endif ()
+
+if (NOT BROKEN_LTO AND CMAKE_VERSION VERSION_GREATER "3.8")
+  # CMake 3.9+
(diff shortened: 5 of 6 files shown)


vitaut:
I wonder why it doesn't happen in CI. Could you provide repro details?

phprus:
Pseudocmake:

```
add_library(mylibrary SHARED lib.cpp)
target_link_libraries(mylibrary PRIVATE libfmt.a) # link with static fmt

add_executable(exe main.cpp)
target_link_libraries(exe PRIVATE mylibrary) # Error: undefined reference to `fmt::v7::detail::basic_data<void>::digits'
```

Because ```extern template struct basic_data<void>;``` and ```template struct detail::basic_data<void>;``` with visibility attribute is add reference to ```fmt::v7::detail::basic_data<void>::digits``` into ```mylibrary.so```.

If link ```exe``` with ```libfmt.a```, there will be no error.

src/format.cc
```diff
@@ -70,6 +70,8 @@ template FMT_API wchar_t detail::decimal_point_impl(locale_ref);
 template FMT_API void detail::buffer<wchar_t>::append(const wchar_t*,
                                                       const wchar_t*);
 
+#ifdef FMT_EXPORT
```
sergiud, line 73: Why is this guard necessary here but not for the explicit instantiations above? This seems to be inconsistent.

sergiud:
Adding the combination to the test suite would certainly help catch problems in the future.

src/format.cc
sergiud: Also, doesn't a compilation unit always export regardless of library type? As such, a static library would instantiate the struct as well. The only case that does not require the explicit instantiation is the header only variant. But then, the instantiations in the compilation unit are irrelevant anyway.

sergiud:
I believe, you are also linking to fmtlib incorrectly because the error occurs not in `mylibrary` but according to your description first in the `exe` target.

Note that libfmt is not linked transitively to `exe` because you specify the `PRIVATE` modifier:
```cmake
add_library(mylibrary SHARED lib.cpp)
target_link_libraries(mylibrary PRIVATE libfmt.a) # link with static fmt
```
while you seem to be using libfmt in your `exe` target as well:
```cmake
add_executable(exe main.cpp)
target_link_libraries(exe PRIVATE mylibrary) # Error: undefined reference to `fmt::v7::detail::basic_data<void>::digits'
```
I think, a clear (non-pseudo) example is necessary to understand the scenario.

phprus:
@sergiud, No.
Now, mylibrary has are reference to ```fmt::v7::detail::basic_data<void>::digits``` always. Any use of mylibrary will result in an error.

The exe target does not use fmt functions anywhere and cannot depend on it.

src/format.cc
phprus: You're right. This guard is unnecessary.

Problem only in a ```__attribute__((visibility("default")))``` and ```extern template``` declaration on static library.

sergiud:
I cannot reproduce the problem at ad97258915e5132a5d3ba98c8e70ea0f7e92fc53. Here's my MWE:

`CMakeLists.txt`
```cmake
cmake_minimum_required (VERSION 3.0)
project (fmt-link)

set (CMAKE_CXX_VISIBILITY_PRESET hidden)

find_package (fmt REQUIRED)

add_library (mylib SHARED mylib.cpp)
target_link_libraries (mylib PRIVATE fmt::fmt)

add_executable (myexe myexe.cpp)
target_link_libraries (myexe PRIVATE mylib)
```

`mylib.cpp`:
```cpp
#include <fmt/format.h>

__attribute__((visibility("default")))
std::string foo()
{
    return fmt::format("foo bar {}", 42);
}
```

`myexe.cpp`:
```cpp
#include <string>
#include <iostream>

extern std::string foo();

int main()
{
    std::cout << foo() << std::endl;
}
```

phprus:
```CMakeLists.txt```:
```
cmake_minimum_required (VERSION 3.9)
project (fmt-link)

set(CMAKE_CXX_STANDARD                  17)
set(CMAKE_VISIBILITY_INLINES_HIDDEN     TRUE)
set(CMAKE_CXX_VISIBILITY_PRESET         "hidden")

include(CheckIPOSupported)
check_ipo_supported()
set(CMAKE_INTERPROCEDURAL_OPTIMIZATION  TRUE)

set(BUILD_SHARED_LIBS           OFF)
add_subdirectory(fmt-master)
#set_property(TARGET fmt PROPERTY POSITION_INDEPENDENT_CODE ON)

add_library (mylib SHARED mylib.cpp)
target_link_libraries (mylib PRIVATE fmt::fmt)

add_executable (myexe myexe.cpp)
target_link_libraries (myexe PRIVATE mylib)
```

```mylib.cpp```:
```
#include <fmt/compile.h>

__attribute__((visibility("default")))
std::string foo()
{
    return fmt::format(FMT_COMPILE("foo bar {}"), 4242);
}
```

Build:
```
cmake -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_BUILD_TYPE=Release . && make
```

Error:
```
/usr/lib64/gcc/x86_64-suse-linux/7/../../../../x86_64-suse-linux/bin/ld: libmylib.so: неопределённая ссылка на «fmt::v7::detail::basic_data<void>::digits»
```

**Comment by phprus** [comment]
openSUSE Leap 15.2
gcc version 7.5.0 (SUSE Linux)

**Comment by sergiud** [comment]
The culprit seems to be the combination `libfmt.a` + `mylib.so` + `-flto`. This sole modification to fix the issue that works for me is
```patch
diff --git a/include/fmt/format.h b/include/fmt/format.h
index 3f6ab357..182cf9da 100644
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -902,9 +902,9 @@ template <typename T = void> struct basic_data {
   FMT_API static constexpr const char right_padding_shifts[] = {0, 31, 0, 1, 0};
 };
 
-#ifndef FMT_HEADER_ONLY
+#if !defined(FMT_HEADER_ONLY) && defined(FMT_EXPORT)
 // Required for -flto, -fivisibility=hidden and -shared to work
-extern template struct basic_data<void>;
+extern template struct FMT_API basic_data<void>;
 #endif
 
 // This is a struct rather than an alias to avoid shadowing warnings in gcc.
```

@phprus Can you verify this? `FMT_API` does not need to be changed.

phprus:
```!defined(FMT_HEADER_ONLY) &&``` - this condition is not needed, because ```FMT_EXPORT``` is defined only on shared fmt.

```FMT_EXPORT``` - is are private cmake definition
```FMT_SHARED``` - is are interface cmake definition

In this case condition ```defined (FMT_EXPORT)``` is not needed and has been removed.

File: include/fmt/core.h [hunk 2]
```diff
@@ -253,9 +253,10 @@
 #  endif
 #else
 #  define FMT_CLASS_API
-#  if defined(__GNUC__) || defined(__clang__)
-#    define FMT_API __attribute__((visibility("default")))
-#    define FMT_EXTERN_TEMPLATE_API FMT_API
```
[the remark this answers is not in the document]
sergiud: Right, "partial incorrect commit" :-1: 

`FMT_EXTERN_TEMPLATE_API` was used before. But you're correct to note that it is not necessary anymore.

sergiud:
Could you also please add the combination to the test suite? This case is currently not being tested.

phprus:
Test added.

sergiud:
Great, thanks! LGTM

phprus:
Don't merge now.
I am making a change to the test.

alexezeder:
Note that you can transform this PR to Draft until you are ready.

phprus:
Changes applied.
@vitaut review my PR please.

test/static-export-test/library.cc
```diff
@@ -0,0 +1,7 @@
+#include <fmt/compile.h>
+
+__attribute__((visibility("default")))
+std::string foo()
+{
+    return fmt::format(FMT_COMPILE("foo bar {}"), 4242);
+}
```
vitaut, line 7: Please apply clang-format here and below.

phprus: Oh! I completely forgot about clang-format.

vitaut:
Merged, thank you both!

PR #2338, opened 2021-06-03
Fixed join_view formatter for wchar_t

crbrz:
Fixed compile error when trying to format a range with a value type formattable only to wide strings.


vitaut:
Thank you

PR #2340, opened 2021-06-04
Feature/complete module

DanielaE:
Add all missing pieces to complete {fmt} as module:

- support non-`char` overloads
- support compile-time argument type checking
- support compile-time format string compilation

This includes the macro-based API `FMT_STRING` and `FMT_COMPILE` by means of a module companion header `fmt.h`

### DIFF
--- a/include/fmt/compile.h
+++ b/include/fmt/compile.h
@@ -164,7 +164,8 @@ struct is_compiled_string : std::is_base_of<compiled_string, S> {};
 #endif
 
 #if FMT_USE_NONTYPE_TEMPLATE_PARAMETERS
-template <typename Char, size_t N, fixed_string<Char, N> Str>
+template <typename Char, size_t N,
+          fmt::detail_exported::fixed_string<Char, N> Str>
 struct udl_compiled_string : compiled_string {
   using char_type = Char;
   constexpr operator basic_string_view<char_type>() const {
@@ -622,7 +623,7 @@ void print(const S& format_str, const Args&... args) {
 
 #if FMT_USE_NONTYPE_TEMPLATE_PARAMETERS
 inline namespace literals {
-template <detail::fixed_string Str>
+template <detail_exported::fixed_string Str>
 constexpr detail::udl_compiled_string<
     remove_cvref_t<decltype(Str.data[0])>,
     sizeof(Str.data) / sizeof(decltype(Str.data[0])), Str>
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -2624,6 +2624,20 @@ template <typename Char>
 void vformat_to(buffer<Char>& buf, basic_string_view<Char> fmt,
                 basic_format_args<buffer_context<type_identity_t<Char>>> args,
                 locale_ref loc) {
+  // workaround for msvc bug regarding name-lookup in module
+  // link names into function scope
+  using detail::arg_formatter;
+  using detail::buffer_appender;
+  using detail::custom_formatter;
+  using detail::default_arg_formatter;
+  using detail::get_arg;
+  using detail::locale_ref;
+  using detail::parse_format_specs;
+  using detail::specs_checker;
+  using detail::specs_handler;
+  using detail::to_unsigned;
+  using detail::type;
+  using detail::write;
   auto out = buffer_appender<Char>(buf);
   if (fmt.size() == 2 && equal2(fmt.data(), "{}")) {
     auto arg = args.get(0);
@@ -2680,13 +2694,12 @@ void vformat_to(buffer<Char>& buf, basic_string_view<Char> fmt,
       begin = parse_format_specs(begin, end, handler);
       if (begin == end || *begin != '}')
         on_error("missing '}' in format string");
-      auto f =
-          detail::arg_formatter<Char>{context.out(), specs, context.locale()};
+      auto f = arg_formatter<Char>{context.out(), specs, context.locale()};
       context.advance_to(visit_format_arg(f, arg));
       return begin;
     }
   };
-  parse_format_string<false>(fmt, format_handler(out, fmt, args, loc));
+  detail::parse_format_string<false>(fmt, format_handler(out, fmt, args, loc));
 }
 
 #ifndef FMT_HEADER_ONLY
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -778,6 +778,33 @@ FMT_INLINE auto make_args_checked(const S& fmt,
   return {args...};
 }
 
+// compile-time support
+namespace detail_exported {
+#if FMT_USE_NONTYPE_TEMPLATE_PARAMETERS
+template <typename Char, size_t N> struct fixed_string {
+  constexpr fixed_string(const Char (&str)[N]) {
+    detail::copy_str<Char, const Char*, Char*>(static_cast<const Char*>(str),
+                                               str + N, data);
+  }
+  Char data[N]{};
+};
+#endif
+
+// Converts a compile-time string to basic_string_view.
+template <typename Char, size_t N>
+constexpr auto compile_string_to_view(const Char (&s)[N])
+    -> basic_string_view<Char> {
+  // Remove trailing NUL character if needed. Won't be present if this is used
+  // with a raw character array (i.e. not defined as a string).
+  return {s, N - (std::char_traits<Char>::to_int_type(s[N - 1]) == 0 ? 1 : 0)};
+}
+template <typename Char>
+constexpr auto compile_string_to_view(detail::std_string_view<Char> s)
+    -> basic_string_view<Char> {
+  return {s.data(), s.size()};
+}
+}  // namespace detail_exported
+
 FMT_BEGIN_DETAIL_NAMESPACE
 
 inline void throw_format_error(const char* message) {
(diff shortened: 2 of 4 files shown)


include/fmt/format.h
```diff
@@ -2673,6 +2673,18 @@ template <typename Char>
...
+  using detail::custom_formatter;
+  using detail::default_arg_formatter;
+  using detail::get_arg;
+  using detail::locale_ref;
+  using detail::parse_format_specs;
+  using detail::specs_checker;
+  using detail::specs_handler;
+  using detail::to_unsigned;
+  using detail::type;
+  using detail::write;
```
DanielaE, line 2687: I'm still not sure, if this is really needed by the standard or if this is just an implementation deficit. At least it looks ugly. 

From reading "C++ Templates - The Definitive Guide" and the standard, it is clear that the local class is a _templated entity_. I need further studies of the standard to find an answer to the question: which names within this so-called 'temploid' are bound at which time.

I'd rather prefer to make this 'temploid' a proper class template.

vitaut (changes requested):
Thanks for the PR.

In general looks good but i don't think we should be introducing a new header for macros because `FMT_STRING` and `FMT_COMPILE` are effectively obsolete with C++20. The former is redundant and the latter is replaced with a UDL.

include/fmt/compile.h
```diff
@@ -11,7 +11,15 @@
 #include "format.h"
 
 FMT_BEGIN_NAMESPACE
-namespace detail {
+FMT_MODULE_EXPORT_BEGIN
+
+// compile-time support
+namespace ct {
```
vitaut, line 17: Why new namespace?

include/fmt/format.h
vitaut: Which local class are you referring to?

test/module-test.cc
```diff
@@ -121,10 +127,23 @@ TEST(module_test, format_to) {
...
+  fmt::format_to(std::back_inserter(w), L"{}", 42);
+  EXPECT_EQ(L"42", w);
+
+  wchar_t wbuffer[4] = {0};
+  fmt::format_to(wbuffer, L"{}", 42);
+  EXPECT_EQ(L"42", std::wstring_view(wbuffer));
+
+  fmt::wmemory_buffer wb;
+  fmt::format_to(wb, L"{}", 42);
+  EXPECT_EQ(L"42", std::wstring_view(wbuffer));
```
vitaut, line 141: As before, I think this does too much functional testing. I suggest having a single wide format_to call.

test/module-test.cc
```diff
@@ -189,18 +233,28 @@ TEST(module_test, vformat_to_n) {
...
   return {reinterpret_cast<const char*>(text.data()),
           text.size() * sizeof(text[0])};
 }
 
+TEST(module_test, print) {
+  EXPECT_WRITE(stdout, fmt::print("{}µ", 42), "42µ");
+  EXPECT_WRITE(stderr, fmt::print(stderr, "{}µ", 4.2), "4.2µ");
+
+  EXPECT_WRITE(stdout, fmt::print(L"{}µ", 42), as_string(L"42µ"));
+  EXPECT_WRITE(stderr, fmt::print(stderr, L"{}µ", 4.2), as_string(L"4.2µ"));
```
vitaut, line 255: These can be problematic because IIRC they may fail if the stream is not in wide mode. I suggest only checking that this compiles without actually running, i.e. wrap in `if (false)`.

include/fmt/format.h
DanielaE: `struct format_handler` within `vformat_to` in format.h, currently line 2649.

The problem is explored with this stripped-down code example here [https://godbolt.org/z/34Ph9dee9](https://godbolt.org/z/34Ph9dee9). Fortunately I got this compiled with gcc (modules-branch) for the first time just a few minutes ago, so that I have a second opinion on the matter now: gcc doesn't require this 'symbolic-linking' of names into the body of `vformat_to` and all of my studies of the standard seem to corroborate that. Therefore I consider this an implementation bug in latest msvc. This symbolic linking is the only workaround that works at all. Qualified name lookup doesn't. And only local classes within function templates are affected.

include/fmt/compile.h
DanielaE: This is stuff that was in namespace `detail` before. It must be made visible to be used from importers, f.e. by the `""_cf` udl. The `compiled_string` tag class can be moved into namespace `fmt` just as well.

test/module-test.cc
DanielaE: These are different overloads in the API. I think this is the minimal set for full coverage.

test/module-test.cc
DanielaE: Ok, will do.

DanielaE:
Even better then. I'm not keen on seeing macros in any API.

`FMT_STRING` being obsolete is news to me. AFAICS the code paths are different with and without `FMT_STRING`. And it is used internally too which was giving me quite some head-scratching how to make these internal uses work, not to mention external uses. Cameron was kindly helping me analyzing the problem.

Both macros are listed in the documentation as official part of the API and this is why I felt like providing them.

vitaut:
`FMT_STRING` will still be supported for pre-C++20 compilers but still modules are C++20 we don't need to export it here.

include/fmt/format.h
DanielaE: This is acknowledged as bug in msvc by realiable sources familiar with the compiler: name-binding is happening in phase 2 instead of phase 1 in this particular case. In other words: too late to find anything in namespace `detail`.

include/fmt/compile.h
vitaut: Why can't we make `detail::compiled_string` visible? It's an implementation detail and not supposed to be directly used by anyone, so `detail` is the correct namespace.

DanielaE: If we make `detail::compiled_string` visible then the namespace `detail` becomes implicitly visible, too. This disenfranchises testing from ever checking if anything from that namespace becomes unwittingly visible that was never meant to be visible int the first place. This is the sole reason why I introduced namespace `ct`: everything in there is the sole minimum of non-public API that needs to be visible. I'd be sad if we'd call namespace `detail` "open season" to visibility. I was bugging Cameron to fix the current issue in msvc of `detail` being visible even if nothing in there is currently visible (he checked this with his BMI explorer tool) and he told me that this should be easy to fix. 🤞

I've just finished reworking the PR and peeked here to check for any news. Part of the rework was moving `compiled_string` to namespace `fmt` because I got the feeling you were objecting to putting it into `ct` where other necessarily visible compile-time related stuff is moved to as well. I will push the PR as it is now. Changing it in either direction is a piece of cake and can be done within a few minutes.

DanielaE:
BTW, I've removed the block in my local repo that prevents msvc from running the `enforce-checks-test`: it passes the test with flying colours. The full CI matrix for Windows is green.

include/fmt/compile.h
```diff
@@ -11,7 +11,12 @@
 #include "format.h"
 
 FMT_BEGIN_NAMESPACE
-namespace detail {
+FMT_MODULE_EXPORT_BEGIN
+
+// A compile-time string which is compiled into fast formatting code.
+class compiled_string {};
```
vitaut, line 17: Does it still need to be public if we don't export the `FMT_COMPILE` macro and only provide a UDL?

DanielaE: Well spotted - thanks!
All related stuff removed.

include/fmt/compile.h
vitaut: Not making `detail` visible makes sense but `ct` looks like part of the public API which is undesirable. I suggest renaming to something like `detail_exported` to make it clear that this is an internal API but exported for module purposes.

include/fmt/format.h
vitaut: Could you add a short comment clarifying that this is a workaround for an msvc bug?

include/fmt/compile.h
DanielaE: Done

include/fmt/compile.h
```diff
@@ -164,7 +164,7 @@ struct is_compiled_string : std::is_base_of<compiled_string, S> {};
 #endif
 
 #if FMT_USE_NONTYPE_TEMPLATE_PARAMETERS
-template <typename Char, size_t N, fixed_string<Char, N> Str>
+template <typename Char, size_t N, fmt::detail_exported::str<Char, N> Str>
```
vitaut, line 167: I suggest keeping the old name (fixed_string) which is more descriptive.

DanielaE: Right.
My concern is the size of these `fmt::v7::detail_exported::fixed_string` linker symbols. This is just the common prefix. If you look at real code, the actual linker symbols with the specialized fixed strings become horribly long. My be I can convince Cameron to use a different mangling, may be derived from the base64-encoded SHA1 of the string value.

Anyway, I will change that.

include/fmt/format.h
```diff
@@ -778,6 +778,32 @@ FMT_INLINE auto make_args_checked(const S& fmt,
...
+constexpr auto string_to_view(const Char (&s)[N]) -> basic_string_view<Char> {
+  // Remove trailing NUL character if needed. Won't be present if this is used
+  // with a raw character array (i.e. not defined as a string).
+  return {s, N - (std::char_traits<Char>::to_int_type(s[N - 1]) == 0 ? 1 : 0)};
+}
+template <typename Char>
+constexpr auto string_to_view(detail::std_string_view<Char> s)
+    -> basic_string_view<Char> {
+  return {s.data(), s.size()};
+}
```
vitaut, line 804: These are only used in macros that are not exported so probably should be in the `detail` namespace, not `detail_exported`. Also please keep the old names (`compile_string_to_view`).

DanielaE: This is not correct. These overloads _must_ be visible or `xchar.h`:231 will fail to compile. And there is _no_ workaround for that.
I will change the name. Even though nothing is compiled here but just a view to the stored data created.

vitaut:
Thank you!

PR #2341, opened 2021-06-04
Fix MSVC Win32 count_digits

upsj:
We are shifting past 32 bit since the integer literals are 32 bits wide.

Additionally, it looks like having the constexpr table inside the function leads to [pretty bad codegen on MSVC](https://godbolt.org/z/sYzqbE4zr), so it might make sense to [move it outside the function](https://godbolt.org/z/4cv4xMoxr), though I didn't include that here since I don't know how you deal with constexpr globals


vitaut:
Thank you!

As for the table, an easy fix is to make it static on MSVC (it cannot be made static unconditionally because of some clang quirk).

upsj:
Yeah, I noticed the same thing. Not sure whether the standard allows it, but `constexpr static` would be what I would ideally want to use

Issue #2319, opened 2021-05-28
Default format prints nothing for std::chrono::system_clock::time_point


toomuchsalt:
Tested with both current head and version 7.1.3( godbolt )

The following code yields an empty string

```cpp
#include <fmt/chrono.h>

#include <chrono>

int main() {
  fmt::print("Default format: {} ", std::chrono::system_clock::now());
}
```

https://godbolt.org/z/E33jW916o

vitaut:
Looks like the default format should be as described in https://eel.is/c++draft/time#clock.system.nonmembers-2, but this is not implemented yet. PRs are welcome.

sunmy2019:
This part is a little bit triky.
In the C++ stardard, we need a locale object to format `year_month_day` and `hh_mm_ss`. Thus, directly using `fmt::print` lacks the information of which locale to use.
Even with `fmt::format(locale, format, args)`, there is a problem: we cannot squeeze the date/time format direcly from locale (unless we use some internal API). So a call to the `time_put` is inevitable if we stick to the standard.

There are some workarounds.
1. We can throw away the locale object, using `{:%Y-%m-%d %H:%M:%S}` as default format.
2. We can keep the locale object, and use the global locale if unsepecified. Making every output go through the function `time_put`. Just like the how we handle `weekday`'s format.

The first is easy to implement. (I already implemented.) While the latter is more complicated, and of course, more time/memory consuming.

vitaut:
The default should be locale-independent (the first option). There is a Library Working Group issue to fix it in the standard as well.

vitaut:
Default format has been added in #2345.

PR #2345, opened 2021-06-07
adding a default format for std::chrono::time_point<std::chrono::syst…

sunmy2019:
For https://github.com/fmtlib/fmt/issues/2319, I add a defalut spec when the format is `{}`.

Here I use function overload to avoid `if constexpr` or different template specializations. If you want to support more char types, you can provide with more overload resolutions, or use some template metaprogramming techniques (to restrict `sizeof(Char)` ).

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -451,10 +451,10 @@ FMT_END_DETAIL_NAMESPACE
 template <typename Char, typename Duration>
 struct formatter<std::chrono::time_point<std::chrono::system_clock, Duration>,
                  Char> : formatter<std::tm, Char> {
-  FMT_CONSTEXPR FMT_INLINE void generate_defalut_spec(char) {
+  FMT_CONSTEXPR FMT_INLINE void generate_default_spec(char) {
     this->specs = {"%Y-%m-%d %H:%M:%S", 17};
   }
-  FMT_CONSTEXPR FMT_INLINE void generate_defalut_spec(wchar_t) {
+  FMT_CONSTEXPR FMT_INLINE void generate_default_spec(wchar_t) {
     this->specs = {L"%Y-%m-%d %H:%M:%S", 17};
   }
 
@@ -465,7 +465,7 @@ struct formatter<std::chrono::time_point<std::chrono::system_clock, Duration>,
     auto end = it;
     while (end != ctx.end() && *end != '}') ++end;
     if (end == it) {
-      generate_defalut_spec(typename ParseContext::char_type{});
+      generate_default_spec(typename ParseContext::char_type{});
     } else {
       this->specs = {it, detail::to_unsigned(end - it)};
     }
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -451,24 +451,16 @@ FMT_END_DETAIL_NAMESPACE
 template <typename Char, typename Duration>
 struct formatter<std::chrono::time_point<std::chrono::system_clock, Duration>,
                  Char> : formatter<std::tm, Char> {
-  FMT_CONSTEXPR FMT_INLINE void generate_default_spec(char) {
-    this->specs = {"%Y-%m-%d %H:%M:%S", 17};
-  }
-  FMT_CONSTEXPR FMT_INLINE void generate_default_spec(wchar_t) {
-    this->specs = {L"%Y-%m-%d %H:%M:%S", 17};
-  }
-
   template <typename ParseContext>
   FMT_CONSTEXPR auto parse(ParseContext& ctx) -> decltype(ctx.begin()) {
     auto it = ctx.begin();
     if (it != ctx.end() && *it == ':') ++it;
     auto end = it;
     while (end != ctx.end() && *end != '}') ++end;
-    if (end == it) {
-      generate_default_spec(typename ParseContext::char_type{});
-    } else {
+    if (end == it)
+      this->specs = {default_spec, 17};
+    else
       this->specs = {it, detail::to_unsigned(end - it)};
-    }
     return end;
   }
 
@@ -478,6 +470,10 @@ struct formatter<std::chrono::time_point<std::chrono::system_clock, Duration>,
     std::tm time = localtime(val);
     return formatter<std::tm, Char>::format(time, ctx);
   }
+  
+  static constexpr Char default_spec[] = {'%', 'Y', '-', '%', 'm', '-',
+                                          '%', 'd', ' ', '%', 'H', ':',
+                                          '%', 'M', ':', '%', 'S'};
 };
 
 template <typename Char> struct formatter<std::tm, Char> {
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -451,9 +451,7 @@ FMT_END_DETAIL_NAMESPACE
 template <typename Char, typename Duration>
 struct formatter<std::chrono::time_point<std::chrono::system_clock, Duration>,
                  Char> : formatter<std::tm, Char> {
-  FMT_CONSTEXPR formatter() {
-    this->specs = {default_specs, 17};
-  }
+  FMT_CONSTEXPR formatter() { this->specs = {default_specs, 17}; }
 
   template <typename ParseContext>
   FMT_CONSTEXPR auto parse(ParseContext& ctx) -> decltype(ctx.begin()) {
@@ -473,8 +471,8 @@ struct formatter<std::chrono::time_point<std::chrono::system_clock, Duration>,
   }
 
   static constexpr Char default_specs[] = {'%', 'Y', '-', '%', 'm', '-',
-                                          '%', 'd', ' ', '%', 'H', ':',
-                                          '%', 'M', ':', '%', 'S'};
+                                           '%', 'd', ' ', '%', 'H', ':',
+                                           '%', 'M', ':', '%', 'S'};
 };
 
 template <typename Char, typename Duration>
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -452,10 +452,10 @@ template <typename Char, typename Duration>
 struct formatter<std::chrono::time_point<std::chrono::system_clock, Duration>,
                  Char> : formatter<std::tm, Char> {
   FMT_CONSTEXPR FMT_INLINE void generate_defalut_spec(char) {
-    this->specs = "%Y-%m-%d %H:%M:%S";
+    this->specs = {"%Y-%m-%d %H:%M:%S", 17};
   }
   FMT_CONSTEXPR FMT_INLINE void generate_defalut_spec(wchar_t) {
-    this->specs = L"%Y-%m-%d %H:%M:%S";
+    this->specs = {L"%Y-%m-%d %H:%M:%S", 17};
   }
 
   template <typename ParseContext>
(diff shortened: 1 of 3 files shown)


sunmy2019 (commented):
I was presuming `basic_string_view` has a constructor that takes in `const char (&)[N]`.

I look up the source today. It turns out there isn't one.

sunmy2019:
I accidently merged the upstream commits. Now I reset to 12aacc2

sunmy2019:
> do you mean default?

fixed, thanks

include/fmt/chrono.h
```diff
@@ -451,6 +451,27 @@ FMT_END_DETAIL_NAMESPACE
 template <typename Char, typename Duration>
 struct formatter<std::chrono::time_point<std::chrono::system_clock, Duration>,
                  Char> : formatter<std::tm, Char> {
+  FMT_CONSTEXPR FMT_INLINE void generate_default_spec(char) {
```
vitaut, line 454: `FMT_INLINE` is not needed here, please remove.

Also it's better to move this function to the detail namespace since it doesn't depend on template parameters and is not part of the public API.

include/fmt/chrono.h
```diff
@@ -451,6 +451,27 @@ FMT_END_DETAIL_NAMESPACE
...
+    this->specs = {"%Y-%m-%d %H:%M:%S", 17};
+  }
+  FMT_CONSTEXPR FMT_INLINE void generate_default_spec(wchar_t) {
```
vitaut, line 457: same here

include/fmt/chrono.h
```diff
@@ -451,6 +451,27 @@ FMT_END_DETAIL_NAMESPACE
...
+  FMT_CONSTEXPR auto parse(ParseContext& ctx) -> decltype(ctx.begin()) {
+    auto it = ctx.begin();
+    if (it != ctx.end() && *it == ':') ++it;
+    auto end = it;
+    while (end != ctx.end() && *end != '}') ++end;
+    if (end == it) {
+      generate_default_spec(typename ParseContext::char_type{});
+    } else {
+      this->specs = {it, detail::to_unsigned(end - it)};
+    }
```
vitaut, line 471: nit: Please remove curly brackets around single statements for consistency with the rest of the code and apply clang-format.

vitaut (changes requested):
Thanks for the PR. Minor comments inline, otherwise LGTM.

foonathan:
> Here I use function overload to avoid if constexpr or different template specializations. If you want to support more char types, you can provide with more overload resolutions, or use some template metaprogramming techniques (to restrict sizeof(Char) ).

Note that in principle, one can use something like `constexpr const CharT format[] = {'%', 'Y', '-', ... }` and automatically supports every character type. I'm not sure if it's worth it, however.

vitaut:
> > Here I use function overload to avoid if constexpr or different template specializations. If you want to support more char types, you can provide with more overload resolutions, or use some template metaprogramming techniques (to restrict sizeof(Char) ).
> 
> Note that in principle, one can use something like `constexpr const CharT format[] = {'%', 'Y', '-', ... }` and automatically supports every character type. I'm not sure if it's worth it, however.

I don't think it'll work in this particular case because we need a static string and `static constexpr` is not supported in `constexpr` functions (sigh).

foonathan:
It could be made a `static constexpr` at class scope.

include/fmt/chrono.h
sunmy2019: But it's a member function. Moving it to the detailed namespace will make it be templated. It looks like

```cpp
template <typename Char, typename Duration>
FMT_CONSTEXPR void generate_default_spec(
    formatter<std::chrono::time_point<std::chrono::system_clock, Duration>,
              Char>* formatter,
    char) {
  formatter->specs = {"%Y-%m-%d %H:%M:%S", 17};
}
```

Maybe making it a private function is more tidier?

```cpp
private:

FMT_CONSTEXPR void generate_default_spec(char) {    
    this->specs = {"%Y-%m-%d %H:%M:%S", 17};
}
```

vitaut:
> It could be made a `static constexpr` at class scope.

Good idea, let's do that instead of my earlier suggestion.

sunmy2019:
But using `static constexpr` in aab4a1b would cause failure on C++11/14, and succeed on C++17/20.
The linker complained about
```
undefined reference to `fmt::v7::formatter<std::chrono::time_point<std::chrono::_V2::system_clock,
std::chrono::duration<long, std::ratio<1l, 1000000000l> > >, wchar_t, void>::default_spec'
```

include/fmt/chrono.h
```diff
@@ -451,24 +451,16 @@ FMT_END_DETAIL_NAMESPACE
...
   FMT_CONSTEXPR auto parse(ParseContext& ctx) -> decltype(ctx.begin()) {
     auto it = ctx.begin();
     if (it != ctx.end() && *it == ':') ++it;
     auto end = it;
     while (end != ctx.end() && *end != '}') ++end;
-    if (end == it) {
-      generate_default_spec(typename ParseContext::char_type{});
-    } else {
+    if (end == it)
+      this->specs = {default_spec, 17};
```
phprus, line 461: Maybe replace constant ```17``` to ```std::end(default_spec) - std::begin(default_spec)``` ?
```std::begin``` and ```std::end``` functions is ```constexpr``` since C++14.

**Comment by sunmy2019** [comment]
> But using `static constexpr` in [aab4a1b](https://github.com/fmtlib/fmt/commit/aab4a1bfa2c3850a3bd83d68976d6d4b6a106648) would cause failure on C++11/14, and succeed on C++17/20.
> The linker complained about
> 
> ```
> undefined reference to `fmt::v7::formatter<std::chrono::time_point<std::chrono::_V2::system_clock,
> std::chrono::duration<long, std::ratio<1l, 1000000000l> > >, wchar_t, void>::default_spec'
> ```

I see. I need to use `static constexpr inline`, since it's implicit inline since C++17.
https://en.cppreference.com/w/cpp/language/constexpr

**Comment by alexezeder** [comment]
> I need to use `static constexpr inline`, since it's implicit inline since C++17.

`static inline` class variable is a C++17 feature. `constexpr` implies `inline` since C++11.
There is an example in {fmt} how to use `static constexpr` class variables and conform to C++11/14/17+.

https://github.com/fmtlib/fmt/blob/36c294822568774629f3d0a3c14ef1bb3d2cd60f/include/fmt/format.h#L842

https://github.com/fmtlib/fmt/blob/36c294822568774629f3d0a3c14ef1bb3d2cd60f/include/fmt/format-inl.h#L148-L149

**Comment by sunmy2019** [comment]
> > I need to use `static constexpr inline`, since it's implicit inline since C++17.
> 
> `static inline` class variable is a C++17 feature. `constexpr` implies `inline` since C++11.
> There is an example in {fmt} how to use `static constexpr` class variables and conform to C++11/14/17+.
> 
> https://github.com/fmtlib/fmt/blob/36c294822568774629f3d0a3c14ef1bb3d2cd60f/include/fmt/format.h#L842
> 
> https://github.com/fmtlib/fmt/blob/36c294822568774629f3d0a3c14ef1bb3d2cd60f/include/fmt/format-inl.h#L148-L149

Does this mean I need to create a `chrono-inl.h` to provide with both header-only library and (non header-only) library? I also need to add a translation unit like `src/format.cc`, and also a change to the building script.

This seems to be a huge change beyond my scope.

**Comment by alexezeder** [comment]
> This seems to be a huge change beyond my scope.

Can't we use here something like:
```cpp
template<typename A, typename B>
class formatter<A, B> {
  static constexpr auto var = { /* initializer here */ };
};
template <typename A, typename B> constexpr B formatter<A, B>::var[]; 
```

For me, it's a big question why this code:
https://github.com/fmtlib/fmt/blob/36c294822568774629f3d0a3c14ef1bb3d2cd60f/include/fmt/core.h#L1686-L1691
does not require any out-of-class definitions.

sunmy2019:
> Can't we use here something like:
> 
> ```c++
> template<typename A, typename B>
> class formatter<A, B> {
>   static constexpr auto var = { /* initializer here */ };
> };
> template <typename A, typename B> constexpr B formatter<A, B>::var[]; 
> ```

Let's have a try.

include/fmt/chrono.h
phprus: My mistake.
Correct code:
```detail::to_unsigned(std::end(default_spec) - std::begin(default_spec))```

**Comment by foonathan** [comment]
> does not require any out-of-class definitions.

You only need a definition if something is "ODR used". Reading a constant integer is not an ODR use, but some part of the code that deals with string literals seems to ODR use the format spec.

> Can't we use here something like:

This should definitely work.

File: include/fmt/chrono.h (hunk 4 above)
Comment (sunmy2019) [reply include/fmt/chrono.h:461]: I don't quite see the point here. Since we are using an array, the result should be certain at compile time, also this sentence is too long. Maybe we can use a comment to explain?

Comment (sunmy2019) [reply include/fmt/chrono.h:461]: I think this is a design choice, not a technical issue. I will leave the choice to @vitaut .

**Comment by sunmy2019** [comment]
Oops, I find my implementation cannot handle **compile time string** correctly for C++17/20.

The format will be parsed empty in
```cpp
fmt::print(FMT_COMPILE("{}"), std::chrono::system_clock::now());
```

But it works correctly in
```cpp
fmt::print(FMT_COMPILE("{:}"), std::chrono::system_clock::now());
```

I don't know what's behind the scene, but I guess the `parse` function was never entered when using `{}`.

So I remove the assignment `this->specs` to the default constructor.

I tested it on my g++-11 with C++11/14/17/20 standard.

include/fmt/chrono.h
```diff
@@ -451,14 +451,37 @@ FMT_END_DETAIL_NAMESPACE
...
+  }
+
   template <typename FormatContext>
   auto format(std::chrono::time_point<std::chrono::system_clock> val,
               FormatContext& ctx) -> decltype(ctx.out()) {
     std::tm time = localtime(val);
     return formatter<std::tm, Char>::format(time, ctx);
   }
+
+  static constexpr Char default_spec[] = {'%', 'Y', '-', '%', 'm', '-',
```
vitaut, line 475: nit: default_spec -> default_specs since it contains more than one specifier.

include/fmt/chrono.h
vitaut: I suggest replacing 17 with `sizeof(default_spec) / sizeof(Char)`.

include/fmt/chrono.h
```diff
@@ -451,14 +451,37 @@ FMT_END_DETAIL_NAMESPACE
...
   template <typename FormatContext>
   auto format(std::chrono::time_point<std::chrono::system_clock> val,
               FormatContext& ctx) -> decltype(ctx.out()) {
     std::tm time = localtime(val);
     return formatter<std::tm, Char>::format(time, ctx);
   }
+
+  static constexpr Char default_specs[] = {'%', 'Y', '-', '%', 'm', '-',
+                                          '%', 'd', ' ', '%', 'H', ':',
+                                          '%', 'M', ':', '%', 'S'};
```
vitaut, line 477: Is it clang-formatted?

sunmy2019: My bad

include/fmt/chrono.h
```diff
@@ -451,14 +451,35 @@ FMT_END_DETAIL_NAMESPACE
 template <typename Char, typename Duration>
 struct formatter<std::chrono::time_point<std::chrono::system_clock, Duration>,
                  Char> : formatter<std::tm, Char> {
+  FMT_CONSTEXPR formatter() { this->specs = {default_specs, 17}; }
```
vitaut, line 454: As commented elsewhere, I suggest replacing 17 with `sizeof(default_spec) / sizeof(Char)`.

vitaut:
Thank you!

Issue #2337, opened 2021-06-03
3 - chrono-test (Failed)


phprus:
Commit: 760ca5ccc0667f671eba928b081f24286af2f0d4

OS: openSUSE Leap 15.2

Console locale:
```
LANG=ru_RU.UTF-8
LC_CTYPE="ru_RU.UTF-8"
LC_NUMERIC="ru_RU.UTF-8"
LC_TIME="ru_RU.UTF-8"
LC_COLLATE="ru_RU.UTF-8"
LC_MONETARY="ru_RU.UTF-8"
LC_MESSAGES="ru_RU.UTF-8"
LC_PAPER="ru_RU.UTF-8"
LC_NAME="ru_RU.UTF-8"
LC_ADDRESS="ru_RU.UTF-8"
LC_TELEPHONE="ru_RU.UTF-8"
LC_MEASUREMENT="ru_RU.UTF-8"
LC_IDENTIFICATION="ru_RU.UTF-8"
LC_ALL=
```

Locale ru_RU.UTF-8 is present:
```
phprus@phprus:~/tmp/fmt/2/fmt-master> date
Чт июн  3 16:03:19 +05 2021
```

Test error:
```
test 3
      Start  3: chrono-test

3: Test command: /home/phprus/tmp/fmt/2/fmt-master/bin/chrono-test
3: Test timeout computed to be: 10000000
3: [==========] Running 25 tests from 1 test suite.
3: [----------] Global test environment set-up.
3: [----------] 25 tests from chrono_test
3: [ RUN      ] chrono_test.format_tm
3: [       OK ] chrono_test.format_tm (0 ms)
3: [ RUN      ] chrono_test.grow_buffer
3: [       OK ] chrono_test.grow_buffer (0 ms)
3: [ RUN      ] chrono_test.format_to_empty_container
3: [       OK ] chrono_test.format_to_empty_container (0 ms)
3: [ RUN      ] chrono_test.empty_result
3: [       OK ] chrono_test.empty_result (0 ms)
3: [ RUN      ] chrono_test.localtime
3: [       OK ] chrono_test.localtime (0 ms)
3: [ RUN      ] chrono_test.gmtime
3: [       OK ] chrono_test.gmtime (0 ms)
3: [ RUN      ] chrono_test.time_point
3: [       OK ] chrono_test.time_point (0 ms)
3: [ RUN      ] chrono_test.format_default
3: [       OK ] chrono_test.format_default (0 ms)
[...]

PR #2349, opened 2021-06-10
Fix chrono_test.locale on libc++

phprus:
Partial fix for issue https://github.com/fmtlib/fmt/issues/2337#issuecomment-854962103 on libc++.
locale: ja_JP.utf8

See: https://datatracker.ietf.org/doc/html/rfc3629

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -291,6 +291,7 @@ inline null<> gmtime_s(...) { return null<>(); }
 inline auto do_write(const std::tm& time, const std::locale& loc, char format,
                      char modifier) -> std::string {
   auto&& os = std::ostringstream();
+  os.imbue(loc);
   using iterator = std::ostreambuf_iterator<char>;
   const auto& facet = std::use_facet<std::time_put<char, iterator>>(loc);
   auto end = facet.put(os, os, ' ', &time, format, modifier);
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -298,8 +298,16 @@ inline auto do_write(const std::tm& time, const std::locale& loc, char format,
   if (end.failed()) FMT_THROW(format_error("failed to format time"));
   auto str = os.str();
   if (!detail::is_utf8() || loc == std::locale::classic()) return str;
-  // char16_t and char32_t codecvts are broken in MSVC (linkage errors).
-  using code_unit = conditional_t<FMT_MSC_VER != 0, wchar_t, char16_t>;
+    // char16_t and char32_t codecvts are broken in MSVC (linkage errors) and
+    // gcc-4.
+#if FMT_MSC_VER != 0 || \
+    (defined(__GLIBCXX__) && !defined(_GLIBCXX_USE_DUAL_ABI))
+  // The _GLIBCXX_USE_DUAL_ABI macro is always defined in libstdc++ from gcc-5
+  // and newer.
+  using code_unit = wchar_t;
+#else
+  using code_unit = char32_t;
+#endif
   auto& f = std::use_facet<std::codecvt<code_unit, char, std::mbstate_t>>(loc);
   auto mb = std::mbstate_t();
   const char* from_next = nullptr;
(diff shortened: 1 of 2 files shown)


include/fmt/chrono.h
```diff
@@ -298,7 +298,7 @@ inline auto do_write(const std::tm& time, const std::locale& loc, char format,
   auto str = os.str();
   if (!detail::is_utf8() || loc == std::locale::classic()) return str;
   // char16_t and char32_t codecvts are broken in MSVC (linkage errors).
-  using code_unit = conditional_t<FMT_MSC_VER != 0, wchar_t, char16_t>;
+  using code_unit = conditional_t<FMT_MSC_VER != 0, wchar_t, char32_t>;
```
vitaut, line 301: Why char32_t?

phprus: ja_JP.utf8 "chars" may don't included in utf-16 without surrogate pairs.

Today I will update the PR with one more fix and tests.

phprus: If apply commit https://github.com/fmtlib/fmt/pull/2349/commits/6ae63a1ecff9824f0096bb6cd58607c98aa8854c to current master next tests do failed: chrono-test, unicode-test.

vitaut: I think this should be `char16_t` + handling of surrogate pairs for compatibility with msvc where `wchar_t` is used.

phprus: gcc-4 does not support char16_t/char32_t conversion. In gcc/clang wchar_t is are 4 bytes (utf-32).
For compatibility with gcc 4 we need three implementations:
- for UTF-16 wchar_t (Windows)
- for UTF-32 wchar_t (gcc-4)
- for UTF-?? char16/32_t (gcc-5+, clang).

I think it is better to choose UTF-32 for gcc/clang.

I will try to add support unicode surrogate pairs for windows, but I don't know how to test it.

vitaut: `char16_t` being unsupported on some platforms is fine because this code is mostly to handle legacy encodings on Windows. In the unsupported case we should just fallback to not transcoding.

phprus: Implemented a universal algorithm (for 2 and 4 bytes ```wchar_t``` and ```char32_t```).

include/fmt/chrono.h
```diff
@@ -311,12 +319,30 @@ inline auto do_write(const std::tm& time, const std::locale& loc, char format,
     FMT_THROW(format_error("failed to format time"));
   str.clear();
   for (code_unit* p = buf; p != to_next; ++p) {
-    auto c = *p;
+    uint32_t c = static_cast<uint32_t>(*p);
+    if (sizeof(code_unit) == 2 && c >= 0xd800 && c <= 0xdfff) {
+      // surrogate pair
+      ++p;
+      if (p == to_next || (c & 0xfffffc00) != 0xd800 ||
```
vitaut, line 326: `0xfffffc00` -> `0xfc00` since code unit only has 2 bytes.

include/fmt/chrono.h
```diff
@@ -311,12 +319,30 @@ inline auto do_write(const std::tm& time, const std::locale& loc, char format,
...
+          (static_cast<uint32_t>(*p) & 0xfffffc00) != 0xdc00) {
```
vitaut, line 327: similarly here

vitaut (changes requested):

include/fmt/chrono.h
```diff
@@ -291,14 +291,22 @@ inline null<> gmtime_s(...) { return null<>(); }
...
   if (end.failed()) FMT_THROW(format_error("failed to format time"));
   auto str = os.str();
   if (!detail::is_utf8() || loc == std::locale::classic()) return str;
-  // char16_t and char32_t codecvts are broken in MSVC (linkage errors).
-  using code_unit = conditional_t<FMT_MSC_VER != 0, wchar_t, char16_t>;
+    // char16_t and char32_t codecvts are broken in MSVC (linkage errors) and
+    // gcc-4.
+#if FMT_MSC_VER != 0 || \
+    (defined(__GLIBCXX__) && !defined(_GLIBCXX_USE_DUAL_ABI))
+  // macro _GLIBCXX_USE_DUAL_ABI is always defined in libstdc++ from gcc-5+
```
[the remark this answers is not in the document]
phprus: ```libstdc++``` provides version macro (```_GLIBCXX_RELEASE```) begin from GCC 7 (https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_macros.html).
```__GLIBCXX__``` is the release date.

For example ```__GLIBCXX__``` value from GCC 5.1 - 20150422 is less value from GCC 4.9.4 - 20160803.

This is a big problem to compare ```libstdc++``` versions. But macro ```_GLIBCXX_USE_DUAL_ABI``` defined only on ```libstdc++``` from GCC 5+, which is necessary for this check.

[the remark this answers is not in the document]
Comment (vitaut) [reply include/fmt/chrono.h:305]: OK, let's keep `_GLIBCXX_USE_DUAL_ABI` then. One nit: please make the comment a proper sentence:

```c++
// Macro _GLIBCXX_USE_DUAL_ABI is always defined in libstdc++ from gcc-5+.
```

include/fmt/chrono.h
phprus: Code using value ```0xfffffc00``` will work always if condition ```sizeof (code_unit) == 2``` is accidentally removed.
I specially made 4-byte masks so that the code always works. And because type of ```c``` is a ```uint32_t```.

If you think it is better to make a change, I will make it. Please write which option to choose.

include/fmt/chrono.h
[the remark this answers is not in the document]
phprus: Sorry, my written English is very, very bad.

> The _GLIBCXX_USE_DUAL_ABI macro is always defined in libstdc++ from gcc-5 and newer.

It is right?
Or please write the correct comment and I will make the change.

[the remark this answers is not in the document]
vitaut: LGTM

include/fmt/chrono.h
vitaut: I suggest using 2-byte masks because surrogate pairs only make sense for UTF-16.

vitaut:
Thank you!

PR #2350, opened 2021-06-10
Fix chrono_test.weekday on legacy glibc

phprus:
The weekday can be three chars long:

```
# cat /etc/SuSE-release
SUSE Linux Enterprise Server 11 (x86_64)
VERSION = 11
PATCHLEVEL = 3

# date
Чтв Июн 10 23:45:13 YEKT 2021
```


vitaut:
Good catch, thanks!

PR #2352, opened 2021-06-11
Fix invalid code and bad codegen (gcc, msvc)

DanielaE:
Found with tests in Windows when `__cplusplus` returns C++20

- `static constexpr var` is invalid within `constexpr` function.
- make lookup tables `static constexpr`
- instruct msvc to report the true value in `__cplusplus` in _some_ tests to improve test coverage and catch similar problems in the future

For details on codegen have a peek here: [https://godbolt.org/z/Evj57ra7r](https://godbolt.org/z/Evj57ra7r)
MSVC and gcc have similar code generation. The macro `LOCAL` toggles between old and new.


DanielaE:
Superseeded by #2353

PR #2356, opened 2021-06-14
Include test sources to pick up functions and classes from the module…

DanielaE:
… rather than from the non-modular library that's baked into the `test-main` library.

This avoids linker problems:
 - strong ownership model: missing linker symbols
 - weak ownership model: duplicate linker symbols


DanielaE:
I observed this linker problem for the first time when experimenting with this `consteval` in msvc.

test/module-test.cc
```diff
@@ -35,18 +35,41 @@
...
+#endif
+#ifndef _WIN32
+#  define FMT_RETRY_VAL(result, expression, error_result) \
+    do {                                                  \
+      (result) = (expression);                            \
+    } while ((result) == (error_result) && errno == EINTR)
+#else
+#  define FMT_RETRY_VAL(result, expression, error_result) result = (expression)
+#endif
+#define FMT_RETRY(result, expression) FMT_RETRY_VAL(result, expression, -1)
```
vitaut, line 53: Why is this copy-pasted here?

DanielaE: These 2 macros are defined in `os.h` which cannot be included because it would reintroduce parts of {fmt} into the TU with different linkage and/or mangling. Avoiding such problems is the whole point of the PR. Unfortunately, pieces of the library are used in the implementation of the test support facilities of the library itself.

vitaut: Copying `FMT_POSIX` is OK but I suggest inlining the retry loop (without the conditional compilation since it should work on Windows too) into the only place where it is used: https://github.com/fmtlib/fmt/blob/889bbf27a2fb258085de5e993e5907b3edeb4a40/test/gtest-extra.cc#L39

Also please remove this while at it:

https://github.com/fmtlib/fmt/blob/889bbf27a2fb258085de5e993e5907b3edeb4a40/test/gtest-extra.cc#L35-L37

DanielaE: Will do.

DanielaE:
The changes are incorporated, the PR is rebased onto HEAD

test/gtest-extra.cc
```diff
@@ -32,11 +32,10 @@ output_redirect::~output_redirect() FMT_NOEXCEPT {
...
 
 void output_redirect::flush() {
-#  if EOF != -1
-#    error "FMT_RETRY assumes return value of -1 indicating failure"
-#  endif
   int result = 0;
-  FMT_RETRY(result, fflush(file_));
+  do {
+    result = fflush(file_);
+  } while (result == -1 && errno == EINTR);
```
vitaut, line 38: Let's use EOF instead of -1 here.

DanielaE: Done

vitaut:
Thank you!

PR #2365, opened 2021-06-18
Fix bug in cmake join function

phprus:
CMake: 3.15.4

Master project cmake output:
```
-- Version: 7.1.4
```

Tests output:

```Start 20: add-subdirectory-test``` (cmake output):
```
20: Version: 7.1.4
```

```Start 21: static-export-test``` (cmake output):
```
21: Version: TRUE7.1.4
```


vitaut:
Thank you!

Issue #2368, opened 2021-06-21
Fix false linter errors for os.h in Visual Studio/VSCode


AidanSun05:
**The Problem:**
When using `fmt::file` and `fmt::output_file` with Visual Studio 2019 and Visual Studio Code on Windows, the linters in both yield the following errors because they can't find the required definitions:

![image](https://user-images.githubusercontent.com/61399657/122692280-14765900-d202-11eb-928d-f3de8f39dc11.png)

However, the actual compile succeeds. This is because a macro (`FMT_HAS_INCLUDE`) is disabled only when IntelliSense is linting, not when MSVC is compiling (this workaround is from a VS 2015 IntelliSense bug: #842). But disabling this macro in newer versions of VS is no longer necessary and leads to the errors mentioned above.

**The Proposed Fix:**
Change the definition of `FMT_HAS_INCLUDE` to:
```cpp
#if defined(__has_include) &&                                               \
    (!defined(__INTELLISENSE__) || (!FMT_MSC_VER || FMT_MSC_VER > 1900)) && \  // <== (CHANGED THIS)
    (!FMT_ICC_VERSION || FMT_ICC_VERSION >= 1600)
#  define FMT_HAS_INCLUDE(x) __has_include(x)
#else
#  define FMT_HAS_INCLUDE(x) 0
#endif
```

In addition to making a decision based on the absence/presence of IntelliSense, this revised code will also check if MSVC's version macro is above `1900` (the value corresponding to VS2015). This retains the original workaround while removing the false errors on newer VS versions.

Is this feasible and can this work in all cases? If so, will a PR be accepted? Thanks!

vitaut:
I can't tell if it will work in all cases but a PR is welcome and if it passes the CI it should be OK.

AidanSun05:
I have opened: #2370

PR #2370, opened 2021-06-21
PR followup for os.h false linter errors

AidanSun05:
This PR implements the changes proposed in issue #2368.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -64,7 +64,8 @@
 #  define FMT_HAS_FEATURE(x) 0
 #endif
 
-#if defined(__has_include) && !defined(__INTELLISENSE__) && \
+#if defined(__has_include) &&                             \
+    (!defined(__INTELLISENSE__) || FMT_MSC_VER > 1900) && \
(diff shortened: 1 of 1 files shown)


DanielaE (commented):
I suggest to squash all three commits into one to keep history clean.

AidanSun05:
Thanks for the suggestion @DanielaE - I have squashed the commits together.

include/fmt/core.h
```diff
@@ -64,7 +64,8 @@
 #  define FMT_HAS_FEATURE(x) 0
 #endif
 
-#if defined(__has_include) && !defined(__INTELLISENSE__) && \
+#if defined(__has_include) &&                                               \
+    (!defined(__INTELLISENSE__) || (!FMT_MSC_VER || FMT_MSC_VER > 1900)) && \
```
vitaut, line 68: I don't think we need the `!FMT_MSC_VER` check here because it will be covered by `!defined(__INTELLISENSE__)`.

AidanSun05: You are right, `!FMT_MSC_VER` is a redundant check. I have removed it.

vitaut:
Thanks

PR #2376, opened 2021-06-22
format: do not use udl_{arg,formatter} return types when UDL is not in use

zx2c4:
The udl_{arg,formatter} structs are only defined when `FMT_USE_USER_DEFINED_LITERALS` is set, so don't try to define things that return that struct when it's not defined.

Fixes https://bugs.gentoo.org/797406


vitaut:
Thank you!

PR #2382, opened 2021-06-23
Eliminate intel compiler warning

gsjaardema:
The intel compiler defines `__GNUC__`, but does not support the gcc `_Pragma`.  This PR filters out the intel compiler to avoid warnings about an unrecognized function.


vitaut:
Thank you!

Issue #2390, opened 2021-06-26
Add consteval format string checks for wide strings


alabuzhev:
`char`-strings are properly rejected, but the same `wchar_t`-strings are happily compiled.

Example:
```C++
#include <string>
#include <string_view>

#include <fmt/core.h>
#include <fmt/xchar.h>

using namespace std::literals;

int main()
{
    fmt::format(L"{:d}", L"I am not a number");
    fmt::format(L"{:d}"s, L"I am not a number");
    fmt::format(L"{:d}"sv, L"I am not a number");

    fmt::format("{:d}", "I am not a number");
    fmt::format("{:d}"s, "I am not a number");
    fmt::format("{:d}"sv, "I am not a number");
}

```

Godbolt:
https://godbolt.org/z/jKr43Tqfs

vitaut:
Yeah, compile-time checks are currently enabled for the `char` overloads only.

vitaut:
We also need to do this for function taking text styles.

vitaut:
This has been implemented in {fmt} v9.

PR #2392, opened 2021-06-26
#2390: consteval format checks for wchar_t and char8_t

alabuzhev:
Implements #2390.


vitaut:
Thanks for the PR but unfortunately this is a breaking change and will have to wait until the next major release.

alabuzhev:
Ok, should I recreate it at some point before the next major release or leave it with you?

vitaut:
I'm planning to do it for all core formatting functions.

PR #2395, opened 2021-06-27
Fix MSVC warning C4819

erengy:
Hi,

I'm using Visual Studio 2019 Version 16.10.2 on Windows with Japanese system locale (code page 932).

Upgrading from fmt 7.1.3 to 8.0.0 resulted in [Compiler Warning (level 1) C4819](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4819):

```
warning C4819: The file contains a character that cannot be represented in the current code page (932). Save the file in Unicode format to prevent data loss
```

The issue was introduced in 7e72673d87cc44340f4b808d0b536e4e191342b5.


vitaut:
Thank you

PR #2397, opened 2021-06-28
silence warning C4100 on MSVC 2019 when exceptions are disabled

mattiasljungstrom:
Removes the following warnings under MSVC 2019 (16.10.2), with W4 enabled and when exceptions are disabled. (/W4 -D_HAS_EXCEPTIONS=0)

fmt\include\fmt\format.h(810,44): warning C4100: 'message': unreferenced formal parameter
fmt\include\fmt/format-inl.h(2553,59): warning C4100: 'message': unreferenced formal parameter

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -337,7 +337,7 @@ struct monostate {
 // `(void)var` method does not work on many intel compilers.  This is
 // from Herb Sutter, "Shutting up compiler warnings",
 // https://herbsutter.com/2009/10/18/mailbag-shutting-up-compiler-warnings/
-template <class T> void ignore_unused(const T&) {}
+template <class... Ts> FMT_CONSTEXPR void ignore_unused(const Ts&...) {}
 
 // An enable_if helper to be used in template parameters which results in much
 // shorter symbols: https://godbolt.org/z/sWw4vP. Extra parentheses are needed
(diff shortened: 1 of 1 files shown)


vitaut:
This is strange because `FMT_THROW` always uses the argument: https://github.com/fmtlib/fmt/blob/a3f762c5aabeeb54b322e1d85e009f8d873b6c85/include/fmt/format.h#L91-L100

Do you have a godbolt repro?

mattiasljungstrom:
I believe that in release mode the FMT_ASSERT is declared ((void)0), which means that the compiler removes everything in the FMT_THROW(x) macro. (Given that exceptions are disabled.)

phprus:
@vitaut, maybe replace https://github.com/fmtlib/fmt/blob/a3f762c5aabeeb54b322e1d85e009f8d873b6c85/include/fmt/core.h#L364
to:
```c++
FMT_INLINE void assert_ignore(bool, const char*) {}
#    define FMT_ASSERT(condition, message) ::fmt::detail::assert_ignore((condition), (message))
```
?

@mattiasljungstrom, please check this solution.

mattiasljungstrom:
@vitaut makes sense!

This works for my case:
```
#    define FMT_ASSERT(condition, message) ((void)message) 
```

Or this works for MSVC (note constexpr needed), haven't tried other compilers:
```
FMT_INLINE constexpr void assert_ignore(bool, const char*) {}
#    define FMT_ASSERT(condition, message) ::fmt::detail::assert_ignore((condition), (message))

```

mattiasljungstrom:
I've updated the PR to use the assert_ignore() solution for FMT_ASSERT.

vitaut:
Can we use the newly added `ignore_unused`

https://github.com/fmtlib/fmt/blob/889bbf27a2fb258085de5e993e5907b3edeb4a40/include/fmt/core.h#L340

instead of introducing a new function?

mattiasljungstrom:
Sure! I've changed the code to use the ::fmt::ignore_unused(), but I had to add FMT_CONSTEXPR since the assert is sometimes used in constexpr functions.

phprus:
Maybe rewrite ```ignore_unused``` function as:
```c++
template <class ...Ts> FMT_CONSTEXPR void ignore_unused(const Ts& ...) {}
```
and remove ```do...while``` block?

vitaut:
> Maybe rewrite ignore_unused function as

Was about to suggest the same =)

mattiasljungstrom:
updated code with @phprus suggestions.

vitaut:
Thank you!

PR #2398, opened 2021-06-28
Remove unneeded `num_result_bigits` decrement

javierhonduco:
For what I see this seems unused?


vitaut:
Good catch, thanks.

javierhonduco:
Thanks @vitaut! Found this internally running infer in our fbcode project cc @jvillard

PR #2399, opened 2021-06-28
Warning removals in test code

mwinterb:
Mostly 0 to nullptr and adding override to virtual function implementations.

### DIFF
--- a/test/gtest-extra.h
+++ b/test/gtest-extra.h
@@ -116,7 +116,7 @@ class output_redirect {
 class suppress_assert {
  private:
   _invalid_parameter_handler original_handler_;
-  FMT_MAYBE_UNUSED int original_report_mode_;
+  int original_report_mode_;
 
   static void handle_invalid_parameter(const wchar_t*, const wchar_t*,
                                        const wchar_t*, unsigned, uintptr_t) {}
@@ -129,6 +129,7 @@ class suppress_assert {
   ~suppress_assert() {
     _set_invalid_parameter_handler(original_handler_);
     _CrtSetReportMode(_CRT_ASSERT, original_report_mode_);
+    (void)original_report_mode_;
   }
 };
 
--- a/test/gtest-extra.h
+++ b/test/gtest-extra.h
@@ -116,7 +116,7 @@ class output_redirect {
 class suppress_assert {
  private:
   _invalid_parameter_handler original_handler_;
-  int original_report_mode_;
+  FMT_MAYBE_UNUSED int original_report_mode_;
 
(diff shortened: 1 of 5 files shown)


mwinterb:
I hadn't tried to compile the module test, I'll see if I can fix that warning up without using FMT_MAYBE_UNUSED.

test/gtest-extra.h
```diff
@@ -129,6 +129,7 @@ class suppress_assert {
   ~suppress_assert() {
     _set_invalid_parameter_handler(original_handler_);
     _CrtSetReportMode(_CRT_ASSERT, original_report_mode_);
+    (void)original_report_mode_;
```
phprus, line 132: Maybe use function ```ignore_unused``` https://github.com/fmtlib/fmt/blob/785908ee37f970513099704386f1befca25de625/include/fmt/core.h#L340 to suppress warning?

Example:
https://github.com/fmtlib/fmt/blob/785908ee37f970513099704386f1befca25de625/include/fmt/core.h#L2746-L2748

vitaut: Yes please.

test/os-test.cc
```diff
@@ -41,9 +41,9 @@ TEST(util_test, utf16_to_utf8_empty_string) {
 }
 
 template <typename Converter, typename Char>
-void check_utf_conversion_error(
-    const char* message,
-    fmt::basic_string_view<Char> str = fmt::basic_string_view<Char>(0, 1)) {
+void check_utf_conversion_error(const char* message,
+                                fmt::basic_string_view<Char> str =
+                                    fmt::basic_string_view<Char>(nullptr, 1)) {
```
vitaut, line 46: Can this be ` = {nullptr, 1}`?

vitaut (changes requested):
Thanks for the PR. LGTM but please address the inline nits.

mwinterb:
Hi, sorry. I've just been busy. Both those seem fine. One question, though: it seems unintentional that ignore_unused is in the fmt namespace instead of fmt::detail? But then so it can be used in this test, presumably actually in fmt::detail_exported?

vitaut:
> it seems unintentional that ignore_unused is in the fmt namespace instead of fmt::detail?

Good catch, could you move it to detail while at it?

> But then so it can be used in this test, presumably actually in fmt::detail_exported?

This is probably an overkill, let's keep cast to `void` since it's just a test.

vitaut:
I think it's good as is, merged.

mwinterb:
Thank you for merging it and taking care of the namespace fix, too.

PR #2410, opened 2021-07-02
Fix compilation when user-defined formatter provided for class instances

kolya7k:
Fix compilation when user-defined formatter provided for class instances like

```
template <>
struct fmt::formatter<Player*>
{
	static constexpr auto parse(format_parse_context &ctx)
	{
		return ctx.end();
	}

	auto format(const Player *player, FormatContext &ctx) const
	{
		return format_to(ctx.out(), "{}, device {}", player->id, player->device);
	}
};
```

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -1300,7 +1300,9 @@ template <typename Context> struct arg_mapper {
   // We use SFINAE instead of a const T* parameter to avoid conflicting with
   // the C array overload.
   template <typename T>
-  FMT_CONSTEXPR auto map(T) -> enable_if_t<std::is_pointer<T>::value, int> {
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR but this is intentional. Formatting arguments are passed by references, not pointers.

kolya7k:
But it's wrong, i don't want to cast *player at any time when i need to print information about it to log, it is ugly.
Is there any reason to deny passing pointers as parameters?

vitaut:
You don't need to cast anything, just dereference. In general {fmt} discourages passing pointers because there is an ambiguity whether they are printed as pointers or as values. That said you can do an implicit dereference in a wrapper function.

kolya7k:
Dereferencing looks very ugly, 1-10 parameters and dereferencing each... So i just wrote patch for Gentoo dev-libs/libfmt.

PR #2412, opened 2021-07-03
Remove unused `exp_`

joakim-brannstrom:
If the test cases
(format_test.format_float, bigint_test.square,
printf_test.fixed_large_exponent, bigint_test.divmod_assign_unaligned,
format_test.format_double) are correct then this line can be removed.


vitaut:
Thanks for the PR but the current code looks correct (although missing a test case). A PR to add a test case covering this path is welcome.

PR #2414, opened 2021-07-05
fix module test odr violations

cdacamar:
As of 16.11 Preview 3 the MSVC compiler will no longer be, erroneously, be injecting header include guard macros when importing a module interface.  The bugfix exposed some issues related to textually including fmt headers in the test code which causes odr violations with the imported interface.

### DIFF
--- a/test/gtest-extra.h
+++ b/test/gtest-extra.h
@@ -12,7 +12,12 @@
 
 #include <string>
 
+#ifdef FMT_MODULE_TEST
+import fmt;
+#else
 #include "fmt/os.h"
+#endif // FMG_MODULE_TEST
+
 #include "gmock/gmock.h"
 
 #define FMT_TEST_THROW_(statement, expected_exception, expected_message, fail) \
--- a/test/module-test.cc
+++ b/test/module-test.cc
@@ -15,6 +15,8 @@
 #  define FMT_HIDE_MODULE_BUGS
 #endif
 
+#define FMT_MODULE_TEST
+
 #include <bit>
 #include <chrono>
 #include <exception>
--- a/test/util.h
+++ b/test/util.h
@@ -10,7 +10,11 @@
 #include <locale>
 #include <string>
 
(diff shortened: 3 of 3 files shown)


DanielaE:
@cdacamar
I'm a bit puzzled regarding the proposed changes. I've upgraded from pre-2 (19.29.30130) to pre-3 (19.29.30132) and can confirm that the include guard macros are no longer injected into the importing TU. To codify this improvement, I changed `module-test.cc` from
```
#define FMT_OS_H_  // don't pull in os.h directly or indirectly

import fmt;

// check for macros leaking from BMI
static bool macro_leaked =
#if defined(FMT_CORE_H_) || defined(FMT_FORMAT_H_)
    true;
#else
    false;
#endif
```
to
```
#define FMT_OS_H_  // don't pull in os.h directly or indirectly

import fmt;

// check for macros leaking from BMI
static bool macro_leaked =
#if defined(FMT_CORE_H_) || defined(FMT_FORMAT_H_) || !defined(FMT_OS_H_)
    true;
#else
    false;
#endif
```
Beginning with pre-3, include guard macros no longer become `#define`d nor `#undef`ed by importing a named module.
By defining the include guard for `fmt/os.h`, no header becomes `#include`d from {fmt} by the code following the `import`. Every required definition is rather imported from the module, e.g. no more ODR violations. I've checked that explicitly with both dumping the preprocessed code (/P) and applying my fake `consteval` emulation.

The only other change to compile `module-test.cc` successfully with pre-3 is in line 80 where the visibility of the non-exported namespace `fmt::detail` is checked for. The compiler version needs to be bumped up to 192930132.

cdacamar:
@DanielaE while your method works it requires knowing the implementation detail of the header include guards by name (which could change in the future).  The proposed solution here relies only on the fact that we understand the contents of `fmtlib` comes from the module alone so we can forget about the mechanics of `#include` altogether and simply import the module which provides the library.

vitaut:
Thanks for the fix!

Issue #1895, opened 2020-09-22
Feature discussion: formatting at compile-time


alexezeder:
Recently I tried to use {fmt} in compile-time, not some `FMT_STRING` compile-time checking or `FMT_COMPILE` compile-time format string, but using `fmt::format_to` with arguments in compile-time, like this:
``` cpp
#include <fmt/format.h>

constexpr auto get_formatted_string() {
    std::array<char, 30> buffer{};
    fmt::format_to(buffer.data(), "The answer = {}", 42);
    return buffer;
}

int main() {
    constexpr auto formatted_string = get_formatted_string();
    printf("%s", formatted_string.data());
    return 0;
}
```
In this example compiler should just format `"The answer = {}"` with argument `42` into `formatted_string` array at compile-time. But since {fmt} probably was not planned to be used that way there are plenty of compilation errors with this example.

By adding a `constexpr` keyword where it's needed, replacing some `std::` entities to `constexpr`ed self-written ones and using C++20 `std::is_constant_evaluated` to eliminate usages of non-`constexpr` functions, I was able to format integers and strings into the buffer at compile-time. But then I tried to format a floating point number and realized that code for floating point formatting, besides of using maybe not compile-time friendly algorithms, uses dynamically growing `memory_buffer`. 

Despite I am pretty sure that I will be able to force floating point formatting to work at compile-time some time later, I decide to pause my work on this to get some feedback in advance.

So, will this ability be useful for everyone, or not? Because I was just experimenting, there is no need for me to do formatting in compile time. Are there any similar or worse pitfalls you can think of?

vitaut:
Floating-point formatting doesn't allocate in the common case. `memory_buffer` is mostly used for convenience. In general, however, it's hard to make FP formatting `constexpr` as you'll need to handle all the weird FP formats. {fmt} has the infra to do this even in the general case (multiprecision arithmetic) but it's still a ton of work.

> So, will this ability be useful for everyone, or not?

I don't think it will be useful for everyone, but it will probably be useful for someone =). I can think of some use cases, e.g. compile-time error reporting.

> Are there any similar or worse pitfalls you can think of?

FP is the most difficult case but pointers and user-defined types will also be problematic.

alexezeder:
> So, will this ability be useful for _everyone_, or not?

supposed to be anyone, not everyone

[...]

Issue #2357, opened 2021-06-14
[PoC] Segmentation fault (possibly due to ODR violation) when mixing format.h with and without ostream.h in different cpp files


devgs:
I won't speculate, it's just a hypothesis that this is an ODR violation. Maybe it's something else. Maybe library makes a static check and interprets the packaged argument one way or another, depending on static conditions, that can result in two different outcomes: depending on whether `fmt/ostream.h` is included. 

I haven't dug too deep into the library. I've just stoppen on a PoC to reproduce the issue. The issue stems from a fact that some class `Foo` has both: **(1)** conversion to `std::string_view` and **(2)** `operator << (std::ostream&)`.

#### foo.hpp

```cpp
#include <string>
#include <string_view>
#include <iostream>

struct Foo
{
    explicit operator std::string_view() const {
        return _data;
    }

    std::string _data{"foo"};
};

inline std::ostream & operator << (std::ostream & stream, const Foo & foo)
{
    return stream.write(foo._data.data(), foo._data.size());
}
```

#### test.cpp

```cpp
#include "foo.hpp"

#include <fmt/format.h>
// By accident, 'ostream.h' is not included
// #include <fmt/ostream.h>

void test()
{
    fmt::format("test {}", Foo{});
}
```

#### main.cpp

```cpp
#include "foo.hpp"

#include <fmt/format.h>
#include <fmt/ostream.h>

// Just a declaration of function from test.cpp
void test();

int main()
{
    fmt::format("test {}", Foo{});
    // Uncomment to force Segfault no matter the compilation order.
    //test();
    return 0;
}
```

#### Compile and run

**Note:** The order of cpp files is important here. We want to instantiate **test.cpp** first and then setfault in **main.cpp**. Though it will work both ways, granted that both usages are invoked (uncomment call to `test();` in main.cpp), one of the will always fail.

##### FreeBSD

`clang --std=c++17 main.cpp test.cpp -I /usr/local/include -L /usr/local/lib -stdlib=libc++ -lc++ -lm -lfmt -o poc && ./poc`

##### Linux

`clang --std=c++17 test.cpp main.cpp -I /usr/local/include -L /usr/local/lib -lstdc++ -lm -lfmt -o poc && ./poc`

The issue is pretty critical, given how hard it is to track all of the possible places where `format()` can accept argument by conversion to `std::string_view`, without a prior include of `<fmt/ostream.h>`.

Update:

[...]

PR #2426, opened 2021-07-16
FP formatting at compile-time

alexezeder:
Few points for this PR, I mentioned some of them in https://github.com/fmtlib/fmt/issues/1895#issuecomment-863585031:
* works only with `FMT_HEADER_ONLY`
* works only with `fast_float`s, even when specs passed, because of the sign check
* separate `compile-fp-test` added with `HEADER_ONLY` argument (tested via G++11 C++20 and MSVC 2019 C++20 configs on CI), the `compile-test` still uses static(/shared?) library


include/fmt/format-inl.h
```diff
@@ -538,7 +579,7 @@ class bigint {
     *this <<= exp;  // Multiply by pow(2, exp) by shifting.
   }
 
-  void square() {
+  FMT_CONSTEXPR20 void square() {
     int num_bigits = static_cast<int>(bigits_.size());
     int num_result_bigits = 2 * num_bigits;
     basic_memory_buffer<bigit, bigits_capacity> n(std::move(bigits_));
```
DanielaE, line 585: With `FMT_CONSTEXPR20` defined as `constexpr`, line 588 no longer compiles on msvc because it calls a function (i.e. constructor) that's not allowed in a `constexpr` context.

include/fmt/format.h
```diff
@@ -1788,12 +1807,32 @@ auto write_float(OutputIt out, const DecimalFP& fp,
 
 template <typename Char, typename OutputIt, typename T,
           FMT_ENABLE_IF(std::is_floating_point<T>::value)>
-auto write(OutputIt out, T value, basic_format_specs<Char> specs,
-           locale_ref loc = {}) -> OutputIt {
+FMT_CONSTEXPR20 auto write(OutputIt out, T value,
```
DanielaE, line 1810: With `FMT_CONSTEXPR20` defined to `constexpr`, line 1842 is no longer portable and causes compile failures when `std::isfinite()` isn't marked as `constexpr`. The C++ standard doesn't mandate this function to be `constexpr`. Other functions from `cmath` are probably affected as well.

I suggest introducing portable, `constexpr`-capable replacements.

test/CMakeLists.txt
```diff
@@ -74,6 +74,7 @@ if (NOT (MSVC AND BUILD_SHARED_LIBS))
 endif ()
 add_fmt_test(ostream-test)
 add_fmt_test(compile-test)
+add_fmt_test(compile-fp-test HEADER_ONLY)
```
DanielaE, line 77: I suggest adding
```
if (MSVC)
  target_compile_options(compile-fp-test PRIVATE /utf-8 /Zc:__cplusplus
                         /Zc:externConstexpr /Zc:inline)
endif ()
```
to catch portability problems like 'module-test' does.

DanielaE (commented):
There seem to be a few general portability issues that need to be addressed.

include/fmt/core.h
```diff
@@ -100,6 +100,13 @@
 #  define FMT_CONSTEXPR_DECL
 #endif
 
+#if __cplusplus >= 202002L || \
```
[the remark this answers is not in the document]
alexezeder: `/Zc:__cplusplus` added, but it's shame that MSVC doesn't follow standards here...

include/fmt/format-inl.h
alexezeder: fixed, somehow I faced with `std::signbit`, but didn't face with `isfinite` and `isinf` 🤔

include/fmt/format.h
alexezeder: fixed, somehow I faced with std::signbit, but didn't face with isfinite and isinf thinking 🤔

test/CMakeLists.txt
alexezeder: done, but only with `/Zc:__cplusplus`, since other options seem unrelated for this, but I can be wrong here

DanielaE: You were asking for full conformance to the standard from Microsoft, disregarding customer demands (i.e. not following stated business policies). This is what you need to get that in recent language modes. In particular when it comes to anything `constexpr` related.

BTW: what is the reason to demand `HEADER_ONLY`? This sounds a little like a portability or conformance problem to me.

DanielaE: I ran my tests again with your updated PR, using the compiler flags as suggested for correct behaviour of `constexpr`:

msvc 16.10 (i.e. VS2019 latest): ✅
msvc 16.11-pre3 (i.e. VS2019 Preview): ✅
msvc 17.0-pre2 (i.e. VS2022 Preview): ✅

All 20 tests compile and pass.
The library also compiles in all modes at /W4 without warnings.

Congratulations!

But I'm not happy about this `HEADER_ONLY` situation.

include/fmt/format-inl.h
```diff
@@ -272,14 +302,17 @@ inline uint64_t multiply(uint64_t lhs, uint64_t rhs) {
...
+}
 
 // Returns a cached power of 10 `c_k = c_k.f * pow(2, c_k.e)` such that its
 // (binary) exponent satisfies `min_exponent <= c_k.e <= min_exponent + 28`.
-inline fp get_cached_power(int min_exponent, int& pow10_exponent) {
+FMT_CONSTEXPR inline fp get_cached_power(int min_exponent,
+                                         int& pow10_exponent) {
   // Normalized 64-bit significands of pow(10, k), for k = -348, -340, ..., 340.
   // These are generated by support/compute-powers.py.
-  static constexpr const uint64_t pow10_significands[] = {
```
DanielaE, line 282: So you ran into the problem of `static` function block scope variables (with initialization on first use a.k.a 'magic statics') being invalid in `constexpr` context. Giving them automatic duration means initialization on every call.

I recommend moving `pow10_significands[]` and `pow10_exponents` into `basic_data` similar to what I have been proposing in #2352

alexezeder: Actually, I thought that compilers are able to optimize this code (based on the fact that this approach is already used in the library - `power_of_10_64` or `bsr2log10`), but I was wrong making this assumption: https://godbolt.org/z/KWYhcqKxn (Clang ❤️).

Quick bench shows that my change made `get_cached_power` _12 times_ slower on GCC, such a great result:

- GCC bench - https://quick-bench.com/q/2p1fSVAl01RjGGSD8a0lan5OPfU
- Clang bench - https://quick-bench.com/q/VB_AUt8Eeb_QivHXd0C1HB5zA2g

So, yeah, I should definitely fix this.

DanielaE: Yeah, it's used elsewhere and both gcc and msvc are still suffering from the problem that I've been pointing out a month ago. I think #2352 should be reopened and reworked.

.github/workflows/linux.yml
```diff
@@ -43,6 +43,11 @@ jobs:
             cxxflags: -stdlib=libc++
             os: ubuntu-20.04
             install: sudo apt install libc++-11-dev libc++abi-11-dev
+          - cxx: g++-11
+            build_type: Debug
+            std: 20
+            os: ubuntu-20.04
+            install: sudo apt install g++-11
```
vitaut, line 50: nit: I suggest grouping gcc configs together by moving this entry right after `g++-10`.

include/fmt/core.h
```diff
@@ -788,23 +795,25 @@ template <typename T> class buffer {
   auto end() const FMT_NOEXCEPT -> const T* { return ptr_ + size_; }
 
   /** Returns the size of this buffer. */
-  auto size() const FMT_NOEXCEPT -> size_t { return size_; }
+  FMT_CONSTEXPR20 auto size() const FMT_NOEXCEPT -> size_t { return size_; }
```
vitaut, line 798: Can this and some other simpler cases be constexpr instead of `FMT_CONSTEXPR20`?

include/fmt/format-inl.h
```diff
@@ -2365,32 +2463,64 @@ int format_float(T value, int precision, float_specs specs, buffer<char>& buf) {
...
+    grisu_shortest_handler handler{buf.data(), 0,
+                                   boundaries.upper - fp_value.f};
+    auto result =
+        grisu_gen_digits(fp(boundaries.upper, fp_value.e),
+                         boundaries.upper - boundaries.lower, exp, handler);
+    if (result == digits::error) {
+      exp += handler.size - cached_exp10 - 1;
+      fallback_format(value, 0, specs.binary32, buf, exp);
+      return exp;
+    }
```
vitaut, line 2493: Looks like this is reverting to some older version of the FP formatter. Please use the most recent implementation.

vitaut (changes requested):
Thanks for the PR!

.github/workflows/linux.yml
alexezeder: I was thinking about replacing `g++-10 -std=c++20` config with the new one `g++-11 -std=c++20` to get less configs, what do you think about that?

include/fmt/core.h
alexezeder: This would probably work for this line.
I was just placing `constexpr` or `FMT_CONSTEXPR` for classes only if these classes can be used at compile-time in C++11 or C++14 respectively. Looks like I had problems with making some of methods of `buffer` class `constexpr` in C++14 or lower, so I chose C++20.
I can just recheck which methods were causing problems here. Or, I can mark these simple methods `constexpr` despite class `constexpr`ness, but I don't see any advantages in this case.

include/fmt/format-inl.h
alexezeder: Actually this code allows to format at compile-time numbers that are formatted using Dragonbox at runtime. It's either this way or `constexpr`ing entire Dragonbox implementation.

alexezeder:
JFI, I will get back on this PR about a week later.

include/fmt/core.h
vitaut: `constexpr` is slightly cleaner so I'd suggest using it when possible (i.e. for simple functions).

include/fmt/format-inl.h
vitaut: We should definitely not reintroduce Grisu here. I suggest either using `fallback_format` or constexprifying Dragonbox.

.github/workflows/linux.yml
vitaut: Sounds good to me.

alexezeder: Replaced and moved to the other GCC configs

include/fmt/core.h
alexezeder: `constexpr` applied where it's possible, but of course, I could miss something

include/fmt/format-inl.h
alexezeder: I removed this Grisu-based FP formatting and reused the existing code, but @vitaut, could you please check this fallback, because I cannot guarantee that it's 100% correct

test/CMakeLists.txt
alexezeder: I'm not happy about `HEADER_ONLY` too, but I don't know how to implement it in the better way.
Also, I just hope it would not introduce some core issues, like the recently discovered issue [when mixing format.h with and without ostream.h in different cpp files caused segfault](https://github.com/fmtlib/fmt/issues/2357) _(which is, IMHO, unsolvable without changing some core ideas)_.

alexezeder:
JFI, I'm back on this PR.
_God, FO4 has too many sidequests..._

include/fmt/format-inl.h
alexezeder: @DanielaE, about your recommendation - should `pow10_significands` and `pow10_exponents` here (and maybe some other variables in other functions) be moved to `basic_data` only, or maybe they can be just moved outside from the function as separate `static constexpr` variables?

DanielaE: It's about possible duplication of tables in different TUs. If you make them members of `basic_data` then you get guaranteed deduplication across the board because of the age of templates and maturity of implementations. With `static constexpr` you may end up with as many copies as there are TUs #including the respective headers. `constexpr` variables by themselves are language entities with _external_ linkage (i.e. generation of linker symbols). Adding `static` changes linkage to _internal_. It's then up to the compiler and sheer happenstance to deduplicate the tables in the final program. A totally valid implementation is neither required nor encouraged to analyse the TUs further to attempt deduplication because there are so many opportunities to trigger ODR-use and implementations aren't required to check for the possibility of ODR-use from a non-`constexpr` context to _guarantee_ no runtime access to these table.

alexezeder: Ok, thank you for such comprehensive description.
I'm just a bit disappointed that we have to move these initializers back to `basic_data` in `format.h`. Maybe we can reuse the same idea for `format-inl.h` since neither `pow10_significands` nor `pow10_exponents` is used in `format.h`.

DanielaE: What's the reason for disappointment? Is it the need to compile the tables? With the definition of these tables in `format.h` in full view, compilers can take advantage of that. Shifting the definition to the point of instantiation in `format-inl.h` may buy you some microseconds of compilation speed but hides the actual values from the compiler. It's a matter of priorities.

alexezeder: No, I'm disappointed that they were the fields of `basic_data` in `format.h`, if I'm correct, later they were moved here, and now we are about to move it back. And since these variables are not needed in `format.h`, then we can probably create some similar `basic_data` that is specific for `format-inl.h`.
So it's not about some compilation-time improvements (also, I'm pretty sure that `constexpr` variables should be defined anyway, it's another big *no* for this kind of optimization), but about scoping variables to only where they are needed.

include/fmt/format.h
```diff
@@ -1798,22 +1818,76 @@ auto write_float(OutputIt out, const DecimalFP& fp,
...
+      using floaty =
+          conditional_t<std::is_same<T, long double>::value, double, T>;
+      using uint = typename dragonbox::float_info<floaty>::carrier_uint;
+      auto bits = bit_cast<uint>(value);
+      is_negative = (bits & (uint(1) << (num_bits<uint>() - 1))) != 0;
+    } else {
+      FMT_ASSERT(false, "floating point type is not supported");
+    }
+  } else
+#endif
```
vitaut, line 1866: `is_constant_evaluated` is available even if `__cpp_if_constexpr` is not defined so please move the `#ifdef __cpp_if_constexpr` block into the if branch to make the formatting less weird.

include/fmt/format.h
```diff
@@ -1798,22 +1818,76 @@ auto write_float(OutputIt out, const DecimalFP& fp,
...
+                           basic_format_specs<Char> specs, locale_ref loc = {})
+    -> OutputIt {
   if (const_check(!is_supported_floating_point(value))) return out;
   float_specs fspecs = parse_float_type_spec(specs);
   fspecs.sign = specs.sign;
-  if (std::signbit(value)) {  // value < 0 is false for NaN so use signbit.
+  bool is_negative;
+#ifdef __cpp_if_constexpr
+  if (is_constant_evaluated()) {
+    // bit check below is valid only for "fast float" types
```
vitaut, line 1855: I don't think this comment adds anything useful considering that we already have an assert, please remove.

include/fmt/format.h
```diff
@@ -1798,22 +1818,76 @@ auto write_float(OutputIt out, const DecimalFP& fp,
...
+    if constexpr (is_fast_float<T>::value) {
+      using floaty =
+          conditional_t<std::is_same<T, long double>::value, double, T>;
+      using uint = typename dragonbox::float_info<floaty>::carrier_uint;
+      auto bits = bit_cast<uint>(value);
+      is_negative = (bits & (uint(1) << (num_bits<uint>() - 1))) != 0;
```
vitaut, line 1861: I think this can be simplified by bit casting into `char[sizeof(T)]` and checking the most-significant bit (using `is_big_endian` to determine it). Then `is_fast_float` can be replaced with `std::numeric_limits<T>::is_iec559`.

Also I suggest moving the sign check into a separate function, `fmt::detail::signbit`.

vitaut (changes requested):
Could you move `bigint` constexprification to a separate PR since it's pretty straightforward?

include/fmt/format.h
alexezeder: Done

include/fmt/format.h
alexezeder: Removed

include/fmt/format.h
alexezeder: > Also I suggest moving the sign check into a separate function, `fmt::detail::signbit`.

Sign check moved to a separate function. But, for example, this code:
https://github.com/fmtlib/fmt/blob/a842d3f84139cb897496c8f8751acac1ac4179d5/include/fmt/format.h#L1931-L1944
does bit casting of float to int twice now, so I think we shouldn't use `fmt::detail::float_signbit` here. And without using it here, there is only one usage of this function left, so I'm not sure that it's worth it at all.

alexezeder: > I think this can be simplified by bit casting into `char[sizeof(T)]` and checking the most-significant bit (using `is_big_endian` to determine it). Then `is_fast_float` can be replaced with `std::numeric_limits<T>::is_iec559`.

`is_fast_float` is used here to omit `long double` as a type, since it can fulfill the requirements of IEC 559 (IEEE 754) standard, but even in this case, the most-significant bit can be not set - https://godbolt.org/z/ao1ab7eb9. As you [pointed out earlier in the discussion](https://github.com/fmtlib/fmt/issues/1895#issuecomment-864410482), the sign bit for most x86 is [79](https://en.wikipedia.org/wiki/Extended_precision#x86_extended_precision_format), not [127](https://en.wikipedia.org/wiki/Quadruple-precision_floating-point_format#IEEE_754_quadruple-precision_binary_floating-point_format:_binary128), but managing the sign bit number for each platform and each compiler seems.. let's say too tricky 😏.

But even with managing the sign bit number with some constants, the idea of bit casting a floating-point number into `char[]` instead of a single unsigned integer number looks more complicated than the current approach, which is already used in the library.

Also, `floaty` alias after `fast_float` check can be eliminated, since there are only IEEE 754 `float` and `double` left.

include/fmt/format-inl.h
alexezeder: They moved to the newly created `basic_inl_data` templated class. I also moved `power_of_10_64` there, since codegen differs between `static`/`constexpr` variable in the function.

alexezeder:
Hmm... looks like we have problems with `digit_grouping` in the `write_float()` since it's used without `is_constant_evaluated()` check and probably because it contains `std::string`, which is not constexpr-friendly (but it should be, right? 🤔). I [tried to solve this](https://github.com/alexezeder/fmt/commit/8a13e2eac78f0f1aa644c3a0126702cc15217989#diff-bdc6f79e8e9f5b4331d66fb785636a87d29f55cf729865e13925b4209424c878L1841-R1882) by using the code from few commits back under `is_constant_evaluated()` check, since it doesn't use `digit_grouping`, but this solution only works for MSVC, while GCC [still has some problems with this function](https://github.com/alexezeder/fmt/runs/3452327985?check_suite_focus=true#step:5:58).

include/fmt/format-inl.h
```diff
@@ -145,6 +145,61 @@ template <> FMT_FUNC int count_digits<4>(detail::fallback_uintptr n) {
   return i >= 0 ? i * char_digits + count_digits<4, unsigned>(n.value[i]) : 1;
 }
 
+template <typename T = void> struct basic_inl_data {
```
vitaut, line 148: I suggest renaming `basic_inl_data` to `basic_impl_data` since this is implementation-specific data while the `-inl` suffix is only relevant to headers.

include/fmt/format-inl.h
```diff
@@ -2347,7 +2356,11 @@ void fallback_format(Double d, int num_digits, bool binary32, buffer<char>& buf,
 }
 
 template <typename T>
-int format_float(T value, int precision, float_specs specs, buffer<char>& buf) {
+#ifdef FMT_HEADER_ONLY
+FMT_CONSTEXPR20
+#endif
```
vitaut, line 2361: Why is it constexpr only in header-only mode?

include/fmt/format-inl.h
```diff
@@ -2358,13 +2371,17 @@ int format_float(T value, int precision, float_specs specs, buffer<char>& buf) {
       return 0;
     }
     buf.try_resize(to_unsigned(precision));
-    std::uninitialized_fill_n(buf.data(), precision, '0');
+    if (is_constant_evaluated()) {
+      fill_n(buf.data(), precision, '0');
+    } else {
+      std::uninitialized_fill_n(buf.data(), precision, '0');
+    }
```
vitaut, line 2378: I think you can call `fill_n` unconditionally here.

include/fmt/format.h
```diff
@@ -1766,12 +1773,13 @@ inline auto write_significand(OutputIt out, T significand, int significand_size,
...
-                 const basic_format_specs<Char>& specs, float_specs fspecs,
-                 locale_ref loc) -> OutputIt {
+FMT_CONSTEXPR20 auto write_float(OutputIt out, const DecimalFP& fp,
+                                 const basic_format_specs<Char>& specs,
+                                 float_specs fspecs, locale_ref loc)
+    -> OutputIt {
   auto significand = fp.significand;
   int significand_size = get_significand_size(fp);
-  static const Char zero = static_cast<Char>('0');
+  constexpr const Char zero = static_cast<Char>('0');
```
vitaut, line 1782: Why change this?

test/CMakeLists.txt
```diff
@@ -84,6 +84,10 @@ if (NOT (MSVC AND BUILD_SHARED_LIBS))
...
+if (MSVC)
+  target_compile_options(compile-fp-test PRIVATE /Zc:__cplusplus)
```
vitaut, line 89: Why add `/Zc:__cplusplus`?

include/fmt/format.h
```diff
@@ -1871,22 +1879,75 @@ auto write_float(OutputIt out, const DecimalFP& fp,
...
+  using floaty =
+      std::conditional_t<std::is_same<T, long double>::value, double, T>;
+  using uint = typename dragonbox::float_info<floaty>::carrier_uint;
+  auto bits = std::bit_cast<uint>(value);
+  return (bits & exponent_mask<floaty>()) != exponent_mask<floaty>();
+}
+#endif
+
+template <typename T, FMT_ENABLE_IF(std::is_floating_point<T>::value)>
+FMT_INLINE FMT_CONSTEXPR bool float_signbit(T value) {
```
vitaut, line 1906: nit: Let's call it just `signbit` without mangling and refer to it as `detail::signbit` to make it clear that it is just a wrapper around `std::signbit`.

(and the same for `is_infinite_fast_float`.

include/fmt/format.h
```diff
@@ -1871,22 +1879,75 @@ auto write_float(OutputIt out, const DecimalFP& fp,
...
+      using floaty =
+          conditional_t<std::is_same<T, long double>::value, double, T>;
+      using uint = typename dragonbox::float_info<floaty>::carrier_uint;
+      auto bits = bit_cast<uint>(value);
+      return (bits & (uint(1) << (num_bits<uint>() - 1))) != 0;
+    } else {
+      FMT_ASSERT(false, "floating point type is not supported");
+    }
+#endif
+  }
```
vitaut, line 1919: Since casts preserve sign, I suggest casting to `double`, checking `numeric_limits::is_iec559` and using `uint64_t` instead of dragonbox internals. This will make the code more future proof.

vitaut (changes requested):

include/fmt/format.h
vitaut: > does bit casting of float to int twice now, so I think we shouldn't use fmt::detail::float_signbit here.

I think it's OK since cast should be optimized away.

vitaut:
> Hmm... looks like we have problems with digit_grouping in the write_float() since it's used without is_constant_evaluated() check and probably because it contains std::string, which is not constexpr-friendly (but it should be, right? 🤔)

We could probably create a fallback `digit_grouping` for the consteval case.

include/fmt/format.h
alexezeder: There is no way to have the `static` variable in the function marked as `constexpr`, and `write_float` is marked so, with C++20. Of course, it could result in worse codegen, but I didn't properly check that.

test/CMakeLists.txt
alexezeder: It was added by [a suggestion from](https://github.com/fmtlib/fmt/pull/2426#discussion_r671811386) @DanielaE, to enable `compile-fp-test` (a check there is based on `__cplusplus`) on MSVC C++20.

include/fmt/format-inl.h
alexezeder: There is an explanation for this - https://github.com/fmtlib/fmt/issues/1895#issuecomment-863585031.

test/CMakeLists.txt
vitaut: Could you clarify this in a comment?

include/fmt/format-inl.h
alexezeder: Done

include/fmt/format-inl.h
alexezeder: Now I call it unconditionally :+1:

test/CMakeLists.txt
alexezeder: Done

include/fmt/format.h
alexezeder: I hope that I got you right here since I change `is_infinite_fast_float` and `is_finite_fast_float` in the same manner, by making them just wrappers, I mean.

include/fmt/format.h
alexezeder: Done.

alexezeder:
> We could probably create a fallback `digit_grouping` for the consteval case.

Without creating a fallback, but I managed to make it compile on GCC 11.

include/fmt/format.h
```diff
@@ -1871,22 +1909,77 @@ auto write_float(OutputIt out, const DecimalFP& fp,
   });
 }
 
+template <typename T, FMT_ENABLE_IF(std::is_floating_point<T>::value)>
+FMT_CONSTEXPR20 bool isinf(T value) {
+  if (is_constant_evaluated()) {
+#if defined(__cpp_lib_bit_cast) && defined(__cpp_if_constexpr)
+    if constexpr (is_fast_float<T>::value) {
+      using floaty =
+          std::conditional_t<std::is_same<T, long double>::value, double, T>;
```
vitaut, line 1918: Let's move this to a new `float_type` trait similar to https://github.com/fmtlib/fmt/blob/117fc67077ed3bbd10ad4ae83f934d49e5da2cec/include/fmt/core.h#L1231 to reduce copy-paste.

include/fmt/format.h
```diff
@@ -1830,28 +1838,58 @@ auto write_float(OutputIt out, const DecimalFP& fp,
...
-      it = write_significand<Char>(it, significand, significand_size,
-                                   fp.exponent, grouping);
-      if (!fspecs.showpoint) return it;
-      *it++ = decimal_point;
-      return num_zeros > 0 ? detail::fill_n(it, num_zeros, zero) : it;
-    });
+    // this is lambda, because GCC 11 cannot handle digit_grouping usages at
+    // compile-time
+    auto runtime_path = [&]() {
+      auto grouping = digit_grouping<Char>(loc, fspecs.locale);
```
vitaut, line 1844: I suggest moving constexpr handling to `digit_grouping`. What exactly is the problem with it?

vitaut (changes requested):
Mostly looks good, a few more comments inline.

include/fmt/format-inl.h
```diff
@@ -145,6 +145,68 @@ template <> FMT_FUNC int count_digits<4>(detail::fallback_uintptr n) {
...
+      0x8bab8eefb6409c1a, 0xd01fef10a657842c, 0x9b10a4e5e9913129,
+      0xe7109bfba19c0c9d, 0xac2820d9623bf429, 0x80444b5e7aa7cf85,
+      0xbf21e44003acdd2d, 0x8e679c2f5e44ff8f, 0xd433179d9c8cb841,
+      0x9e19db92b4e31ba9, 0xeb96bf6ebadf77d9, 0xaf87023b9bf0ee6b,
+  };
+
+#if FMT_GCC_VERSION && FMT_GCC_VERSION < 409
+  FMT_GCC_PRAGMA("GCC diagnostic push")
+  FMT_GCC_PRAGMA("GCC diagnostic ignored \"-Wnarrowing\"")
+#endif
```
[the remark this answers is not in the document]
alexezeder: Yes, it does exactly that - https://godbolt.org/z/45jYhMj4o 🤦

include/fmt/format.h
vitaut: LGTM

include/fmt/format.h
alexezeder: If I understand this correctly, GCC fails when `digit_grouping` is used in any of `if` branch directly, doesn't matter if it's under `is_constant_evaluated()` check and in `else` clause. Why does it fail - probably because `std::string` inside `digit_grouping` is not constexpr. Why does it fail in a constant-evaluated context with `digit_grouping` not reachable - IDK, from my experience, a simple `is_constant_evaluated()` check was enough to avoid this kind of problem. Even MSVC handles code correctly with `digit_grouping` usage under a `is_constant_evaluated()` check.

vitaut: Let's parameterize `digit_grouping` on `is_constant_evaluated` and provide a fallback that does no grouping if it's true. This will keep uses of `digit_grouping` simple and once `std::string` becomes fully constexpr we'll be able to drop parametrization.

alexezeder: If you mean that `digit_grouping` should have a template parameter, it probably won't work because of the nature of `is_constant_evaluated` - it always returns `true` when invoked at compile-time. That's why we cannot write
`if constexpr (is_constant_evaluated())`.

include/fmt/format-inl.h
alexezeder: It's actually something like this in not header-only mode: https://godbolt.org/z/a5vbjaaev

include/fmt/format.h
```diff
@@ -1871,22 +1909,77 @@ auto write_float(OutputIt out, const DecimalFP& fp,
...
+template <typename T, FMT_ENABLE_IF(std::is_floating_point<T>::value)>
+FMT_INLINE FMT_CONSTEXPR bool signbit(T value) {
+  if (is_constant_evaluated()) {
+#ifdef __cpp_if_constexpr
+    if constexpr (is_fast_float<T>::value) {
+      auto bits = detail::bit_cast<uint64_t>(static_cast<double>(value));
+      return (bits & (uint64_t(1) << (num_bits<uint64_t>() - 1))) != 0;
+    } else {
+      FMT_ASSERT(false, "floating point type is not supported");
+    }
```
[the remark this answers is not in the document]
alexezeder: It looks like no because the code under `is_constant_evaluated()` check is still a part of function code. Without optimization, it's not even omitted, according to Compiler Explorer. So this constant-evaluated code has to be valid, and when it contains static assert, which is false, then the "runtime" version of this function becomes ill-formed.

include/fmt/format.h
```diff
@@ -1871,22 +1909,77 @@ auto write_float(OutputIt out, const DecimalFP& fp,
...
+#endif
+  }
+  return std::isfinite(value);
+}
+
+template <typename T, FMT_ENABLE_IF(std::is_floating_point<T>::value)>
+FMT_INLINE FMT_CONSTEXPR bool signbit(T value) {
+  if (is_constant_evaluated()) {
+#ifdef __cpp_if_constexpr
+    if constexpr (is_fast_float<T>::value) {
```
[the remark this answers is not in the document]
alexezeder: Done

include/fmt/format.h
alexezeder: I simplified `isinf` and `isfinite`, applying the same ideas as for `signbit`, so it's probably irrelevant now.

include/fmt/format.h
vitaut: You are right, parametrization on `is_constant_evaluated` won't work but we could still have some dispatch mechanism that chooses whether to use the fallback or real `digit_grouping`. Something along the lines of https://godbolt.org/z/5rbcEEK87.

include/fmt/format-inl.h
vitaut: I see. Let's at least define it in one place, i.e. something like 

```c++
#ifdef FMT_HEADER_ONLY
#define FMT_HEADER_ONLY_CONSTEXPR20 FMT_CONSTEXPR20
#endif
```

include/fmt/format.h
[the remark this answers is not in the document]
vitaut: Sure but at the very least we could drop the else branch because non-IEEE is very uncommon and we would get a reasonable error anyway from `signbit` below.

include/fmt/format.h
alexezeder: Don't underestimate GCC with its ability to find usages of `std::string` in this particular case - https://godbolt.org/z/7fTdhnsPd. Let's also thank GCC for this glorious error message. 😉

include/fmt/format-inl.h
alexezeder: Done

include/fmt/format.h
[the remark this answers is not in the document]
alexezeder: Done

include/fmt/format.h
vitaut: Interesting but I think something like this should work: https://godbolt.org/z/9he3Yobo4

include/fmt/format.h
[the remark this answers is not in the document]
vitaut: I see you dropped it from isinf and isfinite but not here.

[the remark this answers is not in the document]
alexezeder: > I see you dropped it from isinf and isfinite but not here.

Only because you wrote:

> we would get a reasonable error anyway from `signbit` below.

and it's `signbit`.

include/fmt/format.h
alexezeder: Done

include/fmt/format.h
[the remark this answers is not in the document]
alexezeder: If this line is about the `std::signbit()` call below, then we won't get a reasonable error from there in some cases when FP numbers are non-IEEE and `std::signbit()` manages to work with them.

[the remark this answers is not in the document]
vitaut: > we won't get a reasonable error from there in some cases when FP numbers are non-IEEE and std::signbit() manages to work with them

That seems fine to me. We'll either get an error if `signbit` is not constexpr or it will return the correct result.

include/fmt/format.h
```diff
@@ -1886,22 +1897,95 @@ auto write_float(OutputIt out, const DecimalFP& fp,
...
           FMT_ENABLE_IF(std::is_floating_point<T>::value)>
-auto write(OutputIt out, T value, basic_format_specs<Char> specs,
-           locale_ref loc = {}) -> OutputIt {
+FMT_CONSTEXPR20 auto write(OutputIt out, T value,
+                           basic_format_specs<Char> specs, locale_ref loc = {})
+    -> OutputIt {
   if (const_check(!is_supported_floating_point(value))) return out;
   float_specs fspecs = parse_float_type_spec(specs);
   fspecs.sign = specs.sign;
-  if (std::signbit(value)) {  // value < 0 is false for NaN so use signbit.
```
vitaut, line 1896: Could you keep the comment since it may not be obvious to the reader why we don't use `< 0` here?

include/fmt/format.h
```diff
@@ -1886,22 +1897,95 @@ auto write_float(OutputIt out, const DecimalFP& fp,
   });
 }
 
+template <typename Char> class fallback_digit_grouping {
+ public:
+  explicit constexpr fallback_digit_grouping(locale_ref, bool) {}
```
vitaut, line 1902: I don't think it needs to be explicit.

vitaut (commented):
A few more nits, otherwise LGTM.

alexezeder:
I will address these nits and resolve merge conflicts tomorrow.

include/fmt/format.h
alexezeder: Yeah, it's just copied from `digit_grouping`, done.

include/fmt/format.h
alexezeder: This comment reverted back.

include/fmt/format.h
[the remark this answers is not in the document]
alexezeder: Removed also here.

alexezeder:
Unfortunately, my continuous [benchmark pages](https://alexezeder.github.io/fmt_bnchmrk/) broke at some point (June 10). I will try to fix them soon since I have access to this Raspberry finally 😄.

vitaut (approved):

vitaut:
Merged, thanks!

PR #2427, opened 2021-07-18
Replace `throw` with `FMT_THROW`

TrebledJ:
Using `throw` results in compile errors with `-fno-exceptions`. gcc seems fine with `throw`, but arm-gcc and clang would complain.


vitaut:
Thank you!

Issue #2431, opened 2021-07-21
Copy a fmt::dynamic_format_arg_store<Context> object


spyridon97:
I have a ``fmt::dynamic_format_arg_store<fmt::format_context> args`` object which i populate with some variables.

I would like to be able to copy ``args`` into a new object ``args2`` so that I can push back new variables while :
1) maintaining all the variables of ``args`` in ``args2``
2) without touching ``args``

Is that possible in any way?

1) There is no copy constructor for ``fmt::dynamic_format_arg_store``. Could one be implemented?
2) Can I convert fmt::dynamic_format_arg_store to fmt::basic_format_args, and extract the variables from there and push them back to a ``fmt::dynamic_format_arg_store`` object?

vitaut:
> There is no copy constructor for fmt::dynamic_format_arg_store. Could one be implemented?

It could be implemented but if you want to capture arguments you could use a `std::tuple` instead. Also note that `dynamic_format_arg_store` is moveable.

> Can I convert fmt::dynamic_format_arg_store to fmt::basic_format_args, and extract the variables from there and push them back to a fmt::dynamic_format_arg_store object?

No

spyridon97:
I don't know beforehand the types of variables that I will ``push_back``, therefore, how could I use a ``std::tuple``?

As far as the copy- constructor, I quickly implemented something like below inside ``dynamic_format_arg_store``:

```c++
    /**
     \rst
     Default constructor
     \endrst
     */
    constexpr FMT_INLINE dynamic_format_arg_store() = default;

    /**
     \rst
     Constructs a `dynamic_format_arg_store` object from
     dynamic_format_arg_store`.
     \endrst
     */
    constexpr FMT_INLINE dynamic_format_arg_store(
            const dynamic_format_arg_store<Context>& store)
    {
        this->data_ = store.data_;
        this->named_info_ = store.named_info_;
    }
```

Which seems like serving my purpose. Do you think that something like that could be added?
It would be highly appreciated.

Thanks in advance!

vitaut:
> I don't know beforehand the types of variables that I will push_back, therefore, how could I use a std::tuple?

Yeah, `tuple` can only be used if types are known at compile-time.

> Do you think that something like that could be added?

Maybe but why is move not sufficient?

spyridon97:
Maybe Ii am not understanding the move concept correctly.

Could you give me an example of how ``move`` could be used to achieve my purpose?

vitaut:
[...]

PR #2432, opened 2021-07-22
Add copy constructor for dynamic_format_arg_store

spyridon97:
The goal of the PR is to develop a copy constructor (and default constructor) for the ``dynamic_format_arg_store``.

The reason behind the need for a copy-constructor as stated in https://github.com/fmtlib/fmt/issues/2431 is that I would like to be able to instantiate a dynamic_format_arg_store object using another one to emulate a scope of arguments based on the class hierarchy of my code.

For the implementation of the copy-constructor, the members: ``data_`` and ``named_info_`` have been copied, while ``dynamic_args_`` remained untouched because it also did not have a copy constructor or an iterator to implement a copy constructor.

Using the implementation included in this PR i was able to successfully do something like below:

```c++
        fmt::dynamic_format_arg_store<fmt::format_context> dArgs;

        std::vector<int> v1 = {5, 1235, 1235123};
        dArgs.push_back(fmt::arg("global_username", std::ref("utkarsh1")));
        dArgs.push_back(fmt::arg("global_hostname", std::ref("miron")));
        dArgs.push_back(fmt::arg("global_vector", v1));

        fmt::dynamic_format_arg_store<fmt::format_context> dArgs2(dArgs);
        dArgs2.push_back(fmt::arg("extra", std::ref("extra-value")));
```

Because of the new copy constructor, ``dArgs`` includes the first 3 variables, and ``dArgs2`` includes the first 3 variables plus the 4th one.


include/fmt/args.h
```diff
@@ -143,6 +143,38 @@ class dynamic_format_arg_store
   }
 
  public:
+  /**
+    \rst
+    Default constructor of `dynamic_format_arg_store`
+    \endrst
+  */
```
vitaut, line 150: The comment doesn't say anything useful, please remove.

include/fmt/args.h
```diff
@@ -143,6 +143,38 @@ class dynamic_format_arg_store
...
+  constexpr FMT_INLINE dynamic_format_arg_store() = default;
```
vitaut, line 151: `FMT_INLINE` is not needed here.

include/fmt/args.h
```diff
@@ -143,6 +143,38 @@ class dynamic_format_arg_store
...
+
+    **Example**::
+
+      fmt::dynamic_format_arg_store<fmt::format_context> store;
+      store.push_back(123);
+      store.push_back("test");
+
+      fmt::dynamic_format_arg_store<fmt::format_context> store2(store);
+    \endrst
+  */
```
vitaut, line 166: Let's remove this too because we don't need to explain what a copy constructor is.

include/fmt/args.h
```diff
@@ -143,6 +143,38 @@ class dynamic_format_arg_store
...
+  constexpr FMT_INLINE dynamic_format_arg_store(
```
vitaut, line 167: `FMT_INLINE` is not needed here.

include/fmt/args.h
```diff
@@ -143,6 +143,38 @@ class dynamic_format_arg_store
...
+      const dynamic_format_arg_store<Context>& store)
+      : data_(store.data_),
+        named_info_(store.named_info_)
+#if FMT_GCC_VERSION && FMT_GCC_VERSION < 409
+        ,
+        basic_format_args<Context>()
+#endif
```
vitaut, line 174: Why is this necessary?

vitaut (changes requested):
Thanks for the PR. A few small comments inline.

vitaut:
Also please add a test case to https://github.com/fmtlib/fmt/blob/master/test/args-test.cc.

include/fmt/args.h
spyridon97: this was necessary(?) because it way saying that the copy constructor needs to initialize the parent class too.

vitaut: I suggest moving this to the beginning of the initialization list to improve formatting and since base classes are normally initialized first.

include/fmt/args.h
```diff
@@ -230,3 +242,4 @@ class dynamic_format_arg_store
 FMT_END_NAMESPACE
 
 #endif  // FMT_ARGS_H_
+
```
vitaut, line 245: Please remove the extra newline.

test/args-test.cc
```diff
@@ -171,3 +171,16 @@ TEST(args_test, throw_on_copy) {
...
   EXPECT_EQ(fmt::vformat("{}", store), "foo");
 }
+
+TEST(args_test, copy_constructor) {
+  auto store = fmt::dynamic_format_arg_store<fmt::format_context>();
+  store.push_back(fmt::arg("test1", "value1"));
+  store.push_back(fmt::arg("test2", "value2"));
+  store.push_back(fmt::arg("test3", "value3"));
+
+  fmt::dynamic_format_arg_store<fmt::format_context> store2(store);
```
vitaut, line 181: nit: let's use the same declaration format here

```c++
auto store2 = fmt::dynamic_format_arg_store<fmt::format_context>(store);
```

vitaut (changes requested):

include/fmt/args.h
spyridon97: Fixed

test/args-test.cc
spyridon97: also fixed

test/args-test.cc
```diff
@@ -171,3 +171,16 @@ TEST(args_test, throw_on_copy) {
...
   EXPECT_EQ(fmt::vformat("{}", store), "foo");
 }
+
+TEST(args_test, copy_constructor) {
+  auto store = fmt::dynamic_format_arg_store<fmt::format_context>();
+  store.push_back(fmt::arg("test1", "value1"));
+  store.push_back(fmt::arg("test2", "value2"));
+  store.push_back(fmt::arg("test3", "value3"));
+
+  auto store2 = fmt::dynamic_format_arg_store<fmt::format_context>(store);
```
vitaut, line 181: To get rid of the "useless cast" gcc warning:

```c++
  auto store2 = store;
```

spyridon97: Fixed too.

spyridon97:
AppVeyor build succeeded

vitaut (approved):

vitaut:
@DanielaE, do you know why msvc complains about `test-module.ifc`?

DanielaE:
The good news: Spiros' changes are fine, everything builds and all tests pass with msvc 16.10, 16.11-pre3 and 17.0-pre2. 🎉

The bad news: CI has a new windows-2019 image with CMake 3.21 instead of 3.20. The compiler is still the same (16.10.2)
The error message looks like the command line parameters for compiling `fmt.cc` as a module interface unit are missing with CMake 3.21. That's just a guess derived from the error messages without knowing the actual command line parameters in the compiler invocation! The builds on my local machine are still on CMake 3.20 (plus Ninja as supplied by msvc) where everything is fine with the default 26 parallel build threads in VSCode.

I will probably not find enough time today though to switch my local build environment to CMake 3.21 and look into differences in `build.ninja` generated by 3.20 and 3.21 🤔 Stay tuned!

DanielaE:
Everything builds locally with CMake 3.21! 😱
The compiler options are the same as before. But everything is switched from relative paths to absolute paths in `build.ninja`.
I have no clue how to proceed from here without seeing details and related files from the CI build ...

spyridon97:
> Everything builds locally with CMake 3.21! 
> The compiler options are the same as before. But everything is switched from relative paths to absolute paths in `build.ninja`.
> I have no clue how to proceed from here without seeing details and related files from the CI build ...

spyridon97:
I accidentally closed this PR, so I reopened it. 

Is there any progress as far as the Cmake issues? @DanielaE

DanielaE:
I just found out that the CMake generators "Visual Studio 16 2019" and "Visual Studio 17 2022" are 'improved' in CMake 3.21 in the sense that they are broken: they swallow the compiler option `/interface` in `CMakeLists.txt` line 31
```
  target_compile_options(${target}
      PRIVATE /interface /ifcOutput ${BMI}
      INTERFACE /reference fmt=${BMI})
```
Whatever I've tried to hide these compiler options from being processed by the generators fell flat. `/interface` is removed and `/ifcOutput ...` is converted into `<ModuleOutputFile> ./fmt/build/test/test-module.ifc</ModuleOutputFile>` in `build/test/test-module.vcxproj`. The latter is ok, but the former is 😱

vitaut:
In that case I suggest temporarily setting `FMT_CAN_MODULE` to `OFF` in https://github.com/fmtlib/fmt/blob/561834650aa77ba37b15f7e5c9d5726be5127df9/CMakeLists.txt#L89 to unblock this PR.

vitaut:
Thank you!

DanielaE:
> I just found out that the CMake generators "Visual Studio 16 2019" and "Visual Studio 17 2022" are 'improved' in CMake 3.21 in the sense that they are broken: they swallow the compiler option `/interface` in `CMakeLists.txt` line 31

A fix for that will hopefully land in CMake 3.21.1:
- [Issue](https://gitlab.kitware.com/cmake/cmake/-/issues/22477)
- [Direction](https://gitlab.kitware.com/cmake/cmake/-/issues/22474)

vitaut:
Thanks, @DanielaE, for looking into it!

spyridon97:
@vitaut I saw in the newest version of fmt this copy-constructor was deleted. Would it be possible to explain why?

vitaut:
IIRC it was broken, you can look up the commit history for more details.

spyridon97:
I looked it up, there was not MR for the commit that did that (https://github.com/fmtlib/fmt/commit/be51ee1ceb0553411cae166335b59e7ae5126527)

cdacamar:
@spyridon97 do you know if the CMake issue was addressed which ran afoul with IFC generation?

spyridon97:
> @spyridon97 do you know if the CMake issue was addressed which ran afoul with IFC generation?

Issue: https://gitlab.kitware.com/cmake/cmake/-/issues/22477 has been merged
Issue: https://gitlab.kitware.com/cmake/cmake/-/issues/22474 has a discussion which you can check

PR #2434, opened 2021-07-25
exclude fallback functions when FMT_BUILTIN_CLZ(LL) is not defined

bodomartin:
Trivial change to format.h to exclude fallback functions when FMT_BUILTIN_CLZ and FMT_BUILTIN_CLZLL
are not available. Happens to me on windows clangcl msvc codegen

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -916,6 +916,7 @@ FMT_CONSTEXPR inline auto count_digits(uint128_t n) -> int {
 }
 #endif
 
+#ifdef FMT_BUILTIN_CLZLL
 // It is a separate function rather than a part of count_digits to workaround
 // the lack of static constexpr in constexpr functions.
 inline auto do_count_digits(uint64_t n) -> int {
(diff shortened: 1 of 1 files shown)


include/fmt/format.h
```diff
@@ -918,6 +918,7 @@ FMT_CONSTEXPR inline auto count_digits(uint128_t n) -> int {
 
 // It is a separate function rather than a part of count_digits to workaround
 // the lack of static constexpr in constexpr functions.
+#ifdef FMT_BUILTIN_CLZLL
```
vitaut, line 921: Please move the `#ifdef` right before the comment:

```
#ifdef FMT_BUILTIN_CLZLL
// It is a separate function rather than a part of count_digits to workaround
// the lack of static constexpr in constexpr functions.
```

include/fmt/format.h
```diff
@@ -964,6 +966,7 @@ template <> auto count_digits<4>(detail::fallback_uintptr n) -> int;
 
 // It is a separate function rather than a part of count_digits to workaround
 // the lack of static constexpr in constexpr functions.
+#ifdef FMT_BUILTIN_CLZ
```
vitaut, line 969: same he4re

vitaut (changes requested):
Thanks for the PR. LGTM but please address inline comments.

include/fmt/format.h
bodomartin: done

vitaut (approved):

vitaut:
Thank you!

PR #2440, opened 2021-07-28
Fix build in platformio by renaming really common parameter name.

rytai:
I don't know if this is an overkill but "B1" is a really common variable/parameter name. In particular it conflicts with .platformio/packages/framework-arduinoespressif8266/cores/esp8266/binary.h, which is used throughout the espressif framework.

Part of the file:
#define B000000 0
#define B0000000 0
#define B00000000 0
#define B1 1
#define B01 1
#define B001 1
#define B0001 1

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -1053,8 +1053,8 @@ FMT_CONSTEXPR FMT_INLINE void init_named_args(std::nullptr_t, int, int,
                                               const Args&...) {}
 
 template <bool B = false> constexpr auto count() -> size_t { return B ? 1 : 0; }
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR but I don't think we should implement a workaround for this. It is possible to undefined the problematic macro after including `binary.h`.

rytai:
I totally didn't know about undefining macros. Thanks.

bmedici:
Hi, facing the same problem (conflict redefining B1), what si the correct workaround to be able to use this library alongside Arduino ?
Thank you !

PR #2447, opened 2021-08-06
Resolve default constructor error in Xcode 7.2.1 and 8.2.1

cquammen:
Fix issue that produces the error 'default initialization of an object
of const type 'const detail::buffer_size' without a user-provided
default constructor'

### DIFF
--- a/include/fmt/os.h
+++ b/include/fmt/os.h
@@ -394,7 +394,9 @@ struct ostream_params {
 
 FMT_END_DETAIL_NAMESPACE
 
-constexpr detail::buffer_size buffer_size;
+// Added {} below to work around default constructor error known to
+// occur in Xcode versions 7.2.1 and 8.2.1.
+constexpr detail::buffer_size buffer_size{};
 
 /** A fast output stream which is not thread-safe. */
 class FMT_API ostream final : private detail::buffer<char> {


cquammen:
cc @mathstuf

sergiud:
#2291

seanm:
I've confirmed adding the `{}` fixes the issue with both macOS 10.11 + Xcode 8.2.1 (8C1002) and macOS 10.10 + Xcode 7.2.1 (7C1002).

include/fmt/os.h
```diff
@@ -394,7 +394,7 @@ struct ostream_params {
 
 FMT_END_DETAIL_NAMESPACE
 
-constexpr detail::buffer_size buffer_size;
+constexpr detail::buffer_size buffer_size{};
```
vitaut, line 397: Could you add a short comment clarifying that this is a workaround for clang and which version.

vitaut (commented):
Thanks for a PR!

include/fmt/os.h
cquammen: Sure. Since Apple has their own numbering for llvm and clang, I've just commented on the Xcode version where this is known to occur.

cquammen:
Thanks!

PR #2461, opened 2021-08-22
fuzzer maintenance

pauldreik:
* auto format the code using clang format 11
* fixes UB in a fuzzer helper function, not triggered by the existing fuzzers but by an experimental fuzzer not yet published
* sets the path to clang in one place instead of several (just used manually, not in any automated scripts that I know of)


vitaut:
Thank you!

PR #2469, opened 2021-08-26
add fuzzers for chrono timepoint and localtime,gmtime

pauldreik:
This adds fuzzers for previously unfuzzed parts of the fmt api. I have run them locally without finding anything.


test/fuzzing/chrono-timepoint.cc
```diff
@@ -0,0 +1,51 @@
...
+      break;
+    case 1:
+        // may be the same as system_clock
+        doit<std::chrono::high_resolution_clock>(data, size);
+      break;
+    case 2:
+      // won't compile
+      // doit<std::chrono::steady_clock>(data,size);
+      break;
+    }
```
vitaut, line 47: Let's remove this switch because only formatting of `system_clock` time points is supported. It doesn't make much sense to fuzz `high_resolution_clock` because it will either be an alias of `system_clock` in which case we'll be fuzzing the same thing or won't compile.

test/fuzzing/chrono-timepoint.cc
```diff
@@ -0,0 +1,51 @@
...
+#include "fuzzer-common.h"
+
+/*
+ * a fuzzer for the chrono timepoints formatters
+ * C is a clock (std::chrono::system_clock etc)
+ */
+template <typename C> void doit(const uint8_t* data, size_t size) {
+  using D = typename C::duration;
+  using TP = typename C::time_point;
+  using Rep = typename TP::rep;
```
vitaut, line 14: I suggest inlining these aliases to improve readability since they are only used once or, in the case of Rep, twice. Or just keep `Rep` and inline the other two.

vitaut (changes requested):

pauldreik:
followed your requests, rebased and force pushed.

vitaut (approved):

PR #2475, opened 2021-08-31
Fix copy_str performance

Roman-Koshelev:
Contrary to expectations, optimized overloading is not called. std::copy is perfectly optimized in std::memmove (like std::uninitialized_copy_n, and other analogs that are used in the project) and in many more cases than this implementation suggests

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -742,18 +742,7 @@ inline auto get_container(std::back_insert_iterator<Container> it)
 template <typename Char, typename InputIt, typename OutputIt>
 FMT_CONSTEXPR auto copy_str(InputIt begin, InputIt end, OutputIt out)
     -> OutputIt {
-  while (begin != end) *out++ = static_cast<Char>(*begin++);
-  return out;
-}
-
-template <typename Char, FMT_ENABLE_IF(std::is_same<Char, char>::value)>
-FMT_CONSTEXPR auto copy_str(const Char* begin, const Char* end, Char* out)
-    -> Char* {
-  if (is_constant_evaluated())
-    return copy_str<Char, const Char*, Char*>(begin, end, out);
-  auto size = to_unsigned(end - begin);
-  memcpy(out, begin, size);
-  return out + size;
+  return std::copy(begin, end, out);
 }
 
 /**


vitaut:
`core.h` doesn't depend on `<algorithm>` so `std::copy` can't be used there. When optimized overload is not called?

vitaut:
Also `copy_str` does character conversion if necessary.

foonathan:
Also `std::copy` is only constexpr in C++20.

Roman-Koshelev:
@vitaut Why close PR without solving the problem and waiting for an answer? I wrote a great working implementation in which optimizations not only work but are also applicable in large cases. (See code)

Roman-Koshelev:
@foonathan also like copy_str (which depends on std::is_constant_evaluated ())

vitaut:
Please explain in more detail what problem you are trying to solve.

Roman-Koshelev:
@vitaut Example:
#include <cstdio>

```
template <typename InputIt, typename OutputIt>
void fun(InputIt begin, InputIt end, OutputIt out) {
  printf("N1\n");
}

void fun(const char* begin, const char* end, char* out) {
  printf("N2\n");
}

int main() {
  char buf[10];
  char buf2[10];
  fun(buf, buf + 10, buf2);
}
```

What overload will be caused by the result? Obviously the first, but we want the second

Roman-Koshelev:
The best way I can think of to fix this behavior is
```
#include <cstdio>
#include <type_traits>

template <typename InputIt, typename OutputIt>
void fun(InputIt begin, InputIt end, OutputIt out) {
  printf("N1\n");
}

template <typename InputIt, typename OutputIt>
std::enable_if_t
<
  std::is_same_v<std::remove_const_t<InputIt>, OutputIt> &&
  std::is_trivially_copy_assignable_v<OutputIt>,
  void
>
fun(InputIt* begin, InputIt* end, OutputIt* out) {
  printf("N2\n");
}

int main() {
  char buf[10];
  char buf2[10];
  fun(buf, buf + 10, buf2);
}
```
This works great for pointer iterators (we need separate body movements for wrappers, but that's enough for us)

alexezeder:
> Obviously the first, but we want the second

Why it's obvious? https://godbolt.org/z/PPx3rve9h

alexezeder:
Ok, so here is the updated usage with `copy_str<char>()` in both cases: https://godbolt.org/z/GcbnEMnYe. I just took
```
  char buf[10];
  char buf2[10];
  fun(buf, buf + 10, buf2);
```
from examples without properly adapting it.
Both of these usages invoke `copy_str` implementation with while-loop inside, and since all usages of {fmt} look like `copy_str<T>()`, then I guess there are some problems with `copy_str`.

alexezeder:
But can we reopen this PR, since the [second one](https://github.com/fmtlib/fmt/pull/2477) is wrong because of incorrect `memcpy` usage anyway?

vitaut:
I can reopen this PR but this would require closing the other one to make github happy.

vitaut:
Since this PR cannot be reopen because of a force-push, please move the discussion to #2477.

PR #2477, opened 2021-08-31
Fix copy_str performance 2

Roman-Koshelev:
Contrary to expectations, optimized overloading is not called

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -325,6 +325,8 @@ template <bool B> using bool_constant = std::integral_constant<bool, B>;
 template <typename T>
 using remove_reference_t = typename std::remove_reference<T>::type;
 template <typename T>
+using remove_const_t = typename std::remove_const<T>::type;
+template <typename T>
 using remove_cvref_t = typename std::remove_cv<remove_reference_t<T>>::type;
 template <typename T> struct type_identity { using type = T; };
 template <typename T> using type_identity_t = typename type_identity<T>::type;
@@ -746,13 +748,14 @@ FMT_CONSTEXPR auto copy_str(InputIt begin, InputIt end, OutputIt out)
   return out;
 }
 
-template <typename Char, FMT_ENABLE_IF(std::is_same<Char, char>::value)>
-FMT_CONSTEXPR auto copy_str(const Char* begin, const Char* end, Char* out)
-    -> Char* {
+template <typename Char, typename T, typename U,
+FMT_ENABLE_IF(std::is_same<remove_const_t<T>, U>::value && is_char<U>::value)>
+FMT_CONSTEXPR auto copy_str(T* begin, T* end, U* out)
+    -> U* {
   if (is_constant_evaluated())
-    return copy_str<Char, const Char*, Char*>(begin, end, out);
+    return copy_str<Char, T*, U*>(begin, end, out);
   auto size = to_unsigned(end - begin);
-  memcpy(out, begin, size);
+  memcpy(out, begin, size * sizeof(U));
   return out + size;
 }
 


vitaut:
> optimized overloading is not called

Could you elaborate?

Roman-Koshelev:
@vitaut answered here https://github.com/fmtlib/fmt/pull/2475

vitaut:
I'll reopen #2475 and close this one since all the discussion is there.

vitaut:
Actually, I cannot reopen #2475 because the master branch was force-pushed.

include/fmt/core.h
```diff
@@ -740,19 +740,32 @@ inline auto get_container(std::back_insert_iterator<Container> it)
 }
 
 template <typename Char, typename InputIt, typename OutputIt>
-FMT_CONSTEXPR auto copy_str(InputIt begin, InputIt end, OutputIt out)
```
vitaut, line 743: I think the proper solution would be to constrain this overload to `InputIt` and `OutputIt` types that are different and both not convertible to `const Char*`.

include/fmt/core.h
```diff
@@ -740,19 +740,32 @@ inline auto get_container(std::back_insert_iterator<Container> it)
...
+typename std::enable_if
+<
+  std::is_same<typename std::remove_const<InputIt>::type, OutputIt>::value &&
+  std::is_trivially_copy_assignable<OutputIt>::value,
+  OutputIt*
+>::type
+copy_str(InputIt* begin, InputIt* end, OutputIt* out)
+{
   if (is_constant_evaluated())
-    return copy_str<Char, const Char*, Char*>(begin, end, out);
```
vitaut, line 753: This will mean that we cannot call `copy_str` here but it's just a one-line, so can be just copy the while loop here (and remove the cast).

vitaut (changes requested):
Thanks for the PR.

include/fmt/core.h
Roman-Koshelev: I don't see a problem here.

include/fmt/core.h
vitaut: It looks to me that the problem you are trying to solve is that this overload is called in cases where it shouldn't. Therefore the easiest fix is to constraint this overload.

Roman-Koshelev: Want to limit this overload? We have this feature that works in all cases and cannot be limited. And there is an optimization that is sometimes better suited

vitaut: We'll still have an optimization in the other overload (just remove `FMT_ENABLE_IF` there). To clarify, I think the overload set should be something like:

```c++
template <
  typename Char, typename InputIt, typename OutputIt,
  FMT_ENABLE_IF(!is_convertible_v<IntputIt, const Char*> && !is_convertible_v<OutputIt, const Char*>)>
FMT_CONSTEXPR auto copy_str(InputIt begin, InputIt end, OutputIt out);

template <typename Char>
FMT_CONSTEXPR auto copy_str(const Char* begin, const Char* end, Char* out)
    -> Char*;
```

Roman-Koshelev: Corrected a little. But I cannot understand why we turn off the first overload instead of improving the second.

alexezeder: The corrected version doesn't allow `memcpy` version of `copy_str` to be called for wide (exotic) characters. https://godbolt.org/z/z9Kf9jKbY

include/fmt/core.h
```diff
@@ -740,19 +740,27 @@ inline auto get_container(std::back_insert_iterator<Container> it)
...
-template <typename Char, FMT_ENABLE_IF(std::is_same<Char, char>::value)>
-FMT_CONSTEXPR auto copy_str(const Char* begin, const Char* end, Char* out)
-    -> Char* {
+template <typename Char, typename InputIt, typename OutputIt>
+FMT_CONSTEXPR auto copy_str(InputIt begin, InputIt end, OutputIt out)
+    -> OutputIt {
+  return copy_str_constexpr<Char>(begin, end, out);
+}
+
+template <typename Char, typename InputIt, typename OutputIt,
```
mwinterb, line 755: This is probably nitpicky, but since InputIt and OutputIt aren't iterators in this overload, can you rename the template parameters? I keep getting concerned when I see the "* sizeof(OutputIt)" in the call to memcpy below.

Roman-Koshelev: Corrected.

include/fmt/core.h
Roman-Koshelev: Corrected.

include/fmt/core.h
mwinterb: Thanks!

Roman-Koshelev:
I understand that we have reached an implementation that suits both of us?

include/fmt/core.h
```diff
@@ -740,19 +740,26 @@ inline auto get_container(std::back_insert_iterator<Container> it)
 }
 
 template <typename Char, typename InputIt, typename OutputIt>
-FMT_CONSTEXPR auto copy_str(InputIt begin, InputIt end, OutputIt out)
+FMT_CONSTEXPR auto copy_str_constexpr(InputIt begin, InputIt end, OutputIt out)
     -> OutputIt {
   while (begin != end) *out++ = static_cast<Char>(*begin++);
   return out;
 }
```
vitaut, line 747: I suggest merging this into the first overload of `copy_str` as before. The second overload can continue calling it as `copy_str<Char, const Char*, Char*>(...)`.

include/fmt/core.h
```diff
@@ -740,19 +740,26 @@ inline auto get_container(std::back_insert_iterator<Container> it)
...
-template <typename Char, FMT_ENABLE_IF(std::is_same<Char, char>::value)>
-FMT_CONSTEXPR auto copy_str(const Char* begin, const Char* end, Char* out)
-    -> Char* {
+template <typename Char, typename InputIt, typename OutputIt>
+FMT_CONSTEXPR auto copy_str(InputIt begin, InputIt end, OutputIt out)
+    -> OutputIt {
+  return copy_str_constexpr<Char>(begin, end, out);
+}
+
+template <typename Char, typename Tp, typename Up,
```
vitaut, line 755: nit: Tp -> T, Up -> U for consistency with naming conventions in the library.

vitaut (changes requested):
Two small comments, otherwise looks good.

include/fmt/core.h
Roman-Koshelev: fix

include/fmt/core.h
Roman-Koshelev: I thought it would lead to recursion, but it doesn't

vitaut (approved):

vitaut:
Thank you!

PR #2483, opened 2021-09-02
Fix an odr violation in ranges.h

palacaze:


test/ranges-odr-test.cc
```diff
@@ -0,0 +1,22 @@
+// Formatting library for C++ - the core API
+//
+// Copyright (c) 2012 - present, Victor Zverovich
+// All rights reserved.
+//
+// For the license information refer to format.h.
+//
+// Copyright (c) 2018 - present, Remotion (Igor Schulz)
+// All Rights Reserved
+// {fmt} support for ranges, containers and types tuple interface.
```
vitaut, line 10: This part only applies to ranges.h, please remove.

vitaut:
Thanks for the PR!

test/ranges-odr-test.cc
palacaze: Removed, and explained why this content.

PR #2489, opened 2021-09-07
Eliminate double copying in vformat_to_n

Roman-Koshelev:
format_to_n () may not be slower than format_to(). To do this, we will format immediately into the resulting buffer (until it ends)


vitaut:
Thanks for the PR. I think this should be merged into the main `iterator_buffer` template since it's almost identical.

Roman-Koshelev:
At first glance, yes, that's what I wrote. But it seemed to me that it is easier to take out into a separate class than to fence all in one

vitaut:
OK, let's keep it as a separate specialization.

vitaut:
Thank you!

PR #2497, opened 2021-09-12
Simplifying format_decimal()

Roman-Koshelev:


vitaut:
Thanks for the PR but this would be a regression because `memcpy(..., 2)` generates better code in gcc than copying individual chars.

foonathan:
Couldn't we then put the `if (is_constant_evaluated())` inside `copy2`?

vitaut:
> Couldn't we then put the if (is_constant_evaluated()) inside copy2?

We could, PRs are welcome! =)

Roman-Koshelev:
@vitaut What you said is only valid for gcc 7 and older. I will fix the PR

vitaut:
It's true for gcc 10 as well, notice two move instructions in https://godbolt.org/z/qb4x39s33:

```asm
        movzx   ecx, BYTE PTR .LC0[rdx+1]
        movzx   edx, BYTE PTR .LC0[rdx]
```

Roman-Koshelev:
https://godbolt.org/z/YKvoshP8K strange. Everything is optimized without inlining

vitaut:
Strange indeed.

Roman-Koshelev:
I fixed the PR

Roman-Koshelev:
(Apparently it is necessary to rediscover for changes to come)

vitaut:
Please submit a new PR.

PR #2498, opened 2021-09-12
Simplifying format_decimal()

Roman-Koshelev:


include/fmt/format.h
```diff
@@ -1073,11 +1073,21 @@ inline auto equal2(const char* lhs, const char* rhs) -> bool {
...
+  *dst = static_cast<Char>(dc1);
+}
+
+FMT_CONSTEXPR20 FMT_INLINE void copy2(char* dst, const char* src) { 
+  if (!is_constant_evaluated()) {
+    memcpy(dst, src, 2); 
+  } else {
+    copy2<char>(dst, src);
+  }
 }
```
vitaut, line 1090: Let's merge these two overloads.

include/fmt/format.h
```diff
@@ -1073,11 +1073,21 @@ inline auto equal2(const char* lhs, const char* rhs) -> bool {
...
 
 // Copies two characters from src to dst.
-template <typename Char> void copy2(Char* dst, const char* src) {
-  *dst++ = static_cast<Char>(*src++);
-  *dst = static_cast<Char>(*src);
+template <typename Char>
+FMT_CONSTEXPR20 void copy2(Char* dst, const char* src) {
+  char dc0 = *src++;
+  char dc1 = *src;
```
vitaut, line 1079: No need to introduce these variables.

vitaut (changes requested):

include/fmt/format.h
Roman-Koshelev: All the same, this gives the compiler the ability to generate more efficient code. (Although he may not use it)

include/fmt/format.h
Roman-Koshelev: fix

include/fmt/format.h
vitaut: Could you give an example on godbolt illustrating this?

Roman-Koshelev: https://godbolt.org/z/YKvoshP8K This is not enough? We untie the compiler's hands without losing anything.

vitaut: Interesting. Could you add a short comment clarifying this because otherwise it might be tempting to remove these seemingly redundant variables?

Roman-Koshelev: logical.

Roman-Koshelev: fix

vitaut (approved):

vitaut:
Thank you!

PR #2500, opened 2021-09-13
Optimize %T in tm formatting

phprus:
Optimization ```%T``` in tm formatting like this: https://github.com/fmtlib/fmt/commit/67cb2dad37c5949d8902ba25e703991eadcc3874

And fix wchar_t tm formatting error:
```
/.../fmt/include/fmt/chrono.h:567:15: error: no match for ‘operator==’ (operand types are ‘fmt::v8::basic_string_view<wchar_t>’ and ‘fmt::v8::string_view {aka fmt::v8::basic_string_view<char>}’)
     if (specs == string_view("%F", 2))
```

Maybe I should add the use of ```write_digit2_separated``` for formats:

- ```%Y-%m-%d```
- ```%H:%M:%S```
- ```%Y-%m-%d %H:%M:%S```
- ```%Y-%m-%dT%H:%M:%S```

?

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -564,10 +564,13 @@ template <typename Char> struct formatter<std::tm, Char> {
     while (end != ctx.end() && *end != '}') ++end;
     auto size = detail::to_unsigned(end - it);
     specs = {it, size};
-    if (specs == string_view("%F", 2))
-      spec_ = spec::year_month_day;
-    else if (specs == string_view("%T", 2))
-      spec_ = spec::hh_mm_ss;
+    // basic_string_view<>::compare isn't constexpr before C++17
+    if (specs.size() == 2 && specs[0] == Char('%')) {
+      if (specs[1] == Char('F'))
+        spec_ = spec::year_month_day;
+      else if (specs[1] == Char('T'))
+        spec_ = spec::hh_mm_ss;
+    }
     return end;
   }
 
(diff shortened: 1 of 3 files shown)


include/fmt/chrono.h
```diff
@@ -563,7 +564,13 @@ template <typename Char> struct formatter<std::tm, Char> {
     while (end != ctx.end() && *end != '}') ++end;
     auto size = detail::to_unsigned(end - it);
     specs = {it, size};
-    if (specs == string_view("%F", 2)) spec_ = spec::year_month_day;
+    // basic_string_view<>::compare isn't constexpr before C++17
```
miscco, line 567: There is no `basic_string_view<>` before C++17.

Could it be that the internal replacement should be updated?

phprus: Problem in ```std::char_traits<>::compare```:
https://github.com/fmtlib/fmt/blob/3d0c7ae385bcff3d0c152c358db67a222dd0628c/include/fmt/core.h#L472-L478

constexpr ```std::char_traits<>::compare``` starts from C++17 only

include/fmt/chrono.h
```diff
@@ -1217,6 +1230,7 @@ class weekday {
 };
 
 class year_month_day {};
+class hh_mm_ss {};
```
vitaut, line 1233: What is this for?

vitaut (commented):
Thanks for the PR!

include/fmt/chrono.h
phprus: Oh! Removed.

vitaut:
> Maybe I should add the use of write_digit2_separated for formats

I'm not sure if we should add more special cases but it would be nice to get rid of `strftime` completely in `formatter::format` by parsing the format string and using the optimization when possible. It would be not as fast as special casing but would be more generally applicable.

PR #2508, opened 2021-09-22
fmt: disable consteval with clang

timblechmann:
clang (at least appleclang 12.5/13) ICEs when consteval is enabled.

reduced test case: https://bugs.llvm.org/show_bug.cgi?id=51938

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -247,7 +247,7 @@
 #endif
 
 #ifndef FMT_CONSTEVAL
-#  if ((FMT_GCC_VERSION >= 1000 || FMT_CLANG_VERSION >= 1101) && \
+#  if ((FMT_GCC_VERSION >= 1000) && \
        __cplusplus > 201703L) ||                                 \
       (defined(__cpp_consteval) &&                               \
        !FMT_MSC_VER)  // consteval is broken in MSVC.


vitaut:
Could you provide a repro that triggers the ICE? The one in the clang but report doesn't seem to compile.

timblechmann:
i had a few cases in my codebase. the smallest to reduce from came from spdlog when compiling https://github.com/gabime/spdlog/blob/v1.x/src/file_sinks.cpp against fmt-8.0.1 in c++2a mode ...

i can try to see if i can come up with something smaller in the next few days

vitaut:
Anyway, I don't think we should disable `consteval` in clang completely considering that clang 11 passes all tests. You can compile with `-DFMT_CONSTEVAL=` to disable `consteval` support in {fmt} in your project.

timblechmann:
> Anyway, I don't think we should disable `consteval` in clang completely considering that clang 11 passes all tests.

fwiw, xcode12.5 and xcode13 are not bringing clang-11, but clang-12 and clang-13

> You can compile with `-DFMT_CONSTEVAL=` to disable `consteval` support in {fmt} in your project.

i agree that it is a shame to disable `consteval`. however there are (a) already workarounds for broken compilers and (b) it is not easy for users to isolate that fmt is the source of this error: appleclang just ICEs, the only the linux-to-macos cross toolchain pointed to code which is slightly related to fmt ... if someone switches a large codebase to c++2a, it's almost a needle in a haystack 😬

vitaut:
Looks like the issue is specific to Apple clang so I disabled `consteval` there in https://github.com/fmtlib/fmt/commit/807ee5ec3152429f7103e825609f084c03a38ff4.

Issue #2506, opened 2021-09-20
Is there a specific reason to not constexprify Dragonbox?


jk-jeon:
Is there a specific reason to not `constexpr`ify Dragonbox? If not, I believe it should be not terribly difficult to do so, because as far as I know the only real obstacle is that some intrinsics are not `constexpr`. But we can just wrap those intrinsics with `if (!is_constant_evaluated())` and provide fallback implementations along with that. I'm experimenting myself with the recent master branch, and I guess Dragonbox is fine while it fails at [this](https://github.com/fmtlib/fmt/blob/master/include/fmt/format.h#L2073) line, I do not know why.

Another issue is that the compiler is whining about `static constexpr` tables inside `constexpr` functions being `static`. (See e.g., https://stackoverflow.com/questions/62458079/static-constexpr-variables-in-a-constexpr-function.) Honestly I think this is a sort of nonsensical language restriction, but well, it is like that. A simple solution is to move those tables out of the functions.

alexezeder:
Since I was responsible for the recently implemented `constexpr`ification of FP formatting, I guess I can try to answer that.

> Is there a specific reason to not `constexpr`ify Dragonbox?

I don't have a specific reason for this, it was just [mentioned](https://github.com/fmtlib/fmt/issues/1895#issuecomment-811281446) by @vitaut that the most feasible approach is to use Dragon4 at compile-time. I agreed with this because Dragonbox FP formatting code seems complicated for me, thus it could have some ideas that cannot be `constexpr`ified easily. Also, this algorithm is used only for numbers without additional formatting specs, if I'm correct. Besides, who knows, maybe Dragonbox wouldn't be fast at compile-time the same way it's fast at runtime, this should be benchmarked, of course.

>  If not, I believe it should be not terribly difficult to do so

Sure, but it requires some work, additional work considering choosing the most feasible approach. I'm open to making Dragonbox available at compile-time 😉, but I don't think it would help {fmt} somehow. 

> Another issue is that the compiler is whining about `static constexpr` tables inside `constexpr` functions being `static`

Yeah, it's a [familiar problem](https://github.com/fmtlib/fmt/pull/2426#discussion_r673108338)... probably for everyone who tried to make their code `constexpr`. 🤦

jk-jeon:
[...]

PR #2509, opened 2021-09-22
Add `static` to a table

jk-jeon:
As mentioned [here](https://github.com/fmtlib/fmt/pull/2426#discussion_r673237672) and [here](https://github.com/fmtlib/fmt/issues/2506#issuecomment-925398738), GCC fails to optimize non-`static` `constexpr` table initialization, so it is better to add `static` keyword here.


vitaut:
Thanks!

PR #2510, opened 2021-09-24
changed detection of Intel Compiler Classic to distinguish MS-Windows

mborn-adi:
The Intel Compiler Classic on MS-Windows doesn't have the intrinsics "__builtin_ctz" nor "__builtin_ctzll". Instead, it emulates MSVC. Unfortunately, "__has_builtin" works for version 2021, but results in a linker error nonetheless. 
This pull request changes the detection of these intrinsics accordingly.
I also took the liberty to remove it from format.h, because it was already defined in core.h.
The Intel compiler also doesn't know "#pragma managed(push/pop", so I guarded those as well to get rid of the resulting warnings.

Best Regards,
Mathias

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -34,13 +34,13 @@
 
 #ifdef __ICL
 #  define FMT_ICC_VERSION __ICL
-#  define FMT_ICC_ON_WINDOWS 1
+#  define FMT_ICC_POSIX 0
 #elif defined(__INTEL_COMPILER)
 #  define FMT_ICC_VERSION __INTEL_COMPILER
-#  define FMT_ICC_ON_WINDOWS 0
+#  define FMT_ICC_POSIX 1
 #else
 #  define FMT_ICC_VERSION 0
-#  define FMT_ICC_ON_WINDOWS 0
+#  define FMT_ICC_POSIX 1
 #endif
 
 #ifdef __NVCC__
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -32,10 +32,15 @@
 #  define FMT_GCC_PRAGMA(arg)
 #endif
 
-#if defined(__INTEL_COMPILER)
+#ifdef __ICL
+#  define FMT_ICC_VERSION __ICL
+#  define FMT_ICC_ON_WINDOWS 1
+#elif defined(__INTEL_COMPILER)
 #  define FMT_ICC_VERSION __INTEL_COMPILER
+#  define FMT_ICC_ON_WINDOWS 0
 #else
 #  define FMT_ICC_VERSION 0
+#  define FMT_ICC_ON_WINDOWS 0
 #endif
 
 #ifdef __NVCC__
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -34,13 +34,13 @@
 
 #ifdef __ICL
 #  define FMT_ICC_VERSION __ICL
-#  define FMT_ICC_POSIX 0
+#  define FMT_ICC_INTRINSIC_BUG 1
 #elif defined(__INTEL_COMPILER)
 #  define FMT_ICC_VERSION __INTEL_COMPILER
-#  define FMT_ICC_POSIX 1
+#  define FMT_ICC_INTRINSIC_BUG 0
 #else
 #  define FMT_ICC_VERSION 0
-#  define FMT_ICC_POSIX 1
+#  define FMT_ICC_INTRINSIC_BUG 0
 #endif
 
 #ifdef __NVCC__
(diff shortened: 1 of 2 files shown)


include/fmt/core.h
```diff
@@ -32,10 +32,15 @@
...
+#ifdef __ICL
+#  define FMT_ICC_VERSION __ICL
+#  define FMT_ICC_ON_WINDOWS 1
+#elif defined(__INTEL_COMPILER)
 #  define FMT_ICC_VERSION __INTEL_COMPILER
+#  define FMT_ICC_ON_WINDOWS 0
 #else
 #  define FMT_ICC_VERSION 0
+#  define FMT_ICC_ON_WINDOWS 0
 #endif
```
vitaut, line 44: Does Intel compiler define both `__ICL` and `__INTEL_COMPILER` on Windows?

mborn-adi:
__ICL is only defined on Windows, __INTEL_COMPILER on all platforms.

I verified this myself, but you can read about it [here](https://software.intel.com/content/www/us/en/develop/documentation/cpp-compiler-developer-guide-and-reference/top/compiler-reference/macros/additional-predefined-macros.html).

include/fmt/core.h
```diff
@@ -32,10 +32,15 @@
 #  define FMT_GCC_PRAGMA(arg)
 #endif
 
-#if defined(__INTEL_COMPILER)
+#ifdef __ICL
+#  define FMT_ICC_VERSION __ICL
+#  define FMT_ICC_ON_WINDOWS 1
```
vitaut, line 37: I suggest replacing `FMT_ICC_ON_WINDOWS` with `FMT_ICC_POSIX` defined to 1 on non-Windows platforms. This will simplify the checks.

mborn-adi: Fine with me.

include/fmt/format.h
```diff
@@ -192,7 +187,9 @@ FMT_BEGIN_NAMESPACE
 namespace detail {
 // Avoid Clang with Microsoft CodeGen's -Wunknown-pragmas warning.
 #  if !defined(__clang__)
-#    pragma managed(push, off)
+#    if !defined(__ICL)
+#       pragma managed(push, off)
+#    endif
```
vitaut, line 192: Let's remove `#pragma managed` completely as we don't support managed extensions.

mborn-adi: So done. Please have a look.

include/fmt/format.h
```diff
@@ -173,10 +165,13 @@ FMT_END_NAMESPACE
     !FMT_MSC_VER
 #  define FMT_BUILTIN_CLZLL(n) __builtin_clzll(n)
 #endif
-#if (FMT_GCC_VERSION || FMT_HAS_BUILTIN(__builtin_ctz) || FMT_ICC_VERSION)
+#if (FMT_GCC_VERSION || FMT_HAS_BUILTIN(__builtin_ctz) || FMT_ICC_VERSION) && \
+    FMT_ICC_POSIX
```
vitaut, line 169: I think you can simplify this to just

```
#if FMT_GCC_VERSION || FMT_HAS_BUILTIN(__builtin_ctz) || FMT_ICC_POSIX
```
and similarly below.

include/fmt/core.h
```diff
@@ -32,10 +32,15 @@
...
-#if defined(__INTEL_COMPILER)
+#ifdef __ICL
+#  define FMT_ICC_VERSION __ICL
+#  define FMT_ICC_POSIX 0
+#elif defined(__INTEL_COMPILER)
 #  define FMT_ICC_VERSION __INTEL_COMPILER
+#  define FMT_ICC_POSIX 1
 #else
 #  define FMT_ICC_VERSION 0
+#  define FMT_ICC_POSIX 1
```
phprus, line 43: Why is ```FMT_ICC_POSIX``` equal to 1 for other compilers?

include/fmt/format.h
mborn-adi: Unfortunately not.
FMT_HAS_BUILTIN(__builtin_ctz) and FMT_HAS_BUILTIN(__builtin_ctzll) evaluate true on the Intel Classic 2021, and using those intrinsics compiles just fine, but results in linker errors. I think that's a bug in the compiler.

include/fmt/core.h
mborn-adi: You are making a good point. It is misleading. But it makes sure the intrinsic detection still works here for other compilers:
https://github.com/fmtlib/fmt/blob/17fb5de4a1fd1cfd4d93b8233c1b5cc03a43f2ef/include/fmt/format.h#L168-L169
Since this is essentially a work around for a compiler bug, perhaps we should introduce something like FMT_ICC_BUILTIN_CTZ_BUG to be crystal clear about it?

mborn-adi:
I have renamed the macro to make the intend clear. It's a work around for a compiler bug.

vitaut:
Thank you. Please report the compiler bug to Intel.

PR #2512, opened 2021-09-25
Properly specify source directory in CMake examples

Smittyvb:
The CMake source directory is in the parent directory, which is `..`, not `...` (one too many dots).


alexezeder:
Maybe `...` here means "other CMake parameters you want to pass to configure {fmt}"? 🤔
Besides, I don't think this is related to examples in any way.

vitaut:
Thanks for the PR but ... is not a path here but a placeholder for cmake parameters.

PR #2516, opened 2021-09-28
Describe a better approach of how to use {fmt} as a depency in a Bazel build

Vertexwahn:
This PR improves the Bazel build-related documentation.

* The old approach of downloading the repository twice is unclean and unnecessary and was replaced by a better approach.
* The Bazel documentation is now very clear about how to use {fmt} as a dependency in a Bazel related project

Preview of changed documentation: https://github.com/Vertexwahn/fmt/tree/bazel-support/support/bazel

### DIFF
--- a/support/bazel/README.md
+++ b/support/bazel/README.md
@@ -1,85 +1,73 @@
 # Bazel support
 
-Bazel is an open-source build tool.
-More information about Bazel can be found [here](https://bazel.build/).
+To get [Bazel](https://bazel.build/) working with {fmt} you can copy the files `BUILD.bazel`, `WORKSPACE.bazel`, `.bazelrc`, and `.bazelversion` from this folder (`support/bazel`) to the root folder of this project. This way {fmt} gets bazelized and can be used with Bazel (e.g. doing a `bazel build //...` on {fmt}). 
 
-## Using the fmt repository with Bazel
+## Using {fmt} as a dependency
 
-Even though the {fmt} repository does not contain a `WORKSPACE` file in its root directory,
-there is an easy approach to use the {fmt} repository with Bazel out of the box.
-This is demonstrated in the following example.
+The following minimal example shows how to use {fmt} as a dependency within a Bazel project.
 
-Add to your `WORKSPACE` file:
+The following file structure is assumed:
 
-```python
-load("@bazel_tools//tools/build_defs/repo:git.bzl", "new_git_repository")
-
-# Fetch all files from fmt including the BUILD file `support/bazel/BUILD.bazel`
-new_git_repository(
-    name = "fmt_workaround",
-    branch = "master",
-    remote = "https://github.com/fmtlib/fmt/",
-    build_file_content = "# Empty build file on purpose"
-)
-
-# Now the BUILD file `support/bazel/BUILD.bazel` can be used:
-new_git_repository(
-    name = "fmt",
-    branch = "master",
-    remote = "https://github.com/fmtlib/fmt/",
-    build_file = "@fmt_workaround//:support/bazel/BUILD.bazel"
-)
+```
+example
+├── BUILD.bazel
+├── main.cpp
+└── WORKSPACE.bazel
 ```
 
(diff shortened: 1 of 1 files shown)


support/bazel/README.md
```diff
@@ -1,85 +1,63 @@
...
     name = "fmt",
-    branch = "bazel-support", # A copy of master where BUILD.bazel, WORKSPACE.bazel, .bazelrc and .bazelversion are moved to root
-    remote = "https://github.com/<user_or_organisation>/fmt", # replace <user_or_organisation> by a valid account
+    branch = "master",
+    patch_cmds = [
+        "mv support/bazel/.bazelrc .bazelrc",
+        "mv support/bazel/.bazelversion .bazelversion",
+        "mv support/bazel/BUILD.bazel BUILD.bazel",
+        "mv support/bazel/WORKSPACE.bazel WORKSPACE.bazel",
+    ],
```
alexezeder, line 43: I don't want to say that it's a bad thing, but looks like [this solution wouldn't work on Windows](https://docs.bazel.build/versions/main/repo/http.html#:~:text=the%20%60patches%60%20attribute.-,patch_cmds,be%20executed%20on%20Windows%2C%20which%20requires%20Bash%20binary%20to%20exist.,-patch_tool) without Bash.

Vertexwahn: Good point!

The assumption here is that MSYS2 x86_64 is installed on Windows. If you follow the [Bazel Windows installation guide](https://docs.bazel.build/versions/4.2.1/install-windows.html) there is this section:

> Depending on which languages you want to build, you will need:
> 
> 
> MSYS2 x86_64
> 
> MSYS2 is a software distro and building platform for Windows. It contains Bash and common Unix tools (like grep, tar, git).
> 
> You will need MSYS2 to build, test, or run targets that depend on Bash. Typically these are genrule, sh_binary, sh_test, but there may be more (e.g. Starlark rules). Bazel shows an error if a build target needs Bash but Bazel could not locate it.
> 

Since I had on every system also MSYS2 x86_64 installed (CI job installed it) it worked without any problems.

I could now add a comment e.g. "On Windows systems make sure that MSYS2 x86_64 is installed according to the Bazel Installation guide..." and/or provide Powershell commands to move the files.

I am even not sure if the rule `git_repository` works without MSYS2 x86_64 being installed. I could imagine that without MSYS2 x86_64 there will be no `git` for Bazel in place (even if you have a "normal" Git installation on your Windows 10 system/environment path) - but this is pure speculation from my side - I have to test this first. Once I have found out how this works I will provide an update to this PR.

Vertexwahn: Okay tested it in the meanwhile. The rule `git_repository` works without MSYS2being installed. `patch_cmds` does not work on Windows if MSYS2 is not installed. Therefore, `patch_cmds_win` has to be provided to make it working without MSYS2 installed. I fixed the PR accordingly.

Vertexwahn:
Please review again

support/bazel/README.md
```diff
@@ -1,85 +1,73 @@
...
+cc_binary(
+    name = "Demo",
+    srcs = ["main.cpp"],
+    deps = ["@fmt"],
+)
+```
+
+The *BUILD* file defines a binary named `Demo` that has a dependency to {fmt}.
+
+To execute the binary you can run`bazel run //:Demo`.
```
alexezeder, line 72: One space is missing here.

alexezeder (commented):
I guess everything is fine, except for this tiny typo:

support/bazel/README.md
Vertexwahn: Oh no! Fixed now.

Vertexwahn:
@alexezeder Typo fixed!

alexezeder:
> @alexezeder Typo fixed!

Nice, let's just wait for @vitaut, maybe he would add more comments, but more importantly, he can merge this PR 😄.

vitaut:
LGTM, thanks!

PR #2517, opened 2021-09-29
Add concepts supoort for ranges formatter; Allow customize prefix, de…

### LINKED ISSUE #2514: Unable to specialize for particular container when ranges.h is included
Hi,

It seems that I am getting ambiguous partial specialization with the following:

```
#include<fmt/ranges.h>

template <typename... Ts> struct formatter<std::unordered_map<Ts...>, char, void> {

    constexpr static auto prefix = '{';
    constexpr static auto postfix = '}';

    constexpr auto parse(ParseContext &ctx) -> decltype(ctx.begin()) {
        return ctx.begin();
    }

    auto format(const std::unordered_map<Ts...> &values, FormatContext &ctx)
        -> decltype(ctx.out()) {
        auto out = detail::copy(prefix, ctx.out());
        size_t i = 0;
        auto it = std::begin(values);
        auto end = std::end(values);
        for (; it != end; ++it) {
            if (i > 0)
                out = detail::write_delimiter(out);
            out = detail::write_range_entry<char>(out, *it);
            ++i;
        }
        return detail::copy(postfix, out);
    }
};
```

Error message:
```
/usr/local/opt/llvm/bin/../include/c++/v1/type_traits:3174: error: ambiguous partial specializations of 'formatter<std::unordered_map<std::string, int>>'
In file included from /Users/ma/Codes/toltec/cpp/tula/tests/formatter/test_container.cpp:1:
In file included from /Users/ma/.conan/data/gtest/cci.20210126/_/_/package/f590ad446c1ebfcd972512dc1298aaecd7e19ffb/include/gtest/gtest.h:56:
In file included from /usr/local/opt/llvm/bin/../include/c++/v1/limits:105:
/usr/local/opt/llvm/bin/../include/c++/v1/type_traits:3174:38: error: ambiguous partial specializations of 'formatter<std::unordered_map<std::string, int>>'
    : public integral_constant<bool, __is_constructible(_Tp, _Args...)>
                                     ^
/Users/ma/.conan/data/fmt/8.0.1/_/_/package/66b135c50c184f8a978c3bd32e1ef114be7886ae/include/fmt/core.h:1258:42: note: in instantiation of template class 'std::is_constructible<fmt::formatter<std::unordered_map<std::string, int>>>' requested here
                !is_string<T>::value && !has_formatter<T, Context>::value &&
                                         ^
/Users/ma/.conan/data/fmt/8.0.1/_/_/package/66b135c50c184f8a978c3bd32e1ef114be7886ae/include/fmt/core.h:1260:33: note: while substituting prior template arguments into non-type template parameter [with T = std::unordered_map<std::string, int>]
  FMT_CONSTEXPR FMT_INLINE auto map(const T& val)
                                ^~~~~~~~~~~~~~~~~
/Users/ma/.conan/data/fmt/8.0.1/_/_/package/66b135c50c184f8a978c3bd32e1ef114be7886ae/include/fmt/core.h:1346:50: note: while substituting deduced template arguments into function template 'map' [with T = std::unordered_map<std::string, int>, $1 = (no value)]
    type_constant<decltype(arg_mapper<Context>().map(std::declval<const T&>())),
                                                 ^
/Users/ma/.conan/data/fmt/8.0.1/_/_/package/66b135c50c184f8a978c3bd32e1ef114be7886ae/include/fmt/core.h:1694:21: note: in instantiation of template type alias 'mapped_type_constant' requested here
            detail::mapped_type_constant<Args, Context>::value>(args)...} {
                    ^
/Users/ma/.conan/data/fmt/8.0.1/_/_/package/66b135c50c184f8a978c3bd32e1ef114be7886ae/include/fmt/core.h:1710:10: note: in instantiation of member function 'fmt::format_arg_store<fmt::basic_format_context<fmt::appender, char>, std::unordered_map<std::string, int>>::format_arg_store' requested here
  return {args...};
         ^
/Users/ma/.conan/data/fmt/8.0.1/_/_/package/66b135c50c184f8a978c3bd32e1ef114be7886ae/include/fmt/core.h:2886:28: note: in instantiation of function template specialization 'fmt::make_format_args<fmt::basic_format_context<fmt::appender, char>, std::unordered_map<std::string, int>>' requested here
  return vformat(fmt, fmt::make_format_args(args...));
                           ^
/Users/ma/Codes/toltec/cpp/tula/tests/formatter/../common.h:19:19: note: in instantiation of function template specialization 'fmt::format<std::unordered_map<std::string, int> &>' requested here
    auto s = fmt::format(fmt::runtime(fmt_str), args...);
                  ^
/Users/ma/Codes/toltec/cpp/tula/tests/formatter/test_container.cpp:29:21: note: in instantiation of function template specialization 'tula::testing::fmtlog<char const (&)[4], std::unordered_map<std::string, int>>' requested here
    EXPECT_NO_THROW(fmtlog(
                    ^
/Users/ma/.conan/data/fmt/8.0.1/_/_/package/66b135c50c184f8a978c3bd32e1ef114be7886ae/include/fmt/ranges.h:333:8: note: partial specialization matches [with T = std::unordered_map<std::string, int>, Char = char]
struct formatter<
       ^
/Users/ma/Codes/toltec/cpp/tula/include/tula/formatter/container.h:41:8: note: partial specialization matches [with Ts = <std::string, int, std::hash<std::string>, std::equal_to<std::string>, std::allocator<std::pair<const std::string, int>>>]
struct formatter<std::unordered_map<Ts...>, char, void> {
       ^
```

Any suggestions? (I am using c++20 so concepts are also available. But it does not work as well)

Jerry-Ma:
…limiter and postfix.

This pull request close #2514 .

Summary of changes:

* (C++20 only) Defined concepts `fmt::TupleLike`, `fmt::FormattableRange`. The formatter specializations are defined in terms of these concepts, to allow custom code providing more specialized version of formatters, see `test/ranges-custom-formatter-test.cc` for example usage.

* Moved the implementation of the `formatter::format` to classes `tuple_formatter` and `range_formatter`, and the actual formatter is defined by inheriting these classes. This allow one re-uses the formatting logic in their own specialization.

* Removed the hard-coded delimiter from `detail:write_delimiter`. Adjusted the `formatter_tuple` and `formatter_range` class to have `prefix`, `delimiter`, and `postfix` member variables (all of them is now `char[]` instead of `char`). Subclass of `*_formatter` can choose to initialize these fields differently to customize the behavior.
* 
I have tested compiling on Clang 13.0 both with C++17 and C++20, all tests passed.


include/fmt/ranges.h
```diff
@@ -526,20 +530,43 @@ OutputIt write_range_entry(OutputIt out, const Arg& v) {
...
+#if FMT_HAS_CPP20_CONCEPTS
+template <typename T>
+concept Range = is_range_<T>::value;
+
+template <typename T>
+concept TupleLike = is_tuple_like_<T>::value;
+
+template <typename T>
+concept StdStringLike = is_std_string_like<T>::value;
+#endif
```
vitaut, line 542: I don't think introducing concepts is worth it. It just complicates the implementation since we still have to support pre-C++20.

Jerry-Ma: I looked at the CI reports and I think the errors on linux and mac are due to my removal of the `detail::write_delimiter` function. The delimiter is now defined in the`fmt::formatting_{tuple, range}` class as `char[8]` and this is passed to `detal::copy()` function to write to the output iterator. However, the original `detail::copy()` does not resolves the overload correctly for char[] to the second `copy(const char*, ...)`, but to the first one `copy(const RangeT&, ...)`. I added some traits check to force the resolution but these traits may not be available in older gcc and clang.

As far as the windows builds, I am not familiar with msvs so I am not sure. Maybe just disable this feature for windows.

 

> I don't think introducing concepts is worth it. It just complicates the implementation since we still have to support pre-C++20.
On the other hand I do think the ability to override existing partial specialization of the fmt::formatter is an important feature for practical use of custom formatting. I tends to think having to maintain a single way of formatting certain types within the fmtlib (ranges.h, chrono.h, etc) is quite harder than allowing custom code to do the thing while fmtlib provides the necessary utility.

Jerry-Ma: @vitaut
I've addressed the errors in the CI report.

vitaut: Overriding is fine but it shouldn't rely on C++20 concepts.

vitaut (changes requested):
Thanks for the PR. Providing an opt-in mechanism sounds reasonable but it shouldn't rely on C++20 concepts because it would make usefulness of this feature very limited. Also please remove unrelated changes to delimiter from this PR.

include/fmt/ranges.h
Jerry-Ma: I am not sure the best way to provide a more specialized template definition for custom formatting of the std range types with SFAINE. I'll look into it when time permit.

Jerry-Ma:
> Thanks for the PR. Providing an opt-in mechanism sounds reasonable but it shouldn't rely on C++20 concepts because it would make usefulness of this feature very limited. Also please remove unrelated changes to delimiter from this PR.

The other option may be is to define a macro that wraps around the ranges.h in its entirely, so user who wish to implement their own ranges.h will define the macro. Do you think this works?

I agree that this PR needs to be separated to several of them:

1. A way to opt-in pre-defined formatters in fmtlib. This can be done with the wrapping macro, and this feature will be available for all C++ versions.

2. A way to provide more specialized formatter than those defined in fmtlib. I only know how to do this with the help of C++20 concepts. Any help to make it work for older C++ standards would be highly appreciated

3. A way to fully customize the behavior of the formatter defined in fmtlib. The change of delimiter handling in this PR and the addition of formatter implementation class would be reasonable belongs to this.

Let me know what is your preference for the above changes then I can work on them.

vitaut:
1. here definitely shouldn't be an opt-in. Ranges should continue to be formattable by default.

2. This sounds OK if it can be done without concepts and not very intrusive. I think the opt-out can be done with a trait.

3. You can already customize the delimiter with `fmt::join`. I don't think we need to change delimiter handling in the built-in range formatter.

Jerry-Ma:
> * There definitely shouldn't be an opt-in. Ranges should continue to be formattable by default.

This sounds good. I was thinking using this as the opt-out mechanism actually. But if opt-out can be done with a trait we should do it with the trait option.

> * This sounds OK if it can be done without concepts and not very intrusive. I think the opt-out can be done with a trait.

I am afraid that I don't know how to do it. Could you pls show an example of opt-out with a trait?

> * You can already customize the delimiter with `fmt::join`. I don't think we need to change delimiter handling in the built-in range formatter.

This is true but the fmt::join does not work with nested structures. For example std::map cannot be formatted as [{1: 2}] unless re-implement the related formatters, which is pretty much like what is implemented in ranges.h but with a different write_delimiter function.

Jerry-Ma:
@vitaut However this only works with custom range types. What I would like to have is to provide customized formatting for the std containers such as std::vector<int>

vitaut:
It works with any container although I wouldn't recommend overriding formatting of types you don't control.

Jerry-Ma:
> It works with any container although I wouldn't recommend overriding formatting of types you don't control.

Just try it and clearly this line does the trick!
```
template <typename ...Ts>
struct is_range<std::vector<Ts...>, char> : std::false_type {};
```

In fact, I am able to combine this with concept to opt-out vectors that satisfies a particular constraint, see example here:

https://godbolt.org/z/fj73jerss

I guess my primary usage is covered here. Thanks a lot for the help!

On a separate node, how would you like the part of the PR that factors out the range_formatter and tuple_formatter? I'll be good if I can reuse the formatting logic in my custom formatter class.

PR #2518, opened 2021-09-29
Suppress warning C4127 in chrono.h (conditional expression is constant)

BrukerJWD:
Quite self-explaining: I want to use /WX in my MSVC builds (with C++17) and got these warnings when using `chrono.h`


vitaut:
Thanks

Issue #2455, opened 2021-08-15
clang 13 C++20 consteval error


englercj:
I'm getting a compilation error with `8.0.1` using clang 13:

```
In file included from dependencies/fmtlib.fmt/fmt-8.0.1/src/format.cc:8:
dependencies/fmtlib.fmt/fmt-8.0.1/include/fmt/format-inl.h:92:54: error: call to consteval function 'fmt::basic_format_string<char>::basic_format_string<char [21], 0>' is not a constant expression
  if (written < count) FMT_THROW(system_error(errno, "cannot write to file"));
                                                     ^
dependencies/fmtlib.fmt/fmt-8.0.1/include/fmt/core.h:463:15: note: non-constexpr function 'strlen' cannot be used in a constant expression
      size_ = std::strlen(reinterpret_cast<const char*>(s));
              ^
dependencies/fmtlib.fmt/fmt-8.0.1/include/fmt/core.h:2835:51: note: in call to 'basic_string_view(&"cannot write to file"[0])'
  FMT_CONSTEVAL basic_format_string(const S& s) : str_(s) {
                                                  ^
dependencies/fmtlib.fmt/fmt-8.0.1/include/fmt/format-inl.h:92:54: note: in call to 'basic_format_string("cannot write to file")'
  if (written < count) FMT_THROW(system_error(errno, "cannot write to file"));
                                                     ^
```

Here is the full invocation:

```
clang++ -Idependencies/fmtlib.fmt/fmt-8.0.1/include -MMD -MP -m64 -Werror -ffast-math -O0 -g -mavx -Wall -Wextra -std=c++20 -fno-exceptions -fno-rtti -mcx16 -fvisibility=hidden -fvisibility-inlines-hidden -Wundef -o $(OBJDIR)/format.o -c dependencies/fmtlib.fmt/fmt-8.0.1/src/format.cc
```

And my full clang version if it helps:

```
$ clang --version
Debian clang version 13.0.0-++20210811062604+6789c4564a4b-1~exp1~20210811163414.34
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

vitaut:
clang 13 hasn't been released yet. Do you mean the current development version of clang? Also do you have a godbolt repro?

englercj:
[...]

PR #2525, opened 2021-09-30
Made basic_string_view's const Char* constructor constexpr in C++17.

mwinterb:
Also added commentary for why std::strlen is directly called in certain situations.
Addresses https://github.com/fmtlib/fmt/issues/2455#issuecomment-931346096.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -422,11 +422,17 @@ template <typename Char> class basic_string_view {
    */
   FMT_CONSTEXPR_CHAR_TRAITS
   FMT_INLINE
-  basic_string_view(const Char* s) : data_(s) {
+  basic_string_view(const Char* s) : data_(s), size_(0) {
+    // This test is needed to ensure that this constructor is constexpr in C++17
+    // modes.
+#ifdef __cpp_lib_is_constant_evaluated
     if (detail::const_check(std::is_same<Char, char>::value &&
-                            !detail::is_constant_evaluated()))
+                            !detail::is_constant_evaluated())) {
+      // Call strlen directly for better code generation in non-optimized
+      // builds.
       size_ = std::strlen(reinterpret_cast<const char*>(s));
-    else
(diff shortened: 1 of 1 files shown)


include/fmt/core.h
```diff
@@ -422,11 +422,17 @@ template <typename Char> class basic_string_view {
    */
   FMT_CONSTEXPR_CHAR_TRAITS
   FMT_INLINE
-  basic_string_view(const Char* s) : data_(s) {
+  basic_string_view(const Char* s) : data_(s), size_(0) {
+    // This test is needed to ensure that this constructor is constexpr in C++17
+    // modes.
+#ifdef __cpp_lib_is_constant_evaluated
```
phprus, line 428: ``__cpp_lib_is_constant_evaluated`` is defined only in C++20. Therefore, code generation optimizations will not work until C++20.

vitaut:
Thanks for the PR but I'd rather not introduce more conditional compilation there. Here's an alternative fix (still need to confirm once we get a repro): https://github.com/fmtlib/fmt/commit/023c2018f78a66c9acb3b9c7522ec431a210eb8c.

mwinterb:
One thing to note is that `constexpr` `char_traits` comes with C++17, and both GCC and MSVC demand initialization of `size_` in the member initializer list in their C++17 modes: https://gcc.godbolt.org/z/s8874qYWe. Since the original report is calling that constructor by a `consteval` function, obviously C++20 constexpr rules apply, so it may not matter at all for that specific repro. If it helps figure out if that constructor still needs to be constexpr in C++17, 664dd88e3175b92cdac96cb811dab94b6f3bdc62 is when that constructor was first made constexpr.

https://gcc.godbolt.org/z/rPbTP433f for a further reduced version to illustrate that both want the initialization in the initializer list, the fact that the initialization is "conditional" in `basic_string_view` isn't the problem.

PR #2529, opened 2021-10-02
Move size_ initialization to initializer list

phprus:
Fix for https://github.com/fmtlib/fmt/pull/2525#issuecomment-932793298 , https://gcc.godbolt.org/z/WPjG8fj4E


vitaut:
Thank you!

PR #2549, opened 2021-10-15
Extending fmt::join to support C++20-only ranges.

brevzin:
The C++20 Ranges model means that you can have a type which is a C++20 range but not a C++17 range (an example of such is `views::join` on a range of prvalue ranges, or `views::iota` if the underlying type doesn't satisfy `incrementable`). 

We just have to use `std::iter_value_t<It>` instead of going through `std::iterator_traits` directly (bonus, it's shorter, and no `typename`) and we have to not use `*it++` and instead manually do `*it` and then `++it`, since C++20 input iterators are allowed to have `it++` be `void`.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -2749,11 +2749,11 @@ template <typename It, typename Sentinel, typename Char>
 struct formatter<join_view<It, Sentinel, Char>, Char> {
  private:
   using value_type =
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR. Looks good but please apply clang-format.

PR #2544, opened 2021-10-13
Replacing strftime with internal implementation

phprus:
Proposal: https://github.com/fmtlib/fmt/pull/2500#issuecomment-920078450
Discussion: Issue #2541 

All locale independent formats are implemented native with significant performance increase (С++11: 2-4 times, C++17: up to 13 times).
For locale dependent formats, strftime is used.

Draft because:

- [x] Need check ISO week/year algorithm.

#### Performance regression on Windows, because ``ucrt`` ``strftime`` is slow by design.

Possible solution for Windows regression: split processing into 2 way: fast path and slow path.
Slow path: unrecognized format string (non-standard extensions) or locale-dependent (except  "C" locale?) formats in the format string.
Else: Fast path.
Pass the slow path formatting string to ``strftime`` or ``std::time_put``. Fast path - internal implementation.

PR #2550, opened 2021-10-17
Replacing strftime with std::time_put

phprus:
Continuation of PR #2544.

~~Draft, because I haven't yet tested the PR in all my environments.~~

@vitaut, what do you think about using the built-in month and weekday names in the case of the ``std::locale::classic()`` locale (without ``std::time_put`` calls)?

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -1393,11 +1393,38 @@ struct tm_format_checker : null_chrono_spec_handler<tm_format_checker> {
   FMT_CONSTEXPR void on_tz_name() {}
 };
 
+inline const char* tm_wday_full_name(int wday) {
+  static constexpr const char* full_name_list[] = {
+      "Sunday",   "Monday", "Tuesday", "Wednesday",
+      "Thursday", "Friday", "Saturday"};
+  return wday >= 0 && wday <= 6 ? full_name_list[wday] : "?";
+}
+inline const char* tm_wday_short_name(int wday) {
+  static constexpr const char* short_name_list[] = {"Sun", "Mon", "Tue", "Wed",
+                                                    "Thu", "Fri", "Sat"};
+  return wday >= 0 && wday <= 6 ? short_name_list[wday] : "???";
+}
+
+inline const char* tm_mon_full_name(int mon) {
+  static constexpr const char* full_name_list[] = {
+      "January", "February", "March",     "April",   "May",      "June",
+      "July",    "August",   "September", "October", "November", "December"};
+  return mon >= 0 && mon <= 11 ? full_name_list[mon] : "?";
+}
+inline const char* tm_mon_short_name(int mon) {
+  static constexpr const char* short_name_list[] = {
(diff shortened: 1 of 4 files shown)


vitaut:
> what do you think about using the built-in month and weekday names in the case of the std::locale::classic() locale (without std::time_put calls)?

We should definitely use built-in names in this case.

include/fmt/chrono.h
```diff
@@ -1561,17 +1575,51 @@ template <typename OutputIt, typename Char> class tm_writer {
...
+  }
 
   void on_datetime(numeric_system ns) {
-    format_localized('c', ns == numeric_system::standard ? '\0' : 'E');
+    if (is_classic_) {
+#ifdef _WIN32
+      on_us_date();
+      *out_++ = ' ';
+      on_iso_time();
+#else
```
vitaut, line 1597: The locale-independent and C locale format should be consistent among platforms so let's remove this.

include/fmt/chrono.h
```diff
@@ -1685,7 +1733,22 @@ template <typename OutputIt, typename Char> class tm_writer {
     write2(tm_sec());
   }
 
-  void on_12_hour_time() { format_localized('r'); }
+  void on_12_hour_time() {
+    if (is_classic_) {
+#ifdef _WIN32
+      on_iso_time();
+#else
```
vitaut, line 1740: Let's remove this as well.

vitaut (changes requested):
Looks great, just minor comments left.

include/fmt/chrono.h
phprus: Done. Corresponding tests for Windows are disabled (https://github.com/fmtlib/fmt/runs/4055476293?check_suite_focus=true)

test/chrono-test.cc
```diff
@@ -245,13 +248,18 @@ TEST(chrono_test, time_point) {
...
-      "%r",  "%R",  "%T",  "%p",  "%z",  "%Z"};
+      "%%",  "%n",  "%t",  "%Y",  "%EY", "%y",  "%Oy", "%Ey", "%C",
+      "%EC", "%G",  "%g",  "%b",  "%h",  "%B",  "%m",  "%Om", "%U",
+      "%OU", "%W",  "%OW", "%V",  "%OV", "%j",  "%d",  "%Od", "%e",
+      "%Oe", "%a",  "%A",  "%w",  "%Ow", "%u",  "%Ou", "%H",  "%OH",
+      "%I",  "%OI", "%M",  "%OM", "%S",  "%OS", "%x",  "%Ex", "%X",
+      "%EX", "%D",  "%F",  "%R",  "%T",  "%p",  "%z",  "%Z"};
   spec_list.push_back("%Y-%m-%d %H:%M:%S");
+#ifndef _WIN32
+  // Disabled on Windows, because this formats is not consistent among platforms
```
vitaut, line 259: this -> these and please add punctuation (.) at the end of the sentence.

vitaut:
Thank you!

PR #2551, opened 2021-10-17
Fix overflow for very big years (>2*10^9)

phprus:
Information: https://github.com/fmtlib/fmt/pull/2544#issuecomment-945140163


phprus:
I don't know how to write a test. The algorithm requires the correct values of ``tm_yday`` and ``tm_wday``, and they calculated is not be simple for the distant future.
I'm not sure if the ``strftime`` function for such arguments will return the correct result.

PR #2555, opened 2021-10-20
Do not detect LLVM based IBMXL compiler (on ppc) as clang

akohlmey:
This change avoids compilation problems on IBM PowerPC based Linux boxes using IBM's "native" compilers which are LLVM based but not fully clang compatible.


vitaut:
Thank you

PR #2573, opened 2021-10-28
FMT_USE_FCNTL can be predefined

timkalu:
'os.h' accepts a predefined FMT_USE_FCNTL override so using FMT with e.g. the NXP toolchain for ARM (e.g. for FreeRTOS) does not have a fcntl() call, but the detection routine does not detect this correctly.


vitaut:
Thank you

PR #2575, opened 2021-10-30
Fix incompatible between docutils 1.18.0 and sphinx 3.3.0

phprus:
See Sphinx issue: https://github.com/sphinx-doc/sphinx/issues/9777


vitaut:
Thanks for the fix!

PR #2576, opened 2021-10-31
Sped up `chrono.h` formatting for cases without providing locale

toughengineer:
Sped up `chrono.h` formatting (compared to the current trunk) in cases without providing locale by the amount roughly equal to three quarters of the time of
```c++
fmt::format_to(buf, FMT_COMPILE("{:%Y}"), t);
```
i.e. that one is sped up roughly four times.

For more elaborate format specifications and for non-compile-time cases the speedup is proportionally smaller, i.e. the speedup is not proportional to spec contents but rather is kinda a fixed amount.

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -309,12 +309,17 @@ auto write(OutputIt out, const std::tm& time, const std::locale& loc,
   return std::copy(str.begin(), str.end(), out);
 }
 
+inline const std::locale& get_classic_locale() {
+  static const auto& locale = std::locale::classic();
+  return locale;
+}
+
 template <typename Char, typename OutputIt,
           FMT_ENABLE_IF(std::is_same<Char, char>::value)>
 auto write(OutputIt out, const std::tm& time, const std::locale& loc,
            char format, char modifier = 0) -> OutputIt {
   auto str = do_write<char>(time, loc, format, modifier);
-  if (detail::is_utf8() && loc != std::locale::classic()) {
+  if (detail::is_utf8() && loc != get_classic_locale()) {
     // char16_t and char32_t codecvts are broken in MSVC (linkage errors) and
     // gcc-4.
 #if FMT_MSC_VER != 0 || \
@@ -1057,9 +1062,14 @@ struct chrono_formatter {
 
   void format_localized(const tm& time, char format, char modifier = 0) {
     if (isnan(val)) return write_nan();
-    const auto& loc = localized ? context.locale().template get<std::locale>()
-                                : std::locale::classic();
-    out = detail::write<char_type>(out, time, loc, format, modifier);
+    if (localized) {
+      out = detail::write<char_type>(
+          out, time, context.locale().template get<std::locale>(), format,
+          modifier);
+    } else {
+      out = detail::write<char_type>(out, time, get_classic_locale(), format,
+                                     modifier);
+    }
   }
 
   void on_text(const char_type* begin, const char_type* end) {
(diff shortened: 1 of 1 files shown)


include/fmt/chrono.h
```diff
@@ -1057,9 +1062,14 @@ struct chrono_formatter {
...
-                                : std::locale::classic();
-    out = detail::write<char_type>(out, time, loc, format, modifier);
+    if (localized) {
+      out = detail::write<char_type>(
+          out, time, context.locale().template get<std::locale>(), format,
+          modifier);
+    } else {
+      out = detail::write<char_type>(out, time, get_classic_locale(), format,
+                                     modifier);
+    }
```
vitaut, line 1072: Here and below: let's keep the conditional expression to avoid repetition.

include/fmt/chrono.h
```diff
@@ -1822,6 +1834,18 @@ template <typename Char> struct formatter<std::tm, Char> {
     return end;
   }
 
+  template <typename It>
+  It do_format(It out, const std::tm& tm, const std::locale& loc) const {
```
vitaut, line 1838: Why introduce this function?

vitaut (changes requested):
Thanks for the PR! Looks good, just minor comments/questions inline.

include/fmt/chrono.h
toughengineer: What do you mean by "conditional expression"?

include/fmt/chrono.h
toughengineer: To avoid repetion in `foram()` method.

include/fmt/chrono.h
vitaut: Yes, ternary.

include/fmt/chrono.h
vitaut: Keeping ternary should avoid the repetition and introducing a new helper funtion.

include/fmt/chrono.h
toughengineer: Currently it looks like this in the trunk:
```c++
const auto& loc = localized ? context.locale().template get<std::locale>()
                            : std::locale::classic();
```
-----
`context.locale().get()` returns locale object by value so this expression results in a temporary object.

Since one of the subexpressions results in a temporary object, the whole conditional operator
```c++
localized ? context.locale().template get<std::locale>() : std::locale::classic()
```
results in a temporary object. The resulting temporary is then bound to a `const` reference.

When the condition is false (e.g. `localized==false` or locale reference is empty and therefore `bool{context.locale()}==false`) the last subexpression is evaluated: `std::locale::classic()` in this case.

Since the conditional operator results in a temporary, the locale object `classic()` returns reference to is copy constructed into a temporary object, and then destructed at the end of the scope. Copy construction and destruction of locale objects are relatively expensive operations because they need to increment and decrement a bunch of reference counters of locale facets. This is significant compared to work done for simple format specifications like `%Y`.

I changed it so these expressions are evaluated in separate scopes. Depending on the condition, `context.locale().get()` results in a temporary which is bound to a function parameter, and a reference returned by `std::locale::classic()` is passed directly as function parameter without introducing a temporary.

Please let me know if you want me to back all of this up by quotes from and/or links to the C++ standard and/or cppreference.

vitaut: Makes sense, thanks for the detailed explanation. One thing that concerns me is that this change can be easily undone by accident with a seemingly innocent refactor. I suggest moving this optimized locale handling logic into a separate function:

```c++
template <typename FormatContext, typename F>
void with_locale(bool localized, FormatContext& ctx, F f) {
  if (localized) f(ctx.locale().template get<std::locale>());
  else f(get_classic_locale());
}
```

and adding a short comment explaining why it is needed.

Then the usage can be something like

```c++
with_locale(localized, context, [&](std::locale&) {
  out = detail::write<char_type>(out, time, locale, format, modifier);
});
```

toughengineer: One other variant can be to lazily construct the requested locale within the context object and return `const` reference to it from `context.locale()`.

Anyway I think such refactoring is out of the scope of this PR.

I can do something like you suggested in the next PR.

vitaut (approved):

include/fmt/chrono.h
vitaut: > I can do something like you suggested in the next PR.

Please do. Thanks!

Issue #128, opened 2015-03-09
Consider migrating from ReadTheDocs to GitHub pages


vitaut:
Comparison of ReadTheDocs and GitHub pages:

1. Time to load similar pages

  * GitHub: 0.70 - 1.57s
  * ReadTheDocs: 2.46 - 2.75s

2. Compatiblity

  * ReadTheDocs interferes with Sphinx builds breaking some of the features (possibly due to their customizations).
  * GitHub Pages don't interfere with Sphinx build process

3. Reliability

  * ReadTheDocs builds can remain broken for days without a way to cancel them.
  * Didn't notice any issues with GitHub pages so far.

4. Versioning

  * ReadTheDocs automates this, but versioning doesn't integrate well with the page theme except for RTD theme. However the latter has major usability issues.
  * With GitHub pages this will have to be done manually, but can be better integrated into the theme

vitaut:
New version of the documentation on GitHub Pages: http://cppformat.github.io/dev/

Need to migrate old docs and setup redirection.

vitaut:
Also there are some artifacts when building on Travis:

```
Increases the buffer capacity to hold at least *size* elements updating ``ptr_`` and ``capacity_``.
```

https://github.com/cppformat/cppformat.github.io/commit/c4caeb7773e945f4eb7c64306fc76136a36fb92c

vitaut:
Migrated the docs from RTD to GitHub pages.

* Latest release: http://cppformat.github.io/latest/
* Development version: http://cppformat.github.io/dev/

PR #2582, opened 2021-11-04
Cleanup warnings with nvhpc/21.9.

olupton:
This squashes a couple of warnings when building fmt 8.0.1 with version 21.9 of the NVIDIA HPC compiler.

```
"{path}/fmt/include/fmt/core.h", line 313: warning #1676-D: unrecognized GCC pragma
  FMT_GCC_PRAGMA("GCC optimize(\"Og\")")
  ^

"{path}/fmt/include/fmt/format.h", line 948: warning #185-D: dynamic initialization in unreachable code
    int num_digits = 0;
        ^
          detected during:
            instantiation of "auto fmt::v8::detail::count_digits<BITS,UInt>(UInt)->int [with BITS=4, UInt=uint32_t]" at line 1499
            instantiation of "auto fmt::v8::detail::write_int(OutputIt, fmt::v8::detail::write_int_arg<T>, const fmt::v8::basic_format_specs<Char> &, fmt::v8::detail::locale_ref)->OutputIt [with Char=char
, OutputIt=fmt::v8::detail::buffer_appender<char>, T=uint32_t]" at line 1540
            instantiation of "auto fmt::v8::detail::write(OutputIt, T, const fmt::v8::basic_format_specs<Char> &, fmt::v8::detail::locale_ref)->OutputIt [with Char=char, OutputIt=fmt::v8::detail::buffer_a
ppender<char>, T=int, <unnamed>=0]" at line 1912
            instantiation of "auto fmt::v8::detail::write(OutputIt, T, const fmt::v8::basic_format_specs<Char> &, fmt::v8::detail::locale_ref)->OutputIt [with Char=char, OutputIt=fmt::v8::detail::buffer_a
ppender<char>, T=bool, <unnamed>=0]" at line 1969
            instantiation of "auto fmt::v8::detail::default_arg_formatter<Char>::operator()(T)->fmt::v8::detail::default_arg_formatter<Char>::iterator [with Char=char, T=bool]" at line 1465 of "{path}/fmt/include/fmt/core.h"
            instantiation of "auto fmt::v8::visit_format_arg(Visitor &&, const fmt::v8::basic_format_arg<Context> &)->decltype((<expression>)) [with Visitor=fmt::v8::detail::default_arg_formatter<char>, C
ontext=fmt::v8::format_context]" at line 2649
            instantiation of "void fmt::v8::detail::vformat_to(fmt::v8::detail::buffer<Char> &, fmt::v8::basic_string_view<Char>, fmt::v8::basic_format_args<fmt::v8::basic_format_context<fmt::v8::detail::
buffer_appender<fmt::v8::type_identity_t<Char>>, fmt::v8::type_identity_t<Char>>>, fmt::v8::detail::locale_ref) [with Char=char]" at line 2895 of "{path}/fmt/include/fmt/core.h"
            instantiation of "auto fmt::v8::vformat_to(OutputIt, fmt::v8::string_view, fmt::v8::format_args)->OutputIt [with OutputIt=std::back_insert_iterator<spdlog::memory_buf_t>, <unnamed>=0]" at line
 2915 of "{path}/fmt/include/fmt/core.h"
            instantiation of "auto fmt::v8::format_to(OutputIt, fmt::v8::format_string<T...>, T &&...)->OutputIt [with OutputIt=std::back_insert_iterator<spdlog::memory_buf_t>, T=<int &>, <unnamed>=0]" at
 line 58 of "{path}/spdlog/include/spdlog/details/fmt_helper.h"
```

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -279,7 +279,7 @@
 
 // Enable minimal optimizations for more compact code in debug mode.
 FMT_GCC_PRAGMA("GCC push_options")
-#ifndef __OPTIMIZE__
+#if !defined(__OPTIMIZE__) && !defined(__NVCOMPILER)
 FMT_GCC_PRAGMA("GCC optimize(\"Og\")")
 #endif
 
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -18,16 +18,31 @@
 // The fmt library version in the form major * 10000 + minor * 100 + patch.
 #define FMT_VERSION 80001
 
+#if defined(__NVCOMPILER)
+#  define FMT_NVHPC_VERSION                                      \
+    (__NVCOMPILER_MAJOR__ * 10000 + __NVCOMPILER_MINOR__ * 100 + \
+     __NVCOMPILER_PATCHLEVEL__)
+#else
+#  define FMT_NVHPC_VERSION 0
+#endif
+
 #if defined (__clang__ ) && !defined(__ibmxl__)
 #  define FMT_CLANG_VERSION (__clang_major__ * 100 + __clang_minor__)
 #else
 #  define FMT_CLANG_VERSION 0
 #endif
 
-#if defined(__GNUC__) && !defined(__clang__) && !defined(__INTEL_COMPILER) && !defined(__NVCOMPILER)
+#if defined(__GNUC__) && !defined(__clang__) && !defined(__INTEL_COMPILER) && \
+    !defined(__NVCOMPILER)
 #  define FMT_GCC_VERSION (__GNUC__ * 100 + __GNUC_MINOR__)
 #else
 #  define FMT_GCC_VERSION 0
+#endif
+
+#if FMT_NVHPC_VERSION > 0
+#  define FMT_NVHPC_PRAGMA(arg) _Pragma(arg)
+#else
+#  define FMT_NVHPC_PRAGMA(arg)
 #endif
 
 #ifndef FMT_GCC_PRAGMA
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -24,7 +24,7 @@
 #  define FMT_CLANG_VERSION 0
 #endif
 
-#if defined(__GNUC__) && !defined(__clang__) && !defined(__INTEL_COMPILER)
+#if defined(__GNUC__) && !defined(__clang__) && !defined(__INTEL_COMPILER) && !defined(__NVCOMPILER)
 #  define FMT_GCC_VERSION (__GNUC__ * 100 + __GNUC_MINOR__)
 #else
 #  define FMT_GCC_VERSION 0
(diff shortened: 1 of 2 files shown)


include/fmt/core.h
```diff
@@ -279,7 +279,7 @@
 
 // Enable minimal optimizations for more compact code in debug mode.
 FMT_GCC_PRAGMA("GCC push_options")
-#ifndef __OPTIMIZE__
+#if !defined(__OPTIMIZE__) && !defined(__NVCOMPILER)
```
vitaut, line 282: This check should probably be moved here: https://github.com/fmtlib/fmt/blob/812733cc963b2e1d96f6ad2cb2d441c6fe8e4a5b/include/fmt/core.h#L35

include/fmt/format.h
```diff
@@ -967,14 +967,17 @@ FMT_CONSTEXPR20 inline auto count_digits(uint64_t n) -> int {
 template <int BITS, typename UInt>
 FMT_CONSTEXPR auto count_digits(UInt n) -> int {
 #ifdef FMT_BUILTIN_CLZ
-  if (num_bits<UInt>() == 32)
+  if (num_bits<UInt>() == 32) {
```
vitaut, line 970: Let's wrap the condition in `const_check` instead.

vitaut (changes requested):
Thanks for the PR.

include/fmt/core.h
olupton: I moved it so `FMT_GCC_VERSION` is 0 for NVHPC.

include/fmt/format.h
olupton:
```c++
template <int BITS, typename UInt>
FMT_CONSTEXPR auto count_digits(UInt n) -> int {
#ifdef FMT_BUILTIN_CLZ
  if (const_check(num_bits<UInt>() == 32))
    return (FMT_BUILTIN_CLZ(static_cast<uint32_t>(n) | 1) ^ 31) / BITS + 1;
#endif
  int num_digits = 0;
```
doesn't squash the warning.

vitaut: That's unfortunate but I think the suppression with extra blocks is too messy. Is there another way to do this?

olupton: eafdd3eb is _another_ way -- I wouldn't say it's a particularly **nice** way...

vitaut: Yeah, this is clearly worse =(. Does moving the definition of `num_digits` to the beginning of the function suppress the warning?

```c++
  int num_digits = 0;
#ifdef FMT_BUILTIN_CLZ
  if (const_check(num_bits<UInt>() == 32))
    return (FMT_BUILTIN_CLZ(static_cast<uint32_t>(n) | 1) ^ 31) / BITS + 1;
#endif
  ```

Comment (olupton) [reply include/fmt/format.h]: Nice idea, but that just gives
```
warning #128-D: loop is not reachable
```
on the `do {` line. And I would be a little worried about other compilers giving an unused variable warning too...

I pushed yet another attempt, with an immediately executed lambda, which seems to work...

vitaut: This looks better but please fix the CI issues.

vitaut (approved):

vitaut:
Thanks

PR #2586, opened 2021-11-05
Reuse tm_writer in other formatters

phprus:
All locale-dependent formatting is now in the ``tm_writer`` class.

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -1412,6 +1412,7 @@ struct chrono_formatter {
   bool negative;
 
   using char_type = typename FormatContext::char_type;
+  using tm_writer_type = tm_writer<OutputIt, char_type>;
 
   chrono_formatter(FormatContext& ctx, OutputIt o,
                    std::chrono::duration<Rep, Period> d)
@@ -1493,10 +1494,13 @@ struct chrono_formatter {
   void write_pinf() { std::copy_n("inf", 3, out); }
   void write_ninf() { std::copy_n("-inf", 4, out); }
 
-  void format_localized(const tm& time, char format, char modifier = 0) {
+  template <typename Callback, typename... Args>
+  void format_tm(const tm& time, Callback cb, Args... args) {
     if (isnan(val)) return write_nan();
-    out = detail::write<char_type>(
-        out, time, get_locale(localized, context.locale()), format, modifier);
+    get_locale loc(localized, context.locale());
+    auto w = tm_writer_type(loc, out, time);
+    (w.*cb)(args...);
+    out = w.out();
   }
 
   void on_text(const char_type* begin, const char_type* end) {
@@ -1537,7 +1541,7 @@ struct chrono_formatter {
     if (ns == numeric_system::standard) return write(hour(), 2);
     auto time = tm();
     time.tm_hour = to_nonnegative_int(hour(), 24);
-    format_localized(time, 'H', 'O');
+    format_tm(time, &tm_writer_type::on_24_hour, ns);
   }
 
   void on_12_hour(numeric_system ns) {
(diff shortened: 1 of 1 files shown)


phprus:
Rebased

vitaut:
Thanks for another PR. It looks like the first commit only moves code around without changing functionality so could you move it to a separate PR to simplify reviewing this one?

phprus:
Yes. To use ``tm_writer``, it must be declared before.
PR #2591

phprus:
@vitaut , PR updated.

include/fmt/chrono.h
```diff
@@ -303,22 +309,17 @@ inline auto do_write(const std::tm& time, const std::locale& loc, char format,
 
 template <typename Char, typename OutputIt,
           FMT_ENABLE_IF(!std::is_same<Char, char>::value)>
-auto write(OutputIt out, const std::tm& time, const std::locale& loc,
-           char format, char modifier = 0) -> OutputIt {
-  auto str = do_write<Char>(time, loc, format, modifier);
+auto write_localized(OutputIt out, const std::tm& time, const std::locale& loc,
```
vitaut, line 312: Let's not rename this function because the "localized" part is clear from the signature (this function takes a locale).

include/fmt/chrono.h
```diff
@@ -1473,16 +1475,21 @@ struct chrono_formatter {
   void write_pinf() { std::copy_n("inf", 3, out); }
   void write_ninf() { std::copy_n("-inf", 4, out); }
 
-  void format_localized(const tm& time, char format, char modifier = 0) {
+  template <class Callback, class... Args>
+  void do_format_tm(const std::locale& loc, const tm& time, Callback cb,
```
vitaut, line 1479: Let's make this a lambda and merge into `format_tm` where it's used.

include/fmt/chrono.h
```diff
@@ -1634,6 +1641,14 @@ template <typename Char> struct formatter<weekday, Char> {
  private:
   bool localized = false;
 
+  template <typename It>
+  auto do_format(It out, const std::tm& time, const std::locale& loc) const
```
vitaut, line 1645: Same here

vitaut (changes requested):

include/fmt/chrono.h
phprus: Done. Based on #2592

include/fmt/chrono.h
phprus: Done. Based on #2592

phprus:
Waiting for #2594 ...

phprus:
@vitaut, PR updated.

include/fmt/chrono.h
```diff
@@ -289,6 +289,11 @@ inline null<> localtime_s(...) { return null<>(); }
 inline null<> gmtime_r(...) { return null<>(); }
 inline null<> gmtime_s(...) { return null<>(); }
 
+inline const std::locale& get_classic_locale() {
+  static const auto& locale = std::locale::classic();
+  return locale;
+}
+
```
vitaut, line 296: I don't think you need to move this function.

phprus:
@vitaut, PR updated.

include/fmt/chrono.h
```diff
@@ -1493,10 +1494,13 @@ struct chrono_formatter {
   void write_pinf() { std::copy_n("inf", 3, out); }
   void write_ninf() { std::copy_n("-inf", 4, out); }
 
-  void format_localized(const tm& time, char format, char modifier = 0) {
+  template <class Callback, class... Args>
```
vitaut, line 1497: class -> typename for consistency with the rest of the library

include/fmt/chrono.h
```diff
@@ -1493,10 +1494,13 @@ struct chrono_formatter {
...
+  void format_tm(const tm& time, Callback cb, Args... args) {
     if (isnan(val)) return write_nan();
-    out = detail::write<char_type>(
-        out, time, get_locale(localized, context.locale()), format, modifier);
+    get_locale loc(localized, context.locale());
+    auto w = tm_writer_type(loc, out, time);
+    (w.*cb)(std::forward<Args>(args)...);
```
vitaut, line 1502: `forward` is not needed because the argument is passed by value, please remove.

vitaut (changes requested):

include/fmt/chrono.h

vitaut (approved):

vitaut:
Merged, thanks!

PR #2592, opened 2021-11-07
Changed locale retrieval to minimize the risk of unwanted accidental refactoring

toughengineer:
This PR is in continuation of this discussion: https://github.com/fmtlib/fmt/pull/2576#discussion_r744275705

This seems to be a tad slower, probably because the lamda is not inlined.

Though I tried to make it pretty, I personally think this is very ugly.

-----
As a side note: `on_iso_date()` seems to be ~4x slower than `on_iso_time()` though the former does not seem to do 4x more work, somebody should look into it.

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -961,6 +961,16 @@ OutputIt format_duration_unit(OutputIt out) {
   return out;
 }
 
+template <typename F>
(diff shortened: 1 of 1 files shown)


toughengineer:
I like #2594 much better than this one.

PR #2594, opened 2021-11-08
Changed locale retrieval way to a fancy one

toughengineer:
This is a quick and dirty one, but I like it much, much better than #2592.
I hope I haven't screwed up anything.

And this one is **not** slowed down compared to the ugly one with lamdas and whatnot.

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -1374,6 +1374,27 @@ OutputIt format_duration_unit(OutputIt out) {
   return out;
 }
 
+struct get_locale {
+  explicit get_locale(bool localized, locale_ref loc_ref)
+      : has_locale{localized} {
+    if (localized)
+      ::new (&locale) std::locale{loc_ref.template get<std::locale>()};
+  }
+  ~get_locale() {
+    if (has_locale) locale.~locale();
+  }
+  operator const std::locale&() const {
+    return has_locale ? locale : get_classic_locale();
+  }
+
+ private:
+  union {
+    monostate dummy{};
+    std::locale locale;
+  };
+  bool has_locale = false;
+};
+
 template <typename FormatContext, typename OutputIt, typename Rep,
           typename Period>
 struct chrono_formatter {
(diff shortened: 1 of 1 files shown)


include/fmt/chrono.h
```diff
@@ -961,6 +961,26 @@ OutputIt format_duration_unit(OutputIt out) {
...
+  ~get_locale() {
+    if (has_locale) locale.~locale();
+  }
+  operator const std::locale&() const {
+    return has_locale ? locale : get_classic_locale();
+  }
+
+ private:
+  union {
+    monostate dummy{};
```
phprus, line 978: ``monostate``  is C++17 only

toughengineer:
Yep, I screwed up, but now it seems to be fixed.

phprus:
I suggested PR #2586, which moves all locale-specific formatters into the ``tm_writer`` class.
I have an idea how to simplify the locale-specific selectors (this PR). I can implement it on top of the PR #2586.

phprus:
No, my idea is perf slower :(

vitaut:
LGTM but please rebase.

include/fmt/chrono.h
vitaut: `{}` seems unnecessary, please remove.

toughengineer: Strictly speaking this is unnecessary.

But without `monospace` default initialized, this is kinda scary because you have a generally unintialized member, and invariants become kinda hard to follow.

All this is OK for experts, but I imagine that this is harder to follow for non-experts.
`monostate` has negligible overhead, but helps to figure out what's happening more easily.
That's why I added `monostate` member.

vitaut: `monostate` is fine but I don't think you need to initialize it explicitly at least because it has no state =).

toughengineer: If you never initialize `monostate dummy`, than there is no point in having it at all.

IMHO, initializing `monostate dummy` by default gives some understanding and peace of mind to the person trying to understand the code.

Initializing it by default says that the lifetime of the `dummy` object is started, and you need to explicitly do something to change it.

In this case when `locale` is initialized using placement `new`, it explicitly says that the lifetime of `dummy` ends and another object starts to occupy the storage simultaneously starting its lifetime.

If there was no `monostate dummy` it would work perfectly correct, but then a person trying to understand the code must keep in mind that the union is generally _uninitialized_.

vitaut:
Merged, thanks!

PR #2602, opened 2021-11-14
Optimize tm formatting

phprus:
Optimize tm formatting for:

- C-locale and alternative representation.
- ``%z`` format (timezone offset from UTC).

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -13,6 +13,7 @@
 #include <ctime>
 #include <locale>
 #include <sstream>
+#include <type_traits>
 
 #include "format.h"
 
@@ -871,6 +872,22 @@ inline const char* tm_mon_short_name(int mon) {
   return mon >= 0 && mon <= 11 ? short_name_list[mon] : "???";
 }
 
+template <typename T, typename = void>
+struct has_member_data_tm_gmtoff : std::false_type {};
+template <typename T>
+struct has_member_data_tm_gmtoff<T, void_t<decltype(T::tm_gmtoff)>>
+    : std::true_type {};
+
+#if defined(_WIN32)
+inline void tzset_once() {
+  static bool init = []() -> bool {
+    _tzset();
+    return true;
+  }();
+  ignore_unused(init);
+}
+#endif
+
 template <typename OutputIt, typename Char> class tm_writer {
  private:
   static constexpr int days_per_week = 7;
@@ -988,6 +1005,36 @@ template <typename OutputIt, typename Char> class tm_writer {
     }
   }
 
+  void write_utc_offset(long offset) {
+    if (offset < 0) {
+      *out_++ = '-';
+      offset = -offset;
+    } else {
+      *out_++ = '+';
+    }
+    offset /= 60;
+    write2(static_cast<int>(offset / 60));
+    write2(static_cast<int>(offset % 60));
+  }
+  void format_utc_offset_impl(std::true_type) {
+    write_utc_offset(tm_.tm_gmtoff);
+  }
+  void format_utc_offset_impl(std::false_type) {
+#if defined(_WIN32)
+    tzset_once();
+    long offset = 0;
+    _get_timezone(&offset);
+    if (tm_.tm_isdst) {
+      long dstbias = 0;
+      _get_dstbias(&dstbias);
+      offset += dstbias;
+    }
+    write_utc_offset(-offset);
+#else
+    format_localized('z');
+#endif
+  }
+
   void format_localized(char format, char modifier = 0) {
     out_ = write<Char>(out_, tm_, loc_, format, modifier);
   }
@@ -1094,7 +1141,9 @@ template <typename OutputIt, typename Char> class tm_writer {
     out_ = copy_str<Char>(std::begin(buf) + offset, std::end(buf), out_);
   }
 
-  void on_utc_offset() { format_localized('z'); }
+  void on_utc_offset() {
+    format_utc_offset_impl(has_member_data_tm_gmtoff<std::tm>{});
+  }
   void on_tz_name() { format_localized('Z'); }
 
   void on_year(numeric_system ns) {
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -1018,13 +1018,16 @@ template <typename OutputIt, typename Char> class tm_writer {
       format_localized('A');
   }
   void on_dec0_weekday(numeric_system ns) {
-    if (ns != numeric_system::standard) return format_localized('w', 'O');
-    write1(tm_wday());
+    if (is_classic_ || ns == numeric_system::standard) return write1(tm_wday());
+    format_localized('w', 'O');
   }
   void on_dec1_weekday(numeric_system ns) {
-    if (ns != numeric_system::standard) return format_localized('u', 'O');
-    auto wday = tm_wday();
-    write1(wday == 0 ? days_per_week : wday);
+    if (is_classic_ || ns == numeric_system::standard) {
+      auto wday = tm_wday();
+      write1(wday == 0 ? days_per_week : wday);
+    } else {
+      format_localized('u', 'O');
+    }
   }
 
   void on_abbr_month() {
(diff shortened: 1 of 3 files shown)


vitaut:
Thanks for the PR. Could you provide any benchmark results showing the effects of this change?

include/fmt/chrono.h
```diff
@@ -1091,21 +1105,51 @@ template <typename OutputIt, typename Char> class tm_writer {
     out_ = copy_str<Char>(std::begin(buf) + offset, std::end(buf), out_);
   }
 
-  void on_utc_offset() { format_localized('z'); }
+  void on_utc_offset() {
+#if defined(_WIN32)
+    _tzset();
```
vitaut, line 1110: We shouldn't call _tzset or tzset because it changes global state.

phprus:
Benchmark: https://github.com/phprus/fmt-bench/tree/optimize-tm-formatting-5

*BSD, Apple, glibc: Acceleration 10-20 times (``%z``).
This is ``std::put_time`` and ostrstream overhead.

include/fmt/chrono.h
phprus: ``_tzset`` / ``tzset`` is used internally in ``std::time_put``.

Without calling them, it is impossible to get the timezone correctly (if the timezone is not stored in ``struct std::tm``).

Windows: https://docs.microsoft.com/en-us/cpp/c-runtime-library/daylight-dstbias-timezone-and-tzname?view=msvc-140 (Remarks section)
Solaris: https://docs.oracle.com/cd/E86824_01/html/E54766/ctime-3c.html

On Solaris and Windows tzset is threadsafe.

phprus:
cc @alexezeder

phprus:
Other formatters (``%EY``, ...) on all OS with C-locale: 10-20 times

alexezeder:
> cc @alexezeder

I'm not 100% sure why I was mentioned here, but here are the results on my machine. 😏

Currently, there is no option to run benchmarks for PRs in [fmt_bnchmrk_gnrtr](https://github.com/alexezeder/fmt_bnchmrk_gnrtr) (I doubt it would be implemented someday). But the only one runner there uses the same version of GCC - 11.

### GCC 11 + libstdc++ (C++20)

| Benchmark             | old time | new time  |
|-----------------------|----------|-----------|
| FMTFormatter_z        | 261 ns   | 26.3 ns   |
| FMTFormatterCompile_z | 246 ns   | 11.6 ns   |
| FMTFormatter_Y        | 263 ns   | 26.1 ns   |
| FMTFormatterCompile_Y | 244 ns   | 10.5 ns   |

<details>
<summary>Raw output</summary>

```
cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_STANDARD=20 ..
```

```
$ ./fmt_test_old
----------------------------------------------------------------
Benchmark                      Time             CPU   Iterations
----------------------------------------------------------------
FMTFormatter_z               261 ns          261 ns      2692446
FMTFormatterCompile_z        246 ns          246 ns      2851517
FMTFormatter_Y               263 ns          263 ns      2662867
FMTFormatterCompile_Y        244 ns          244 ns      2867342
```

```
$ ./fmt_test_new
----------------------------------------------------------------
Benchmark                      Time             CPU   Iterations
----------------------------------------------------------------
FMTFormatter_z              26.3 ns         26.3 ns     26136296
FMTFormatterCompile_z       11.6 ns         11.6 ns     61292230
FMTFormatter_Y              26.1 ns         26.1 ns     26920284
FMTFormatterCompile_Y       10.5 ns         10.5 ns     67130168
```
</details>


### Clang 13 + libc++ (C++20)

| Benchmark             | old time | new time |
|-----------------------|----------|----------|
| FMTFormatter_z        | 110 ns   | 30.6 ns  |
| FMTFormatterCompile_z | 90.0 ns  | 10.4 ns  |
| FMTFormatter_Y        | 110 ns   | 30.5 ns  |
| FMTFormatterCompile_Y | 88.4 ns  | 12.7 ns  |

<details>
<summary>Raw output</summary>

```
CC='clang' CXX='clang++' CXXFLAGS='-stdlib=libc++' cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_STANDARD=20 ..
```

```
$ ./fmt_test_old
----------------------------------------------------------------
Benchmark                      Time             CPU   Iterations
----------------------------------------------------------------
FMTFormatter_z               110 ns          110 ns      6403692
FMTFormatterCompile_z       90.0 ns         90.0 ns      7797814
FMTFormatter_Y               110 ns          110 ns      6353394
FMTFormatterCompile_Y       88.4 ns         88.4 ns      7909662
```

```
$ ./fmt_test_new
----------------------------------------------------------------
Benchmark                      Time             CPU   Iterations
----------------------------------------------------------------
FMTFormatter_z              30.6 ns         30.6 ns     23115238
FMTFormatterCompile_z       10.4 ns         10.4 ns     67346905
FMTFormatter_Y              30.5 ns         30.4 ns     22898049
FMTFormatterCompile_Y       12.7 ns         12.7 ns     55214092
```
</details>

phprus:
@alexezeder, I mentioned you since you wrote a comment about performance regression (https://github.com/fmtlib/fmt/commit/1031eedf27fe55fb7e85d7ce47dcbd3ecebb290e#commitcomment-60128058)

This PR improves performance in those cases (C-locale and timezone offset).

include/fmt/chrono.h
phprus: @vitaut, What do you think about this?

vitaut: I think we should limit system-specific logic here and suggest just detecting if `tm_gmtoff` is available and otherwise go through `format_localized` as before. The presence of `tm_gmtoff` can be detected in a portable way using SFINAE.

include/fmt/chrono.h
```diff
@@ -1091,21 +1105,51 @@ template <typename OutputIt, typename Char> class tm_writer {
...
+    format_localized('z');
+#endif
+  }
   void on_tz_name() { format_localized('Z'); }
 
   void on_year(numeric_system ns) {
-    if (ns != numeric_system::standard) return format_localized('Y', 'E');
+    if (!is_classic_ && ns != numeric_system::standard)
+      return format_localized('Y', 'E');
     write_year(tm_year());
```
vitaut, line 1136: Let's swap the branches to get rid of negation:

```c++
    if (is_classic_ || ns == numeric_system::standard)
      return write_year(tm_year());
    format_localized('Y', 'E');
```

and the same below.

test/chrono-test.cc
```diff
@@ -49,7 +49,14 @@ std::string system_strftime(const std::string& format, const std::tm* timeptr,
   os.imbue(loc);
   facet.put(os, os, ' ', timeptr, format.c_str(),
             format.c_str() + format.size());
+#ifdef _WIN32
+  // Workaround to early ucrt bug
```
vitaut, line 53: early?

vitaut (changes requested):

vitaut:
@phprus, thanks for the benchmark. Could you add `strftime` there for comparison?

test/chrono-test.cc
phprus: Early version.

I have not tested all versions. But in the latest version of the Windows SDK I have installed with MSVS, '%z' returns '+0000'. In the old version: '-0000'.

If remove platform-dependent code, this change will not be necessary.

include/fmt/chrono.h

phprus: ``tm_gmtoff`` and SFINAE - Done.
Solaris specific code is removed.
Windows specific code I suggest to include in fmt, because Windows is more popular.

phprus:
@vitaut, I will do a performance test (on Windows and with raw ``strftime``, ``std::time_put``) in the coming days.

include/fmt/chrono.h
phprus: I was wrong.
Inside ``ucrt``, the internal function __tzset () is used, which is called only once. When using _tzset (), performance is 10 times slower.

I've added a workaround, but I'm not sure if I need to switch to custom code.

Perf with workaround:

```
2021-11-22T00:37:43+05:00
Running fmt_test_old
Run on (2 X 2494 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x2)
  L1 Instruction 32 KiB (x2)
  L2 Unified 4096 KiB (x2)
  L3 Unified 16384 KiB (x1)
------------------------------------------------------------------
Benchmark                        Time             CPU   Iterations
------------------------------------------------------------------
FMTFormatter_z                1384 ns         1360 ns       448000
FMTFormatterCompile_z         1321 ns         1318 ns       497778
FMTFormatter_z_strftime        312 ns          314 ns      2240000
FMTFormatter_z_time_put       1217 ns         1221 ns       640000
FMTFormatter_Y                1385 ns         1381 ns       497778
FMTFormatterCompile_Y         1300 ns         1311 ns       560000
FMTFormatter_Y_strftime        301 ns          300 ns      2240000
FMTFormatter_Y_time_put       1214 ns         1221 ns       640000
2021-11-22T00:37:50+05:00
Running fmt_test_new
Run on (2 X 2494 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x2)
  L1 Instruction 32 KiB (x2)
  L2 Unified 4096 KiB (x2)
  L3 Unified 16384 KiB (x1)
------------------------------------------------------------------
Benchmark                        Time             CPU   Iterations
------------------------------------------------------------------
FMTFormatter_z                 123 ns          123 ns      5600000
FMTFormatterCompile_z         71.1 ns         69.8 ns      8960000
FMTFormatter_z_strftime        312 ns          314 ns      2240000
FMTFormatter_z_time_put       1214 ns         1221 ns       640000
FMTFormatter_Y                 121 ns          120 ns      5600000
FMTFormatterCompile_Y         67.9 ns         68.4 ns     11200000
FMTFormatter_Y_strftime        303 ns          305 ns      2357895
FMTFormatter_Y_time_put       1203 ns         1200 ns       560000
```

phprus:
@vitaut, benchmark updated: https://github.com/phprus/fmt-bench/tree/optimize-tm-formatting-5

include/fmt/chrono.h
```diff
@@ -871,6 +872,22 @@ inline const char* tm_mon_short_name(int mon) {
...
 }
 
+template <typename T, typename = void>
+struct has_member_data_tm_gmtoff : std::false_type {};
+template <typename T>
+struct has_member_data_tm_gmtoff<T, void_t<decltype(T::tm_gmtoff)>>
+    : std::true_type {};
+
+#if defined(_WIN32)
+void tzset_once() {
```
vitaut, line 882: This function should be inline because it's defined in a header.

include/fmt/chrono.h
```diff
@@ -871,6 +872,22 @@ inline const char* tm_mon_short_name(int mon) {
...
+  static bool init = ([]() -> bool {
+    _tzset();
+    return true;
+  })();
```
vitaut, line 886: Parentheses around the lambda are redundant, please remove.

test/chrono-test.cc
vitaut: Since we are keeping this, let's tweak it a bit for clarity:

```c++
// Workaround a bug in older versions of ucrt.
```

BTW is "ucrt" the correct term here? Is it Microsoft C runtime library? Maybe calling it Microsoft CRT here would be clearer?

vitaut (changes requested):

vitaut:
> benchmark updated

Awesome, thanks!

phprus:
Sorry for delay.

UCRT is are official name for Microsoft C Runtime start from Visual Studio 2015 (https://docs.microsoft.com/en-us/cpp/porting/upgrade-your-code-to-the-universal-crt?view=msvc-170)

I replaced the abbreviation UCRT to another version of the name: Universal CRT.

vitaut (approved):

vitaut:
Merged, thank you for another great PR!

PR #2604, opened 2021-11-19
Avoid MSVC 2013 workaround used in other compilers

ibmibmibm:
https://github.com/gabime/spdlog/pull/2180


vitaut:
Thanks for the PR but could you elaborate why you want to make it conditionally compiled?

vitaut:
Closing as enabling the workaround on other compilers is harmless.

PR #2606, opened 2021-11-19
Broken link in README.rst

ericcurtin:
Added src directory to link


vitaut:
Thank you

ericcurtin:
That you for creating such a neat library. I thought I'd never be converted from printf, etc. Next goal, beat puts performance 🙂

PR #2611, opened 2021-11-24
fix int -> uint warning

Acretock:
in my project i tried to mute warnings from fmt, but this one persist, should be harmless.

### DIFF
--- a/src/os.cc
+++ b/src/os.cc
@@ -218,7 +218,7 @@ int buffered_file::fileno() const {
 
 #if FMT_USE_FCNTL
 file::file(cstring_view path, int oflag) {
-  int mode = S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH;
+  unsigned int mode = S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH;
 #  if defined(_WIN32) && !defined(__MINGW32__)
   fd_ = -1;
   FMT_POSIX_CALL(sopen_s(&fd_, path.c_str(), oflag, _SH_DENYNO, mode));
--- a/src/os.cc
+++ b/src/os.cc
@@ -20,9 +20,6 @@
 
 #  ifndef _WIN32
 #    include <unistd.h>
-#    ifndef mode_t
-#      define mode_t unsigned short
-#    endif
 #  else
 #    ifndef WIN32_LEAN_AND_MEAN
 #      define WIN32_LEAN_AND_MEAN
@@ -44,12 +41,12 @@
 #    ifndef S_IROTH
 #      define S_IROTH 0
 #    endif
-#    ifndef mode_t
-#      define mode_t int
-#    endif
+
 
 #    ifdef __MINGW32__
 #      define _SH_DENYNO 0x40
+#    else
+       using mode_t = int;
 #    endif
 #  endif  // _WIN32
 #endif    // FMT_USE_FCNTL
@@ -367,8 +364,5 @@ long getpagesize() {
 FMT_API void ostream::grow(size_t) {
   if (this->size() == this->capacity()) flush();
 }
-#    ifdef mode_t
-#        undef mode_t
-#    endif
 #endif  // FMT_USE_FCNTL
 FMT_END_NAMESPACE
--- a/src/os.cc
+++ b/src/os.cc
@@ -21,7 +21,7 @@
 #  ifndef _WIN32
 #    include <unistd.h>
 #    ifndef mode_t
-#      define mode_t unsigned int
+#      define mode_t unsigned short
 #    endif
 #  else
 #    ifndef WIN32_LEAN_AND_MEAN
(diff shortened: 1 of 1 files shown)


vitaut:
Could you post the complete warning message and platform details?

Acretock:
> Could you post the complete warning message and platform details?

warnings, build with Android Studio
```
warning: implicit conversion loses integer precision: 'int' to 'mode_t' (aka 'unsigned short') [-Wimplicit-int-conversion]
warning: implicit conversion changes signedness: 'int' to 'mode_t' (aka 'unsigned int') [-Wsign-conversion]
```

vitaut (changes requested):

src/os.cc
```diff
@@ -218,7 +218,7 @@ int buffered_file::fileno() const {
 
 #if FMT_USE_FCNTL
 file::file(cstring_view path, int oflag) {
-  int mode = S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH;
+  unsigned int mode = S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH;
```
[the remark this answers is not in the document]
AlexGuteniev: Wouldn't warning disappear if `constexpr` is used, so that the compiler will have to substitute the value and not being afraid of narrowing. Then it could be `constexpr auto` I guess.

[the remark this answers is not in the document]
vitaut: I don't think `constexpr` is relevant here and `auto` will just result in `int`. Note that the warning is not on this line.

src/os.cc
```diff
@@ -41,6 +44,9 @@
 #    ifndef S_IROTH
 #      define S_IROTH 0
 #    endif
+#    ifndef mode_t
+#      define mode_t int
+#    endif
```
AlexGuteniev, line 49: This looks scary to me. What if `mode_t` is previously defined as a typedef, and you redefine it to a different type and never restore it.

AlexGuteniev: This is also scary if `mode_t` is actually a define.

Maybe have checks as
```
# if !defined(mode_t) && !defined(__mode_t_defined)
```

in both copies. I see the headers define `__mode_t_defined` macro when they typedef `mode_t`

src/os.cc
```diff
@@ -20,6 +20,9 @@
 
 #  ifndef _WIN32
 #    include <unistd.h>
+#    ifndef mode_t
+#      define mode_t unsigned short
+#    endif
```
phprus, line 25: The ``mode_t`` type is defined in the POSIX standard (https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_types.h.html).

Please, remove.

src/os.cc
phprus, line 49: For win32:
```c++
#ifndef __MINGW32__
using mode_t = int;
#endif
```

src/os.cc
```diff
@@ -361,5 +367,8 @@ long getpagesize() {
 FMT_API void ostream::grow(size_t) {
   if (this->size() == this->capacity()) flush();
 }
+#    ifdef mode_t
+#        undef mode_t
+#    endif
```
phprus, line 372: Please, remove.

src/os.cc
vitaut: If mode_t is a macro this will be just a warning, no need to do anything.

phprus:
@vitaut , @Acretock 

Mingw always defines the ``_mode_t`` type. And in addition, the ``mode_t`` type is defined (if the ``NO_OLDNAMES`` macro is not defined).

See:
https://github.com/mirror/mingw-w64/blob/19ecabd807e3429ecf88263107911958b63d4e76/mingw-w64-headers/crt/sys/types.h#L72-L79


Maybe define the type like this:
```c++

#ifdef _WIN32
#  ifdef __MINGW32__
using fmt_mode_t = _mode_t;
#  else
using fmt_mode_t = int;
#  endif
#else
using fmt_mode_t = mode_t;
#endif
```

And then use the ``fmt_mode_t`` type?

src/os.cc
```diff
@@ -42,8 +42,11 @@
 #      define S_IROTH 0
 #    endif
 
+
 #    ifdef __MINGW32__
 #      define _SH_DENYNO 0x40
+#    else
+       using mode_t = int;
```
vitaut, line 49: Let's put `mode_t` in the `fmt::detail` namespace and in the `file` constructor add using `namespace fmt::detail`. Then it will pick up our `fmt::detail::mode_t` if available and the global one otherwise. Also no need to check for `__MINGW32__`.

src/os.cc
```diff
@@ -42,8 +42,11 @@
 #      define S_IROTH 0
 #    endif
 
+
```
vitaut, line 45: Extra newline, please remove.

src/os.cc
```diff
@@ -107,6 +107,8 @@ int detail::utf16_to_utf8::convert(basic_string_view<wchar_t> s) {
 
 namespace detail {
 
+using mode_t = int;
```
vitaut, line 110: Shouldn't this be wrapped in `#ifdef _WIN32`?

Acretock: this detrail is already wrapped in _WIN32 on line 79

vitaut (approved):

vitaut:
Thank you

PR #2612, opened 2021-11-24
Annotate fmt::format and fmt::formatted_size as [[nodiscard]]

0x8000-0000:
This prevents accidentally writing fmt::format when fmt::print was
intended. Other than running tests, there's not a good use case for
discarding the formatted output.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -18,7 +18,7 @@
 // The fmt library version in the form major * 10000 + minor * 100 + patch.
 #define FMT_VERSION 80001
 
-#if defined (__clang__ ) && !defined(__ibmxl__)
+#if defined(__clang__) && !defined(__ibmxl__)
 #  define FMT_CLANG_VERSION (__clang_major__ * 100 + __clang_minor__)
 #else
 #  define FMT_CLANG_VERSION 0
@@ -187,6 +187,14 @@
 #  define FMT_FALLTHROUGH
 #endif
 
+#ifndef FMT_NODISCARD
+#  if FMT_HAS_CPP17_ATTRIBUTE(nodiscard)
+#    define FMT_NODISCARD [[nodiscard]]
+#  else
+#    define FMT_NODISCARD
+#  endif
+#endif
+
 #ifndef FMT_USE_FLOAT
 #  define FMT_USE_FLOAT 1
 #endif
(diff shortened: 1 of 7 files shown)


include/fmt/core.h
```diff
@@ -3071,7 +3071,7 @@ FMT_API auto vformat(string_view fmt, format_args args) -> std::string;
   \endrst
 */
 template <typename... T>
-FMT_INLINE auto format(format_string<T...> fmt, T&&... args) -> std::string {
+[[nodiscard]] FMT_INLINE auto format(format_string<T...> fmt, T&&... args) -> std::string {
```
0x8000-0000, line 3074: Probably needs a macro for ancient compilers which can't ignore unknown attributes.

include/fmt/core.h
```diff
@@ -187,6 +187,12 @@
 #  define FMT_FALLTHROUGH
 #endif
 
+#if FMT_HAS_CPP17_ATTRIBUTE(nodiscard)
+#  define FMT_NODISCARD [[nodiscard]]
+#else
+#  define FMT_NODISCARD
+#endif
```
vitaut, line 194: Let's wrap this in `#ifndef FMT_NODISCARD` to allow users override autodetection.

vitaut (changes requested):

include/fmt/core.h
0x8000-0000: Ack

vitaut: Please apply clang-format to your changes.

0x8000-0000: Sorry about that; done.

(Would it be worthwhile to have an automated checker for this?)

vitaut (approved):

include/fmt/core.h
vitaut: > Would it be worthwhile to have an automated checker for this?

Sure, a PR is welcome =)

vitaut:
Merged, thank you!

include/fmt/core.h
0x8000-0000: > Sure, a PR is welcome =)

Challenge accepted (https://github.com/fmtlib/fmt/pull/2613), but the tools are somewhat crude. Is there a time budget for the CI ?

In another project we grep the output of running git-clang-format, but perhaps I should just contribute/port --dry-run from clang-format to git-clang-format.

PR #2614, opened 2021-11-25
Reformat all source code; no functional changes

0x8000-0000:
Before adding the format checker, the mainline should be clean, to
avoid false failures.


vitaut:
Thank you

PR #2615, opened 2021-11-25
Make MSVC use [[nodiscard]]

AlexGuteniev:
Uniformly detect attributes for __cplusplus and _MSVC_LANG

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -18,7 +18,7 @@
 // The fmt library version in the form major * 10000 + minor * 100 + patch.
 #define FMT_VERSION 80001
 
-#if defined (__clang__ ) && !defined(__ibmxl__)
+#if defined(__clang__) && !defined(__ibmxl__)
 #  define FMT_CLANG_VERSION (__clang_major__ * 100 + __clang_minor__)
 #else
 #  define FMT_CLANG_VERSION 0
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -18,7 +18,7 @@
 // The fmt library version in the form major * 10000 + minor * 100 + patch.
 #define FMT_VERSION 80001
 
-#if defined(__clang__) && !defined(__ibmxl__)
+#if defined (__clang__ ) && !defined(__ibmxl__)
 #  define FMT_CLANG_VERSION (__clang_major__ * 100 + __clang_minor__)
 #else
 #  define FMT_CLANG_VERSION 0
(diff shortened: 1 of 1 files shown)


include/fmt/core.h
```diff
@@ -82,11 +82,17 @@
...
+#  define FMT_CPLUSPLUS __cplusplus
+#endif
+
 #define FMT_HAS_CPP14_ATTRIBUTE(attribute) \
-  (__cplusplus >= 201402L && FMT_HAS_CPP_ATTRIBUTE(attribute))
+  (FMT_CPLUSPLUS >= 201402L && FMT_HAS_CPP_ATTRIBUTE(attribute))
 
 #define FMT_HAS_CPP17_ATTRIBUTE(attribute) \
-  (__cplusplus >= 201703L && FMT_HAS_CPP_ATTRIBUTE(attribute))
+  (FMT_CPLUSPLUS >= 201703L && FMT_HAS_CPP_ATTRIBUTE(attribute))
```
AlexGuteniev, line 95: I'm not even sure why guard against C++17 attributes in C++14 mode if compiler exposes them unconditionally. Compilers are happy to warn even in C++14 mode: https://godbolt.org/z/TrrKhq355

vitaut:
```
<source>:3:3: warning: use of the 'nodiscard' attribute is a C++17 extension [-Wc++17-extensions]
[[nodiscard]] int f();
  ^
```
that's why

include/fmt/core.h
```diff
@@ -82,11 +82,17 @@
 #  define FMT_HAS_CPP_ATTRIBUTE(x) 0
 #endif
 
+#ifdef _MSVC_LANG
+#  define FMT_CPLUSPLUS _MSVC_LANG
+#else
+#  define FMT_CPLUSPLUS __cplusplus
+#endif
```
vitaut, line 89: What is going on here? Doesn't MSVC use the standard `__cplusplus` macro?

AlexGuteniev: [/Zc:__cplusplus](https://docs.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170) enables updated C++ macro. By default `__cplusplus` is defined to some ancient value.

As I read in MSVC STL chat, the reason for this default behavior is the compatibility with some open-source code, in which higher `__cplusplus` values enable not only legitimate C++ features, but also gcc extensions.

You already had `_MSVC_LANG` for `[[fallthrough]]`, I'm just propagating it a bit further.

include/fmt/core.h
AlexGuteniev: uh, ok

vitaut:
Thank you

PR #2617, opened 2021-11-26
Optimize tm formatting (Non C-locales and %Z)

phprus:
Replaced ``std::basic_ostringstream`` to ``std::basic_ostream`` with ``fmt::detail::formatbuf``.

Benchmark: https://github.com/phprus/fmt-bench/tree/optimize-tm-formatting-6

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -11,8 +11,9 @@
 #include <algorithm>
 #include <chrono>
 #include <ctime>
+#include <iterator>
 #include <locale>
-#include <sstream>
+#include <ostream>
 #include <type_traits>
 
 #include "format.h"
@@ -291,23 +292,24 @@ inline null<> gmtime_r(...) { return null<>(); }
 inline null<> gmtime_s(...) { return null<>(); }
 
 template <typename Char>
-inline auto do_write(const std::tm& time, const std::locale& loc, char format,
-                     char modifier) -> std::basic_string<Char> {
-  auto&& os = std::basic_ostringstream<Char>();
+inline void do_write(buffer<Char>& buf, const std::tm& time,
+                     const std::locale& loc, char format, char modifier) {
+  auto&& format_buf = formatbuf<std::basic_streambuf<Char>>(buf);
+  auto&& os = std::basic_ostream<Char>(&format_buf);
   os.imbue(loc);
   using iterator = std::ostreambuf_iterator<Char>;
   const auto& facet = std::use_facet<std::time_put<Char, iterator>>(loc);
   auto end = facet.put(os, os, Char(' '), &time, format, modifier);
   if (end.failed()) FMT_THROW(format_error("failed to format time"));
-  return std::move(os).str();
 }
 
 template <typename Char, typename OutputIt,
           FMT_ENABLE_IF(!std::is_same<Char, char>::value)>
 auto write(OutputIt out, const std::tm& time, const std::locale& loc,
            char format, char modifier = 0) -> OutputIt {
-  auto str = do_write<Char>(time, loc, format, modifier);
-  return std::copy(str.begin(), str.end(), out);
+  auto&& buffer = get_buffer<Char>(out);
+  do_write<Char>(buffer, time, loc, format, modifier);
+  return buffer.out();
 }
 
 inline const std::locale& get_classic_locale() {
@@ -319,7 +321,8 @@ template <typename Char, typename OutputIt,
           FMT_ENABLE_IF(std::is_same<Char, char>::value)>
 auto write(OutputIt out, const std::tm& time, const std::locale& loc,
            char format, char modifier = 0) -> OutputIt {
-  auto str = do_write<char>(time, loc, format, modifier);
+  auto&& buffer = basic_memory_buffer<Char>();
+  do_write<char>(buffer, time, loc, format, modifier);
   if (detail::is_utf8() && loc != get_classic_locale()) {
     // char16_t and char32_t codecvts are broken in MSVC (linkage errors) and
     // gcc-4.
@@ -347,11 +350,11 @@ auto write(OutputIt out, const std::tm& time, const std::locale& loc,
     code_unit* to_next = nullptr;
     constexpr size_t buf_size = 32;
     code_unit buf[buf_size] = {};
-    auto result = f.in(mb, str.data(), str.data() + str.size(), from_next, buf,
-                       buf + buf_size, to_next);
+    auto result = f.in(mb, buffer.data(), buffer.data() + buffer.size(),
+                       from_next, buf, buf + buf_size, to_next);
     if (result != std::codecvt_base::ok)
       FMT_THROW(format_error("failed to format time"));
-    str.clear();
+    buffer.clear();
     for (code_unit* p = buf; p != to_next; ++p) {
       uint32_t c = static_cast<uint32_t>(*p);
       if (sizeof(code_unit) == 2 && c >= 0xd800 && c <= 0xdfff) {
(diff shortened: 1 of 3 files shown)


vitaut:
Thanks for the PR. I wonder why `strftime` is so much faster in the benchmark?

phprus:
glibc and BSD libc has a ``tm.tm_zone`` field. I will add using of the ``tm.tm_zone`` to this PR.
But why ostringstream is so slower I don't know. I will do this research on Monday.

phprus:
@vitaut,
``glibc`` ``strftime`` does not use locales.
``libstdc++`` uses ``strftime_l``  and ``strftime_l`` uses locales.

New algorithm for ``Z`` format is implemented. Benchmark is updated.

include/fmt/format.h
```diff
@@ -265,6 +266,37 @@ FMT_END_NAMESPACE
...
+  explicit formatbuf(buffer<Char>& buf) : buffer_(buf) {}
+
+ protected:
+  // The put area is always empty. This makes the implementation simpler and has
+  // the advantage that the streambuf and the buffer are always in sync and
+  // sputc never writes into uninitialized memory. A disadvantage is that each
+  // call to sputc always results in a (virtual) call to overflow. There is no
+  // disadvantage here for sputn since this always results in a call to xsputn.
+
+  auto overflow(int_type ch = traits_type::eof()) -> int_type override {
```
vitaut, line 287: Default can be removed here.

include/fmt/format.h
```diff
@@ -38,6 +38,7 @@
 #include <limits>        // std::numeric_limits
 #include <memory>        // std::uninitialized_copy
 #include <stdexcept>     // std::runtime_error
+#include <streambuf>     // std::basic_streambuf
```
vitaut, line 41: Adding a dependency on `streambuf` is undesirable esp. since `formatbuf` is not even used in `format.h`. Let's drop it by making `formatbuf` parameterized on `streambuf`.

vitaut (changes requested):
Thanks for the PR!

include/fmt/chrono.h
```diff
@@ -290,36 +291,49 @@ inline null<> localtime_s(...) { return null<>(); }
...
+}
+
 template <typename Char>
-inline auto do_write(const std::tm& time, const std::locale& loc, char format,
-                     char modifier) -> std::basic_string<Char> {
-  auto&& os = std::basic_ostringstream<Char>();
+inline void do_write(buffer<Char>& buf, const std::tm& time,
+                     const std::locale& loc, char format, char modifier) {
+  auto&& format_buf = formatbuf<Char>(buf);
+  auto&& os = std::basic_ostream<Char>(&format_buf);
```
vitaut, line 303: Why `&&`?

include/fmt/chrono.h
```diff
@@ -290,36 +291,49 @@ inline null<> localtime_s(...) { return null<>(); }
...
+inline const std::locale& get_classic_locale() {
+  static const auto& locale = std::locale::classic();
+  return locale;
+}
+
 template <typename Char>
-inline auto do_write(const std::tm& time, const std::locale& loc, char format,
-                     char modifier) -> std::basic_string<Char> {
-  auto&& os = std::basic_ostringstream<Char>();
+inline void do_write(buffer<Char>& buf, const std::tm& time,
```
vitaut, line 300: nit: I think we can drop `do_` here and below because this prefix is only used in cases with identical signatures. Here signature is specific enough.

include/fmt/chrono.h
```diff
@@ -332,27 +346,13 @@ auto write(OutputIt out, const std::tm& time, const std::locale& loc,
...
-#if FMT_CLANG_VERSION
-#  pragma clang diagnostic push
-#  pragma clang diagnostic ignored "-Wdeprecated"
-    auto& f = std::use_facet<codecvt>(loc);
-#  pragma clang diagnostic pop
-#else
-    auto& f = std::use_facet<codecvt>(loc);
-#endif
+    constexpr size_t unit_buf_size = 32;
+    code_unit unit_buf[unit_buf_size] = {};
```
vitaut, line 350: Let's move buffer management to `[do_]write_codecvt` by making it return a struct, something like:

```c++
struct codecvt_result {
  code_unit buf[32];
  code_unit* end;
};
```

include/fmt/chrono.h
phprus: Copy from:
https://github.com/fmtlib/fmt/blob/6ab73113fc39e43329b109d57cd079e15465eacd/include/fmt/ostream.h#L100-L101

include/fmt/chrono.h
phprus: Why? ``write_codecvt`` does not depend on the the output buffer size. So I moved the buffer management out of ``write_codecvt``.

vitaut: Because otherwise it has to be done by the callers effectively duplicating the logic.

include/fmt/format.h
phprus: Done

include/fmt/chrono.h
phprus: Done

include/fmt/chrono.h

phprus:
@vitaut, PR updated.

phprus:
Unexpected runtime error: https://github.com/phprus/fmt/runs/4412073708?check_suite_focus=true
I convert PR to draft...

include/fmt/format.h
```diff
@@ -265,6 +265,37 @@ FMT_END_NAMESPACE
 
 FMT_BEGIN_NAMESPACE
 namespace detail {
+
+template <typename Char, typename Streambuf, typename Streamsize>
```
vitaut, line 269: Both `Char` and `Streamsize` can be determined from `Streambuf`: `Char` is `Streambuf::char_type` and `Streamsize` is the return type of `xsputn`.

include/fmt/chrono.h
phprus: ``end`` value is dependent of ``buf``.
Returning ``codecvt_result`` from a function corrupts ``end``.

include/fmt/chrono.h
vitaut: I guess it has something to do with noncopyability.

phprus:
@vitaut, PR updated. Error on Windows is fixed.

include/fmt/chrono.h
```diff
@@ -290,36 +291,39 @@ inline null<> localtime_s(...) { return null<>(); }
...
 }
 
-template <typename Char, typename OutputIt,
-          FMT_ENABLE_IF(std::is_same<Char, char>::value)>
-auto write(OutputIt out, const std::tm& time, const std::locale& loc,
-           char format, char modifier = 0) -> OutputIt {
-  auto str = do_write<char>(time, loc, format, modifier);
+template <typename CodeUnit> struct codecvt_result {
+  CodeUnit buf[32];
+  CodeUnit* end{};
```
vitaut, line 301: nit: I suggest replacing `{}` with ` = nullptr` for clarity or removing it altogether since `buf` is not initialized.

include/fmt/chrono.h
```diff
@@ -290,36 +291,39 @@ inline null<> localtime_s(...) { return null<>(); }
...
+};
+
+template <typename CodeUnit>
+inline void write_codecvt(codecvt_result<CodeUnit>& out, string_view in_buf,
```
vitaut, line 305: Why not return `codecvt_result` instead of passing it as an output parameter?

Also inline is probably not needed here.

include/fmt/chrono.h
```diff
@@ -332,56 +336,88 @@ auto write(OutputIt out, const std::tm& time, const std::locale& loc,
...
+  return copy_str<Char>(unit.buf, unit.end, out);
+}
+
+template <typename Char, typename OutputIt,
+          FMT_ENABLE_IF(std::is_same<Char, char>::value)>
+auto write_tm_str(OutputIt out, string_view sv, const std::locale& loc)
+    -> OutputIt {
+  auto&& buf = basic_memory_buffer<Char>();
+  buf.append(sv.begin(), sv.end());
+  return write_char_buffer(out, buf, loc);
```
vitaut, line 390: I think it's better to replace `write_char_buffer` with a function that takes `string_view` instead of a buffer to avoid copy.

include/fmt/chrono.h
```diff
@@ -332,56 +336,88 @@ auto write(OutputIt out, const std::tm& time, const std::locale& loc,
...
-    code_unit buf[buf_size] = {};
-    auto result = f.in(mb, str.data(), str.data() + str.size(), from_next, buf,
-                       buf + buf_size, to_next);
-    if (result != std::codecvt_base::ok)
-      FMT_THROW(format_error("failed to format time"));
-    str.clear();
-    for (code_unit* p = buf; p != to_next; ++p) {
+    codecvt_result<code_unit> unit;
+    write_codecvt(unit, string_view(buf.data(), buf.size()), loc);
+    buf.clear();
```
vitaut, line 341: I think this function should use a new buffer instead of reusing the one passed externally. This will also allow to pass input as a string_view suggested below.

vitaut (changes requested):
In general looks good, just a few more comments.

include/fmt/chrono.h
phprus: ``codecvt_result.end`` is depends on ``codecvt_result.buf``
Returning ``codecvt_result`` from a function may result in ``codecvt_result`` being copied and ``end`` being corrupted.
It is possible to replace the pointer ``end`` with the length ``len``, but this does not guarantee that there is no copying.

vitaut: Makes sense, let's leave as out parameter then.

include/fmt/chrono.h

include/fmt/chrono.h

vitaut (approved):

vitaut:
Thank you!

PR #2623, opened 2021-11-28
Implement c++20 std::chrono::duration subsecond formatting

matrackif:
Hello,

I have returned with a solution to #2207, after dealing with licensing issues in PR #2554. This time I am 100% the author of the code. The solution is simplified, and could be adapted to use if-constexpr if C++17 is detected at compile-time. In fact, the solution I have proposed is working better than the current <format> MSVC implementation which leads to template overflow when `std::atto` is used as a precision. I will open an issue there. 

1. Subsecond values that are not representable in `std::intmax_t` (the standard type for chrono tick counts) cause program termination with `FMT_ASSERT()`. The standard requires us to support up to 18 decimal digits, and `std::int64_t` can store a maximum value of 9,223,372,036,854,775,807 or, 9.2e+18,. Therefore it can represent all 18 digit numbers. However the following assert test had to be removed:
```
(void)fmt::format("{:%S}",
                    std::chrono::duration<float, std::atto>(1.79400457e+31f));
```
2. The decimal point is still not formatted according to the locale but that is a general problem.

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -1393,12 +1393,11 @@ template <class Duration> class subsecond_helper {
   // Returns the amount of digits according to the c++ 20 spec
   // In the range [0, 18], if more than 18 fractional digits are required,
   // then we return 6 for microseconds precision
-  static constexpr int num_digits(std::intmax_t num, std::intmax_t den,
-                                  int n = 0) {
+  static constexpr int num_digits(long long num, long long den, int n = 0) {
     return num % den == 0 ? n : (n > 18 ? 6 : num_digits(num * 10, den, n + 1));
   }
 
-  static constexpr std::intmax_t pow10(std::uint32_t n) {
+  static constexpr long long pow10(std::uint32_t n) {
     return n == 0 ? 1 : 10 * pow10(n - 1);
   }
 
@@ -1406,7 +1405,13 @@ template <class Duration> class subsecond_helper {
             FMT_ENABLE_IF(std::numeric_limits<Rep>::is_signed)>
   static constexpr std::chrono::duration<Rep, Period> abs(
       std::chrono::duration<Rep, Period> d) {
-    return d >= d.zero() ? d : -d;
+    // We need to compare the duration using the count() method directly
+    // due to a compiler bug in clang-11 regarding the spaceship operator,
+    // when -Wzero-as-null-pointer-constant is enabled.
+    // In clang-12 the bug has been fixed. See
+    // https://bugs.llvm.org/show_bug.cgi?id=46235 and the reproducible example:
+    // https://www.godbolt.org/z/Knbb5joYx
+    return d.count() >= d.zero().count() ? d : -d;
   }
 
   template <class Rep, class Period,
@@ -1591,8 +1596,8 @@ struct chrono_formatter {
   template <typename RepType> void write(RepType value, int width) {
     write_sign();
     if (isnan(value)) return write_nan();
-    uint32_or_64_or_128_t<std::intmax_t> n =
-        to_unsigned(to_nonnegative_int(value, max_value<std::intmax_t>()));
+    uint32_or_64_or_128_t<long long> n =
+        to_unsigned(to_nonnegative_int(value, max_value<long long>()));
     int num_digits = detail::count_digits(n);
     if (width > num_digits) out = std::fill_n(out, width - num_digits, '0');
     out = format_decimal<char_type>(out, n, num_digits).end;
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -1420,17 +1420,6 @@ static constexpr std::chrono::duration<Rep, Period> abs(
   return d;
 }
 
-template <class Rep, class Period, class ToDuration>
-static constexpr typename ToDuration::rep get_subseconds(
-    std::chrono::duration<Rep, Period> d) {
-  return std::chrono::treat_as_floating_point<typename ToDuration::rep>::value
-             ? (abs(d) - std::chrono::duration_cast<std::chrono::seconds>(d))
-                   .count()
-             : std::chrono::duration_cast<ToDuration>(
-                   abs(d) - std::chrono::duration_cast<std::chrono::seconds>(d))
-                   .count();
-}
-
 template <typename Char, typename Rep, typename OutputIt,
           FMT_ENABLE_IF(std::is_integral<Rep>::value)>
 OutputIt format_duration_value(OutputIt out, Rep val, int) {
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -1595,20 +1595,21 @@ struct chrono_formatter {
     static constexpr auto fractional_width =
         detail::num_digits(Duration::period::num, Duration::period::den);
 
-    using precision = std::chrono::duration<
+    using subsecond_precision = std::chrono::duration<
         typename std::common_type<typename Duration::rep,
                                   std::chrono::seconds::rep>::type,
         std::ratio<1, detail::pow10(fractional_width)>>;
     // We could use c++ 17 if constexpr here.
-    if (std::ratio_less<typename precision::period,
+    if (std::ratio_less<typename subsecond_precision::period,
                         std::chrono::seconds::period>::value) {
       *out++ = '.';
       const auto subseconds =
-          std::chrono::treat_as_floating_point<typename precision::rep>::value
+          std::chrono::treat_as_floating_point<
+              typename subsecond_precision::rep>::value
               ? (detail::abs(d) -
                  std::chrono::duration_cast<std::chrono::seconds>(d))
                     .count()
-              : std::chrono::duration_cast<precision>(
+              : std::chrono::duration_cast<subsecond_precision>(
                     detail::abs(d) -
                     std::chrono::duration_cast<std::chrono::seconds>(d))
                     .count();
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -1389,59 +1389,47 @@ inline std::chrono::duration<Rep, std::milli> get_milliseconds(
 #endif
 }
 
-template <class Duration> class subsecond_helper {
-  // Returns the amount of digits according to the c++ 20 spec
-  // In the range [0, 18], if more than 18 fractional digits are required,
-  // then we return 6 for microseconds precision
-  static constexpr int num_digits(long long num, long long den, int n = 0) {
-    return num % den == 0 ? n : (n > 18 ? 6 : num_digits(num * 10, den, n + 1));
-  }
+// Returns the amount of digits according to the c++ 20 spec
+// In the range [0, 18], if more than 18 fractional digits are required,
+// then we return 6 for microseconds precision.
+static constexpr int num_digits(long long num, long long den, int n = 0) {
+  return num % den == 0 ? n : (n > 18 ? 6 : num_digits(num * 10, den, n + 1));
+}
 
-  static constexpr long long pow10(std::uint32_t n) {
-    return n == 0 ? 1 : 10 * pow10(n - 1);
-  }
+static constexpr long long pow10(std::uint32_t n) {
+  return n == 0 ? 1 : 10 * pow10(n - 1);
+}
 
-  template <class Rep, class Period,
-            FMT_ENABLE_IF(std::numeric_limits<Rep>::is_signed)>
-  static constexpr std::chrono::duration<Rep, Period> abs(
-      std::chrono::duration<Rep, Period> d) {
-    // We need to compare the duration using the count() method directly
-    // due to a compiler bug in clang-11 regarding the spaceship operator,
-    // when -Wzero-as-null-pointer-constant is enabled.
-    // In clang-12 the bug has been fixed. See
-    // https://bugs.llvm.org/show_bug.cgi?id=46235 and the reproducible example:
-    // https://www.godbolt.org/z/Knbb5joYx
-    return d.count() >= d.zero().count() ? d : -d;
-  }
+template <class Rep, class Period,
+          FMT_ENABLE_IF(std::numeric_limits<Rep>::is_signed)>
+static constexpr std::chrono::duration<Rep, Period> abs(
+    std::chrono::duration<Rep, Period> d) {
+  // We need to compare the duration using the count() method directly
+  // due to a compiler bug in clang-11 regarding the spaceship operator,
+  // when -Wzero-as-null-pointer-constant is enabled.
+  // In clang-12 the bug has been fixed. See
+  // https://bugs.llvm.org/show_bug.cgi?id=46235 and the reproducible example:
+  // https://www.godbolt.org/z/Knbb5joYx
+  return d.count() >= d.zero().count() ? d : -d;
+}
 
-  template <class Rep, class Period,
-            FMT_ENABLE_IF(!std::numeric_limits<Rep>::is_signed)>
-  static constexpr std::chrono::duration<Rep, Period> abs(
-      std::chrono::duration<Rep, Period> d) {
-    return d;
-  }
+template <class Rep, class Period,
+          FMT_ENABLE_IF(!std::numeric_limits<Rep>::is_signed)>
+static constexpr std::chrono::duration<Rep, Period> abs(
+    std::chrono::duration<Rep, Period> d) {
+  return d;
+}
 
- public:
-  static constexpr auto fractional_width =
-      num_digits(Duration::period::num, Duration::period::den);
-
-  using precision = std::chrono::duration<
-      typename std::common_type<typename Duration::rep,
-                                std::chrono::seconds::rep>::type,
-      std::ratio<1, pow10(fractional_width)>>;
-
-  template <class Rep, class Period>
-  static constexpr typename precision::rep get_subseconds(
-      std::chrono::duration<Rep, Period> d) {
-    return std::chrono::treat_as_floating_point<typename precision::rep>::value
-               ? (abs(d) - std::chrono::duration_cast<std::chrono::seconds>(d))
-                     .count()
-               : std::chrono::duration_cast<precision>(
-                     abs(d) -
-                     std::chrono::duration_cast<std::chrono::seconds>(d))
-                     .count();
-  }
-};
+template <class Rep, class Period, class ToDuration>
+static constexpr typename ToDuration::rep get_subseconds(
+    std::chrono::duration<Rep, Period> d) {
+  return std::chrono::treat_as_floating_point<typename ToDuration::rep>::value
+             ? (abs(d) - std::chrono::duration_cast<std::chrono::seconds>(d))
+                   .count()
+             : std::chrono::duration_cast<ToDuration>(
+                   abs(d) - std::chrono::duration_cast<std::chrono::seconds>(d))
+                   .count();
+}
 
 template <typename Char, typename Rep, typename OutputIt,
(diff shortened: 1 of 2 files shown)


matrackif:
@vitaut Hi could you rerun the CI tests.

Thanks

matrackif:
@vitaut I locally used -Wconversion instead of -Wsign-converison so line of code still had to be fixed. If you could run the workflow one more time it would be much appreciated.

Thanks

include/fmt/chrono.h
```diff
@@ -1550,11 +1588,11 @@ struct chrono_formatter {
     }
   }
 
-  void write(Rep value, int width) {
+  template <typename RepType> void write(RepType value, int width) {
```
vitaut, line 1591: Why introduce a template parameter instead of using `Rep`?

include/fmt/chrono.h
```diff
@@ -1550,11 +1588,11 @@ struct chrono_formatter {
...
     write_sign();
     if (isnan(value)) return write_nan();
-    uint32_or_64_or_128_t<int> n =
-        to_unsigned(to_nonnegative_int(value, max_value<int>()));
+    uint32_or_64_or_128_t<std::intmax_t> n =
+        to_unsigned(to_nonnegative_int(value, max_value<std::intmax_t>()));
```
vitaut, line 1595: Let's use `long long` instead of `intmax_t` because it's a built-in type and can also handle 18 digits of precision.

include/fmt/chrono.h
```diff
@@ -1319,20 +1319,20 @@ inline bool isfinite(T) {
 }
 
 // Converts value to int and checks that it's in the range [0, upper).
```
vitaut, line 1321: int -> Int

include/fmt/chrono.h
```diff
@@ -1389,16 +1389,54 @@ inline std::chrono::duration<Rep, std::milli> get_milliseconds(
...
-  using common_type = typename std::common_type<Rep, std::intmax_t>::type;
-  auto ms = mod(d.count() * static_cast<common_type>(Period::num) /
-                    static_cast<common_type>(Period::den) * 1000,
-                1000);
-  return std::chrono::duration<Rep, std::milli>(static_cast<Rep>(ms));
-}
+template <class Duration> class subsecond_helper {
+  /// Returns the amount of digits according to the c++ 20 spec
+  /// In the range [0, 18], if more than 18 fractional digits are required,
+  /// then we return 6 for microseconds precision
```
vitaut, line 1395: Please replace `///` with `//` since this is an internal API.

include/fmt/chrono.h
```diff
@@ -1389,16 +1389,54 @@ inline std::chrono::duration<Rep, std::milli> get_milliseconds(
...
-          FMT_ENABLE_IF(std::is_floating_point<Rep>::value)>
-inline std::chrono::duration<Rep, std::milli> get_milliseconds(
-    std::chrono::duration<Rep, Period> d) {
-  using common_type = typename std::common_type<Rep, std::intmax_t>::type;
-  auto ms = mod(d.count() * static_cast<common_type>(Period::num) /
-                    static_cast<common_type>(Period::den) * 1000,
-                1000);
-  return std::chrono::duration<Rep, std::milli>(static_cast<Rep>(ms));
-}
+template <class Duration> class subsecond_helper {
```
vitaut, line 1392: Having a utility struct seems unnecessary. I suggest converting it into a function (e.g. `write_fractional_seconds`) or merging the logic into `write` where it is used.

vitaut (changes requested):
Thanks for the PR!

test/chrono-test.cc
```diff
@@ -585,4 +583,46 @@ TEST(chrono_test, weekday) {
   }
 }
 
+TEST(chrono_test, cpp20_duration_subsecond_support) {
+  using attoseconds = std::chrono::duration<std::intmax_t, std::atto>;
+  // Check that 18 digits of subsecond precision are supported
+  EXPECT_EQ(fmt::format("{:%S}", attoseconds{999999999999999999}),
+            "00.999999999999999999");
+  EXPECT_EQ(fmt::format("{:%S}", attoseconds{673231113420148734}),
```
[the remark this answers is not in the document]
matrackif: It is an arbitrary 18 digit number, perhaps it is redundant.

test/chrono-test.cc
```diff
@@ -585,4 +583,46 @@ TEST(chrono_test, weekday) {
...
+  {
+    // Check that {:%H:%M:%S} is equivalent to {:%T}
+    auto dur = std::chrono::milliseconds{3601234};
+    auto formatted_dur = fmt::format("{:%T}", dur);
+    EXPECT_EQ(formatted_dur, "01:00:01.234");
+    EXPECT_EQ(fmt::format("{:%H:%M:%S}", dur), formatted_dur);
+  }
+  using nanoseconds_dbl = std::chrono::duration<double, std::nano>;
+  EXPECT_EQ(fmt::format("{:%S}", nanoseconds_dbl{-123456789.123456789}),
+            "-00.123456789");
```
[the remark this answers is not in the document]
matrackif: According to the spec: 

> If the number of seconds is less than 10, the result is prefixed with 0

The number of seconds is `0`, which is less than `10`, hence we write `00`.

The MSVC implementation of `std::format` also interprets the standard this way. The check below passes locally in my instance of Visual Studio 2022

`std::format("{:%S}", std::chrono::duration<double, std::nano{-123456789.123456789}) == std::string("-00.123456789")`

test/chrono-test.cc
```diff
@@ -544,14 +544,12 @@ TEST(chrono_test, negative_durations) {
 
 TEST(chrono_test, special_durations) {
   EXPECT_EQ(
-      "40.",
-      fmt::format("{:%S}", std::chrono::duration<double>(1e20)).substr(0, 3));
+      "40",
+      fmt::format("{:%S}", std::chrono::duration<double>(1e20)).substr(0, 2));
```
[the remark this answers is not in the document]
matrackif: According to the spec, the decimal point should be printed only if the duration "period" type is more precise than seconds:

> If the precision of the input cannot be exactly represented with seconds, then the format is a decimal floating-point number with a fixed format and a precision matching that of the precision of the input

 This can be determined statically at compile time and does not depend on runtime values of the duration. The test I modified has seconds precision, so I removed the decimal point.

In my local instance of Visual Studio 2022, the following test passes:

`std::format("{:%S}", std::chrono::duration<double>{1.234}) == std::string("01")`

include/fmt/chrono.h
matrackif:
```suggestion
  // Returns the amount of digits according to the c++ 20 spec
  // In the range [0, 18], if more than 18 fractional digits are required,
  // then we return 6 for microseconds precision
```

test/chrono-test.cc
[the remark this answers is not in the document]
vitaut: Oh, I misread the test case and thought that the integral part `123456789.` is seconds rather than nanoseconds. I suggest changing the fractional part to something different to prevent confusion.

test/chrono-test.cc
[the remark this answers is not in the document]
vitaut: Shall we check that there is no decimal point then?

include/fmt/chrono.h
matrackif: As you wish, std::chrono::duration types use std::intmax_t internally that's why I went along with it.

include/fmt/chrono.h
matrackif: `Rep` is currently assumed to be an unsigned type. I wanted the proper type to propagate through the code and be deduced in the template. Otherwise there could be an unsigned to signed conversion here which would lead to a -Wconversion error.

vitaut: > `Rep` is currently assumed to be an unsigned type.

I don't think this is correct. `Rep` is the representation type of a duration and it can be signed (in fact it's normally signed).

matrackif:
@vitaut I have implemented the suggestions you have requested. If you have any further remarks I would gladly hear them. The workflows still need to be run :)

test/chrono-test.cc
```diff
@@ -543,15 +543,14 @@ TEST(chrono_test, negative_durations) {
 }
 
 TEST(chrono_test, special_durations) {
-  EXPECT_EQ(
-      "40.",
-      fmt::format("{:%S}", std::chrono::duration<double>(1e20)).substr(0, 3));
+  auto value = fmt::format("{:%S}", std::chrono::duration<double>(1e20));
+  // No decimal point is printed so size() is 2.
+  EXPECT_EQ(value.size(), 2);
+  EXPECT_EQ("40", value.substr(0, 2));
```
vitaut, line 549: Let's simplify this:

```c++
EXPECT_EQ(value, "40");
```
Then we don't need a comment because it's clear that we don't emit a decimal point.

include/fmt/chrono.h
```diff
@@ -1319,20 +1319,20 @@ inline bool isfinite(T) {
 }
 
 // Converts value to int and checks that it's in the range [0, upper).
-template <typename T, FMT_ENABLE_IF(std::is_integral<T>::value)>
-inline int to_nonnegative_int(T value, int upper) {
+template <typename T, typename Int, FMT_ENABLE_IF(std::is_integral<T>::value)>
```
vitaut, line 1322: Please update the comment in the line above because we are converting to `Int` now (with capital "I").

test/chrono-test.cc
matrackif: Of course. I recall the size of this string being platform dependent but after the changes in this PR it should always be consistent. Done

include/fmt/chrono.h
matrackif: Done

include/fmt/chrono.h
matrackif: Sorry, lowercase `rep` is used to store the internal duration value, and `rep` is unsigned to avoid overflow. At first I wanted to use the existing logic in the `write()`  function to format a duration value of type lowercase `rep`. This is irrelevant now as I reverted this function to its previous state

test/chrono-test.cc
```diff
@@ -585,4 +583,46 @@ TEST(chrono_test, weekday) {
   }
 }
 
+TEST(chrono_test, cpp20_duration_subsecond_support) {
+  using attoseconds = std::chrono::duration<std::intmax_t, std::atto>;
+  // Check that 18 digits of subsecond precision are supported
```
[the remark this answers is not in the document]
matrackif: Done

test/chrono-test.cc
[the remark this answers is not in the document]
matrackif: Done

include/fmt/chrono.h
```diff
@@ -1389,15 +1389,35 @@ inline std::chrono::duration<Rep, std::milli> get_milliseconds(
 #endif
 }
 
-template <typename Rep, typename Period,
-          FMT_ENABLE_IF(std::is_floating_point<Rep>::value)>
-inline std::chrono::duration<Rep, std::milli> get_milliseconds(
+// Returns the amount of digits according to the c++ 20 spec
```
vitaut, line 1392: nit: amount -> number

include/fmt/chrono.h
```diff
@@ -1389,15 +1389,35 @@ inline std::chrono::duration<Rep, std::milli> get_milliseconds(
...
+// In the range [0, 18], if more than 18 fractional digits are required,
+// then we return 6 for microseconds precision.
+static constexpr int num_digits(long long num, long long den, int n = 0) {
```
vitaut, line 1395: I don't think static is needed here (and similarly in a functions below).

include/fmt/chrono.h
```diff
@@ -1389,15 +1389,35 @@ inline std::chrono::duration<Rep, std::milli> get_milliseconds(
 #endif
 }
 
-template <typename Rep, typename Period,
-          FMT_ENABLE_IF(std::is_floating_point<Rep>::value)>
-inline std::chrono::duration<Rep, std::milli> get_milliseconds(
+// Returns the amount of digits according to the c++ 20 spec
+// In the range [0, 18], if more than 18 fractional digits are required,
+// then we return 6 for microseconds precision.
```
vitaut, line 1394: Is returning 6 actually required by the standard? Can we turn it into a compile-time error instead?

include/fmt/chrono.h
```diff
@@ -1560,6 +1580,38 @@ struct chrono_formatter {
     out = format_decimal<char_type>(out, n, num_digits).end;
   }
 
+  template <class Duration> void write_fractional_seconds(Duration d) {
+    static constexpr auto fractional_width =
```
vitaut, line 1584: This doesn't need to be static because it's just an int.

vitaut (changes requested):
A few more, mostly minor comments.

include/fmt/chrono.h
matrackif: Good point, static made sense when it was a member function, now it is no longer is needed.

include/fmt/chrono.h
matrackif: Ok

include/fmt/chrono.h
matrackif: From the standard:

> ...and a precision matching that of the precision of the input (or to a microseconds precision if the conversion to floating-point decimal seconds cannot be made within 18 fractional digits)

Microseconds precision is 6 digits. The MSVC implementation of `std::format` also behaves this way.

include/fmt/chrono.h
matrackif: Good point, lightweight integral types probably do not need to be static, even if `constexpr`

include/fmt/chrono.h
```diff
@@ -1389,15 +1389,35 @@ inline std::chrono::duration<Rep, std::milli> get_milliseconds(
...
+  // when -Wzero-as-null-pointer-constant is enabled.
+  // In clang-12 the bug has been fixed. See
+  // https://bugs.llvm.org/show_bug.cgi?id=46235 and the reproducible example:
+  // https://www.godbolt.org/z/Knbb5joYx
+  return d.count() >= d.zero().count() ? d : -d;
+}
+
+template <class Rep, class Period,
+          FMT_ENABLE_IF(!std::numeric_limits<Rep>::is_signed)>
+static constexpr std::chrono::duration<Rep, Period> abs(
```
vitaut, line 1418: One more static to remove here.

vitaut:
A fuzzer has discovered a UB which seems to be introduced by this PR:

```c++
fmt::print("{:%S}", std::chrono::duration<long double, std::ratio<1, 1000000> >(-1.7591538808701083784E+4865L));
```

```
% c++ test.cc -std=c++11 -I include src/format.cc -fsanitize=undefined
% ./a.out
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/v1/chrono:906:28: runtime error: 1.75915e+4859 is outside the range of representable values of type 'long long'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior 
```

matrackif:
Subseconds are always an integral value no longer than 18 digits. The implementation will cast all values to a long long. If the value is not representable by a long long like this huge double UBsan will report this error. 

It is not clear how to handle values not representable by intmax_t/long long, the spec doesn't seem to mention what to do. I guess it's up to the implementation how to truncate such values.

vitaut:
I think we should add a range check when converting from an FP number.

matrackif:
And throw an exception if the value is out of range? What method do you propose for range checking floats. I assume such checks exist already in the codebase

vitaut:
I implemented a fix for the UB in https://github.com/fmtlib/fmt/commit/784e2a7b4227657229fcac4532977b99508ee5cb.

matrackif:
> I implemented a fix for the UB in [784e2a7](https://github.com/fmtlib/fmt/commit/784e2a7b4227657229fcac4532977b99508ee5cb).

This commit does not work. I believe it can be safely reverted because currently the above example throws an exception "invalid value". If exceptions are disabled then we need an alternative approach.

`fmt::print("{:%S}\n", std::chrono::duration<long double, std::milli >(1.0));` correctly prints `00.001` 

Whereas if we change the type to `long double` then it incorrectly prints `00.000`

[The offending line is here](https://github.com/fmtlib/fmt/commit/784e2a7b4227657229fcac4532977b99508ee5cb#diff-5d08eff445a9a7793fbd22d873b2284f0ae69f1550aff06f673de49768bccca0R1670)

`detail::abs(d) - std::chrono::duration_cast<sec>(d)` where `sec` is `std::chrono::duration<long double>` will subtract away the subsecond part which we want to keep. The integer cast to `std::chrono::seconds` was intentional in the implementation exactly for this reason. 

[Link to a godbolt example](https://godbolt.org/z/T58broGea)

vitaut:
Could you provide a PR with a better workaround for the UB?

matrackif:
At first I thought of the following solution:

1. Check if the duration type is floating_point and if it is greater than `max_value<long long>`

 a. If yes, then cast it to its seconds representation, and **truncate it**. This way we don't subtract away the valuable subsecond part of the number.  Something like this:

`detail::abs(d) - Duration{std::trunc(std::chrono::duration_cast<std::chrono::duration<typename D::rep>>(d).count())}`

b. If not, then cast to `std::chrono::seconds` as before and call it a day.

But then I realized that in practice such code will never be reached because 

`uint32_or_64_or_128_t<long long> n =
          to_unsigned(to_nonnegative_int(subseconds, max_value<long long>()));`

Will thrown an exception if the value is larger than `max_value<long long>` anyways. The simple code suddenly becomes much more complicated if we want to handle floating point values in the range: `(max_value<long, long>, max_value<long double>]`

A possible workaround would be to avoid using `detail::count_digits()` to count the number of digits for floating point types and simply delegate the work to the fmt floating point formatter and specifying precision to be equal to `num_fractional_digits`:

Any thoughts?

vitaut:
> A possible workaround would be to avoid using detail::count_digits() to count the number of digits for floating point types and simply delegate the work to the fmt floating point formatter and specifying precision to be equal to num_fractional_digits

This sounds reasonable. We can always optimize this path later and correctness is more important.

matrackif:
Ok, I am trying to format floating point subsecond values using the floating point formatter, but as always there are issues :).

If the duration type is floating point, then we do our calculations using floating point arithmetic. Let's say we calculate the subseconds remaining to be `std::chrono::duration<double, std::micro>(123.456)`. We call:

`format_to(out, "{:{}.{}f}", subseconds, num_fractional_digits, num_fractional_digits);`, but this yields `123.456000` which is not what we want. We want the result to be simply `123`, because we don't care about the fractional part of the subseconds themselves.  After calling `std::trunc()` on the value we obtain `123.000000`., still doesn't help because

So we could cast to an integral type to remove the fractional part, but then we end up with the same problem that we started with, since some floating point values are not representable by integral types and we get UB.

In other words, how can we remove the fractional part of floating point numbers without casting to an integral type? Or how can we format a FP number without its fractional part (and decimal point), using existing facilities?

matrackif:
The easy way out would be to simply throw an exception if the formatted value is larger than `max_value<long long>`. Then the code would technically be correct and there would be no UB, the downside is that floating point values larger than `max_value<long long>` would not be supported by the facility.

vitaut:
I think we should format integral and fractional parts together. I took a stab at this as part of fixing another UB: https://github.com/fmtlib/fmt/commit/d9f045fba18c6897ae0931a931450638560e3fd4.

matrackif:
I am a fan of this approach. We simply format the duration using the floating point formatting facility, otherwise we use the integral implementation.

Issue #2628, opened 2021-12-03
Current fmt dosn't build with UWP as platform


Acretock:
On 8.0.1 builds normally 

Build with cmake with this
`cmake -G "Visual Studio 16 2019" -DCMAKE_SYSTEM_NAME=WindowsStore -DCMAKE_SYSTEM_VERSION=10.0 -DCMAKE_BUILD_TYPE=Release -A ARM64 -DUWP=TRUE ..`
and 
`cmake --build . --config Release --target Project`

error code is 
```cpp
fmt/chrono.h(884,5): error C3861: '_tzset':
identifier not found project\build\Project.vcxproj]
```

Acretock:
i found out that `_tzset` is not supported in UWP, [post](https://social.msdn.microsoft.com/Forums/sqlserver/en-US/f5c99f2a-912e-4489-bf9f-3b74e609b18d/uwpwin32-functions-in-c-libraries-build-for-uwp-arm?forum=wpdevelop)

vitaut:
Should be fixed by #2633.

PR #2629, opened 2021-12-03
check for win desktop app

Acretock:
Should fix this #2628 
this should checks for Windows desktop app, and disable for UWP, wich dont support `_tzset`

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -878,7 +878,7 @@ template <typename T>
 struct has_member_data_tm_gmtoff<T, void_t<decltype(T::tm_gmtoff)>>
     : std::true_type {};
 
-#if defined(_WIN32)
+#if defined(_WIN32) && (!defined(WINAPI_FAMILY) || (WINAPI_FAMILY == WINAPI_FAMILY_DESKTOP_APP))
 inline void tzset_once() {
   static bool init = []() -> bool {
     _tzset();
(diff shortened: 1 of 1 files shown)


Acretock:
found about this check in [this](https://social.microsoft.com/Forums/partner/zh-CN/51f29cfe-d07f-4898-9962-945ed7b66d7f/predefined-macro-to-identify-uwp?forum=wpdevelop)

include/fmt/chrono.h
```diff
@@ -878,7 +878,7 @@ template <typename T>
 struct has_member_data_tm_gmtoff<T, void_t<decltype(T::tm_gmtoff)>>
     : std::true_type {};
 
-#if defined(_WIN32)
+#if defined(_WIN32) && (!defined(WINAPI_FAMILY) || (WINAPI_FAMILY == WINAPI_FAMILY_DESKTOP_APP))
```
vitaut, line 881: Let's merge `tzset_once` into `format_utc_offset_impl` to avoid repeating the check and since it's only used there.

vitaut (changes requested):
Thanks for the PR.

include/fmt/chrono.h
phprus: @vitaut, after the proposed merge the ``_tzset`` function will be called multiple times. Once for each ``tm_writer`` template instantiation.
It won't throw an error. But there may be a slowdown if present many  ``tm_writer`` template instantiation with differents args.

vitaut: OK, let's keep this function but introduce a macro `FMT_USE_TZSET` instead to avoid repetition.

vitaut:
Closing in favor of #2633.

PR #2633, opened 2021-12-06
Enable tzset only on Windows desktop app

phprus:
More complete fix for #2628
Original solution: #2629

@Acretock, check please this solution.


vitaut:
Thanks!

PR #2636, opened 2021-12-07
The enable_if argument should be 3rd.

BRevzin:

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -278,7 +278,7 @@ You can also write a formatter for a hierarchy of classes::
   };
 
   template <typename T>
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR. Does having `enable_if` as the second argument break anything?

BRevzin:
You know what, I mis-parsed the example. It says:

```cpp
struct fmt::formatter<T, std::enable_if_t<std::is_base_of<A, T>::value, char>>
```

but I read it as:

```cpp
struct fmt::formatter<T, std::enable_if_t<std::is_base_of<A, T>::value>, char>
```

So I thought it was just wrong (i.e. the 2nd argument is `void` and the 3rd argument is `char`, instead of the reverse). But it's actually fine as is.

PR #2643, opened 2021-12-10
Avoid qualifying by inline namespace. Fixes #2642.

### LINKED ISSUE #2642: Support for 'user-defined' namespace
We're using fmt in a library https://github.com/DARMA-tasking/vt, which can be next linked to an app, that itself uses fmt for its own purposes. We'd like to protect ourselves from any kind of ODR violations or ABI mismatch problems, or anything alike. It seems that the simplest way would be to just use our own namespace inside of the fmt namespace. 

At first I wanted to simply define `FMT_BEGIN_NAMESPACE` to something like
```cpp
#define FMT_BEGIN_NAMESPACE \
  namespace fmt {           \
  FMT_INLINE_NAMESPACE vt {
#endif
```

but in `core.h` there's one place where `fmt::<version>` is explicitly used
```cpp
void to_string_view(...);
using fmt::v8::to_string_view;
```

What I did in the end was adding another macro, and replacing explicit references to version, with that macro

```cpp
#ifndef FMT_INLINE_NAMESPACE_NAME
#  define FMT_INLINE_NAMESPACE_NAME v7
#endif
```

Could something like this be added to fmt, to add possibility to easily define own namespace for fmt?

For reference, here's a PR to our library with described changes:
DARMA-tasking/vt#1618

mkurdej:


vitaut:
Thanks!

PR #2645, opened 2021-12-12
Make `buffered_file` move assignment `noexcept`

jerryct:


jerryct:
In general move operations shall be `noexcept`. Instead of directly closing
via `close()` and setting `other.file_` to `nullptr` this PR swaps the member
`file_` and delays closing until `other` goes out of scope.

This change is not complete because the following test fails

```
[ RUN      ] buffered_file_test.move_assignment_closes_file
../test/os-test.cc:210: Failure
Value of: isclosed(old_fd)
  Actual: false
Expected: true
```

Would it be fine to change the behavior? If yes, I would adapt the tests.

vitaut:
Thanks for the PR but changing move assignment in such way results in surprising behavior (and test failures that cannot be fixed).

Issue #2640, opened 2021-12-09
Compile checks in c++20 are kind of lacking


YarikTH:
I massively rewrote ostream based logging in my codebase using _format literal. And suddenly I found out that this is not checked at all. I really like the idea of compile-time checked format because it allows seeing all the problems immediately at the compilation stage, instead of waiting for a crash somewhen in future when the code with {fmt} call will be executed. And it's sad that currently it's almost doesn't working.

Here is a list of compile-time checks that works:

https://godbolt.org/z/cc8MPrGc9

```c++
#include <fmt/format.h>

using namespace fmt::literals;

void func() {
    fmt::format( "{}", 1, 2 );                      // (1) UNCHECKED
    fmt::format( "{} {} {}", 1, 2 );                // (2) checked
    fmt::format( "{:d} {}", "", 2 );                // (3) checked
    fmt::format( "{addr2}", fmt::arg("addr", 1) );  // (4) UNCHECKED
    fmt::format( "{addr2}", "addr"_a = 1 );         // (5) UNCHECKED

    "{}"_format( 1, 2 );                            // (6) UNCHECKED
    "{} {} {}"_format( 1, 2 );                      // (7) UNCHECKED
    "{:d} {}"_format( "", 2 );                      // (8) UNCHECKED
    "{addr2}"_format( fmt::arg("addr", 1) );        // (9) UNCHECKED
    "{addr2}"_format( "addr"_a = 1 );               // (10) UNCHECKED
}
```

- 1,6 - too many arguments. Unchecked at all
- 2,7 - too few arguments. Checked only in fmt::format
- 3,8 - incorrect format flags. Checked only in fmt::format
- 4,9 - incorrectly named arguments using fmt::arg. Unchecked at all
- 5,10 - incorrectly named arguments using _a literal. Unchecked at all

Is there a chance that some unchecked scenarios, especially with _format literals would become checked? Now it seems quite inconsistent and confusing.

P.S. I checked only gcc10 and gcc11 with trunk {fmt}. So maybe they don't work only under some list of compilers.

vitaut:
All of these are expected. It's OK to have unused arguments, named arguments cannot be checked due to C++ limitations and UDL-based is not recommended at all and will likely be deprecated.

alexezeder:
> All of these are expected

@vitaut, I'm afraid I have to disagree with you here since only some of them are expected:

[...]

PR #2646, opened 2021-12-13
Deprecate `_format` UDL in code using `FMT_DEPRECATED`

alexezeder:
Originated from discussion in https://github.com/fmtlib/fmt/issues/2640.

The `_format`-related test moved to separate test suite, since tests are not allowed to have warnings (with `FMT_PEDANTIC` and `FMT_WERROR`) and `FMT_DEPRECATED` produces one. This way, we still can be sure that `_format` works fine.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -3056,7 +3056,7 @@ constexpr auto operator"" _a(const char* s, size_t) -> detail::udl_arg<char> {
     using namespace fmt::literals;
     std::string message = "The answer is {}"_format(42);
  */
-constexpr auto operator"" _format(const char* s, size_t n)
+FMT_DEPRECATED constexpr auto operator"" _format(const char* s, size_t n)
     -> detail::udl_formatter<char> {
   return {{s, n}};
 }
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR. Mostly looks good but let's not introduce a new test for a deprecated function. Let's suppress the warning instead.

alexezeder:
> Let's suppress the warning instead.

That's a bit trickier since there are too many compilers. Of course, we could use some trivial checks and pragmas for GCC, Clang, and MSVC. But to get that warning suppressed in most compilers, something like this is needed - [hedley/hedley.h](https://github.com/nemequ/hedley/blob/8fb0604a8095f6c907378cc3f0391520ae843f6f/hedley.h#L881-L923).

vitaut:
> That's a bit trickier since there are too many compilers.

Then I would suggest removing the test altogether or limit it to a single compiler where it's easy to suppress the deprecation warning. The test is pretty rudimentary anyway.

vitaut:
Thank you!

PR #2647, opened 2021-12-13
Fix `throw` with exceptions disabled

alexezeder:
Originated from https://github.com/fmtlib/fmt/commit/215f21a0382d325efa66df53fbfbfddb020a2234#r61421336.

`throw` replaced with `FMT_THROW`, nothing more.

But the most interesting part here is why it was not detected by tests - it's described in https://github.com/fmtlib/fmt/pull/2648.


vitaut:
Thank you!

Issue #2644, opened 2021-12-11
Include directories should be marked as system includes


hadrielk:
When compiling `fmt` by the direct CMake `add_subdirectory(fmt)` method (as recommended in the docs), the `fmt` build inherits the same compilation flags as the project including it.

But it fails for various errors/warnings, in `clang` at least. 

For example `missing-noreturn`, `signed-enum-bitfield`, and `documentation-unknown-command`:

```
In file included from /fmt/include/fmt/format.h:48:
/fmt/include/fmt/core.h:451:5: error: unknown command tag name [-Werror,-Wdocumentation-unknown-command]
    \rst
    ^~~~
/fmt/include/fmt/core.h:604:38: error: function 'on_error' could be declared with attribute 'noreturn' [-Werror,-Wmissing-noreturn]
  void on_error(const char* message) { throw_format_error(message); }
                                     ^
/fmt/include/fmt/core.h:2075:9: error: enums in the Microsoft ABI are signed integers by default; consider giving the enum type an unsigned underlying type to make this code portable [-Werror,-Wsigned-enum-bitfield]
        align(align::none),
        ^
```
There were more instances than the above, but this gives the idea.

From my perspective, users of `fmt` would normally consider it a system include, not a regular part of their codebase. So I suggest that for the `target_include_directories()` in `fmt/CMakeLists.txt`, the `SYSTEM` cmake attribute be used, or at least a CMake `OPTION` be available to do so.

vitaut:
This should definitely not be the default because it would just hide potentially useful diagnostic but a PR to add a CMake option would be welcome.

daquexian:
FYI, CMake 3.25 [introduces](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/7399) `add_subdirectory(... SYSTEM)` to mark 3rd libraries as system includes: https://cmake.org/cmake/help/latest/command/add_subdirectory.html

PR #2651, opened 2021-12-13
Expose headers depending on special configuration option

alexezeder:
Originated from https://github.com/fmtlib/fmt/issues/2644.

Special `FMT_SYSTEM_HEADERS` (I didn't come up with a better name) CMake option added. It's disabled by default.


vitaut (changes requested):
Thanks for the PR. Looks good but I don't think the option should be tied to {fmt}'s use as a subproject. I suggest making it off by default unconditionally instead.

vitaut (approved):

PR #2658, opened 2021-12-17
Add %I64d/u for printf-style on windows.

bakurise:


vitaut:
Thanks for the PR but this is non-standard (https://en.cppreference.com/w/cpp/io/c/fprintf) and `ll` should be used instead.

PR #2662, opened 2021-12-17
Don't touch begin iterator (pointer) if begin == end.

### LINKED ISSUE #2661: Undefined behavior adding `1` to `nullptr`
On `master` I am seeing some UB here:

```
../fmt/include/fmt/core.h:2485:13: runtime error: applying non-zero offset 1 to null pointer
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../fmt/include/fmt/core.h:2485:13 in
```

which is this line:

```cpp
if (begin + 1 < end && begin[1] == '}' && is_ascii_letter(*begin) &&
      *begin != 'L') {
```

when `begin` is `nullptr`.  But in that case it should be the case that `end` is `nullptr` as well, and so we can get around this by only adding `1` to `begin` when `begin != end`, which logically makes sense anyway, since if `begin == end` then `begin + 1 < end` will always be `false` anyway.  That seems to prevent the `begin + 1` from executing when `begin == nullptr`.  I will send a PR for this.

dpacbach:
Fixes #2661.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -2482,6 +2482,8 @@ FMT_CONSTEXPR FMT_INLINE auto parse_format_specs(const Char* begin,
                                                  const Char* end,
                                                  SpecHandler&& handler)
     -> const Char* {
+  if (begin == end) return begin;
+
   if (begin + 1 < end && begin[1] == '}' && is_ascii_letter(*begin) &&
       *begin != 'L') {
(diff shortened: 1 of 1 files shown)


vitaut:
Fixed in a slightly different way in https://github.com/fmtlib/fmt/commit/659de779e6eceeee66b4b3165a84c928220ef378.

dpacbach:
Thx

PR #2663, opened 2021-12-18
Remove incorrect C++20 check from test/CMakeLists.txt

alexezeder:
This check disables entire branch of tests declaration unconditionally because `CXX_STANDARD` is not defined there. But even if we use `CMAKE_CXX_STANDARD` there, these tests should work for standards ≥20 too, IMHO.

Here are the tests:
* `std-format-test` (now removed)
* `noexception-test`
* `nolocale-test`
* `compile-error-test`

`std-format-test` does not compile after enabling it, since it's either outdated or should still be disabled. @vitaut, could you explain what should be done with it?

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -137,7 +137,7 @@ endif ()
 
 message(STATUS "FMT_PEDANTIC: ${FMT_PEDANTIC}")
 
-if (FMT_PEDANTIC AND CXX_STANDARD LESS 20)
+if (FMT_PEDANTIC)
   # Test that the library can be compiled with exceptions disabled.
   # -fno-exception is broken in icc: https://github.com/fmtlib/fmt/issues/822.
   if (NOT CMAKE_CXX_COMPILER_ID STREQUAL "Intel")


vitaut (changes requested):
`std-format-test` is incompatible with C++20 so the check should be preserved but possibly moved to a place where it only applies to this test.

alexezeder:
I've disabled `std-format-test` with standards ≥ C++20, but there are still errors in `format` header.

alexezeder:
@vitaut, I suggest merging this PR to some temporary branch, where you can apply needed changes for `std-format-test` since I'm not sure that I can adapt it to the latest version of the library. Of course, you could guide me here by discussions, but it would be much faster for you to fix the remaining problems. What do you think about that?
Or we can simply disable `std-format-test` for now. In this case, I will create an issue to enable it again.

vitaut:
I removed `std-format-test` since the API it has been testing has already been merged into the standard.

alexezeder:
Rebased.

vitaut (approved):

vitaut:
Thank you

PR #2664, opened 2021-12-18
Don't explicitly delete copy ctor of dynamic_format_arg_store

lucpelletier:
Explicitly deleting the copy ctor causes the move constructor to not be
implicitly generated. This behaviour is different than what was in
v8.0.1 and causes code that relied on the move ctor of
dynamic_format_arg_store to break.

### DIFF
--- a/test/args-test.cc
+++ b/test/args-test.cc
@@ -171,3 +171,21 @@ TEST(args_test, throw_on_copy) {
   }
   EXPECT_EQ(fmt::vformat("{}", store), "foo");
 }
+
+TEST(args_test, move_constructor) {
+  const int test_integer = 42;
+  const char* const test_c_string = "foo";
+
+  auto store_uptr =
+      std::make_unique<fmt::dynamic_format_arg_store<fmt::format_context>>();
+  store_uptr->push_back(test_integer);
+  store_uptr->push_back(std::string(test_c_string));
+  store_uptr->push_back(fmt::arg("a1", test_c_string));
+
+  fmt::dynamic_format_arg_store<fmt::format_context> moved_store(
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR. Could you elaborate why you need a move ctor?

lucpelletier:
Yes, of course. I want to use Quill, which is a low-latency asynchronous logging library that uses fmtlib. The following class has a member of type dynamic_format_arg_store, and needs to support move operations:

https://github.com/odygrd/quill/blob/1386bff42eea9411600e58b9ddd682f1ec685c5a/quill/include/quill/detail/backend/BackendWorker.h#L211

vitaut (changes requested):
Makes sense but please add a unit test for the move ctor in https://github.com/fmtlib/fmt/blob/master/test/args-test.cc.

lucpelletier:
Added test for move ctor. Did you also want a test for move assignment?

test/args-test.cc
```diff
@@ -171,3 +171,21 @@ TEST(args_test, throw_on_copy) {
   }
   EXPECT_EQ(fmt::vformat("{}", store), "foo");
 }
+
+TEST(args_test, move_constructor) {
+  const int test_integer = 42;
+  const char* const test_c_string = "foo";
+
+  auto store_uptr =
+      std::make_unique<fmt::dynamic_format_arg_store<fmt::format_context>>();
```
phprus, line 180: ``std::make_unique`` - C++14+

lucpelletier: fixed

vitaut (approved):

vitaut:
Thank you

PR #2668, opened 2021-12-22
Remove two expressions which had no effect (reported by LGTM)

stweil:
Signed-off-by: Stefan Weil <sw@weilnetz.de>


stweil:
See https://lgtm.com/projects/g/fmtlib/fmt/ for the alerts which are fixed by this pull requests and some remaining alerts.

vitaut:
Thanks

PR #2673, opened 2021-12-24
Supporting nested format specs for ranges.

BRevzin:
Basic support for nested formatters for ranges from P2286. This does not add any top-level range format specifiers (pad/align/width/empty/delimiter), but does add the placeholder colon for them for the future.

### DIFF
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -633,7 +633,7 @@ struct formatter<
     if (it == end || *it == '}') return it;
 
     if (*it != ':')
-      throw format_error("no top-level range formatters supported");
+      FMT_THROW(format_error("no top-level range formatters supported"));
 
     custom_specs_ = true;
     ++it;
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -601,12 +601,12 @@ struct formatter<
   template <typename T,
             FMT_ENABLE_IF(has_formatter<remove_cvref_t<T>, context>::value)>
   static auto map(T&& value) -> T&& {
-    return (T &&) value;
+    return static_cast<T&&>(value);
   }
   template <typename T,
             FMT_ENABLE_IF(!has_formatter<remove_cvref_t<T>, context>::value)>
-  static auto map(T&& value) -> decltype(mapper().map((T &&) value)) {
-    return mapper().map((T &&) value);
+  static auto map(T&& value) -> decltype(mapper().map(static_cast<T&&>(value))) {
+    return mapper().map(static_cast<T&&>(value));
   }
 
   using formatter_type = conditional_t<
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -222,8 +222,13 @@ template <class Tuple, class F> void for_each(Tuple&& tup, F&& f) {
 }
 
 template <typename Range>
-using uncvref_type =
-    remove_cvref_t<decltype(*detail::range_begin(std::declval<Range>()))>;
+using range_reference_type =
+    decltype(*detail::range_begin(std::declval<Range>()));
+
+// We don't use the Range's value_type for anything, but we do need the Range's
+// reference type, with cv-ref stripped
+template <typename Range>
+using uncvref_type = remove_cvref_t<range_reference_type<Range>>;
 
 template <typename OutputIt> OutputIt write_delimiter(OutputIt out) {
   *out++ = ',';
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -221,9 +221,23 @@ template <class Tuple, class F> void for_each(Tuple&& tup, F&& f) {
   for_each(indexes, std::forward<Tuple>(tup), std::forward<F>(f));
 }
 
+#if FMT_MSC_VER
+// older MSVC doesn't get the reference type correctly for arrays
+template <typename R> struct range_reference_type_impl {
+  using type = decltype(*detail::range_begin(std::declval<R&>()));
+};
+
+template <typename T, std::size_t N> struct range_reference_type_impl<T[N]> {
+  using type = T&;
+};
+
+template <typename T>
+using range_reference_type = typename range_reference_type_impl<T>::type;
+#else
 template <typename Range>
 using range_reference_type =
-    decltype(*detail::range_begin(std::declval<Range>()));
+    decltype(*detail::range_begin(std::declval<Range&>()));
+#endif
 
 // We don't use the Range's value_type for anything, but we do need the Range's
 // reference type, with cv-ref stripped
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -222,7 +222,7 @@ template <class Tuple, class F> void for_each(Tuple&& tup, F&& f) {
 }
 
 template <typename Range>
-using value_type =
+using uncvref_type =
     remove_cvref_t<decltype(*detail::range_begin(std::declval<Range>()))>;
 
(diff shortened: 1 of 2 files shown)


include/fmt/ranges.h
```diff
@@ -589,34 +589,71 @@ template <typename T, typename Char> struct is_range {
...
+            FMT_ENABLE_IF(!has_formatter<remove_cvref_t<T>, context>::value)>
+  static auto map(T&& value) -> decltype(mapper().map((T &&) value)) {
+    return mapper().map((T &&) value);
+  }
+
+  using formatter_type = conditional_t<
+      is_formattable<element, Char>::value,
+      formatter<remove_cvref_t<decltype(map(std::declval<element>()))>, Char>,
+      detail::fallback_formatter<element, Char>>;
+  formatter_type underlying_;
```
BRevzin, line 623: I more or less copied this from `join`, but I feel like there's probably a more direct way of getting the right `formatter_type`?

include/fmt/ranges.h
```diff
@@ -229,7 +229,7 @@ template <class Tuple, class F> void for_each(Tuple&& tup, F&& f) {
 }
 
 template <typename Range>
-using value_type =
+using uncvref_type =
```
vitaut, line 232: I think `value_type` is more descriptive because this gives the range value type. The fact that cvref is removed is less important.

include/fmt/ranges.h
```diff
@@ -589,34 +589,72 @@ template <typename T, typename Char> struct is_range {
...
+  bool custom_specs_ = false;
+
   template <typename ParseContext>
   FMT_CONSTEXPR auto parse(ParseContext& ctx) -> decltype(ctx.begin()) {
-    return ctx.begin();
+    auto it = ctx.begin();
+    auto end = ctx.end();
+    if (it == end || *it == '}') {
+      return it;
+    }
```
vitaut, line 633: Throughout the PR: we don't use braces for single-line if statements.

include/fmt/ranges.h
```diff
@@ -589,34 +589,72 @@ template <typename T, typename Char> struct is_range {
...
+
+    if (*it != ':') {
+      // no top-level range formatters yet
```
vitaut, line 636: The comment basically repeats the error message, let's remove.

include/fmt/ranges.h
```diff
@@ -589,34 +589,72 @@ template <typename T, typename Char> struct is_range {
...
+            FMT_ENABLE_IF(!has_formatter<remove_cvref_t<T>, context>::value)>
+  static auto map(T&& value)
+      -> decltype(mapper().map(static_cast<T&&>(value))) {
+    return mapper().map(static_cast<T&&>(value));
+  }
+
+  using formatter_type = conditional_t<
+      is_formattable<element, Char>::value,
+      formatter<remove_cvref_t<decltype(map(std::declval<element>()))>, Char>,
+      detail::fallback_formatter<element, Char>>;
```
vitaut, line 623: Please move this to the detail namespace since it doesn't need to be parameterized on R.

vitaut (changes requested):
Thanks for the PR! Overall looks good but please address the inline comments.

include/fmt/ranges.h
BRevzin: It doesn't give the `value_type` though, it gives you the range's `reference` - with cvref stripped. A lot of the time, that _is_ the `value_type`, but not always. For example: for `vector<bool>` the `value_type` is `bool` but this gives you `vector<bool>::reference` (which is what we would actually use to format). Or `zip`ping two `vector<int>`s has a `value_type` of `pair<int, int>` but this gives you `pair<int&, int&>`.

I wanted to change it mainly because calling it `value_type` is misleading - we never use the range's `value_type` for anything.

include/fmt/ranges.h
BRevzin: As you wish.

include/fmt/ranges.h
BRevzin: Done.

BRevzin:
Uh, I'm failing CI because of... throwing exceptions? I don't understand.

include/fmt/ranges.h
```diff
@@ -589,34 +589,77 @@ template <typename T, typename Char> struct is_range {
...
+
   template <typename ParseContext>
   FMT_CONSTEXPR auto parse(ParseContext& ctx) -> decltype(ctx.begin()) {
-    return ctx.begin();
+    auto it = ctx.begin();
+    auto end = ctx.end();
+    if (it == end || *it == '}') return it;
+
+    if (*it != ':')
+      throw format_error("no top-level range formatters supported");
```
phprus, line 643: Replace ``throw`` to ``FMT_THROW(...)``.

include/fmt/ranges.h
vitaut: OK but could you clarify this in the comment then?

include/fmt/ranges.h
vitaut: Not if we want to support implicit conversions. I suggest moving it to a new trait in `detail` and using in both places.

vitaut:
> Uh, I'm failing CI because of... throwing exceptions? I don't understand.

Not sure what you are referring to but the error is

> 'parse': is not a member of 'fmt::v8::detail::fallback_formatter<int [3][2],Char,void>

BRevzin:
> > Uh, I'm failing CI because of... throwing exceptions? I don't understand.
> 
> Not sure what you are referring to but the error is
> 
> > 'parse': is not a member of 'fmt::v8::detail::fallback_formatter<int [3][2],Char,void>

@phprus pointed out that it was because I was doing `throw format_error(...)` instead of `FMT_THROW(format_error(...))`.

This one I don't understand. Only fails on Windows too?

include/fmt/ranges.h
BRevzin: Sure.

vitaut:
Looks like on MSVC the `formatter` specialization for `int (&)[3][2]` is somehow disabled (`has_formatter` returns false).

BRevzin:
> Looks like on MSVC the `formatter` specialization for `int (&)[3][2]` somehow becomes disabled with these changes (`has_formatter` returns false).

This seems like an MSVC bug. Here's [an example](https://godbolt.org/z/dq8cvf1hf):

```cpp
#include <utility>

namespace fmt2 {
    template <typename T, size_t N>
    auto range_begin(T(&)[N]) -> T*;

    template <typename R>
    using iterator_t = decltype(range_begin(std::declval<R&>()));

    template <typename R>
    using reference_t = decltype(*std::declval<iterator_t<R>&>());
}

using R = int[3][2];

static_assert(
    std::is_same_v<
        fmt2::iterator_t<R>, int(*)[2]
        >,
    "!"
);

static_assert(
    std::is_same_v<
        fmt2::reference_t<R>, int(&)[2]
        >,
    "!"
);
```

The reference type of `int[3][2]` should be `int(&)[2]`. But MSVC v19.16 thinks it's `int(*&&)[2]`. Which is not a range. It's like they totally ignore the `*`. MSVC v19.21 still miscompiles, MSVC v19.22 gets this right. 

If it's just arrays, I guess could add special handling for arrays for MSVC?

BRevzin:
I got an email about checks failing, but it looks like all the checks passed? I don't even know.

vitaut (approved):

vitaut:
Merged, thanks!

vitaut:
@BRevzin could you by any chance add a section documenting this new feature (similarly to https://github.com/fmtlib/fmt/blob/master/doc/syntax.rst#chrono-format-specifications)?

PR #2674, opened 2021-12-24
Supporting ? as a string presentation type

BRevzin:
Formatting ranges prints strings escaped, this allows the user to do so directly outside of a range context (if so desired). 

Most of this PR is simply moving the logic from fmt/ranges.h to fmt/core.h. The width-counting (to support like... `"{:*^10?}"`) using a counting output iterator is... probably not the right way to do this, but I'm not sure what the right way is.


BRevzin:
So the issue here (as far as I can tell) is that when I moved the implementation from ranges.h to format.h, now it's suddenly getting used for `detail::char8_type` too in some contexts. In C++20 (where I was building locally), that's `char8_t`, which is convertible to `char`, so calls like this:
```cpp
out = format_to(out, "\\x{:02x}", escape.cp);
```
work. But in C++11 (like some of the CI), it doesn't, complaining about how you can't call that:
```cpp
../include/fmt/format.h:1661:26: error: no matching function for call to ‘format_to(std::back_insert_iterator<fmt::v8::detail::buffer<fmt::v8::detail::char8_type> >&, const char [9], uint32_t&)’
 1661 |           out = format_to(out, "\\U{:08x}", escape.cp);
      |                 ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
because the overload checks to see if the iterator is an `output_iterator` for `char`, and it's not if `char8_type` is just some enum.

The just-make-it-work solution (that I just pushed) is to only do the escaping logic for specifically `char`. That works on C++11 too. But like... obviously not the Right Thing To Do.

include/fmt/format.h
```diff
@@ -1387,13 +1387,361 @@ auto write_ptr(OutputIt out, UIntPtr value,
...
+  \rst
+  Constructs a compile-time format string from a string literal *s*.
+
+  **Example**::
+
+    // A compile-time error because 'd' is an invalid specifier for strings.
+    std::string s = fmt::format(FMT_STRING("{:d}"), "foo");
+  \endrst
+ */
+#define FMT_STRING(s) FMT_STRING_IMPL(s, fmt::compile_string, )
```
vitaut, line 1645: Why move this?

include/fmt/format.h
```diff
@@ -1387,13 +1387,361 @@ auto write_ptr(OutputIt out, UIntPtr value,
...
+    xsigned -= len;
+    if (xsigned < 0) break;
+    current = !current;
+  }
+  return current;
+}
+
+// Returns true iff the code point cp is printable.
+// This code is generated by support/printable.py.
+inline auto is_printable(uint32_t cp) -> bool {
```
vitaut, line 1428: Let's move the definition to `format-inl.h` and remove `inline`.

include/fmt/format.h
```diff
@@ -1387,13 +1387,361 @@ auto write_ptr(OutputIt out, UIntPtr value,
...
+#define FMT_STRING(s) FMT_STRING_IMPL(s, fmt::compile_string, )
+
+template <typename Char, typename OutputIt>
+auto write_escaped_string(OutputIt out, basic_string_view<Char> str) -> OutputIt {
+  return copy_str<Char>(str.data(), str.data() + str.size(), out);
+}
+
+template <typename OutputIt, typename Char>
+auto write_escaped_cp(OutputIt out, find_escape_result<Char> const& escape) -> OutputIt
+{
```
vitaut, line 1654: Please apply clang-format.

include/fmt/format.h
```diff
@@ -1387,13 +1387,361 @@ auto write_ptr(OutputIt out, UIntPtr value,
...
+ */
+#define FMT_STRING(s) FMT_STRING_IMPL(s, fmt::compile_string, )
+
+template <typename Char, typename OutputIt>
+auto write_escaped_string(OutputIt out, basic_string_view<Char> str) -> OutputIt {
+  return copy_str<Char>(str.data(), str.data() + str.size(), out);
+}
+
+template <typename OutputIt, typename Char>
+auto write_escaped_cp(OutputIt out, find_escape_result<Char> const& escape) -> OutputIt
```
vitaut, line 1653: `const` is in the wrong place =)

include/fmt/format.h
```diff
@@ -1387,13 +1387,361 @@ auto write_ptr(OutputIt out, UIntPtr value,
...
 FMT_CONSTEXPR auto write_char(OutputIt out, Char value,
                               const basic_format_specs<Char>& specs)
     -> OutputIt {
+  bool is_debug = specs.type == presentation_type::debug;
   return write_padded(out, specs, 1, [=](reserve_iterator<OutputIt> it) {
-    *it++ = value;
-    return it;
+    if (is_debug) {
+      return write_escaped_char(it, value);
+    } else {
```
vitaut, line 1741: nit: `else` is not needed because of early return, please remove. Also we don't use braces for single-line `if` statements.

include/fmt/format.h
```diff
@@ -1657,18 +2005,68 @@ FMT_CONSTEXPR FMT_INLINE auto write(OutputIt out, T value,
...
   auto size = s.size();
   if (specs.precision >= 0 && to_unsigned(specs.precision) < size)
     size = code_point_index(s, to_unsigned(specs.precision));
-  auto width =
-      specs.width != 0 ? compute_width(basic_string_view<Char>(data, size)) : 0;
+  bool is_debug = specs.type == presentation_type::debug;
+  size_t width = 0;
+  if (specs.width != 0) {
+    if (is_debug) {
+      width = write_escaped_string(counting_iterator{}, s).count();
```
vitaut, line 2058: Width is in code points, not code units.

vitaut (changes requested):
Thanks for the PR!

vitaut:
@BRevzin, do you still plan working on this?

BRevzin:
> @BRevzin, do you still plan working on this?

I do intend on getting to it eventually, but if you (or someone else) wants to take over, I'm totally fine with that. As you can tell, I'm not really sure how to actually do some of these things :) (well, except for the formatting comments - I can format the formatting PR at least. much meta. such wow).

vitaut:
I'm OK leaving the switch to code point for later and will merge the PR if you address the other (minor) comments.

include/fmt/format.h
BRevzin: Punting on this per your comment.

include/fmt/format.h
BRevzin: Done as requested.

include/fmt/format.h
BRevzin: Done.

include/fmt/format.h
BRevzin: Done.

include/fmt/format.h
BRevzin: `write_escaped_string` uses it, which was previously only used in `ranges.h` - but now is used by `?` so had to move it earlier.

support/printable.py
```diff
@@ -171,7 +171,7 @@ def main():
     normal1 = compress_normal(normal1)
 
     print("""\
-inline auto is_printable(uint32_t cp) -> bool {\
+FMT_PRINT auto is_printable(uint32_t cp) -> bool {\
```
vitaut, line 174: You probably meant `FMT_FUNC`.

BRevzin: Yeah lol. I made the same mistake in the real code, but of course that one I actually compiled... Fixed.

vitaut (approved):

vitaut:
Thank you!

PR #2677, opened 2021-12-26
Fix throw with exceptions disabled and improve noexception test

phprus:
``throw`` replaced with ``FMT_THROW`` in ``chrono.h``.
Improve noexception test.


alexezeder:
We can probably change `noexception-test` to include all headers. This way, it would also fail for commits like https://github.com/fmtlib/fmt/commit/eab2ea9fc2c183dfa2d45b196fd8002fb41aa489.

alexezeder:
It looks like GCC doesn't care about `throw` in headers with exceptions disabled with no usage in cpp file. But at least Clang cares.

vitaut:
Thank you!

PR #2686, opened 2022-01-02
Improve detail::write_digit2_separated

bdejean:
Hi,
I felt little bit anxious when I saw the call to detail::write_digit2_separated with no buffer size in https://github.com/fmtlib/fmt/pull/2500/files and I was wondering what would happen if the tm struct was invalid.
Then I read its code and I think it can be made little bit safer for future usage by making sure it is only given a char[8] as first argument.


vitaut:
Thanks for the PR but I don't think it's necessary and there are some build failures. If you are concerned about some corner cases such as invalid tm a test case would be welcome.

Issue #2687, opened 2022-01-03
When C4061 is enabled in Visual Studio, usage of fmt fails


dconnet:
I have warnings-as-errors and have enabled 4061 (technically, I have enabled /Wall). This produces many warnings such as:
core.h(2775,1): warning C4061: enumerator 'fmt::v8::presentation_type::hex_upper' in switch of enum 'fmt::v8::presentation_type' is not explicitly handled by a case label
(format.h also is hit by this)

Compilation of the fmt library works. But using it in a program results in the above.

vitaut:
A PR to suppress the warning would be welcome.

dconnet:
I know how to suppress it with pragma for Visual Studio. But that won't work with other compilers. Would adding all the enum items to the switch be acceptable?

vitaut:
Adding unused branches is undesirable. The warning itself is meaningless considering that there is a default.

dconnet:
The warning is extremely useful in large projects when one person adds an enum value and misses updating a switch statement in one of hundreds of dependent projects. (We've caught errors like that at work)

I forgot to mention in my original description that I'm compiling libfmt at /W4 (plus /WX) - compiling at /Wall doesn't work.

vitaut:
The switch in question has a default though so the warning doesn't make much sense. I recommend reporting to Microsoft as a false positive.

dconnet:
It's not a false positive: https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4061?view=msvc-170
This warning is explicitly for letting you know that all values do not have a case (note: C4062 is the warning for when you don't have a default case). Yes, it's off by default. But /Wall enables it.

vitaut:
Regardless of whether it's false positive or not a PR to suppress it is welcome provided that it's not too intrusive. Note that fixing all warnings in existence is a non-goal so this **is** intrusive I recommend suppressing the warnings with `FMT_SYSTEM_HEADERS` or other means that don't require changes to the library.

dconnet:
https://github.com/fmtlib/fmt/pull/2688

vitaut:
As commented on the PR, since this is not even level 3 warning, I suggest suppressing by other means such as `FMT_SYSTEM_HEADERS` unless there is some clean way to suppress it in the library.

PR #2688, opened 2022-01-03
Fix VS C4061 warning when using /Wall

dconnet:
Fix C4061 by adding the missing enum values to the switch statements.

(Issue 2687)


vitaut (changes requested):
Thanks for the PR but as commented in #2687 adding unused branches is undesirable.

vitaut:
Thanks again but that's too messy for such a low quality warning. Since this is not even level 3 warning, I suggest suppressing by other means such as `FMT_SYSTEM_HEADERS`.

Issue #2689, opened 2022-01-03
Chrono formatter does not work if FMT_STATIC_THOUSANDS_SEPARATOR is defined


gsjaardema:
In fmt-8.1.0, I get an undefined symbol warning if I use `#define FMT_STATIC_THOUSANDS_SEPARATOR ','`

For example:
```
#include "fmt/chrono.h"

std::string time_stamp(const std::string &format)
{
  if (format == "") {
    return std::string("");
  }

  time_t      calendar_time = std::time(nullptr);
  struct tm * local_time    = std::localtime(&calendar_time);
  std::string time_string   = fmt::format(format, *local_time);
  return time_string;
}

int main()
{
  std::string tsFormat = "[{:%H:%M:%S}] ";
  auto ts = time_stamp(tsFormat);
  fmt::print("{}\n", ts);
}
```

If I compile with `g++ -DFMT_HEADER_ONLY -I include time_stamp.C`, it works.  

If I compile with `g++ -DFMT_HEADER_ONLY -DFMT_STATIC_THOUSANDS_SEPARATOR=\',\' -I include time_stamp.C`, I get:
```
Undefined symbols for architecture x86_64:
  "__ZNK3fmt2v86detail10locale_ref3getISt6localeEET_v", referenced from:
      __ZN3fmt2v86detail10get_localeC1EbNS1_10locale_refE in ccDqSfBU.o
ld: symbol(s) not found for architecture x86_64
collect2: error: ld returned 1 exit status
```

vitaut:
Chrono formatting requires locale support while `FMT_STATIC_THOUSANDS_SEPARATOR` disables it so the error is expected although it could probably be diagnosed better.

gsjaardema:
Up until 8.1.0, this worked fine.  It only broke with 8.1.0 release.  So did the requirement change at 8.1.0 and is this documented somewhere that you cannot now use chrono if `FMT_STATIC_THOUSANDS_SEPARATOR` is defined?

vitaut:
The implicit requirement was always there but the locale support was not implemented until 8.1.0 which is why it happened to work. I don't think `FMT_STATIC_THOUSANDS_SEPARATOR` is documented anywhere at all (a PR to document it would be welcome).

delaitre-manfrotto:
we have similar issue
we use FMT_STATIC_THOUSANDS_SEPARATOR because we target embedded and using locale just make the binary size way too big
[...]

PR #2693, opened 2022-01-03
Allow chrono to work with defined thousands separator

gsjaardema:
This allows the 8.1.0 and later chrono.h implementation to work (i.e. do not emit any undefined symbol warnings) if the user has `FMT_STATIC_THOUSANDS_SEPARATOR` defined.  The chrono implementation is relying on this definition of `locale_ref` to be defined.  

Moving the `#ifndef` line to still define `locale_ref` and only select which version of `thousands_sep_impl` gets defined.

This solves the issue referenced in #2689

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -17,10 +17,7 @@
 #include <cstring>  // std::memmove
 #include <cwchar>
 #include <exception>
-
-#ifndef FMT_STATIC_THOUSANDS_SEPARATOR
-#  include <locale>
-#endif
+#include <locale>
 
 #ifdef _WIN32
 #  include <io.h>  // _isatty
@@ -96,7 +93,6 @@ inline void fwrite_fully(const void* ptr, size_t size, size_t count,
   if (written < count) FMT_THROW(system_error(errno, "cannot write to file"));
 }
 
-#ifndef FMT_STATIC_THOUSANDS_SEPARATOR
 template <typename Locale>
 locale_ref::locale_ref(const Locale& loc) : locale_(&loc) {
   static_assert(std::is_same<Locale, std::locale>::value, "");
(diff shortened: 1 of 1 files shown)


godsswaggz:
Wow

vitaut:
Thanks for the PR but I think this is undesirable because it will effectively revert https://github.com/fmtlib/fmt/commit/abde38b4fba8c9578d503ea561e1bcc153669d0d. Note that defining `FMT_STATIC_THOUSANDS_SEPARATOR` doesn't make much sense unless you also want to get rid of the locale dependency because the default formatting is already locale-independent.

vitaut:
Thinking more of it, the term "C locale" is ambiguous and may refer either to the locale accessed via the standard C API (which is what I thought you are referring to) or the locale called "C". {fmt}'s default is compatible with the "C" locale and you can use `group_digits` do additionally add (compatible) grouping.

gsjaardema:
Thanks.  I will go back and see why I originally had to use the defined `FMT_STATIC_THOUSANDS_SEPARATOR` and see if I still see the problem I was seeing at that time....  If so, I will try to formulate a better issue report or PR.  Thanks.

gsjaardema:
Just to make sure I understand.  If I want to have numbers output with comma-separated digits no matter what locale is specified, then the correct way to do this with {fmt} is using `group_digits` instead of using the `{:L}`. 
For example:
```
fmt::format("{}\n", fmt::group_digits(variable));
```
Instead of:
```
fmt::format("{:L}\n", variable);
```

vitaut:
> If I want to have numbers output with comma-separated digits no matter what locale is specified, then the correct way to do this with {fmt} is using group_digits instead of using the {:L}.

Yes.

jbongseo:
`fmt::group_digits` does not work with chrono.

```
fmt::format("{}\n", fmt::group_digits(std::chrono::milliseconds{12345}));  // compile error.
```

It there any workaround?

vitaut:
You can extract representation and format it with digit grouping.

PR #2696, opened 2022-01-04
Restore FMT_API on error_handler::on_error()

### LINKED ISSUE #2695: Undefined symbol _ZN3fmt2v86detail13error_handler8on_errorEPKc
Upgrading fmt from 8.0.1 to 8.1.0 breaks a CI run here: https://github.com/flatsurf/flatsurf/runs/4698212878?check_suite_focus=true#step:6:554

In that CI run, we are trying to load a library, libintervalxt, which was [built against fmt 8.0.1](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=380913&view=logs&j=656edd35-690f-5c53-9ba3-09c10d0bea97&t=e5c8ab1d-8ff9-5cae-b332-e15ae582ed2d&l=688) with [fmt 8.1.0 installed](https://github.com/flatsurf/flatsurf/runs/4698212878?check_suite_focus=true#step:6:432).

While the missing symbol is not part of the public API of fmt, it used to have `FMT_API` set in 8.0.1 but not anymore in 8.1.0. At least in the setup that the conda-forge distribution uses, this made that symbol disappear from the library. Since libraries such as libintervalxt build against the old ABI (without using this symbol explicitly in the code of course) this seems like a breaking ABI change was introduced in the 8.1.0 release.

Or maybe conda-forge's libintervalxt and fmt are using the wrong compiler flags somehow?

saraedum:
this fixes a breaking ABI change that was introduce in the upgrade from
8.0.1 to 8.1.0.

Fixes #2695.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -601,7 +601,8 @@ struct error_handler {
   constexpr error_handler(const error_handler&) = default;
 
   // This function is intentionally not constexpr to give a compile-time error.
-  void on_error(const char* message) { throw_format_error(message); }
+  // This function is marked as FMT_API for backwards compatibility, see #2695.
+  FMT_NORETURN FMT_API void on_error(const char* message) { throw_format_error(message); }
 };
 FMT_END_DETAIL_NAMESPACE
 


vitaut:
Thank you!

PR #2701, opened 2022-01-04
Make colored print handle UTF-8

### LINKED ISSUE #2681: UTF8 doesn't work with text styles
I'm using vs2022 and have set `/utf-8` option.
Thr example code 
`fmt::print(
  "┌{0:─^{2}}┐\n"
  "│{1: ^{2}}│\n"
  "└{0:─^{2}}┘\n", "", "Hello, world!", 20);`
works fine.
But when I try to add some text style,
`fmt::print(fmt::fg(fmt::color::red),
  "┌{0:─^{2}}┐\n"
  "│{1: ^{2}}│\n"
  "└{0:─^{2}}┘\n", "", "Hello, world!", 20);`
the output get garbled.

AlexGuteniev:
Fixes #2681 

The working flow:
 * With `is_utf8()` calls `detail::print` that specifically converts to UTF-16 and calls `WriteConsoleW` 
 * Without `is_utf8()` calls `fwrite_fully`

The broken flow:
 * Always call `fputs`

The fix:
 * With `is_utf8()` call `detail::print` 
 * Without `is_utf8()` keep calling `fputs`, as `fwrite_fully` is not exported

This is tested on Windows only. On Windows `fputs` does not do any magic, and color sequences are processed by `WriteConsole`. I'm not sure if on other platform `fputs` processes these ANSI escape sequences, or `fwrite` will also do.

### DIFF
--- a/include/fmt/color.h
+++ b/include/fmt/color.h
@@ -530,7 +530,10 @@ void vprint(std::FILE* f, const text_style& ts, const S& format,
   basic_memory_buffer<Char> buf;
   detail::vformat_to(buf, ts, to_string_view(format), args);
   buf.push_back(Char(0));
-  detail::fputs(buf.data(), f);
+  if (detail::is_utf8())
+    detail::print(f, basic_string_view<Char>(buf.begin(), buf.size()));
+  else
+    detail::fputs(buf.data(), f);
 }
 
 /**
--- a/include/fmt/color.h
+++ b/include/fmt/color.h
@@ -474,26 +474,27 @@ FMT_CONSTEXPR ansi_color_escape<Char> make_emphasis(emphasis em) FMT_NOEXCEPT {
   return ansi_color_escape<Char>(em);
 }
 
-template <typename Char>
-inline void fputs(const Char* chars, FILE* stream) FMT_NOEXCEPT {
-  std::fputs(chars, stream);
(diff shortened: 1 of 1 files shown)


AlexGuteniev:
I actually observe inconsistency that non-colored format throws I/O errors but colored directly calls `fputs` without checking. I'm not creating the issue yet, maybe it should be fixed within this change.

include/fmt/color.h
```diff
@@ -529,8 +529,12 @@ void vprint(std::FILE* f, const text_style& ts, const S& format,
             basic_format_args<buffer_context<type_identity_t<Char>>> args) {
   basic_memory_buffer<Char> buf;
   detail::vformat_to(buf, ts, to_string_view(format), args);
-  buf.push_back(Char(0));
-  detail::fputs(buf.data(), f);
+  if (detail::is_utf8())
+    detail::print(f, basic_string_view<Char>(buf.begin(), buf.size()));
```
vitaut, line 533: Please wrap the body of the if statement in curly braces since else is braced.

vitaut (changes requested):
Thanks for the PR!

> I actually observe inconsistency that non-colored format throws I/O errors but colored directly calls fputs without checking. I'm not creating the issue yet, maybe it should be fixed within this change.

Good catch, please add an error check here or in a separate PR.

vitaut (approved):

PR #2711, opened 2022-01-12
Use FMT_NOEXCEPT macro instead of noexcept

tohammer:
Replace all occurrences in non-test code.

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -965,37 +965,37 @@ template <typename OutputIt, typename Char> class tm_writer {
   OutputIt out_;
   const std::tm& tm_;
 
-  auto tm_sec() const noexcept -> int {
+  auto tm_sec() const FMT_NOEXCEPT -> int {
     FMT_ASSERT(tm_.tm_sec >= 0 && tm_.tm_sec <= 61, "");
     return tm_.tm_sec;
   }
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR but is there any problem with using `noexcept` directly?

phprus:
@vitaut 
https://github.com/fmtlib/fmt/blob/a126b4d888dbdf254bc2b11fab8a5195951549f5/include/fmt/core.h#L148-L168
If I'm not mistaken, then ``FMT_NOEXCEPT`` is always ``noexcept``. ``FMT_NOEXCEPT`` was needed for older compilers that are not supported by the fmt now.

vitaut:
We can definitely get rid of `FMT_DETECTED_NOEXCEPT` since all the supported compilers have `noexcept` but I wonder if the ability to disable `FMT_NOEXCEPT` is still used anywhere.

phprus:
The code (core.h, lines 148-168) may be reduced to:
```c++
#ifndef FMT_NOEXCEPT 
# define FMT_NOEXCEPT noexcept
/* optional warning: */
#else
# warning FMT_NOEXCEPT is deprecated
#endif 
```
?

vitaut:
Let's hear from @tohammer first.

Issue #2715, opened 2022-01-13
overflow in constant expression in count_fractional_digits with non-decimal durations


BrukerJWD:
- fmtlib 8.1.1
- gcc 10.2

see https://godbolt.org/z/rMvqK5Y99

When trying to format a duration that has non-decimal denominator (e.g. `std::chrono::duration<int, std::ratio<3, 7>>`), we get compiler errors in gcc due to an overflow when counting the fractional digits:

    chrono.h:1669:32:   in 'constexpr' expansion of 'fmt::v8::detail::count_fractional_digits(((long long int)((intmax_t)std::ratio<3, 7>::num)), ((long long int)((intmax_t)std::ratio<3, 7>::den)), 0)'
    chrono.h:1479:53:   in 'constexpr' expansion of 'fmt::v8::detail::count_fractional_digits((num * 10), den, (n + 1))'
    chrono.h:1479:53:   in 'constexpr' expansion of 'fmt::v8::detail::count_fractional_digits((num * 10), den, (n + 1))'
    [...]

It is a quite uncommon use case, but in general fmtlib does support it, e.g.  `std::chrono::duration<int, std::ratio<3, 10>>` is output as `[3/10]s`.

vitaut:
cc @matrackif

matrackif:
C++11 constexpr functions force us to use recursion instead of loops.  

We also cannot have multiple return statements or variables declarations. Therefore we use the ternary operator everywhere, but ternary operator expressions are always evaluated, even when the "false" case is chosen. 

The only solution I came up with is to use classic C++98 style recursive template instantiation, where our partial specialization  handles the base case.

vitaut:
Thanks, @matrackif, for a quick fix!

PR #2722, opened 2022-01-17
Fix overflow for chrono durations

matrackif:
Use template instantiation to handle recursion overflow in C++11 constexpr functions. 

Related to #2715


vitaut:
Thank you!

PR #2724, opened 2022-01-18
Fix -Wconversion warning 

Tobi823:
Thanks for your great work and the great fmt library.

I have a small problem. When I cross-compile (with dockcross/linux-armv7 https://github.com/dockcross/dockcross to Raspberry Pi ARMv7) my source code with your fmt-library, I receive this warning:

```
/work/build_armv7/_deps/fmt-src/include/fmt/chrono.h:1217:37: warning: conversion from 'std::make_unsigned<long long int>::type' {aka 'long long unsigned int'} to 'size_t' {aka 'unsigned int'} may change value [-Wconversion]
 1217 |       copy2(buf, digits2(to_unsigned(year / 100)));
      |                          ~~~~~~~~~~~^~~~~~~~~~~~
```

Not using the -Wconversion flag is not an option because I want to use the flag to detect as many errors as possible.
And I'm not able to configure CMake, CPMAddPackage and `target_include_directories`+`SYSTEM` correctly to ignore warnings from your library 🙈.

My solution (adding const_cast<size_t>) works for me but I don't know if this is a clean solution. I'm not a C++ expert.

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -1214,7 +1214,7 @@ template <typename OutputIt, typename Char> class tm_writer {
     char buf[10];
     size_t offset = 0;
     if (year >= 0 && year < 10000) {
-      copy2(buf, digits2(static_cast<size_t>(to_unsigned(year / 100))));
+      copy2(buf, digits2(static_cast<size_t>(year / 100)));
     } else {
       offset = 4;
       write_year_extended(year);
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -1214,7 +1214,7 @@ template <typename OutputIt, typename Char> class tm_writer {
     char buf[10];
     size_t offset = 0;
     if (year >= 0 && year < 10000) {
-      copy2(buf, digits2(to_unsigned(year / 100)));
+      copy2(buf, digits2(static_cast<size_t>(to_unsigned(year / 100))));
     } else {
       offset = 4;
       write_year_extended(year);


include/fmt/chrono.h
```diff
@@ -1214,7 +1214,7 @@ template <typename OutputIt, typename Char> class tm_writer {
     char buf[10];
     size_t offset = 0;
     if (year >= 0 && year < 10000) {
-      copy2(buf, digits2(to_unsigned(year / 100)));
+      copy2(buf, digits2(static_cast<size_t>(to_unsigned(year / 100))));
```
vitaut, line 1217: `to_unsigned` is kinda redundant now that we have a cast, let's remove it.

vitaut (changes requested):
Thanks for the PR.

include/fmt/chrono.h
Tobi823: But `to_unsigned()` contains a ">= 0" check.

Would it not be possible, that:
-  `tm_year()` returns a negative number (line 1213)
- negative number divided by 100 results in a negative number (line 1217)
- casting a negative number to size_t results in a big positive number
- the big positive number would overrun (line 1217) the char array in the function `digits2` (format.h line 919/925)?

phprus: No, this function (line 1217) is only called for ``year`` in range [0, 10000). Check on line 1216.

Tobi823: ok

vitaut (approved):

Issue #1684, opened 2020-05-15
Cannot build tests with cross-compiler (Ubuntu/MXE)


ambitslix:
Trying to build under MXE shared: 

```
Linking CXX executable ../bin/os-test.exe
cd /build/Linux_x86_64_Debug_x86_64-w64-mingw32.shared-cross-compiler/submodules/fmt/test && /mxe/usr/x86_64-pc-linux-gnu/bin/cmake -E cmake_link_script CMakeFiles/os-test.dir/link.txt --verbose=1
/mxe/usr/x86_64-pc-linux-gnu/bin/cmake -E remove -f CMakeFiles/os-test.dir/objects.a
/mxe/usr/bin/x86_64-w64-mingw32.shared-ar cr CMakeFiles/os-test.dir/objects.a @CMakeFiles/os-test.dir/objects1.rsp
/mxe/usr/x86_64-pc-linux-gnu/bin/x86_64-w64-mingw32.shared-g++ -g   -Wl,--whole-archive CMakeFiles/os-test.dir/objects.a -Wl,--no-whole-archive  -o ../bin/os-test.exe -Wl,--out-implib,libos-test.dll.a -Wl,--major-image-version,0,--minor-image-version,0 @CMakeFiles/os-test.dir/linklibs.rsp
CMakeFiles/os-test.dir/objects.a(os-test.cc.obj): In function `fmt::v6::internal::count_digits(unsigned long long)':
/submodules/fmt/include/fmt/format.h:762: undefined reference to `__imp__ZN3fmt2v68internal10basic_dataIvE23zero_or_powers_of_10_64E'
CMakeFiles/os-test.dir/objects.a(os-test.cc.obj): In function `fmt::v6::internal::count_digits(unsigned int)':
/submodules/fmt/include/fmt/format.h:820: undefined reference to `__imp__ZN3fmt2v68internal10basic_dataIvE23zero_or_powers_of_10_32E'
CMakeFiles/os-test.dir/objects.a(os-test.cc.obj): In function `void fmt::v6::internal::basic_writer<fmt::v6::buffer_range<char> >::write<float, 0>(float, fmt::v6::basic_format_specs<char>)':
/submodules/fmt/include/fmt/format.h:1708: undefined reference to `__imp__ZN3fmt2v68internal10basic_dataIvE5signsE'
/submodules/fmt/include/fmt/format.h:1717: undefined reference to `__imp__ZN3fmt2v68internal10basic_dataIvE5signsE'
[...]

PR #1716, opened 2020-06-04
Fixes for MINGW build.

ambitslix:
I agree that my contributions are licensed under the {fmt} license, and agree to future changes to the licensing.

PR #2728, opened 2022-01-22
Making target_compile_options PRIVATE, fix #2726, fix #2507

### LINKED ISSUE #2507: Building with cmake's fetch content using a cross compiler
So I have the following in my CMakeLists.txt file:
```
include(FetchContent)

set(FMT_TEST FF CACHE INTERNAL "disabling fmt tests")

FetchContent_Declare(
        fmt
        GIT_REPOSITORY  https://github.com/fmtlib/fmt.git
        GIT_TAG         8.0.1
        GIT_PROGRESS    TRUE
        USES_TERMINAL_DOWNLOAD TRUE
)

...
target_link_libraries(mybinary
        PRIVATE fmt::fmt
)

```

When cross-compiling with say an arm compiler, i get undefined reference errors like this:
```
<artificial>:(.text.startup+0x1588): undefined reference to `fmt::v8::vprint(fmt::v8::basic_string_view<char>, ...
```

I'm cross-compiling like this:
```
cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=arm-openwrt-linux-gcc -DCMAKE_CXX_COMPILER=arm-openwrt-linux-g++ ..
```
So i'm not specifying a toolchain file.

Now, is this a cmake bug in the FetchContent module, or is this a libfmt bug?

Cheers

### LINKED ISSUE #2726: `-mbig-obj` is incompatible with `-flto` on mingw compiler and cause problems linking the library in a project
I think this is a bug in mingw compiler, but this library adds the switch to its interface https://github.com/fmtlib/fmt/blob/c5aafd8f9003c5f0ac1d9bc00d566fefe60996d7/CMakeLists.txt#L251-L256 and this trigger errors "plugin needed to handle lto object"

My system is Ubuntu 20.04 and I am compiling with 86_64-w64-mingw32

May be the requirement can be relaxed for 64 bit projects? What was the reason to include the flag into the config?

VasiliPupkin256:
I weren't able to reach @ambitslix in the issue #1684 so I am not sure what exactly caused the problem for him. I think that making compile options private is a good compromise. It is always easier to add an option to the project using the library than removing an option set by an included library and unconditionally propagated to the whole project using it. This should close #2726 and also solve an already closed #2507. Probably this target_compile_options can be removed completely, I don't think there was an issue building the library itself, but I don't have an environment to reproduce the error that this line solved on the first place.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -251,7 +251,7 @@ endif ()
 if (MINGW)
   check_cxx_compiler_flag("-Wa,-mbig-obj" FMT_HAS_MBIG_OBJ)
   if (${FMT_HAS_MBIG_OBJ})
-    target_compile_options(fmt PUBLIC "-Wa,-mbig-obj")
+    target_compile_options(fmt PRIVATE "-Wa,-mbig-obj")
   endif()
 endif ()
 


CMakeLists.txt
```diff
@@ -251,7 +251,7 @@ endif ()
 if (MINGW)
   check_cxx_compiler_flag("-Wa,-mbig-obj" FMT_HAS_MBIG_OBJ)
   if (${FMT_HAS_MBIG_OBJ})
-    target_compile_options(fmt PUBLIC "-Wa,-mbig-obj")
+    target_compile_options(fmt PRIVATE "-Wa,-mbig-obj")
   endif()
```
vitaut, line 255: I suggest removing this block altogether because this is some old workaround that may not even be applicable any more. We can always reintroduce the workaround as an opt-in if it happens again.

vitaut (changes requested):
Thanks for the PR.

CMakeLists.txt
VasiliPupkin256: The issue #1684 is less than 2 years old...

vitaut: Yeah, but it's unclear what purpose `-Wa,-mbig-obj` serves here. The library itself is very small and unlikely to cause linker issues so it might be solving a problem elsewhere.

vitaut: BTW this option was added in #1716. Maybe @ambitslix can shed light why it was needed.

vitaut (approved):

vitaut:
Anyway, let's merge this as an immediate fix for now.

PR #2732, opened 2022-01-25
Fixes NVIDIA HPC compiler and Intel ICC compatibility

federico-busato:


include/fmt/core.h
```diff
@@ -49,6 +49,12 @@
 #  define FMT_ICC_VERSION 0
 #endif
 
+#ifdef __NVCOMPILER
+#  define FMT_NVHPC_VERSION (__NVCOMPILER_MAJOR__ * 100 + __NVCOMPILER_MINOR__)
+#else
+#  define FMT_NVHPC_VERSION 0
+#endif
```
vitaut, line 56: What is the difference between `__NVCOMPILER` and `__NVCC__`? Can we use the latter instead of introducing another macro?

vitaut (commented):
Thanks for the PR

include/fmt/core.h
federico-busato: NVCC (`__NVCC__`) and NVC++ (`__NVCOMPILER`) are two entirely different tools.
NVCC is a front-end for separating CUDA code from standard C/C++ and sending it to standard compilers, e.g. clang and GCC. NVC/C++ is the NVIDIA C/C++ compiler that replaces clang and GCC. See https://docs.nvidia.com/hpc-sdk/compilers/hpc-compilers-user-guide/index.html 
In the end, we need two separate macros for them...

vitaut: OK, thanks for clarification. I suggest renaming `FMT_NVHPC_VERSION` to `FMT_NVCOMPILER_VERSION` for consistency with `__NVCOMPILER_*__`.

federico-busato: sure, I will do it

Issue #2530, opened 2021-10-02
access mode of files created by output_file is surprising


berthubert:
Hi,

In 8.0.1 at least, on POSIX, files created by fmt::output_file explicitly set themselves to be readable and writable by the user only. 

This led to some issues after I moved to fmt::output_file since all of a sudden my files could no longer be read by the web server.

POSIX custom is to set a broader default permission (0660 for example), and let the operator tone that down using their umask setting. 

Now it may be that fmt has a good reason for 0600 of course, but it is a bit surprising. There is also no way to change this mode.

Thoughts? Can I help?

   Bert (formerly powerdns_bert)

vitaut:
Fixed in https://github.com/fmtlib/fmt/commit/4a85db1ce15cbf232ddb9a8a6640b74dbcbc2cbd, now the default mode is `-rw-r--r--` which is consistent with `fopen`. Thanks for reporting.

berthubert:
Thanks!

arogge:
Sorry for digging this one up again.
If the intent is to mimic behaviour of `fopen()` then the patch is wrong. 
While FMT is setting `S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH`, POSIX mandates that it should be  `S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH`.

[fopen() in the Open Group Specification](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fopen.html) reads:
> If mode is w, wb, a, ab, w+, wb+, w+b, a+, ab+, or a+b, and the file did not previously exist, the fopen() function shall create a file as if it called the creat() function with a value appropriate for the path argument interpreted from pathname and a value of S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH for the mode argument.

The system will usually strip `S_IWGRP` and `S_IWOTH` based on the process' umask, which will result in the observed `-rw-r--r--` if umask was set to `022`. However with a umask of `002` `fopen()` would produce `-rw-rw-r--` and fmt should probably do the same.

I can do a pull request for that if you want.

vitaut:
@arogge, good catch, thanks.

> I can do a pull request for that if you want.

Please, do.

arogge:
See PR #2733

PR #2733, opened 2022-01-26
Fix access mode of files created (#2530)

arogge:
The previous fix for this in 4a85db1 was incomplete. The intent was to
mimic what `fopen()` is doing. As per standard[1] `fopen()` also sets
`S_IWGRP` and `S_IWOTH` and lets the umask handle the rest.

[1] https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/functions/fopen.html

### DIFF
--- a/src/os.cc
+++ b/src/os.cc
@@ -35,8 +35,14 @@
 #    ifndef S_IRGRP
 #      define S_IRGRP 0
 #    endif
+#    ifndef S_IWGRP
+#      define S_IWGRP 0
+#    endif
 #    ifndef S_IROTH
 #      define S_IROTH 0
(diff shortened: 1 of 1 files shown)


arogge:
Do we need automated test for that and if so: how would you like it implemented?

vitaut:
Thanks for the PR! A test would be nice, please add one to https://github.com/fmtlib/fmt/blob/master/test/os-test.cc (it can be POSIX-specific/disabled on Windows for simplicity).

vitaut:
Anyway, this is a straightforward fix so I'm merging this in but a PR with a test would be welcome.

PR #2740, opened 2022-01-30
Adding a put API to basic_format_context.

BRevzin:
This is inspired by Peter Dimov's email. tl;dr This PR adds an API onto basic_format_context that makes it more convenient and, more importantly, potentially more performant.

---

A lot of the time when we're formatting, we're either copying a single character or a string_view into the output iterator. So you get a bunch of either:

```cpp
auto out = ctx.out();
*out++ = c;
out = std::ranges::copy("whatever"sv, out).out;
```

This works fine. But it could be better. The C++ output iterator model is very limited to just writing one thing at a time and doesn't have range support -- but if we (the implementation) know what our iterator is, we can do better than that if we just expose that kind of ability to the user.

{fmt} already has `copy_str` which takes advantage of that knowledge.

This PR simply exposes that to the user by way of a member functions on the format_context that take either a char or a string_view (maybe this should be an arbitrary range instead or something?). I named it `put`  after D's output range model (and this library already has enough things named write). The `put` API should be at least as efficient as what the user could write.

As an added bonus, occasionally you need to also go through an underlying formatter. At that point, you have to carefully:

```cpp
ctx.advance_to(out);
out = underlying.format(some_object, ctx);
*out++ = '};
return out;
```

This back and forth is a little jarring, so I also added a `put` overload that takes a `formatter<T>` and a `T`/`const T` so that you can just do:
```cpp
ctx.put(c);
ctx.put("whatever"sv); // or even just "whatever" if we stick with string_view
ctx.put(underlying, some_object);
ctx.put('}');
return ctx.out();
```

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -1776,6 +1776,15 @@ template <typename OutputIt, typename Char> class basic_format_context {
     if (!detail::is_back_insert_iterator<iterator>()) out_ = it;
   }
 
+  FMT_CONSTEXPR auto put(char_type c) -> iterator {
+    *out_++ = c;
+    return out;
+  }
+  FMT_CONSTEXPR auto put(basic_string_view<char_type> sv) -> iterator {
+    detail::copy_str<char_type>(sv.data(), sv.data() + sv.size(), out);
+    return out;
+  }
+
   FMT_CONSTEXPR auto locale() -> detail::locale_ref { return loc_; }
 };
 
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -1778,11 +1778,18 @@ template <typename OutputIt, typename Char> class basic_format_context {
 
   FMT_CONSTEXPR auto put(char_type c) -> iterator {
     *out_++ = c;
-    return out;
+    return out_;
   }
   FMT_CONSTEXPR auto put(basic_string_view<char_type> sv) -> iterator {
-    detail::copy_str<char_type>(sv.data(), sv.data() + sv.size(), out);
-    return out;
+    detail::copy_str<char_type>(sv.data(), sv.data() + sv.size(), out_);
+    return out_;
+  }
+
+  template <typename T, typename U,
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR but it doesn't belong in the context and an optimized version of copy should be provided instead.

pdimov:
It's not (just) a matter of optimization; formatter specializations are much easier to write with this API, as the test case shows.

vitaut:
Maybe with the nested formatter but that's very uncommon. The rest is much cleaner with iterators IMHO. Users don't have to learn yet another API to copy things =).

pdimov:
And when you said "an optimized version of copy should be provided instead" what exactly did you mean? Who should provide the optimized copy, and in what form?

pdimov:
To pick a concrete example to discuss: https://godbolt.org/z/Mn8oMsxzo

In principle this simple case can be done with a single `format_to` call, but it's simplified from a generic struct printer that obtains the member names using reflection or "reflection".

It's clear how using the proposed `put` simplifies it. Is there a more idiomatic approach? And is `std::copy` supposed to contain the optimized path?

vitaut:
We could provide a drop-in replacement for `std::copy` which is optimized for buffer iterators.

> In principle this simple case can be done with a single format_to call

And that's what should normally be used. If we had a public buffer API then we could potentially add some operations there but it has been decided that the buffer should be hidden behind the iterator facade so that's what should be used.

> And is std::copy supposed to contain the optimized path?

Eventually yes.

pdimov:
`format_to` is what I ended up using: https://github.com/boostorg/describe/blob/develop/example/fmtlib_formatter.cpp

The codegen will not win any awards but it's good enough for government use.

Extending this to handle `:L` or `:?` correctly - that is, pass them down to the member when its formatter supports it, and not when not - is decidedly nontrivial, primarily because there's no way to ask a formatter whether it supports something.

PR #2742, opened 2022-02-01
qualify unqualified calls to format in compile.h

barcharcraz:
Now that both libfmt and the STL have implemented P2418 our formats are no longer worse (w.r.t. overload resolution) than fmt's formats. Hilarity ensues. (libfmt probably gets `<format>` via `<chrono>`, I'm working on fixing that too, but even with the fix these calls should be qualified in case someone wants to use fmt and gets `<format>` through some other path).

Fully qualifies the problematic functions calls to avoid ambiguity.


vitaut:
Thank you!

PR #2750, opened 2022-02-09
Reflect the new paper on Dragonbox

jk-jeon:
The new paper is here: https://github.com/jk-jeon/dragonbox/blob/master/other_files/Dragonbox.pdf
This PR is mostly the reflection of the new paper that has not been done in the previous PR.

1. Implemented the "divisibility check and divide" trick based on Lemire's paper mentioned in the previous PR.
2. Magic numbers for log / division computations are changed. I removed binary expansions of log10(2) and friends because now the magic numbers a priori do not need to have anything to do with the binary expansions. More comprehensive explanation can be found in the paper.
3.  In the new paper, we define beta to be one less of beta from the previous paper. So I renamed beta_minus_1 into beta.
4. The branch that removes the trailing zeros is the branch chosen for most short inputs, so it would be better to check for that branch first.
5. I did some experiments on trailing zero removal, and concluded that the implementation in this PR is the fastest. Separating the divisibility checks for powers of 2 and powers of 5 does not seem to be a good idea, so I removed the usage of `ctz` and friends, and instead introduced `rotr`. I think Dragonbox is the only one who uses `ctz` in fmt, so probably you can remove `ctz` and friends from fmt now. Also, interestingly simple loop seems to outperform binary search. I think this makes a lot of sense for binary64 because 64-bit constants in general must be first loaded into register and cannot be used as immediates, but it is quite surprising that even for binary32 still loop is better. Looking at the assembly, it seems that the reason is because for the binary search, the compiler is trying too hard to get rid of branches and ends up generating more instructions. However, it should be noted that now `remove_trailing_zeros` traps into infinite loop if the input is `0`. I believe `0` cannot be fed into `remove_trailing_zeros` in the current implementation, but you should make sure nobody feeds `0` into Dragonbox.


include/fmt/format-inl.h
```diff
@@ -819,6 +817,16 @@ struct uint128_wrapper {
   }
 };
 
+// Compilers should be able to optimize this into the ror instruction.
+inline std::uint32_t rotr(uint32_t n, uint32_t r) noexcept {
```
vitaut, line 821: nit: I suggest dropping `std::` in the return type for consistency (and the same in the overload below).

include/fmt/format-inl.h
```diff
@@ -895,86 +903,72 @@ inline uint64_t umul96_lower64(uint32_t x, uint64_t y) noexcept {
 // Computes floor(log10(pow(2, e))) for e in [-1700, 1700] using the method from
```
vitaut, line 903: [-1700, 1700] -> [-2620, 2620]

include/fmt/format-inl.h
```diff
@@ -895,86 +903,72 @@ inline uint64_t umul96_lower64(uint32_t x, uint64_t y) noexcept {
...
 // https://fmt.dev/papers/Grisu-Exact.pdf#page=5, section 3.4.
```
vitaut, line 904: I guess we should link to the Dragonbox paper now.

include/fmt/format-inl.h
```diff
@@ -1845,115 +1836,86 @@ bool is_left_endpoint_integer_shorter_interval(int exponent) noexcept {
...
-    if (n * mod_inv2 > max_quotient2) break;
-    n *= mod_inv2;
+  while (true) {
+    auto q = rotr(n * mod_inv_25, 2);
+    if (q <= std::numeric_limits<uint32_t>::max() / 100) {
+      n = q;
+      s += 2;
+    } else {
+      break;
+    }
```
vitaut, line 1851: Why not use early exit, i.e.

```c++
      if (q > std::numeric_limits<uint32_t>::max() / 100) break;
      n = q;
      s += 2;
```
?

vitaut (commented):
Thanks a lot for the PR! Looks great, just a few minor comments inline.

include/fmt/format-inl.h
jk-jeon: @vitaut Sorry, I think I was in rush for some reason 😅 corrected now.
@miscco Added, thanks for pointing out.

include/fmt/format-inl.h
jk-jeon: Corrected.

include/fmt/format-inl.h
jk-jeon: Yeah, I linked to your copy of the paper (fmt.dev/papers/Dragonbox.pdf) assuming you will replace the copy in a near future.

include/fmt/format-inl.h
jk-jeon: Great suggestion, modified the code according to it.

include/fmt/format-inl.h
phprus: @jk-jeon
``FMT_CONSTEXPR`` - macro for C++14 constexpr.

jk-jeon: @phprus  oooohhh didn't know that, thanks for letting me know=)

vitaut:
Merged, thank you!

vitaut:
@jk-jeon, have you thought about handling fixed precision in Dragonbox? Is it even possible, at least for small precision?

jk-jeon:
>@jk-jeon, have you thought about handling fixed precision in Dragonbox?

I haven't, at least recently.

>Is it even possible, at least for small precision?

Might be, but I'm not sure. I don't even know how do you do that with Grisu 😋.

jk-jeon:
Hi Victor,

I'm recently giving some shots on fixed-precision formatting in my spare time. A good news is that I think it is possible to come up with a configurable method of trading between the cache table size and performance. What do you think is the maximum size of the table (which will be there in addition to the Dragonbox table) which `fmt` can afford?

vitaut:
Ideally the smaller the better but I think we can live with a few more kiB of data. There is no exact budget.

jk-jeon:
So I've been thinking about what can I do with the fixed-precision case. Here is what I thought.

The algorithm I recently advertised is composed of two parts, one for the first few digits (covered by the Dragonbox table) and another for further digits (covered by an additional table). I think it is probably too early to adopt the second part, but meanwhile we can discuss the adoption of the first part.

Assuming `double`, for normal inputs, the first part can generate at least 18 digits (up to 19 digits, which is the largest number of digits for 64-bit integers), and for subnormal inputs, it can generate at least 3 digits. It is actually possible to make it to generate 18 digits even for the subnormal inputs, if we augment the Dragonbox table a little bit. (I think the compressed version of the table only need to have one more entry. Or maybe two more.)

I don't know if this adoption will make the performance better or worse though. Grisu uses fewer precision so it might be faster. But we do not need to rely on Dragon4 fallback anymore if the number of digits is at most 18. Also we can get rid of the Grisu table.

What do you think?

Thanks.

vitaut:
This sounds like a good plan. The performance will be more predictable without the fallback for the common case and we could get rid of Grisu tables which is an improvement. A regression for the non-fallback case is OK if it's not too big.

Issue #2746, opened 2022-02-08
Static Assert failure with intel-19


gsjaardema:
There seems to be an issue with the intel-19 c++ compiler and the use of `fmt::group_digits()`.  A simple reproducer is below:
```
>>> cat test-1.C
#include <fmt/format.h>
#include <fmt/ostream.h>
#include <string>
#include <sstream>

int main()
{
  std::ostringstream errmsg;
  int age = 56;
  fmt::print(errmsg, "I am {} years old\n", fmt::group_digits(age));
}
```
I try to compile:
```
>>>> icpc -I../fmt_copy test-1.C
../fmt_copy/fmt/core.h(1706): error: static assertion failed with "Cannot format a const argument."
    static_assert(formattable_const, "Cannot format a const argument.");
    ^
          detected during:
            instantiation of "auto fmt::v8::detail::make_arg<IS_PACKED,Context,<unnamed>,T,<unnamed>>(T &&)->fmt::v8::detail::value<Context> [with IS_PACKED=true, Context=fmt::v8::basic_format_context<fmt::v8::detail::buffer_appender<char>, char>, <unnamed>=fmt::v8::detail::type::custom_type, T=const fmt::v8::remove_cvref_t<const fmt::v8::remove_reference_t<fmt::v8::group_digits_view<int>> &> &, <unnamed>=0]" at line 1848
            instantiation of "fmt::v8::format_arg_store<Context, Args...>::format_arg_store(T &&...) [with Context=fmt::v8::basic_format_context<fmt::v8::detail::buffer_appender<char>, char>, Args=<fmt::v8::group_digits_view<int>>, T=<const fmt::v8::remove_reference_t<fmt::v8::group_digits_view<int>> &>]" at line 850 of "../fmt_copy/fmt/format.h"
            instantiation of "auto fmt::v8::make_args_checked<Args...,S,Char>(const S &, const fmt::v8::remove_reference_t<Args> &...)->fmt::v8::format_arg_store<fmt::v8::buffer_context<Char>, fmt::v8::remove_reference_t<Args>...> [with Args=<fmt::v8::group_digits_view<int>>, S=char [19], Char=char]" at line 131 of "../fmt_copy/fmt/ostream.h"
            instantiation of "void fmt::v8::print(std::basic_ostream<Char, std::char_traits<Char>> &, const S &, Args &&...) [with S=char [19], Args=<fmt::v8::group_digits_view<int>>, Char=char]" at line 10 of "test-1.C"

[...]

PR #2760, opened 2022-02-12
Replace ``make_args_checked`` with ``make_format_args``

phprus:
Replace ``make_args_checked`` with ``make_format_args`` and remove legacy ``make_args_checked``.

See: https://github.com/fmtlib/fmt/issues/2746#issuecomment-1033347178

@vitaut, please review these changes. I am not sure of their correctness.

### DIFF
--- a/include/fmt/color.h
+++ b/include/fmt/color.h
@@ -547,7 +547,7 @@ template <typename S, typename... Args,
 void print(std::FILE* f, const text_style& ts, const S& format_str,
            const Args&... args) {
   vprint(f, ts, format_str,
-         fmt::make_args_checked<Args...>(format_str, args...));
+         fmt::make_format_args<buffer_context<char_t<S>>>(args...));
 }
 
 /**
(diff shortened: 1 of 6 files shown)


include/fmt/color.h
```diff
@@ -547,7 +547,8 @@ template <typename S, typename... Args,
 void print(std::FILE* f, const text_style& ts, const S& format_str,
            const Args&... args) {
   vprint(f, ts, format_str,
-         fmt::make_args_checked<Args...>(format_str, args...));
+         fmt::make_format_args<buffer_context<type_identity_t<char_t<S>>>>(
```
vitaut, line 550: I don't think `type_identity_t` is needed here and in similar cases below.

include/fmt/format.h
```diff
@@ -825,27 +825,6 @@ class FMT_API format_error : public std::runtime_error {
...
-                                  const remove_reference_t<Args>&... args)
-    -> format_arg_store<buffer_context<Char>, remove_reference_t<Args>...> {
-  static_assert(
-      detail::count<(
-              std::is_base_of<detail::view, remove_reference_t<Args>>::value &&
-              std::is_reference<Args>::value)...>() == 0,
-      "passing views as lvalues is disallowed");
-  detail::check_format_string<Args...>(fmt);
-  return {args...};
-}
```
vitaut, line 847: Let's mark it as deprecated to avoid breaking user code until the next major release.

phprus:
@vitaut done

vitaut (approved):

PR #2770, opened 2022-02-16
Apply 2746 fix for NVidia compiler also

gsjaardema:
The same fix that was applied to address #2746 for the intel compiler is also needed for NVidia nvcc compiler.  Tested on nvcc-11 and 10.


vitaut:
Thank you

PR #2772, opened 2022-02-16
add ability to build library as Apple framework

kambala-decapitator:
Building as Apple framework is achieved via `-DCMAKE_FRAMEWORK` flag on the command line. Tested by building for macOS and iOS.

Besides that I also modified the way public headers are installed because without that there's a configure time warning:

> INSTALL TARGETS - target fmt has PUBLIC_HEADER files but no PUBLIC_HEADER DESTINATION.


vitaut:
Thank you

PR #2780, opened 2022-02-19
Update format.h

JurajX:
To prevent the following compiler message:
```
[...]/include/fmt/format.h:392:20: warning: conversion to 'long unsigned int' from 'int' may change the sign of the result [-Wsign-conversion]
  392 |     unsigned value[size];
      |                    ^~~~
```

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -389,7 +389,7 @@ template <typename To, typename From, FMT_ENABLE_IF(sizeof(To) > sizeof(From))>
 inline auto bit_cast(const From& from) -> To {
   constexpr auto size = static_cast<int>(sizeof(From) / sizeof(unsigned));
   struct data_t {
-    unsigned value[static_cast<size_t>(size)];
+    unsigned value[to_unsigned(size)];
   } data = bit_cast<data_t>(from);
   auto result = To();
   if (const_check(is_big_endian())) {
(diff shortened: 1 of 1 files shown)


include/fmt/format.h
```diff
@@ -389,7 +389,7 @@ template <typename To, typename From, FMT_ENABLE_IF(sizeof(To) > sizeof(From))>
 inline auto bit_cast(const From& from) -> To {
   constexpr auto size = static_cast<int>(sizeof(From) / sizeof(unsigned));
   struct data_t {
-    unsigned value[size];
+    unsigned value[static_cast<size_t>(size)];
```
vitaut, line 392: Let's use to_unsigned instead of static_cast

vitaut (commented):
Thank you for the PR.

include/fmt/format.h
JurajX: Sure :)
In case I did something badly, feel free to fix it directly and close this PR. ;)

phprus: @vitaut ``to_unsigned`` is constexpr only since C++14.
I think simple ``static_cast`` is better here.

vitaut: Right. I switched back to cast and merged.

vitaut:
Merged in https://github.com/fmtlib/fmt/commit/70dc3de053b6814c023faebbd21525aab575d25d, thanks.

PR #2787, opened 2022-02-22
Fixing formatting of certain kinds of ranges of ranges.

brevzin:
If you had a range of `R`, where `R` is a const-iterable range, then the `formatter` would expect its argument to be passed as a `const R&` but the implementation would end up passing an `R&` instead (the `fmt_ref_view` case fails to compile on master). 

But we don't need to deduce the range and check that it has the right const-ness, we can just declare it to have the expected type instead.  If that type is `const R&`, then that just works.

### DIFF
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -367,9 +367,9 @@ struct formatter<
 #endif
         >> {
 
-  using const_range = detail::maybe_const_range<R>;
+  using range_type = detail::maybe_const_range<R>;
   using formatter_type =
-      detail::range_formatter_type<Char, detail::uncvref_type<const_range>>;
+      detail::range_formatter_type<Char, detail::uncvref_type<range_type>>;
   formatter_type underlying_;
   bool custom_specs_ = false;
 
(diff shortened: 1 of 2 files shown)


include/fmt/ranges.h
```diff
@@ -340,29 +340,36 @@ template <typename Context, typename Element> struct range_mapper {
...
+        &&
+        (is_formattable<detail::uncvref_type<detail::maybe_const_range<R>>,
+                        Char>::value ||
+         detail::has_fallback_formatter<
+             detail::uncvref_type<detail::maybe_const_range<R>>, Char>::value)
 #endif
-                >> {
+        >> {
 
+  using const_range = detail::maybe_const_range<R>;
```
vitaut, line 370: `const_range` seems a bit misleading considering that it can be non-const. Maybe rename to `range_type`?

vitaut (commented):
Thanks for the PR.

include/fmt/ranges.h
brevzin: Agreed. Renamed.

vitaut:
Please rebase, I've fixed the `apt install` issue.

brevzin:
> Please rebase, I've fixed the `apt install` issue.

Done 👍

vitaut:
Thank you!

PR #2789, opened 2022-02-23
fix typo

setoye:


OnurKader:
It was changed to be square brackets, maybe the godbolt link should be updated? [godbolt](https://godbolt.org/z/MxM1YqjE7)

setoye:
fixed.

vitaut:
Thank you

PR #2793, opened 2022-03-02
Implement styled arguments

rbrugo:
Hello,
this PR implements a way to apply different styles to different arguments in a single `format` (or `print`) call.
It is done by introducing a new function, `fmt::styled`, which takes an argument of arbitrary type and a color, an `emphasis` or a `text_style`, and creates a wrapper around the argument that will be formatted it in the requested style.

A simple example of usage:
```
void log(std::string_view caller, std::string_view message, bool error = false)
{
    fmt::print("[{}] {}: {}\n",
                fmt::styled(std::chrono::system_clock::now(), fmt::emphasis::bold),
                fmt::styled(caller, error ? fmt::color::red : fmt::color::green),
                message
    );
}
```

### DIFF
--- a/include/fmt/color.h
+++ b/include/fmt/color.h
@@ -644,29 +644,29 @@ struct formatter<detail::styled_arg<T>, Char> : formatter<T, Char> {
     const auto& value = arg.value;
     auto out = ctx.out();
 
-    using detail::get_buffer;
-    auto&& buf = get_buffer<Char>(out);
-
     bool has_style = false;
     if (ts.has_emphasis()) {
       has_style = true;
       auto emphasis = detail::make_emphasis<Char>(ts.get_emphasis());
-      buf.append(emphasis.begin(), emphasis.end());
+      out = std::copy(emphasis.begin(), emphasis.end(), out);
     }
     if (ts.has_foreground()) {
       has_style = true;
       auto foreground =
           detail::make_foreground_color<Char>(ts.get_foreground());
-      buf.append(foreground.begin(), foreground.end());
+      out = std::copy(foreground.begin(), foreground.end(), out);
     }
     if (ts.has_background()) {
       has_style = true;
       auto background =
           detail::make_background_color<Char>(ts.get_background());
-      buf.append(background.begin(), background.end());
+      out = std::copy(background.begin(), background.end(), out);
     }
     out = formatter<T, Char>::format(value, ctx);
-    if (has_style) detail::reset_color<Char>(buf);
+    if (has_style) {
+      auto reset_color = string_view("\x1b[0m");
+      out = std::copy(reset_color.begin(), reset_color.end(), out);
+    }
     return out;
   }
 };
--- a/include/fmt/color.h
+++ b/include/fmt/color.h
@@ -643,22 +643,7 @@ template <typename Arg> struct styled_arg {
 FMT_END_DETAIL_NAMESPACE
 
 template <typename Arg, typename Char>
-struct formatter<detail::styled_arg<Arg>, Char> {
- private:
-  using value_type = Arg;
-  using formatter_type =
-      conditional_t<is_formattable<value_type, Char>::value,
-                    formatter<remove_cvref_t<value_type>, Char>,
-                    detail::fallback_formatter<value_type, Char>>;
-
-  formatter_type value_formatter_;
-
- public:
-  template <typename ParseContext>
-  FMT_CONSTEXPR auto parse(ParseContext& ctx) -> decltype(ctx.begin()) {
-    return value_formatter_.parse(ctx);
-  }
-
+struct formatter<detail::styled_arg<Arg>, Char> : formatter<Arg, Char> {
   template <typename FormatContext>
   auto format(detail::styled_arg<Arg> const& arg, FormatContext& ctx) const
       -> decltype(ctx.out()) {
@@ -687,7 +672,7 @@ struct formatter<detail::styled_arg<Arg>, Char> {
           detail::make_background_color<Char>(ts.get_background());
       buf.append(background.begin(), background.end());
     }
-    out = value_formatter_.format(value, ctx);
+    out = formatter<Arg, Char>::format(value, ctx);
     if (has_style) detail::reset_color<Char>(buf);
     return out;
   }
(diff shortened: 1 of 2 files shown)


include/fmt/color.h
```diff
@@ -630,6 +630,118 @@ inline auto format_to(OutputIt out, const text_style& ts, const S& format_str,
...
+
+template <typename Arg> struct styled_arg {
+  FMT_CONSTEXPR styled_arg(Arg const& argument, text_style style)
+      : argument(argument), style(style) {}
+
+  const Arg& argument;
+  text_style style;
+};
+
+FMT_END_DETAIL_NAMESPACE
```
vitaut, line 643: Could you move the definition of `styled_arg` to one of the `detail` blocks earlier in this file instead of introducing another one?

include/fmt/color.h
```diff
@@ -630,6 +630,118 @@ inline auto format_to(OutputIt out, const text_style& ts, const S& format_str,
...
+
+template <typename Arg, typename Char>
+struct formatter<detail::styled_arg<Arg>, Char> {
+ private:
+  using value_type = Arg;
+  using formatter_type =
+      conditional_t<is_formattable<value_type, Char>::value,
+                    formatter<remove_cvref_t<value_type>, Char>,
+                    detail::fallback_formatter<value_type, Char>>;
```
vitaut, line 652: `fallback_formatter` is being removed so please don't use it here.

vitaut (changes requested):
Thanks for the PR! Overall looks good but please rebase past the CI fix and address inline comments.

include/fmt/color.h
```diff
@@ -630,6 +630,118 @@ inline auto format_to(OutputIt out, const text_style& ts, const S& format_str,
...
+
+  formatter_type value_formatter_;
```
vitaut, line 654: I suggest inheriting from `formatter<Arg>`. Then we won't need `parse` below.

rbrugo:
Everything should be solved now!
Do you think something else needs to be changed?

include/fmt/color.h
```diff
@@ -630,6 +638,91 @@ inline auto format_to(OutputIt out, const text_style& ts, const S& format_str,
                     fmt::make_format_args<buffer_context<char_t<S>>>(args...));
 }
 
+template <typename Arg, typename Char>
+struct formatter<detail::styled_arg<Arg>, Char> : formatter<Arg, Char> {
+  template <typename FormatContext>
+  auto format(detail::styled_arg<Arg> const& arg, FormatContext& ctx) const
```
alexezeder, line 644: I'm not sure about `const` on the right side, @vitaut was not very happy about that, IIRC 😏

rbrugo: Good catch!

include/fmt/color.h
```diff
@@ -492,6 +492,14 @@ template <typename Char> inline void reset_color(buffer<Char>& buffer) {
   buffer.append(reset_color.begin(), reset_color.end());
 }
 
+template <typename Arg> struct styled_arg {
+  FMT_CONSTEXPR styled_arg(const Arg& format_argument, text_style format_style)
+      : argument(format_argument), style(format_style) {}
```
vitaut, line 497: I don't think the constructor is needed because you could use aggregate initialization `styled_arg{...}` instead.

include/fmt/color.h
```diff
@@ -492,6 +492,14 @@ template <typename Char> inline void reset_color(buffer<Char>& buffer) {
...
+
+  const Arg& argument;
```
vitaut, line 499: nit: I'd rename `Arg` and `argument` to `T` and `value` respectively because these are just arbitrary types values. What makes them args is the fact that they are passed into a formatting function.

include/fmt/color.h
rbrugo: You're right. I'm changing this right now

include/fmt/color.h
```diff
@@ -630,6 +635,91 @@ inline auto format_to(OutputIt out, const text_style& ts, const S& format_str,
...
+/**
+  \rst
+  Returns an argument that will be formatted using ANSI escape sequences,
+  to be used in a formatting function.
+
+  **Example**::
+
+    fmt::print("Elapsed time: {s:.2f} seconds",
+               fmt::styled(1.23, fmt::fg(fmt::colors::green) |
+  fmt::bg(fmt::color::blue))); \endrst
```
vitaut, line 683: Here and below: formatting is broken.

vitaut (changes requested):

include/fmt/color.h
```diff
@@ -630,6 +635,91 @@ inline auto format_to(OutputIt out, const text_style& ts, const S& format_str,
...
+  **Example**::
+
+    fmt::print("Elapsed time: {s:.2f} seconds", fmt::styled(1.23,
+  fmt::colors::green)); \endrst
+ */
+template <typename T>
+FMT_CONSTEXPR auto styled(const T& value, detail::color_type color)
+    -> detail::styled_arg<remove_cvref_t<T>> {
+  return detail::styled_arg<remove_cvref_t<T>>{value, fg(color)};
+}
```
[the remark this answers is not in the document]
rbrugo: I am aware of `fg`/`bg`, I decided to add this overload anyway just to make what I believe to be the most common use-case a little shorter.
I'm going to remove it anyway if you think it is not worth it

include/fmt/color.h
```diff
@@ -630,6 +635,59 @@ inline auto format_to(OutputIt out, const text_style& ts, const S& format_str,
...
+  to be used in a formatting function.
+
+  **Example**::
+
+    fmt::print("Elapsed time: {s:.2f} seconds",
+               fmt::styled(1.23, fmt::fg(fmt::colors::green) | fmt::bg(fmt::color::blue)));
+  \endrst
+ */
+template <typename T>
+FMT_CONSTEXPR auto styled(const T& value, text_style ts = {})
```
vitaut, line 686: And finally I don't think ts should be defaulted.

vitaut (approved):

vitaut:
Thank you!

vitaut:
Could you also add it to the docs (https://github.com/fmtlib/fmt/blob/master/doc/api.rst)?

rbrugo:
Sure!

PR #2801, opened 2022-03-07
Misc warnings

seanm:


include/fmt/compile.h
```diff
@@ -36,7 +36,7 @@ template <typename OutputIt> class truncating_iterator_base {
   using difference_type = std::ptrdiff_t;
   using pointer = void;
   using reference = void;
-  using _Unchecked_type =
+  using Unchecked_type_ =
```
vitaut, line 39: This will introduce warnings in MSVC, please revert.

include/fmt/core.h
```diff
@@ -1498,7 +1498,7 @@ class appender : public std::back_insert_iterator<detail::buffer<char>> {
  public:
   using std::back_insert_iterator<detail::buffer<char>>::back_insert_iterator;
   appender(base it) noexcept : base(it) {}
-  using _Unchecked_type = appender;  // Mark iterator as checked.
+  using Unchecked_type_ = appender;  // Mark iterator as checked.
```
vitaut, line 1501: same here

vitaut (changes requested):
Thanks for the PR.

include/fmt/compile.h
seanm: What warning would it introduce?

Underscore followed by uppercase is reserved in C++: https://en.cppreference.com/w/cpp/language/identifiers

vitaut: Warnings about (un)checked iterators.

seanm: Is `_Unchecked_type` some kind of Microsoft extension keyword?

vitaut: Sort of. It's not a keyword but some mechanism to opt out iterators.

seanm: Could we somewhere do something like:

```c++
#if defined(_MSC_VER)
  #define fmt_unchecked_type _Unchecked_type
#else
  #define fmt_unchecked_type
#endif
```

test/test-main.cc
```diff
@@ -27,12 +24,12 @@ int main(int argc, char** argv) {
   // doesn't help.
   SetErrorMode(SEM_FAILCRITICALERRORS | SEM_NOGPFAULTERRORBOX |
                SEM_NOOPENFILEERRORBOX);
-#endif
   // Disable message boxes on assertion failures.
   _CrtSetReportMode(_CRT_ERROR, _CRTDBG_MODE_FILE | _CRTDBG_MODE_DEBUG);
   _CrtSetReportFile(_CRT_ERROR, _CRTDBG_FILE_STDERR);
   _CrtSetReportMode(_CRT_ASSERT, _CRTDBG_MODE_FILE | _CRTDBG_MODE_DEBUG);
   _CrtSetReportFile(_CRT_ASSERT, _CRTDBG_FILE_STDERR);
+#endif
```
phprus, line 32: Different conditions in ``#if``.

Line 16:
```
#ifdef _MSC_VER
```

Line 24:
```
#ifdef _WIN32
```

include/fmt/compile.h
vitaut: If you have access to MSVC 2015, I'd recommend checking if removing `_Unchecked_type` altogether still produces warnings. It might be no longer the case as IIRC they have been moving away from the checked iterators. You could even submit a test PR to the fmt repo and check warnings in appveyor: https://ci.appveyor.com/project/vitaut/fmt.

vitaut: Actually, we can see on this PR that there are definitely more warnings: https://ci.appveyor.com/project/vitaut/fmt/builds/42816914/job/briwktqjiiw9oa34. The macro approach seems OK but not that macros should be uppercase.

test/test-main.cc
seanm: OK, changed.

include/fmt/compile.h
seanm: OK, see new attempt.

include/fmt/core.h
```diff
@@ -2002,11 +2008,11 @@ using format_args = basic_format_args<format_context>;
 // We cannot use enum classes as bit fields because of a gcc bug
 // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61414.
 namespace align {
-enum type { none, left, right, center, numeric };
+enum type : unsigned { none, left, right, center, numeric };
```
vitaut, line 2011: Why unsigned?

test/posix-mock-test.cc
```diff
@@ -6,7 +6,7 @@
 // For the license information refer to format.h.
 
 // Disable bogus MSVC warnings.
-#ifndef _CRT_SECURE_NO_WARNINGS
+#if !defined(_CRT_SECURE_NO_WARNINGS) && defined(_MSC_VER)
```
vitaut, line 9: Why check if _MSC_VER is defined?

vitaut (changes requested):

test/posix-mock-test.cc
seanm: As per commit message: to fix the -Wreserved-identifier warnings.  Underscore + uppercase is reserved, and #defineing such a name causes a warning.  No need to define in on unix since it won't do anything, thus the warning is avoided.

include/fmt/core.h
seanm: As per commit message: to fix the -Wsigned-enum-bitfield warnings.

include/fmt/compile.h
```diff
@@ -8,6 +8,7 @@
 #ifndef FMT_COMPILE_H_
 #define FMT_COMPILE_H_
 
+#include "core.h"
```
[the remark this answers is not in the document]
seanm: Removed.

include/fmt/core.h
```diff
@@ -217,6 +217,12 @@
 #  endif
 #endif
 
+#ifdef _MSC_VER
+#  define FMT_UNCHECKED_TYPE _Unchecked_type
+#else
+#  define FMT_UNCHECKED_TYPE DummyName
+#endif
```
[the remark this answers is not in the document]
seanm: Done.

include/fmt/compile.h
```diff
@@ -36,7 +36,7 @@ template <typename OutputIt> class truncating_iterator_base {
   using difference_type = std::ptrdiff_t;
   using pointer = void;
   using reference = void;
-  using _Unchecked_type =
+  using FMT_UNCHECKED_TYPE =
       truncating_iterator_base;  // Mark iterator as checked.
```
vitaut, line 40: This should be `FMT_UNCHECKED_TYPE(truncating_iterator_base);

include/fmt/core.h
```diff
@@ -217,6 +217,12 @@
 #  endif
 #endif
 
+#ifdef _MSC_VER
+#  define FMT_UNCHECKED_ITERATOR(It) using _Unchecked_type = It; // Mark iterator as checked.
+#else
+#  define FMT_UNCHECKED_ITERATOR(It) using DummyTypeName = It;
```
vitaut, line 223: Let's remove trailing semicolons since those are added at call sites.

test/posix-mock-test.cc
vitaut: But it's not an identifier. What compiler gives this warning?

vitaut (changes requested):

test/posix-mock-test.cc
seanm: clang.

```
% cat test.cxx
#define _FOO

int main (void)
{
	return 0;
}


% clang -fsyntax-only -Weverything ~/Desktop/test.cxx 
/Users/sean/test.cxx:1:9: warning: macro name is a reserved identifier [-Wreserved-macro-identifier]
#define _FOO
        ^
```

include/fmt/compile.h
seanm: Doh, sorry I missed that one. Fixed.

include/fmt/core.h
seanm: Ah, good point. Fixed.

include/fmt/compile.h
seanm: ...Oh, wait, it's a different keyword...

vitaut:
Please fix build failures.

seanm:
> Please fix build failures.

I don't understand that error.  I played on godbolt, and it seems to me like a gcc bug fixed between 9.2 and 9.3:

https://godbolt.org/z/58aEv8zEq

Interesting, the warning goes away if I change the enum from `unsigned` to `unsigned char`.  Shall we do that?

vitaut:
Do you still plan working on this PR?

seanm:
> Do you still plan working on this PR?

Yes. But I'm at a loss as to why that old gcc complains.  I don't have a means to reproduce it locally.

I guess I could split that one commit out so the rest can be merged...

vitaut:
Thanks

PR #2802, opened 2022-03-07
Eliminate intel compiler warnings

gsjaardema:
The intel compiler emits  `declaration hides parameter "loc" ..."`  and similar warnings for "out" and "args" since the function arguments are the same as other visible symbols.  This is for intel-2021.3.0 20210609.


vitaut:
Thanks

PR #2803, opened 2022-03-08
Document output_file default behavior correctly

rtobar:
The documentation for output_file incorrectly says that opening an output file happens with `file::WRONLY | file::CREATE` by default, but this includes `file::TRUNC` is also included since 119f7dc3d.


vitaut:
Thank you

PR #2804, opened 2022-03-08
Add `styled` in documentation

rbrugo:
As requested in #2793


vitaut:
There are seem to be a bunch of older commits. Could you rebase?

PR #2807, opened 2022-03-11
Fix compilation error for ranges with ADL `begin`/`end`

rbrugo:
This PR should fix an issue that occurs when the range `formatter` is unable to use `begin` and `end` found by ADL.
For example, this code currently fails to compile:
```cpp
namespace adl
{
struct box
{
    int value;
};

int * begin(box & b) noexcept { return std::addressof(b.value); }
int * end(box & b)   noexcept { return std::addressof(b.value) + 1; }

int const * begin(box const & b) noexcept { return std::addressof(b.value); }
int const * end(box const & b)   noexcept { return std::addressof(b.value) + 1; }
} // namespace adl

int main()
{
    auto x = adl::box{42};
    
    // this works
    for (auto elem : x) {
        fmt::print("{}\n", elem);
    }
    
    // this compiles correctly
    static_assert(fmt::is_range<adl::box, char>::value, "");

    // but this does not compile
    fmt::print("{}\n", x);
}
```
but with the proposed patch works as expected.

### DIFF
--- a/test/ranges-test.cc
+++ b/test/ranges-test.cc
@@ -65,32 +65,27 @@ TEST(ranges_test, format_set) {
             "{\"one\", \"two\"}");
 }
 
-namespace adl
-{
-    struct box {
-        int value;
-    };
+namespace adl {
+struct box {
+  int value;
+};
 
-    auto begin(box & b) noexcept -> int * {
-        return std::addressof(b.value);
-    }
+auto begin(box& b) noexcept -> int* { return std::addressof(b.value); }
 
-    auto end(box & b) noexcept -> int * {
-        return std::addressof(b.value) + 1;
-    }
+auto end(box& b) noexcept -> int* { return std::addressof(b.value) + 1; }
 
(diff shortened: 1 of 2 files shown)


phprus:
please apply clang-format

rbrugo:
I completely forgot to run it on the tests. Thank you for pointing out.

test/ranges-test.cc
```diff
@@ -65,6 +65,29 @@ TEST(ranges_test, format_set) {
...
+  int value;
+};
+
+auto begin(box& b) noexcept -> int* { return std::addressof(b.value); }
+
+auto end(box& b) noexcept -> int* { return std::addressof(b.value) + 1; }
+
+auto begin(const box& b) noexcept -> const int* {
+  return std::addressof(b.value);
+}
```
vitaut, line 79: noexcept is unnecessary, this is just a test. Also let's replace addressof with `&` for simplicity.

vitaut (approved):

vitaut:
Thank you!

Issue #1077, opened 2019-03-12
Support systems without wchar.h / std::wstring


RedDwarf69:
I know it's hard to believe, but I do have a system with C++11 support... but no wchar.h in its C standard library, which makes GNU libstdc++ build correctly but without wstring.

Would it be possible to add a flag to disable wchar/wstring support from fmt, so it can be used in such a system?

vitaut:
> Would it be possible to add a flag to disable wchar/wstring support from fmt, so it can be used in such a system?

If you are willing to do the work then sure =). It should be mostly a matter of adding a conditionally compiled type alias for `wchar_t` in https://github.com/fmtlib/fmt/blob/master/include/fmt/core.h.

vitaut:
Closing as there are no current plans to support this, but PRs are welcome.

PR #2820, opened 2022-03-18
djgpp compatibility

jwt27:
These two patches make the library compile on djgpp (using gcc 10.3.0).  Everything seems to work, although the test cases don't compile, due to excessive usage of `std::wstring`.  They would probably work if you replaced all those, but I don't really see the purpose in doing that.

I've wanted to use fmtlib for a long time, but making it compatible with djgpp always seemed like a lot of work, so I kept putting it off.  I thought everything related to wide character strings would have to be `#ifdef`'d out.
Turns out the only problem is `std::wstring`, which is only ever used on *two* lines. :)

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -244,7 +244,7 @@
 #else
 #  define FMT_CLASS_API
 #  if defined(FMT_EXPORT) || defined(FMT_SHARED)
-#    if defined(__GNUC__) || defined(__clang__)
+#    if (defined(__GNUC__) || defined(__clang__)) && !defined(DJGPP)
 #      define FMT_API __attribute__((visibility("default")))
 #    endif
 #  endif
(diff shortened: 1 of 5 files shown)


jwt27:
This fixes #1077.

vitaut (changes requested):
Thanks for the PR but missing `wstring` and `wostream` is non-conformant and we shouldn't be providing workarounds for this. You can define those aliases in your code before including {fmt}.

jwt27:
Could those typedefs be provided by fmt (as is the case for `wstring_view`)?  Technically adding anything in namespace `std` is undefined behavior, according to the standard.

vitaut:
> Could those typedefs be provided by fmt (as is the case for wstring_view)?

This is essentially the same workaround that I don't think is worth doing considering a very niche nonconformant target.

jwt27:
It's been reported there before and nothing was ever done.   I think it'll be significantly easier to keep fmt compatible with djgpp rather than the other way around - a whole bunch of `wchar_t` string functions would need to be implemented in libc, that likely no one will ever use, only to gain two typedefs in libstdc++.

Issue #2653, opened 2021-12-14
Undefined behaviour when copying dynamic_format_arg_store


lucpelletier:
The copy constructor of dynamic_format_arg_store completely ignores the dynamic_args_ private member variable. This can unexpectedly cause undefined behaviour when a dynamic_format_arg_store instance is copied, since anything stored in dynamic_args_ will not get copied over. See the following for an example:

https://godbolt.org/z/8n7rnM55E

I can think of a couple of ways of fixing this problem, and I'd be happy to provide a PR with a fix, but I first wanted to get input from library maintainers on what would be the best solution.

vitaut:
The easiest solution is to mark the copy ctor as deleted. A PR would be welcome.

vitaut:
I think we should disable the copy ctor because it's currently broken. If someone has a use case they can contribute a fix later.

vitaut:
Disable the broken copy ctor in https://github.com/fmtlib/fmt/commit/be51ee1ceb0553411cae166335b59e7ae5126527. I don't think there is much reason to copy `dynamic_format_arg_store` at all, it's better to copy the actual arguments and construct the argument list only before a formatting function call.

vitaut:
Thanks for reporting.

spyridon97:
@vitaut The actual arguments cannot be copied from an `dynamic_format_arg_store` (there is no public fuction for such functionality) so i believe the copy constructor is the way to go.

I did not copy the `dynamic_args_` inside `dynamic_format_arg_store` because `dynamic_arg_list` also does not have a copy constructor. I can add one but it's not gonna be constexpr. Are you fine with that?

spyridon97:
@vitaut Any suggested solution?

PR #2832, opened 2022-03-24
[Draft]Add copy constructor for dynamic_format_arg_store

spyridon97:
The goal of this PR is to implement properly what https://github.com/fmtlib/fmt/pull/2432/ proposed but as detected in https://github.com/fmtlib/fmt/issues/2653, it produced undefined behavior, and that's why it was deleted in https://github.com/fmtlib/fmt/commit/be51ee1ceb0553411cae166335b59e7ae5126527.

It should be noted that this functionality is needed in https://github.com/Kitware/ParaView.

dynamic_format_arg_store has the following members which need to be copied over:

-  detail::dynamic_arg_list dynamic_args_

-  std::vector<basic_format_arg<Context>> data_

-  std::vector<detail::named_arg_info<char_type>> named_info_

For this to work we need the following classes/structs/enums to have a copy constructor  
- [x] detail::dynamic_arg_list (Done in this PR)

- [ ] basic_format_arg<Context> which has the following member
    - [x] detail::type (Default copy constructor is good enough)
    - [ ] detail::value<Context> (Done in this PR, but leads to wrong results)

- [ ] detail::named_arg_info<char_type> (Default copy constructor prints correct result when iterating over, but leads to segfault)

I would like your view on what i have already implemented and how could i go about implementing the remaining copy constructors if it's possible of course.

### DIFF
--- a/include/fmt/args.h
+++ b/include/fmt/args.h
@@ -33,11 +33,14 @@ class dynamic_arg_list {
   // unit for placing vtable. So storage_node_base is made a fake template.
   template <typename = void> struct node {
     virtual ~node() = default;
+    virtual std::unique_ptr<node<>> clone() const = 0;
     std::unique_ptr<node<>> next;
   };
 
   template <typename T> struct typed_node : node<> {
     T value;
+
+    FMT_CONSTEXPR typed_node(const typed_node<T>& node) : value(node.value) {}
 
     template <typename Arg>
     FMT_CONSTEXPR typed_node(const Arg& arg) : value(arg) {}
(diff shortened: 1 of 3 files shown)


spyridon97:
Note: copy_constructor, move_constructor tests fail for the time being until all the copy constructors are implemented correctly.

That's because all the const char* variables when store.reset() is called become null. Would you like me to create unique_ptr to handle them?

vitaut:
Could you remind why you want to add a copy constructor in the first place? In general `dynamic_format_arg_store` is supposed to be used for argument passing, not arbitrary management of user-defined objects. For the latter you should probably use a tuple.

spyridon97:
> Could you remind why you want to add a copy constructor in the first place? In general `dynamic_format_arg_store` is supposed to be used for argument passing, not arbitrary management of user-defined objects. For the latter you should probably use a tuple.

When ParaView is initialized we store in an argument list certain enviromental variables like os, username, hostname .etc that are available to the user in each formattable string. On top of that, they are certain views (like render view) that we want to add extra variable like time or title, and after we close the view, we want to remove those variables and use only the previous ones as a baseline for new ones.

To do that we are using a copy constructor and we have a stack of dynamic_format_arg_store and each we push a new dynamic_format_arg_store we initializing using the previous top dynamic_format_arg_store Once we finish using the new variables included in the top dynamic_format_arg_store, we pop the dynamic_format_arg_store from the stack.

Another approach to accomplish that is to be able to remove a variable from the dynamic_format_arg_store instead of using a copy constructor.

Both options seems like they can achieve the same thing, but i believe the second option seems more easy to be implemented.

What do you think?

spyridon97:
> Could you remind why you want to add a copy constructor in the first place? In general `dynamic_format_arg_store` is supposed to be used for argument passing, not arbitrary management of user-defined objects. For the latter you should probably use a tuple.

When ParaView is initialized we store in an argument list certain enviromental variables that are available to the user in each formattable string. On top of that, they are certain views (like render view) )that we want to add extra variable like time or title, and after we close the view, we want to remove those variables and use only the previous ones as a baseline.

To do that we are using a copy constructor and we have a stack of dynamic_format_arg_store and each we push a new dynamic_format_arg_store we initializing using the previous top dynamic_format_arg_store Once we finish using the new variables included in the top dynamic_format_arg_store, we pop the dynamic_format_arg_store from the stack.

An other approach to accomplish that is to be able to remove a variable from the dynamic_format_arg_store instead of using a copy constructor.

vitaut:
> What do you think?

Thanks for the explanation. I think it's a misuse of the `dynamic_format_arg_store` and you should use a real container instead only constructing the store at the time of doing the actual formatting.

PR #2833, opened 2022-03-24
Fixed all clang -Wsigned-enum-bitfield warnings

seanm:
Made enums involved in bitfields unsigned.

Used unsigned char specifically because gcc before 9.3 seems to have a bug where it warns about bitfields not being big enough to hold the enum, even though they are.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -2014,11 +2014,11 @@ using format_args = basic_format_args<format_context>;
 // We cannot use enum classes as bit fields because of a gcc bug
 // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61414.
 namespace align {
-enum type { none, left, right, center, numeric };
+enum type : unsigned char { none, left, right, center, numeric };
 }
 using align_t = align::type;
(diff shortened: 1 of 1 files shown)


seanm:
So I'm just headed on vacation, but when back I'll take another stab at making the old gcc happy...

vitaut:
Closing for now to reduce noise but feel free to reopen once it's ready for review.

seanm:
@vitaut I don't think I have permission to reopen this issue, but I have an updated commit.  Can you reopen this please?

vitaut:
Unfortunately I cannot reopen it either because the branch was force-pushed so you'll have to create a new PR.

seanm:
I pushed my updated commit, which I thought would automatically reopen this PR, but I guess that's not how github works?  Must one first reopen a PR before updating it?

Anyway, new PR coming...

seanm:
https://github.com/fmtlib/fmt/pull/2882

vitaut:
> Must one first reopen a PR before updating it?

I think so. GitHub is weird in this respect.

PR #2836, opened 2022-03-26
Fix incompatible between Jinja2 >= 3.1 and sphinx 3.3.0

phprus:
Jinja2 >= 3.1 incompatible with sphinx 3.3.0
See: https://github.com/sphinx-doc/sphinx/issues/10291


vitaut:
Thank you

PR #2839, opened 2022-03-27
Suppress clang-tidy warnings in format-inl.h

fghzxm:
Suppress clang-tidy warnings about non-inline functions in format-inl.h
(if the user is using clang 14+).

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -28,6 +28,14 @@
 
 #include "format.h"
 
+// In non-header-only mode, functions in this file are compiled by format.cc
+// once and exclusively, so we define functions without `inline`;
+// this triggers a clang-tidy warning (misc-definitions-in-headers).
+// If the user is using clang-tidy 14+, we can suppress it.
+#ifndef FMT_HEADER_ONLY
+// NOLINTBEGIN(misc-definitions-in-headers)
+#endif
+
 FMT_BEGIN_NAMESPACE
 namespace detail {
 
@@ -2589,5 +2597,9 @@ FMT_FUNC auto is_printable(uint32_t cp) -> bool {
 }  // namespace detail
 
 FMT_END_NAMESPACE
+
+#ifndef FMT_HEADER_ONLY
+// NOLINTEND(misc-definitions-in-headers)
+#endif
 
 #endif  // FMT_FORMAT_INL_H_


vitaut (changes requested):
Please post the warnings you are trying to suppress and how to reproduce them.

fghzxm:
Steps to reproduce:
1. Install clangd (version 14+) from clang-tools-extra;
2. Install the clangd extension for VSCode;
3. Add
   ```json
   "clangd.arguments": [
     "--clang-tidy"
   ]
   ```
   to the settings;
4. Generate a CMake cache and soft-link to the generated `compile_commands.json` from the project's root folder;
5. Open `format-inl.h` in VSCode; you should see the `clang-tidy(misc-definitions-in-headers)` warning.

Sorry about the havoc; I added some explainers to the code and the commit message.

vitaut:
Thanks but could you also post (a sample of) the warnings for future reference?

fghzxm:
```
$ clang-tidy --checks=misc-definitions-in-headers -p compile_commands.json include/fmt/format-inl.h
15 warnings generated.
/path/to/fmt/include/fmt/format-inl.h:34:15: warning: function 'assert_fail' defined in a header file; function definitions in header files can lead to ODR violations [misc-definitions-in-headers]
FMT_FUNC void assert_fail(const char* file, int line, const char* message) {
              ^
/path/to/fmt/include/fmt/format-inl.h:34:15: note: make as 'inline'
FMT_FUNC void assert_fail(const char* file, int line, const char* message) {
              ^
         inline
```

include/fmt/format-inl.h
```diff
@@ -28,6 +28,14 @@
 
 #include "format.h"
 
+// In non-header-only mode, functions in this file are compiled by format.cc
+// once and exclusively, so we define functions without `inline`;
+// this triggers a clang-tidy warning (misc-definitions-in-headers).
+// If the user is using clang-tidy 14+, we can suppress it.
+#ifndef FMT_HEADER_ONLY
+// NOLINTBEGIN(misc-definitions-in-headers)
+#endif
```
vitaut, line 37: I think this should be moved to format.cc and made unconditional.

vitaut (changes requested):

include/fmt/format-inl.h
fghzxm: If I move it to format.cc, it doesn't look like it works as intended:
```
$ clang-tidy --checks=misc-definitions-in-headers -p compile_commands.json src/format.cc
15 warnings generated.
Suppressed 15 warnings (15 in non-user code).
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.
$ clang-tidy --checks=misc-definitions-in-headers -p compile_commands.json -header-filter=.* src/format.cc
15 warnings generated.
/path/to/fmt/include/fmt/format-inl.h:34:15: warning: function 'assert_fail' defined in a header file; function definitions in header files can lead to ODR violations [misc-definitions-in-headers]
FMT_FUNC void assert_fail(const char* file, int line, const char* message) {
              ^
```
which is the same behavior as when the `NOLINT` is not present.  It appears that clang-tidy is hiding the warnings not because they're suppressed, but because it hides warnings in headers by default; and that `NOLINT` is not effective cross-`#include`.

If the `NOLINT` is in format-inl.h, the warnings appear suppressed as intended:
```
$ clang-tidy --checks=misc-definitions-in-headers -p compile_commands.json -header-filter=.* src/format.cc
Suppressed 15 warnings (15 NOLINT).
```

vitaut: Can you run clang-tidy on {fmt} in the header-only mode (with `FMT_HEADER_ONLY` defined)?

vitaut:
As commented, I suggest defining `FMT_HEADER_ONLY` in compile commands passed to clang-tidy.

PR #2847, opened 2022-03-30
Fix negative zero output when formatting flaoting point values close to zero

ToolsDevler:

I work for the Profidata AG which is using the libfmt version 6. Since we want to switch to version 8 the contribution I made earlier ([PullRequest](https://github.com/fmtlib/fmt/pull/1093)) is no longer available. To solve the issue I spend some time to figure out how to achieve the same behavior with the new libfmt version. I hope I came up with an acceptable answer since we need this library in production within a year from now.

Brief explanation of my change:
I altered the default behavior of printf and format in a way, that it checks for arguments which values would be printed as "-0.0". For every argument that would be printed like this to this, I set the value to zero to prevent the negative sign to be printed. This happens when a negative value is close to zero, but the fraction is not revealed due to the precision.

All tests are green, but I still hope I covered all edge cases.

Looking forward hearing from you.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -2554,10 +2554,29 @@ template <typename Char> struct arg_formatter {
   FMT_CONSTEXPR FMT_INLINE auto operator()(T value) -> iterator {
     return detail::write(out, value, specs, locale);
   }
+
+  FMT_CONSTEXPR FMT_INLINE auto operator()(float value) -> iterator {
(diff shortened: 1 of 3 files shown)


ToolsDevler:
Seems I was too easy on the test side. I'm fixing them right now.

ToolsDevler:
I'll try to find another solution outside libfmt then. Thanks for your fast feedback!

Issue #2848, opened 2022-03-31
Clang Compiler warning: [-Wfloat-equal]


abdes:
```
third_party_deps/fmtlib-src/include/fmt/format.h:2299:53: warning: comparing floating point with == or != is unsafe [-Wfloat-equal]
  if (is_constant_evaluated()) return value - value == 0;
                                      ~~~~~~~~~~~~~ ^  ~
```

please consider disabling the warning in fmtlib itself or defining a `float_equal_no_warning` function or similar, or using std::equal_to, or comparing to `epsilon`, etc.

vitaut:
A PR to silence this warning would be welcome.

abdes:
As it stands today in the current implementation, it does not look like you are doing very accurate floating point comparison. Such comparison is much more complicated, and would require a lot more code, similar to what boost.Math is doing for example.

Let me know if just silencing the warning is ok, or if you intend to have very accurate floating point comparison.

vitaut:
It is not a comparison but a fallback implementation of `isfinite`. The warning should simply be suppressed.

vitaut:
Suppressed this false positive in https://github.com/fmtlib/fmt/commit/ef54f9aa3858d7d10591719615bce337fbe49c40.

abdes:
Sorry, but you only partially solved the problem:

```
[ 33%] Building CXX object test/CMakeFiles/format-test.dir/format-test.cc.o
In file included from /home/xxx/projects/fmt/test/format-test.cc:13:
/home/xxx/projects/fmt/include/fmt/format.h: In instantiation of ‘constexpr bool fmt::v8::detail::isfinite(T) [with T = __float128; typename std::enable_if<std::is_same<T, __float128>::value, int>::type <anonymous> = 0]’:
/home/xxx/projects/fmt/test/format-test.cc:80:3:   required from ‘void check_isfinite() [with Float = __float128]’
/home/xxx/projects/fmt/test/format-test.cc:95:40:   required from here
/home/xxx/projects/fmt/include/fmt/format.h:2308:24: warning: comparing floating-point with ‘==’ or ‘!=’ is unsafe [-Wfloat-equal]
 2308 |   return value - value == 0;  // std::isfinite doesn't support __float128.
      |          ~~~~~~~~~~~~~~^~~~
```

Additionally, not in the API, but in a test file:
```
[ 83%] Building CXX object test/CMakeFiles/xchar-test.dir/xchar-test.cc.o
[...]

PR #2849, opened 2022-04-01
fix: silence warning: [-Wfloat-equal] (#2848)

abdes:


vitaut:
Closing in favor of a simpler solution (https://github.com/fmtlib/fmt/commit/ef54f9aa3858d7d10591719615bce337fbe49c40) but thanks anyway.

PR #2852, opened 2022-04-06
MSVC CMake generation optimization (13.6s -> 5.7s)

mattiasljungstrom:
The following CMake changes reduce project generation on Win11 with MSVC 2022 from 13.6s to 5.7s. (Using a deskop system with Ryzen 3900X / M.2 SSD).

Motivation: 
* Tests in cxx14.cmake are using Clang/GCC style flags, these are all invalid on MSVC.
* The flag SUPPORTS_USER_DEFINED_LITERALS is only used for tests that are disabled on WIN32.
* The flag FMT_HAS_VARIANT is not used at all from what I can see. (?)
* check_cxx_compiler_flag(-fno-exceptions) is always invalid on MSVC.
* check_cxx_compiler_flag(-fno-delete-null-pointer-checks) is always invalid on MSVC.

With FMT_TEST=OFF, project generation takes 4.3s and is almost instant when included as sub-project.

Note, correct check on MSVC would be like this:
```
check_cxx_compiler_flag("/std:c++20" has_std_20_flag)
```

But it might be better to use CMake checks instead of testing flags, like 'CMAKE_CXX_COMPILE_FEATURES' (available from CMake 3.1).

### DIFF
--- a/support/cmake/cxx14.cmake
+++ b/support/cmake/cxx14.cmake
@@ -1,7 +1,11 @@
 # C++14 feature support detection
 
-include(CheckCXXSourceCompiles)
 include(CheckCXXCompilerFlag)
+function (fmt_check_cxx_compiler_flag flag result)
+  if (NOT MSVC)
+    check_cxx_compiler_flag("${flag}" ${result})
+  endif ()
+endfunction ()
 
 if (NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 11)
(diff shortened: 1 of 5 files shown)


mattiasljungstrom:
this style also works on MSVC: (note ':' instead of '=')
```
check_cxx_compiler_flag(-std:c++20 has_std_20_flag)
```

vitaut (changes requested):
Thanks for the PR.

Instead of patching every invocation of `check_cxx_compiler_flag` let's move it to one place, a new `fmt_check_cxx_compiler_flag` function, that is a noop for MSVC.

vitaut:
> The flag FMT_HAS_VARIANT is not used at all from what I can see.

Let's remove the check completely then.

mattiasljungstrom:
Pushed an update with your change requests.

Note: it's possible to run Clang under MSVC, in which case _some_ clang/gcc flags are supported. But not the "-std=c++XX" flags. These still use the MSVC format instead.

support/cmake/cxx14.cmake
```diff
@@ -1,43 +1,50 @@
 # C++14 feature support detection
 
-include(CheckCXXSourceCompiles)
 include(CheckCXXCompilerFlag)
+macro (fmt_check_cxx_compiler_flag _FLAG _RESULT)
```
vitaut, line 4: Let's make it a function instead of a macro unless there is a strong reason not to and use normal naming convention, i.e. `flag` and `result` for arguments.

test/CMakeLists.txt
```diff
@@ -145,9 +143,7 @@ if (FMT_PEDANTIC)
     target_include_directories(
       noexception-test PRIVATE ${PROJECT_SOURCE_DIR}/include)
     target_compile_options(noexception-test PRIVATE -fno-exceptions)
-    if (FMT_PEDANTIC)
```
vitaut, line 148: Why is it removed?

mattiasljungstrom: This 'if (FMT_PEDANTIC)' is inside another 'if (FMT_PEDANTIC)' block.

mattiasljungstrom:
Pushed a new update based on your feedback.

test/CMakeLists.txt
```diff
@@ -171,8 +167,7 @@ if (FMT_PEDANTIC AND NOT WIN32)
     "-DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}"
     "-DCMAKE_CXX_FLAGS=${CMAKE_CXX_FLAGS}"
     "-DCMAKE_CXX_STANDARD=${CMAKE_CXX_STANDARD}"
-    "-DFMT_DIR=${CMAKE_SOURCE_DIR}"
-    "-DSUPPORTS_USER_DEFINED_LITERALS=${SUPPORTS_USER_DEFINED_LITERALS}")
```
phprus, line 175: ``CXX_STANDARD_FLAG`` is not defined in ``test/compile-error-test/CMakeLists.txt``

Suggestion:
``+    "-DCXX_STANDARD_FLAG=${CXX_STANDARD_FLAG}"``

vitaut (approved):

vitaut:
Thank you

Issue #2860, opened 2022-04-18
-Wliteral-range warning on Apple M1


phprus:
Commit: https://github.com/fmtlib/fmt/commit/5d804ee7fe9b197b29d2012191f7af74bcc1a837

Warning:
```
test/format-test.cc:1012:36: warning: magnitude of floating-point constant too small for type 'long double'; minimum is 4.9406564584124654E-324 [-Wliteral-range]
    EXPECT_EQ(fmt::format("{:0g}", 5.02957084971264961283E-4940L),
                                   ^
1 warning generated.
```

PR #2861, opened 2022-04-18
Suppress a -Wliteral-range warning on Apple M1

phprus:
Fix issue #2860

``std::numeric_limits<long double>::digits`` is equal to ``LDBL_MANT_DIG ``
https://en.cppreference.com/w/cpp/types/numeric_limits/digits

### DIFF
--- a/test/format-test.cc
+++ b/test/format-test.cc
@@ -1009,8 +1009,9 @@ TEST(format_test, precision) {
   if (std::numeric_limits<long double>::digits == 64) {
     auto ld = (std::numeric_limits<long double>::min)();
     EXPECT_EQ(fmt::format("{:.0}", ld), "3e-4932");
-    EXPECT_EQ(fmt::format("{:0g}", 5.02957084971264961283E-4940L),
-              "5.02957e-4940");
+    EXPECT_EQ(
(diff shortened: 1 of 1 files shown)


test/format-test.cc
```diff
@@ -1006,12 +1007,12 @@ TEST(format_test, precision) {
   EXPECT_THAT(outputs,
               testing::Contains(fmt::format("{:.838A}", -2.14001164E+38)));
 
-  if (std::numeric_limits<long double>::digits == 64) {
-    auto ld = (std::numeric_limits<long double>::min)();
-    EXPECT_EQ(fmt::format("{:.0}", ld), "3e-4932");
-    EXPECT_EQ(fmt::format("{:0g}", 5.02957084971264961283E-4940L),
-              "5.02957e-4940");
-  }
+#if LDBL_MANT_DIG == 64
```
vitaut, line 1010: I suggest keeping the normal if statement but replacing `5.02957084971264961283E-4940L` with `std::numeric_limits<long double>::denorm_min()` because the exact denormal value doesn't matter for this test.

vitaut (changes requested):
Thanks for the PR.

test/format-test.cc
phprus: Done

PR #2869, opened 2022-04-24
Add .gitattributes to have uniform line endings

Vertexwahn:
Uniform line endings


vitaut:
Thanks for the PR but this look unnecessary. You can probably do the same in your local config if you like.

PR #2875, opened 2022-04-30
Fix Unicode handling when writing to an ostream on GCC on Windows

dimztimz:
I noticed that you recently added this 8751a03a042850854c6cdfbab510758e7f9d0f89 which works just on MSVC. This patch does the same on GCC on Windows (Mingw).

The relevant info about the internals of GCC's standard library is here https://github.com/gcc-mirror/gcc/blob/master/libstdc++-v3/src/c++98/ios_init.cc

### DIFF
--- a/include/fmt/ostream.h
+++ b/include/fmt/ostream.h
@@ -10,6 +10,10 @@
 
 #include <fstream>
 #include <ostream>
+#if __GLIBCXX__
+#  include <ext/stdio_filebuf.h>
+#  include <ext/stdio_sync_filebuf.h>
+#endif
 
 #include "format.h"
 
@@ -81,6 +85,7 @@ template class filebuf_access<filebuf_access_tag,
                               decltype(&filebuf_type::_Myfile),
                               &filebuf_type::_Myfile>;
 
+#if FMT_MSC_VER
 inline bool write(std::filebuf& buf, fmt::string_view data) {
   print(get_file(buf), data);
   return true;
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR but the commit you are referring to is just an experiment, it doesn't need to be ported anywhere.

dimztimz:
> Thanks for the PR but the commit you are referring to is just an experiment, it doesn't need to be ported anywhere.

The experiment shipped in v9.0.0.

vitaut:
It is included in the release but none of this is recommended. There are better APIs that don't involve iostreams.

vitaut:
The divergence is preexisting but now it's more limited. In any case this is mostly a proof of concept and

> There are better APIs that don't involve iostreams.

dimztimz:
Ok, let me understand this better. The patch I linked, 8751a03a042850854c6cdfbab510758e7f9d0f89, is POC implementation for the following function.
```cpp
void vprint_unicode(ostream& os, string_view fmt, format_args args);
```
The function is proposed in your paper [Formatted output](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2093r14.html). Only Windows needs special handing to implement this function. Until that patch, it did not had a reference implementation, yet it was proposed. There are 3 standard libraries that work with Windows. Your patch implements the function only for MSVC and uses some unorthodox tricks to access some private members of the `streambuf` in `cout`. There is no other way.

My patch additionally implements the same function when GCC+MinGW is used. It uses completely documented interfaces by GCC, no tricks here. If you have something against this function then really don't propose it at all the paper. But if its there it would be good to work as described on all compilers that are available on Windows, not just on MSVC.

vitaut:
> It uses completely documented interfaces by GCC

If this is the case then I'm not opposed to adding the functionality. My main concern was adding more hacks like in the MSVC case. So feel free to resubmit the PR (I can't reopen this one for some reason) but please check that the mojibake problem exists in libstdc++ case on Windows and this fix is actually needed.

PR #2882, opened 2022-05-05
Fixed all clang -Wsigned-enum-bitfield warnings

seanm:
Made enums involved in bitfields unsigned.

Due to a bug, gcc < 9.3 warns about bitfields not being big enough to hold the enum, even though they are. So keep the plain enum for old gcc.


include/fmt/core.h
```diff
@@ -2029,15 +2029,25 @@ template <typename Context> class basic_format_args {
...
 }
+#else
+namespace align {
+enum type : unsigned char { none, left, right, center, numeric };
+}
+using align_t = align::type;
+namespace sign {
+enum type : unsigned char { none, minus, plus, space };
+}
+#endif
```
phprus, line 2050: Maybe simplify:
```c++
#if FMT_GCC_VERSION && FMT_GCC_VERSION < 903
# define FMT_ENUM_PARENT_TYPE(type)
#else
# define FMT_ENUM_PARENT_TYPE(type) : type
#endif


namespace align {
enum type FMT_ENUM_PARENT_TYPE(unsigned char) { none, left, right, center, numeric };
}
using align_t = align::type;
namespace sign {
enum type FMT_ENUM_PARENT_TYPE(unsigned char) { none, minus, plus, space };
}
```

?

seanm: Sure, if you prefer...

seanm: Done.

include/fmt/core.h
```diff
@@ -2029,14 +2029,20 @@ template <typename Context> class basic_format_args {
 // between clang and gcc on ARM (#1919).
 using format_args = basic_format_args<format_context>;
 
-// We cannot use enum classes as bit fields because of a gcc bug
+// We cannot use enum classes as bit fields because of a gcc bug before 9.3
```
vitaut, line 2032: This seems to be about  underlying types rather than enum classes because we don't use the former in both branches below.

vitaut:
Could you provide a repro (ideally a gobolt link) for future reference.

seanm:
> Could you provide a repro (ideally a gobolt link) for future reference.

There's one in the previous PR: https://github.com/fmtlib/fmt/pull/2801#issuecomment-1068335456  here: https://godbolt.org/z/58aEv8zEq

Or do you mean to add that link in the code comments?

vitaut:
> Or do you mean to add that link in the code comments?

No, I meant in the description of this PR so that it contains all the relevant info.

include/fmt/core.h
seanm: I'm not so sure they are different bugs in fact.  But in any case I've improved both the comments in the code and the commit message.

vitaut: Thanks!

vitaut:
Merged, thank you!

seanm:
@vitaut any chance there is a release / tag coming soon?  These warnings I fixed I'd like to integrate into another project I work on, but they prefer to reference official tags / releases of their upstream projects (like fmtlib)...

vitaut:
Yes by some definition of "soon". I don't have any specific dates though.

PR #2892, opened 2022-05-12
VS2022 17.2: C4189: 'zero': local variable is initialized but not referenced #2891

chronoxor:


vitaut:
The warning looks wrong because `zero` is referenced here: https://github.com/fmtlib/fmt/blob/f6f920a1a877271ab36114bf4b6b107bc02c73e3/include/fmt/format.h#L2244. Could you report to Microsoft?

chronoxor:
Sure

chronoxor:
Reported to Microsoft with a sample to reproduce - https://godbolt.org/z/cT6sKs3dY

include/fmt/format.h
```diff
@@ -2203,7 +2203,7 @@ FMT_CONSTEXPR20 auto do_write_float(OutputIt out, const DecimalFP& fp,
     -> OutputIt {
   auto significand = fp.significand;
   int significand_size = get_significand_size(fp);
-  constexpr Char zero = static_cast<Char>('0');
+  Char zero = static_cast<Char>('0');
```
vitaut, line 2206: Let's make it `const` so that it doesn't add to lambda capture.

chronoxor:
Updated PR with `const Char zero = static_cast<Char>('0');`

vitaut:
Thank you

PR #2902, opened 2022-05-18
Improve std formatters

phprus:
1) Improve ``std::filesystem::path`` formatter (https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1636r2.pdf).
2) Improve xchar support for std formatters.

Discussion:
https://github.com/fmtlib/fmt/commit/f0903ad9df429f3a610da45de5161c076cf2d268

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -19,14 +19,9 @@
 #ifdef __cpp_lib_filesystem
 #  include <filesystem>
 
-template <>
-struct fmt::formatter<std::filesystem::path> : formatter<string_view> {
-  template <typename FormatContext>
-  auto format(const std::filesystem::path& p, FormatContext& ctx) const ->
-      typename FormatContext::iterator {
-    return formatter<string_view>::format(p.string(), ctx);
-  }
-};
+FMT_BEGIN_NAMESPACE
+template <> struct formatter<std::filesystem::path> : ostream_formatter {};
+FMT_END_NAMESPACE
 #endif
 
 FMT_BEGIN_NAMESPACE
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -20,8 +20,35 @@
 #  include <filesystem>
 
 FMT_BEGIN_NAMESPACE
+
+namespace detail {
+
 template <typename Char>
-struct formatter<std::filesystem::path, Char> : basic_ostream_formatter<Char> {
+void write_escaped_path(basic_memory_buffer<Char>& quoted,
+                        const std::filesystem::path& p) {
+  write_escaped_string<Char>(std::back_inserter(quoted), p.string<Char>());
+}
+template <>
+void write_escaped_path<std::filesystem::path::value_type>(
+    basic_memory_buffer<std::filesystem::path::value_type>& quoted,
+    const std::filesystem::path& p) {
+  write_escaped_string<std::filesystem::path::value_type>(
+      std::back_inserter(quoted), p.native());
+}
+
+}  // namespace detail
+
+template <typename Char>
+struct formatter<std::filesystem::path, Char>
+    : formatter<basic_string_view<Char>> {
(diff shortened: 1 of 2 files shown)


include/fmt/std.h
```diff
@@ -19,18 +19,16 @@
...
-template <>
-struct fmt::formatter<std::filesystem::path> : formatter<string_view> {
-  template <typename FormatContext>
-  auto format(const std::filesystem::path& p, FormatContext& ctx) const ->
-      typename FormatContext::iterator {
-    return formatter<string_view>::format(p.string(), ctx);
-  }
+FMT_BEGIN_NAMESPACE
+template <typename Char>
+struct formatter<std::filesystem::path, Char> : basic_ostream_formatter<Char> {
```
vitaut, line 24: Let's not use iostreams here and do escaping with `write_escaped_string` instead.

vitaut (changes requested):
Thanks for the PR!

include/fmt/std.h
phprus: ``write_escaped_string`` does not support ``xchar`` and does unnecessary escaping.
For std-compatible formatting of ``std::filesystem::path``, only the quote needs to be escaped.

We can implement a separate escaping function for ``std::filesystem::path``. But using iostreams allows to add support for different character types (char/wchar_t/...).

Using a separate escaping function for ``std::filesystem::path`` will allow to make implementations that do not require double memory allocation for matching character types (for conversion and escaping).  Because the ``formatter<std::filesystem::path, Char>`` has string formatting semantics, a single memory allocation for escaping will be required anyway.


@vitaut Which way to choose?

phprus: @vitaut, ``path`` escaping is implemented in separated commit.

vitaut: I don't think it does unnecessary escaping. Note that SG16 requested the removal of path formatting from p1636 and it will likely be reintroduced in a separate paper consistent with range escaping which is what `write_escaped_string` does.

phprus: Does this paper already exist?

Using ``write_escaped_string`` breaks compatibility with ``operator>>(std::basic_istream<CharT,Traits>& is, path& p)``. If we are ready to break the compatibility of the result with ``operator>>``, then I will rewrite the implementation to using ``write_escaped_string``.

P.S. Why ``write_escaped_string`` escape chars only for ``basic_string_view<char> ``:

https://github.com/fmtlib/fmt/blob/5c7d315ded7bdb6cc5bd65daef091eefec512f2e/include/fmt/format.h#L1692-L1693

vs

https://github.com/fmtlib/fmt/blob/5c7d315ded7bdb6cc5bd65daef091eefec512f2e/include/fmt/format.h#L1639-L1643
?

vitaut: > Does this paper already exist?

No.

Compatibility with `operator>>` is not critical.

> Why write_escaped_string escape chars only for basic_string_view<char> ... ?

It hasn't been implemented yet for other types.

phprus: I added ``xchar`` support for ``write_escape_string`` in a separate PR #2904.

After merging #2904, I will make changes to this PR.

include/fmt/std.h
```diff
@@ -19,18 +19,43 @@
 #ifdef __cpp_lib_filesystem
 #  include <filesystem>
 
+FMT_BEGIN_NAMESPACE
+
+namespace detail {
+
+template <typename Char>
+void write_escaped_path(basic_memory_buffer<Char>& quoted,
```
vitaut, line 27: What is this function for? Avoid string construction if native path's code unit type is the same as Char?

phprus: Yes.

vitaut (approved):

vitaut:
Thanks!

Issue #2808, opened 2022-03-11
Compile error in `format-inl.h` when including FMT using CMake's `FetchContent` on Windows 


szymonj99:
I'm using CMake and using GitHub Actions to compile the code remotely. I'm getting the same compile error locally.

Using this commit: https://github.com/fmtlib/fmt/commit/a8fe8becf45b520c03250bda6e6914786ee91d83

I like to think the options below might be of use (e.g. The C++ standard I'm using.) 🚀 
### CMakeLists.txt
```CMAKE
...
set(CMAKE_CXX_STANDARD 20)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
set(CMAKE_CXX_EXTENSIONS ON)
set(CMAKE_POSITION_INDEPENDENT_CODE ON)
add_subdirectory(src)
```

### src/CMakeLists.txt
```CMAKE
...
FetchContent_Declare(fmt
  GIT_REPOSITORY https://github.com/fmtlib/fmt.git
  GIT_TAG a8fe8becf45b520c03250bda6e6914786ee91d83
)
FetchContent_MakeAvailable(... fmt ...)
...
set(BUILD_SHARED_LIBS OFF)
...
set(SPDLOG_FMT_EXTERNAL_HO ON)
set(SPDLOG_FMT_EXTERNAL OFF)
...
target_link_libraries(${PROJECT_NAME} PRIVATE ... fmt::fmt-header-only ...)
...
target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_20)
...
```

### Error log
```
...\build\_deps\fmt-src\include\fmt\format-inl.h(1823,44): warning C4003: not enough arguments for function-like macro invocation 'max'
...\build\_deps\fmt-src\include\fmt\format-inl.h(1823,44): error C2589: '(': illegal token on right side of '::'
...\build\_deps\fmt-src\include\fmt\format-inl.h(1823): error C2062: type 'unknown-type' unexpected
...\build\_deps\fmt-src\include\fmt\format-inl.h(1823,55): error C2059: syntax error: ')'
...\build\_deps\fmt-src\include\fmt\format-inl.h(1828,43): warning C4003: not enough arguments for function-like macro invocation 'max'
...\build\_deps\fmt-src\include\fmt\format-inl.h(1828,43): error C2589: '(': illegal token on right side of '::'
...\build\_deps\fmt-src\include\fmt\format-inl.h(1828): error C2062: type 'unknown-type' unexpected
[...]

PR #2903, opened 2022-05-18
Fix Windows max mix-up in chrono

frithrah:
Similar to #2808, fixed by 9a1beab, this addresses an occurrence in chrono.


vitaut:
Thank you!

PR #2904, opened 2022-05-22
Add xchar support for write_escaped_string.

phprus:
Add ``xchar`` support for ``write_escaped_string``.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -1570,10 +1570,14 @@ inline auto find_escape(const char* begin, const char* end)
  */
 #define FMT_STRING(s) FMT_STRING_IMPL(s, fmt::compile_string, )
 
-template <typename Char, typename OutputIt>
-auto write_escaped_string(OutputIt out, basic_string_view<Char> str)
-    -> OutputIt {
-  return copy_str<Char>(str.data(), str.data() + str.size(), out);
+template <size_t width, typename Char, typename OutputIt>
+auto write_codepoint(OutputIt out, char prefix, uint32_t cp) -> OutputIt {
+  *out++ = static_cast<Char>('\\');
+  *out++ = static_cast<Char>(prefix);
+  Char buf[width];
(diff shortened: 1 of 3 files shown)


include/fmt/format.h
```diff
@@ -1570,91 +1570,107 @@ inline auto find_escape(const char* begin, const char* end)
...
       if (escape.cp < 0x100) {
-        return format_to(out, FMT_STRING("\\x{:02x}"), escape.cp);
+        *out++ = static_cast<Char>('\\');
+        *out++ = static_cast<Char>('x');
+        constexpr size_t len = 2;
+        Char buf[len];
+        fill_n(buf, len, static_cast<Char>('0'));
+        format_uint<4>(buf, escape.cp, len);
+        out = copy_str<Char>(buf, buf + len, out);
+        return out;
```
vitaut, line 1607: Let's move this logic into a separate function and reuse here and three places below (prefix x/u and width can be parameters of this function).

vitaut (changes requested):
Thanks for the PR.

include/fmt/format.h
```diff
@@ -1582,79 +1586,74 @@ auto write_escaped_cp(OutputIt out, const find_escape_result<Char>& escape)
...
     FMT_FALLTHROUGH;
   case '\\':
-    *out++ = '\\';
+    *out++ = static_cast<Char>('\\');
     break;
   default:
     if (is_utf8()) {
       if (escape.cp < 0x100) {
-        return format_to(out, FMT_STRING("\\x{:02x}"), escape.cp);
+        return write_codepoint<2>(out, static_cast<Char>('x'), escape.cp);
```
vitaut, line 1610: Let's pass prefix as char to have the cast in one place (`write_codepoint`) instead of each call site.

vitaut (approved):

vitaut:
Thank you!

PR #2914, opened 2022-05-30
Add support for std.h in Bazel build

Vertexwahn:
This PR only affects the Bazel support.

The header file `fmt/locale.h` has been removed and `fmt/std.h` was added. This change is now reflected in the Bazel `BUILD` file. Furthermore, the directory `src` needs not to be part of the include dirs.

Bumped also Bazel version.


vitaut:
Thank you

Issue #2917, opened 2022-05-31
UTF-8 ready std::filesystem::path formatter on Windows


phprus:
On Windows ``std::filesystem::path`` formatter will use the system encoding, not utf-8.

Should we change the formatter so that when using ``char`` (not ``char8_t``) on Windows it uses UTF-8 encoding?
``libstdc++`` and ``libc++`` on linux expect utf-8 path:

https://github.com/gcc-mirror/gcc/blob/7ca388565af176bd4efd4f8db1e5e9e11e98ef45/libstdc%2B%2B-v3/include/bits/fs_path.h#L1060-L1065

I think output in utf-8 would be better.

PR #2918, opened 2022-05-31
Make std::filesystem::path formatter utf-8 compatible.

phprus:
Motivation: #2917


phprus:
@vitaut if you have time, review my PR please.

test/CMakeLists.txt
```diff
@@ -78,6 +78,10 @@ add_fmt_test(printf-test)
 add_fmt_test(ranges-test ranges-odr-test.cc)
 add_fmt_test(scan-test)
 add_fmt_test(std-test)
+add_fmt_test(std-filesystem-test)
+if (MSVC)
+  target_compile_options(std-filesystem-test PRIVATE /source-charset:utf-8)
+endif ()
```
vitaut, line 84: Let's not split `std-test`. You can apply the compiler options directly to `std-test` instead. Also it should probably be `/utf-8`, not `/source-charset:utf-8`.

include/fmt/std.h
```diff
@@ -28,6 +28,18 @@ void write_escaped_path(basic_memory_buffer<Char>& quoted,
                         const std::filesystem::path& p) {
   write_escaped_string<Char>(std::back_inserter(quoted), p.string<Char>());
 }
+#  ifdef _WIN32
+template <>
+void write_escaped_path<char>(basic_memory_buffer<char>& quoted,
+                              const std::filesystem::path& p) {
+  auto s = p.u8string();
+  // This is safe reinterpret_cast:
+  // https://stackoverflow.com/a/57453713/590688
```
vitaut, line 37: Let's remove this comment cast to `const char*` is obviously fine.

vitaut (changes requested):
Thanks for the PR.

include/fmt/std.h

test/CMakeLists.txt
phprus: When using ``/utf-8`` everything work by default (but I'm not sure). ``/utf-8`` includes utf-8 for source code and UCRT runtime encoding.
A separate test with ``/source-charset:utf-8`` makes this behavior reproducible regardless of compiler and runtime settings. Since this is a specific option, I moved the test into a separate file.

Failed test after set ``/source-charset:utf-8`` options:
https://github.com/phprus/fmt/runs/6674766235?check_suite_focus=true

vitaut: I don't think you need this option at all because the only non-ascii unescaped text is `L"Файл.txt"` which should either work or could be changed to use `\u...` escapes. In any case splitting tests is an overkill and we should find another solution.

phprus: Without the ``/source-charset:utf-8`` option, visual studio STL uses the default encoding CP_ACP (The system default Windows ANSI code page). ``CP_ACP`` can be utf-8 (by default or when using ``/utf-8``). It depends on the locale settings, and I'm not very good at this.

See commit https://github.com/phprus/fmt/commit/22a7b25db02c625220ae9b5bb5d474a8fa1a66a0 (with only L"Файл.txt").
All Windows tests completed successfully:
https://github.com/phprus/fmt/actions/runs/2439942944

But, this behavior depends on the settings. With the ``/source-charset:utf-8`` option, ``CP_ACP`` is no longer utf-8 and the problem always reproduces.
Commit: https://github.com/phprus/fmt/commit/70ed57c5054802aa06bb1200dc01eb7c21bcd2d0
Test: https://github.com/phprus/fmt/actions/runs/2439967401
```
test 16
      Start 16: std-filesystem-test
16: Test command: D:\a\fmt\build\bin\Debug\std-filesystem-test.exe
16: Test timeout computed to be: 10000000
16: [==========] Running 1 test from 1 test suite.
16: [----------] Global test environment set-up.
16: [----------] 1 test from std_filesystem_test
16: [ RUN      ] std_filesystem_test.path
16: unknown file: error: C++ exception with description "No mapping for the Unicode character exists in the target multi-byte code page." thrown in the test body.
16: [  FAILED  ] std_filesystem_test.path (1 ms)
16: [----------] 1 test from std_filesystem_test (1 ms total)
16: 
16: [----------] Global test environment tear-down
16: [==========] 1 test from 1 test suite ran. (1 ms total)
16: [  PASSED  ] 0 tests.
16: [  FAILED  ] 1 test, listed below:
16: [  FAILED  ] std_filesystem_test.path
16: 
16:  1 FAILED TEST
16/21 Test #16: std-filesystem-test ..............***Failed    0.01 sec
```

The problem is not in ``L"Файл.txt"`` string. The problem is in the ``path.string()`` algorithm and the default locale parameters.

I can remove the ``/source-charset:utf-8`` option, but then with the default settings, the old code using ``path.string()`` will not throw a test error. And this is bad.

phprus: I understand the reason!
If utf-8 encoded text is found in the source file, then Visual Studio starts using UTF-8 encoding by default for ``CP_ACP`` and ``path.string()`` works like ``path.u8string()``.



**WORKAROUND**
Commit: https://github.com/phprus/fmt/commit/e9c77dfda65d30de4862c0dc864dfac718ceeee1#diff-ed8a5a8747dfee863d65fc1ad1c2d10083df05419f6645bcf5043a4bb9a66fc4R18-R24
Test failed as expected: 
https://github.com/phprus/fmt/runs/6738989031?check_suite_focus=true

```
test 16
      Start 16: std-filesystem-test
16: Test command: D:\a\fmt\build\bin\Debug\std-filesystem-test.exe
16: Test timeout computed to be: 10000000
16: [==========] Running 1 test from 1 test suite.
16: [----------] Global test environment set-up.
16: [----------] 1 test from std_filesystem_test
16: [ RUN      ] std_filesystem_test.path
16: unknown file: error: C++ exception with description "No mapping for the Unicode character exists in the target multi-byte code page." thrown in the test body.
16: [  FAILED  ] std_filesystem_test.path (0 ms)
16: [----------] 1 test from std_filesystem_test (1 ms total)
16: 
16: [----------] Global test environment tear-down
16: [==========] 1 test from 1 test suite ran. (1 ms total)
16: [  PASSED  ] 0 tests.
16: [  FAILED  ] 1 test, listed below:
16: [  FAILED  ] std_filesystem_test.path
16: 
16:  1 FAILED TEST
16/21 Test #16: std-filesystem-test ..............***Failed    0.01 sec

```

BUT in this case the test source file (``std-test.cc``) can only contain ASCII characters.

I think writing a test that depends on the presence of utf-8 characters in the source code is a bad idea.
And it is better to put it in a separate file with the ``/source-charset:utf-8`` compiler option.

vitaut: If you really need `/source-charset:utf-8` why not apply it to `std-test.cc` directly? There is nothing else there that is affected by this option.

phprus: Done

vitaut (approved):

vitaut:
Thanks!

Issue #2920, opened 2022-06-01
Visual Studio 22 (17.2.3) + Module compilation results in ambiguous format_to call


Agga:
### Setup
```cpp
# CMakeLists.txt

set(FMT_CAN_MODULE OFF)
if (CMAKE_CXX_STANDARD GREATER 17 AND
    # msvc 16.10-pre4
    MSVC AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 19.29.30035)
  set(FMT_CAN_MODULE ON) # << this is a modification from me to get module compilation going
endif ()
```
`cmake -DCMAKE_CXX_STANDARD=20 -DFMT_MODULE=ON`

### Bug

The `format_to` call in `os.cc` function `format_windows_error` becomes ambiguous. https://github.com/fmtlib/fmt/blob/9d6039595302a6d6fe4cb2b8bfe98e9b09ed559b/src/os.cc#L172

```
2>fmt.cc
2>D:\projects\fmt-master\src\os.cc(172,78): error C2668: 'fmt::v8::format_to': ambiguous call to overloaded function
2>D:\projects\fmt-master\include\fmt\core.h(3139,17): message : could be 'OutputIt fmt::v8::format_to<fmt::v8::appender,const char*&,fmt::v8::detail::utf16_to_utf8&,0>(OutputIt,fmt::v8::basic_format_string<char,const char *&,fmt::v8::detail::utf16_to_utf8 &>,const char *&,fmt::v8::detail::utf16_to_utf8 &)'
2>        with
2>        [
2>            OutputIt=fmt::v8::appender
2>        ]
2>C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.32.31326\include\format(3046,11): message : or       '_OutputIt std::format_to<fmt::v8::appender,const char*&,fmt::v8::detail::utf16_to_utf8&>(_OutputIt,const std::_Basic_format_string<char,const char *&,fmt::v8::detail::utf16_to_utf8 &>,const char *&,fmt::v8::detail::utf16_to_utf8 &)' [found using argument-dependent lookup]
2>        with
2>        [
2>            _OutputIt=fmt::v8::appender
2>        ]
2>D:\projects\fmt-master\src\os.cc(172,78): message : while trying to match the argument list '(fmt::v8::appender, const char [7], const char *, fmt::v8::detail::utf16_to_utf8)'
2>Generating Code...
[...]

PR #2922, opened 2022-06-01
use qualified format_to call to avoid ADL conflict with std::format_to

Agga:
As discussed here: https://github.com/fmtlib/fmt/issues/2920#issuecomment-1144210328


vitaut:
Thank you

PR #2927, opened 2022-06-05
Fix 'duplicate symbol' error.

phprus:
Error:
```
duplicate symbol 'void fmt::v8::detail::write_escaped_path<char>(fmt::v8::basic_memory_buffer<char, 500ul, std::__1::allocator<char> >&, std::__1::__fs::filesystem::path const&)' in:
    CMakeFiles/src1.cpp.o
    CMakeFiles/src2.cpp.o
ld: 2 duplicate symbols for architecture arm64
```

The bug was introduced in PR #2918 (I am sorry...).


vitaut:
Thanks

PR #2932, opened 2022-06-09
Make the tests pass on a CHERI system.

davidchisnall:
Opening this for discussion because I don't particularly like this as the 'fix':

On a CHERI system (and, indeed, any system where pointers are not simply addresses), the size of `void*` does not define the number of bytes of an address.  The C++ standard conflates pointer (a thing that authorises access to an object) and an address (a location in some memory space that contains the object) in various ways.

With this change, the test suite all passes on Arm's Morello system but I would prefer a cleaner solution.


vitaut:
Thanks for the PR. Could you elaborate why this test is failing on CHERI and in particular what `sizeof(void*)` and `sizeof(size_t)` are on this system?

davidchisnall:
> Thanks for the PR. Could you elaborate why this test is failing on CHERI and in particular what sizeof(void*) and sizeof(size_t) are on this system?

On Morello, they are 16 and 8, respectively.  On CHERI systems with 32-bit address spaces, they are 8 and 4.

> I think that on CHERI there is extra data associated with a pointer, so the address is only a part of the value of a pointer. That is, sizeof(void*) is larger than the size needed to just store the address.

Correct, a pointer stores an address, distance to the top and bottom and permissions.

> On a CHERI system how do you find the number of bytes of an address? Using size_t seems weird to me but is that how the people writing compilers etc. for CHERI expect things to be done? I suppose sizeof(intptr_t) is also the wrong number of bytes (that is, not the same as the number of bytes in an address) on a CHERI system?

There is a `ptraddr_t` (it's called [`vaddr_t`](https://www.cl.cam.ac.uk/techreports/UCAM-CL-TR-947.pdf) in the older C/C++ reference) that represents an address.  Practically, this is always the same as `size_t` because that must be large enough to represent the size of any object.  It's possible that CHERI version for a 128-bit address space might have `sizeof(void) == 32`, `sizeof(ptraddr_t) == 16` and `sizeof(size_t) == 8`, if it doesn't allow a single object to be more than 2^64 bytes (which is probably a reasonable restriction).

`uintptr_t` must be able to round trip a pointer and so must be (at least) as big as `void*` unless a pointer has a load of unused bytes.  

The root of the problem is that C conflates the size of a type and the range of a type.  I'd like to see something like `std::numeric_limits` expose this.

test/format-test.cc
```diff
@@ -1493,8 +1493,13 @@ TEST(format_test, format_pointer) {
   EXPECT_EQ("0x0", fmt::format("{0}", static_cast<void*>(nullptr)));
   EXPECT_EQ("0x1234", fmt::format("{0}", reinterpret_cast<void*>(0x1234)));
   EXPECT_EQ("0x1234", fmt::format("{0:p}", reinterpret_cast<void*>(0x1234)));
+#ifdef __CHERI_PURE_CAPABILITY__
+  EXPECT_EQ("0x" + std::string(sizeof(size_t) * CHAR_BIT / 4, 'f'),
+            fmt::format("{0}", reinterpret_cast<void*>(~uintptr_t())));
```
vitaut, line 1498: I suggest removing conditional compilation by switching to `size_t` unconditionally as a more portable version. `uintptr_t` should probably be changed to `size_t` as well.

vitaut: And please add a short comment clarifying why we use `size_t` here.

davidchisnall: You can't use `size_t` instead of `uintptr_t` for the cast, they're different sizes on CHERI, if you want to cast a pointer to a `size_t` it needs to be `static_cast<size_t>(reinterpret_cast<uintptr_t>(ptr))`.

vitaut: > You can't use size_t instead of uintptr_t for the cast

Right, please ignore this part of my comment.

davidchisnall: Done.

vitaut:
Thank you!

PR #2937, opened 2022-06-11
fix -Wsign-conversion warnings

JurajX:
Fixing the two following warnings:
include/fmt/core.h:2890:40: warning: conversion to 'long unsigned int' from 'const int' may change the sign of the result [-Wsign-conversion]
 2890 |   parse_func parse_funcs_[num_args > 0 ? num_args : 1];
include/fmt/core.h:2891:28: warning: conversion to 'long unsigned int' from 'const int' may change the sign of the result [-Wsign-conversion]
 2891 |   type types_[num_args > 0 ? num_args : 1];

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -2887,8 +2887,8 @@ class format_string_checker {
   using parse_func = const Char* (*)(parse_context_type&);
 
   parse_context_type context_;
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks

PR #2938, opened 2022-06-13
Remove ``/source-charset:utf-8`` compile option.

phprus:
Integrating ``fmt`` into my projects revealed the issue (https://godbolt.org/z/GqKEGexvo):
```
cl : Command line error D8016 : '/source-charset:utf-8' and '/utf-8' command-line options are incompatible
```

Compile options ``/utf-8``, ``/source-charset:utf-8``, ``/execution-charset:utf-8`` are not compatible in any combination.
I'm using the globally set ``/utf-8`` option, which results in an error when adding fmt with tests.

We can solve this problem globally by removing the '/source-charset:utf-8' option and replacing the string literal ``L"Файл.txt"`` with a  ``wchar_t`` representation.

@vitaut, What do you think about this?


fredemmott:
`/utf-8` *is* `/source-charset:utf-8` and `/execution-charset:utf-8`; perhaps specifying those separately for your project would work for you?

wchar_t's size (and UTF-16'ness) is Windows-specific; AIUI this approach isn't portable to the other platforms fmt supports

ilyapopov:
> wchar_t's size (and UTF-16'ness) is Windows-specific; AIUI this approach isn't portable to the other platforms fmt supports

This code is already guarded with `#ifdef _WIN32`

https://github.com/fmtlib/fmt/pull/2938/files#diff-d6fdd38fa0cc7a99925e7925eeae577c399747b4f8deb61a3aed03c0d56c007bR18

test/std-test.cc
```diff
@@ -16,7 +16,11 @@ TEST(std_test, path) {
             "\"foo\\\"bar.txt\"");
 
 #  ifdef _WIN32
-  EXPECT_EQ(fmt::format("{}", std::filesystem::path(L"Файл.txt")),
+  // File.txt in Russian.
```
vitaut, line 19: You could drop ".txt" because we only care about Unicode transcoding here.

test/std-test.cc
```diff
@@ -16,7 +16,11 @@ TEST(std_test, path) {
...
+  // Do not use utf-8 encoded chars in this file.
```
vitaut, line 20: This comment is a bit misleading because we actually use UTF-8 below.

test/std-test.cc
```diff
@@ -16,7 +16,11 @@ TEST(std_test, path) {
...
+  const wchar_t cyrillic_path[] = {0x424, 0x430, 0x439, 0x43b, 0x2e,
+                                   0x74,  0x78,  0x74,  0};
+  EXPECT_EQ(fmt::format("{}", std::filesystem::path(cyrillic_path)),
             "\"\xd0\xa4\xd0\xb0\xd0\xb9\xd0\xbb.txt\"");
```
vitaut, line 24: I think you can use UTF-8 without escape sequences here as we do in `format-test`: https://github.com/fmtlib/fmt/blob/2471875867fbf6febdf119f287296a93843ae66b/test/format-test.cc#L175

test/std-test.cc
```diff
@@ -16,7 +16,11 @@ TEST(std_test, path) {
             "\"foo\\\"bar.txt\"");
 
 #  ifdef _WIN32
-  EXPECT_EQ(fmt::format("{}", std::filesystem::path(L"Файл.txt")),
+  // File.txt in Russian.
+  // Do not use utf-8 encoded chars in this file.
+  const wchar_t cyrillic_path[] = {0x424, 0x430, 0x439, 0x43b, 0x2e,
```
vitaut, line 21: nit: cyrillic_path -> unicode_path because the test is for unicode, not just cyrillic text.

vitaut (commented):
In general looks good, a few comments inline.

phprus:
@fredemmott,

fmt uses the ``/utf-8`` flag in tests:

https://github.com/fmtlib/fmt/blob/fb991e9d3b853c5bbc1a19412bda2dfa49f57d24/test/CMakeLists.txt#L84-L87

Therefore, I cannot set the ``/source-charset:utf-8 /execution-charset:utf-8`` option, it will conflict with ``/utf-8`` from the tests (and several other third party libraries in my projects). And I cannot set the ``/utf-8`` option, it will conflict with ``/source-charset:utf-8`` from this test.

My mistake was choosing the ``/source-charset:utf-8`` (PR ) option instead of using the ``wchar_t`` array.

> wchar_t's size (and UTF-16'ness) is Windows-specific;
This test is only required for Windows. ``libstdc++`` and ``libc++`` use utf-8 in the ``std::filesystem::*``.
This test will not reproduce the error when using the ``/utf-8`` option or when there are utf-8 characters in the test source file.

test/std-test.cc
phprus: Unexpected Linux CI error:
```
E: Failed to fetch http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu/pool/main/g/gcc-11/libgomp1_11.1.0-1ubuntu1~20.04_amd64.deb  Cannot initiate the connection to ppa.launchpad.net:80 (2620:2d:4000:1::3e). - connect (101: Network is unreachable) Could not connect to ppa.launchpad.net:80 (185.125.190.52), connection timed out [IP: 185.125.190.52 80]
```

test/std-test.cc
phprus: Comment updated.

vitaut:
Merged, thanks!

PR #2941, opened 2022-06-18
Add support for 'std::variant' in C++17

jehelset:
Quick variant implementation based on same machinery as tuple-formatting in #2940.

I just wanted to submit it to see if it was acceptable.


phprus:
My suggestions:

Move ``variant.h`` content into ``std.h``.
Use  ``#ifdef __cpp_lib_variant`` instad of ``#if FMT_HAS_VARIANT``.

vitaut:
Thanks for the PR. This should definitely go to `fmt/std.h` and please fix build failures.

jehelset:
i rewrote `write_range_entry` from `fmt/range.h` as `write_variant_alternative`. since variant is C++17, i hoped i can could get away with if-constexpr instead, and compilers didn't complain. i just used `is_string` from `fmt/core.h` instead of this `is_std_string_like` in `fmt/ranges.h`, to not have to include or move code around. i also took the liberty of using `std::integral_sequence`, `std::void_t` and `std::conjuction`, and instead of checking for those library features i used the `__cpp_lib_variant` as a proxy. that's not quite correct, but i'm not sure what is the best approach.

doc/api.rst
```diff
@@ -447,6 +449,35 @@ The format syntax is described in :ref:`chrono-specs`.
...
+
+Formatting Variants
+-------------------
+
+A ``std::variant`` is only formattable if every variant alternative is formattable, and requires the
+``__cpp_lib_variant`` `library feature <https://en.cppreference.com/w/cpp/feature_test>`_.
+  
+**Example**::
+
+  #include <fmt/variant.h>
```
phprus, line 472: ``#include <fmt/std.h>``

jehelset: Thanks, fixed!

doc/api.rst
```diff
@@ -447,6 +449,35 @@ The format syntax is described in :ref:`chrono-specs`.
...
+
+A ``std::variant`` is only formattable if every variant alternative is formattable, and requires the
+``__cpp_lib_variant`` `library feature <https://en.cppreference.com/w/cpp/feature_test>`_.
+  
+**Example**::
+
+  #include <fmt/std.h>
+
+  std::variant<char, float> v0{'x'};
+  // Prints "<'x'>"
```
vitaut, line 475: Since variants don't have an established notation, I suggest using something more verbose here e.g. `variant('x')` instead of `<'x'>`.

doc/api.rst
```diff
@@ -447,6 +449,35 @@ The format syntax is described in :ref:`chrono-specs`.
...
+  fmt::print("{}", v0);
+
+  std::variant<std::monostate, char> v1{};
```
vitaut, line 478: nit: `{}` is not needed here.

doc/api.rst
```diff
@@ -447,6 +449,35 @@ The format syntax is described in :ref:`chrono-specs`.
...
+  // Prints "< >"
```
vitaut, line 479: Empty variant could be represented as `variant()`.

include/fmt/std.h
```diff
@@ -68,4 +70,95 @@ template <typename Char>
...
+template <typename T> struct is_variant_like {
+  static constexpr const bool value = detail::is_variant_like_<T>::value;
+};
+
+template <typename T, typename C> struct is_variant_formattable {
+  static constexpr const bool value =
+      detail::is_variant_formattable_<T, C>::value;
+};
+
+template <typename VariantT, typename Char>
```
vitaut, line 137: nit: VariantT -> Variant for consistency with `Char`.

include/fmt/std.h
```diff
@@ -68,4 +70,95 @@ template <typename Char>
...
+struct formatter<
+    VariantT, Char,
+    std::enable_if_t<std::conjunction_v<
+        is_variant_like<VariantT>, is_variant_formattable<VariantT, Char>>>> {
+  template <typename ParseContext>
+  FMT_CONSTEXPR auto parse(ParseContext& ctx) -> decltype(ctx.begin()) {
+    return ctx.begin();
+  }
+
+  template <typename FormatContext = format_context>
```
vitaut, line 147: I don't think you need a default here.

vitaut (changes requested):

vitaut:
Overall looks good but I'd suggest changing the default representation to `variant(...)`. Also a few comments inline.

include/fmt/std.h
```diff
@@ -68,4 +70,95 @@ template <typename Char>
 struct formatter<std::thread::id, Char> : basic_ostream_formatter<Char> {};
 FMT_END_NAMESPACE
 
+#ifdef __cpp_lib_variant
+#  include <variant>
+FMT_BEGIN_NAMESPACE
```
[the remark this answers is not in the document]
powel-jeh: Fixed.

jehelset:
> Overall looks good but I'd suggest changing the default representation to `variant(...)`. Also a few comments inline.

If we can be verbose I would prefer to print `std::monostate` as `monostate`, and a variant holding a monostate as just `variant(monostate)`. Probably could do with some format specifiers, but I will leave that for someone more proficient.

vitaut (approved):

vitaut:
Thank you!

PR #2944, opened 2022-06-20
Fix enable_ifs for map formatter: map formatter requires std::pair formatter but never uses it

BrukerJWD:
I am referring to L455 in the formatter<T> definition if `is_map<T>` is true:

https://github.com/fmtlib/fmt/blob/568233889171e0ee91bf28879b1482b038e84cc2/include/fmt/ranges.h#L449-L456

Please note: `T` is the map type. We are using `detail::uncvref_type<T>` here, which gives the type of a dereferenced iterator of `T`, i.e. a `std::pair<const K,V>` in case of a `std::map<K, V>`.
This formatter is never used, though, as key and value are formatted separately:

https://github.com/fmtlib/fmt/blob/568233889171e0ee91bf28879b1482b038e84cc2/include/fmt/ranges.h#L476-L479

Thus, a formatter for the `std::pair` is not needed, or more generally speeking: L455 isn't checking the correct thing.

I stumbled upon this with a custom map which isn't using `std::pair` internally, and doesn't have a formatter for the internal `pair` type.  But this shouldn't hinder this formatter from being used.
Additionally, the `pair<K, V>` formatter is no guarantee that we have formatters for `K` and `V` of the map (although it probably would be bad code not to use them in the pair formatter).

### DIFF
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -269,6 +269,14 @@ using range_reference_type =
 template <typename Range>
 using uncvref_type = remove_cvref_t<range_reference_type<Range>>;
 
+template <typename Range>
+using uncvref_first_type = remove_cvref_t<
+    decltype(std::declval<range_reference_type<Range>>().first)>;
+
+template <typename Range>
+using uncvref_second_type = remove_cvref_t<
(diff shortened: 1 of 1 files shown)


vitaut:
Thank you!

PR #2951, opened 2022-06-28
Fix std::variant, std::filesystem::path tests on GCC-8, Clang-7,8.

phprus:
``libstdc++`` from GCC-8 support std::variant.
``libstdc++`` from GCC-8 has no ``<version>`` file.
``libstdc++`` from GCC-8 defines the macro ``__cpp_lib_variant`` in ``<variant>``.

https://github.com/fmtlib/fmt/blob/6a775e95604da91d0b97799039e52860964b7dea/test/gtest/gtest/gtest.h#L2621-L2627

Because ``gtest.h`` include ``<variant>``, then the macro ``__cpp_lib_variant`` is defined in the test, but is not defined in ``"fmt/std.h"``.

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -17,10 +17,17 @@
 #if FMT_HAS_INCLUDE(<version>)
 #  include <version>
 #endif
+// Checking FMT_CPLUSPLUS for warning suppression in MSVC.
+#if FMT_CPLUSPLUS >= 201703L
+#  if FMT_HAS_INCLUDE(<filesystem>)
+#    include <filesystem>
+#  endif
+#  if FMT_HAS_INCLUDE(<variant>)
+#    include <variant>
+#  endif
+#endif
 
 #ifdef __cpp_lib_filesystem
-#  include <filesystem>
-
 FMT_BEGIN_NAMESPACE
 
 namespace detail {
@@ -71,8 +78,6 @@ struct formatter<std::thread::id, Char> : basic_ostream_formatter<Char> {};
 FMT_END_NAMESPACE
 
 #ifdef __cpp_lib_variant
-#  include <variant>
-
 FMT_BEGIN_NAMESPACE
 template <typename Char> struct formatter<std::monostate, Char> {
   template <typename ParseContext>
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -78,6 +78,14 @@ add_fmt_test(printf-test)
 add_fmt_test(ranges-test ranges-odr-test.cc)
 add_fmt_test(scan-test)
 add_fmt_test(std-test)
+try_compile(compile_result_unused
+            ${CMAKE_CURRENT_BINARY_DIR}
+            SOURCES ${CMAKE_CURRENT_LIST_DIR}/detect-stdfs.cc
+            OUTPUT_VARIABLE RAWOUTPUT)
+string(REGEX REPLACE ".*libfound \"([^\"]*)\".*" "\\1" STDLIBFS "${RAWOUTPUT}")
+if (STDLIBFS)
+  target_link_libraries(std-test ${STDLIBFS})
+endif ()
 add_fmt_test(unicode-test HEADER_ONLY)
 if (MSVC)
   target_compile_options(unicode-test PRIVATE /utf-8)
--- a/test/detect-stdfs.cc
+++ b/test/detect-stdfs.cc
@@ -1,0 +1,18 @@
+// Formatting library for C++ - tests of formatters for standard library types
+//
+// Copyright (c) 2012 - present, Victor Zverovich
+// All rights reserved.
+//
+// For the license information refer to format.h.
+
+#include <exception>  // _GLIBCXX_RELEASE & _LIBCPP_VERSION
+
+#if defined(_GLIBCXX_RELEASE) && _GLIBCXX_RELEASE == 8
+#  error libfound "stdc++fs"
+#elif !defined(__apple_build_version__) && defined(_LIBCPP_VERSION) && \
+    _LIBCPP_VERSION >= 7000 && _LIBCPP_VERSION < 9000
+#  error libfound "c++fs"
+#else
+// none if std::filesystem does not require additional libraries
+#  error libfound ""
+#endif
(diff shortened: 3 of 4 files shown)


vitaut (changes requested):
Thanks for the PR. Unfortunately it only masks the issue so I suggest including `<variant>` in `fmt/std.h` when the header exists instead. This can be checked with `FMT_HAS_INCLUDE`.

phprus:
@vitaut, Done.
I have added two new CI configurations to detect such errors.

include/fmt/std.h
```diff
@@ -17,10 +17,16 @@
 #if FMT_HAS_INCLUDE(<version>)
 #  include <version>
 #endif
+#if FMT_CPLUSPLUS >= 201703L
```
vitaut, line 20: What is this check for? Shouldn't `FMT_HAS_INCLUDE` and feature test macro checks be enough?

test/CMakeLists.txt
```diff
@@ -78,6 +78,15 @@ add_fmt_test(printf-test)
...
 add_fmt_test(std-test)
+try_compile(compile_result_unused
+            ${CMAKE_CURRENT_BINARY_DIR}
+            SOURCES ${CMAKE_CURRENT_LIST_DIR}/detect-stdfs.cc
+            CMAKE_FLAGS CMAKE_OSX_ARCHITECTURES=${CMAKE_OSX_ARCHITECTURES}
+            OUTPUT_VARIABLE RAWOUTPUT)
+string(REGEX REPLACE ".*libfound \"([^\"]*)\".*" "\\1" STDLIBFS "${RAWOUTPUT}")
+if (STDLIBFS)
+  target_link_libraries(std-test ${STDLIBFS})
+endif ()
```
vitaut, line 89: Is it possible to simplify this?

include/fmt/std.h
vitaut: Could you add a short comment clarifying that this check is for MSVC?

test/CMakeLists.txt
phprus: The FS-library name is determined by the version of the C++ Standard Library.
CMake only determine the compiler version.
Clang supports ``libstdc++`` and ``libcxx``.
Intel C++ supports ``libstdc++`` on Linux, ``libcxx`` on macOS, and Microsoft STL on Windows.
Therefore, the only way to determine which standard library is actually used is by compiling the code.

The code I wrote is the easiest way I know.

include/fmt/std.h
phprus: @vitaut Done

test/CMakeLists.txt
vitaut: Can this information be derived from `CMAKE_CXX_STANDARD_LIBRARIES` or `CMAKE_CXX_IMPLICIT_LINK_LIBRARIES`?

phprus: The variable ``CMAKE_CXX_IMPLICIT_LINK_LIBRARIES`` allows to detect the standard library name, but not its version.

Example:
```
GCC-8:
    CMAKE_CXX_IMPLICIT_LINK_LIBRARIES: stdc++;m;gcc_s;gcc;c;gcc_s;gcc

GCC-9:
    CMAKE_CXX_IMPLICIT_LINK_LIBRARIES: stdc++;m;gcc_s;gcc;c;gcc_s;gcc

Clang 13 + libstdc++ (from GCC-11):
    CMAKE_CXX_IMPLICIT_LINK_LIBRARIES: stdc++;m;gcc_s;gcc;c;gcc_s;gcc

Clang 13 + libc++:
    CMAKE_CXX_IMPLICIT_LINK_LIBRARIES: c++;m;gcc_s;gcc;c;gcc_s;gcc
```

``CMAKE_CXX_STANDARD_LIBRARIES`` - CMake 3.6+ and empty in most configurations

test/detect-stdfs.cc
```diff
@@ -0,0 +1,18 @@
+// Formatting library for C++ - tests of formatters for standard library types
+//
+// Copyright (c) 2012 - present, Victor Zverovich
+// All rights reserved.
+//
+// For the license information refer to format.h.
+
+#include <exception>
```
vitaut, line 8: This include looks unused.

phprus: The macros ``_GLIBCXX_RELEASE`` and ``_LIBCPP_VERSION`` is defined in standard library header files.

``<exception>`` is the smallest of the standard library header files.

vitaut: I see. Maybe include `<initializer_list>` which is slightly cheaper according to https://artificial-mind.net/projects/compile-health/ and add a comment that this is for `_GLIBCXX_RELEASE` & `_LIBCPP_VERSION`?

test/CMakeLists.txt
```diff
@@ -78,6 +78,15 @@ add_fmt_test(printf-test)
 add_fmt_test(ranges-test ranges-odr-test.cc)
 add_fmt_test(scan-test)
 add_fmt_test(std-test)
+try_compile(compile_result_unused
+            ${CMAKE_CURRENT_BINARY_DIR}
+            SOURCES ${CMAKE_CURRENT_LIST_DIR}/detect-stdfs.cc
+            CMAKE_FLAGS CMAKE_OSX_ARCHITECTURES=${CMAKE_OSX_ARCHITECTURES}
```
vitaut, line 84: What is this for?

phprus: Compiling ``detect-stdfs.cc`` and read compiler output to get fslib info from error message.

test/detect-stdfs.cc
phprus: ``<initializer_list>`` is C++11.
Specifying the version of the standard will complicate the call to ``try_compile``.

Comment added.

test/CMakeLists.txt
vitaut: I mean specifically the above line that forwards `CMAKE_OSX_ARCHITECTURES`.

phprus: ``CMAKE_OSX_ARCHITECTURES`` affects the compiler but I'm not sure if it affects the standard library as well.
Removed it.

vitaut (approved):

vitaut:
Merged, thanks!

Issue #2954, opened 2022-06-30
[fmtlib trunk] Including both "fmt/std.h" and "fmt.ranges.h" causes Visual Studio compiler error


Dani-Hub:
Tested on fmtlib trunk downloaded at 2022-06-29. The following complete code leads to a compiler error for Visual Studio compilers (Tested for VS 2019 and VS 2022 targeting C++17):
```
#include "fmt/ranges.h"
#include "fmt/std.h"

int main()
{
  std::filesystem::path p("data/somewhat");
  fmt::format("Path={}", p);
}
```
The produced error is:
```
1>XXX\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30133\include\type_traits(593,1): error C2752: 'fmt::v8::formatter<std::filesystem::path,char,void>': more than one partial specialization matches the template argument list
1>XXX\fmt-master\include\fmt\ranges.h(396,1): message : could be 'fmt::v8::formatter<R,Char,std::enable_if<fmt::v8::is_range<T,Char>::value,void>::type>'
1>XXX\fmt-master\include\fmt\std.h(54,1): message : or       'fmt::v8::formatter<std::filesystem::path,Char,void>'
1>XXX\fmt-master\include\fmt\core.h(1485): message : see reference to class template instantiation 'std::is_constructible<fmt::v8::formatter<std::filesystem::path,char,void>>' being compiled
1>XXX\fmt-master\include\fmt\core.h(1679): message : see reference to alias template instantiation 'fmt::v8::detail::mapped_type_constant<std::filesystem::path,fmt::v8::format_context>' being compiled
1>XXX\fmt-master\include\fmt\core.h(1840): message : see reference to function template instantiation 'unsigned __int64 fmt::v8::detail::encode_types<Context,std::filesystem::path,>(void)' being compiled
1>        with
1>        [
1>            Context=fmt::v8::format_context
1>        ]
1>XXX\fmt-master\include\fmt\core.h(3224): message : see reference to class template instantiation 'fmt::v8::format_arg_store<fmt::v8::format_context,std::filesystem::path>' being compiled
1>XXX\Cpp\vs2019\fmtlib-composition\fmt-issue.cpp(7): message : see reference to function template instantiation 'void fmt::v8::print<std::filesystem::path&>(fmt::v8::basic_format_string<char,std::filesystem::path &>,std::filesystem::path &)' being compiled
[...]

PR #2957, opened 2022-07-02
Fix partial specialization problem for filesystem for Visual Studio

Dani-Hub:
PULL request for issue #2954


phprus:
@Dani-Hub
Please resolve merge conflicts.

include/fmt/core.h
```diff
@@ -292,6 +292,32 @@ template <typename T> using type_identity_t = typename type_identity<T>::type;
...
+template <>
+struct disjunction<> : std::false_type {};
+template <typename P>
+struct disjunction<P> : P {};
+template <typename P1, typename P2>
+struct disjunction<P1, P2> : std::conditional<P1::value, P1, P2>::type {};
+template <typename P1, typename P2, typename P3, typename... Pn>
+struct disjunction<P1, P2, P3, Pn...> 
+  : std::conditional<P1::value, P1, disjunction<P2, P3, Pn...>>::type {
+};
```
vitaut, line 306: Why not use a simpler implementation like the one from [cppreference](https://en.cppreference.com/w/cpp/types/disjunction)?

```c++
template<class...> struct disjunction : [std::false_type](http://en.cppreference.com/w/cpp/types/integral_constant) { };
template<class B1> struct disjunction<B1> : B1 { };
template<class B1, class... Bn>
struct disjunction<B1, Bn...> 
    : [std::conditional_t](http://en.cppreference.com/w/cpp/types/conditional)<bool(B1::value), B1, disjunction<Bn...>>  { };
 ```

And the same for conjunction.

File: test/CMakeLists.txt [hunk 2]
```diff
@@ -78,13 +78,15 @@ add_fmt_test(printf-test)
 add_fmt_test(ranges-test ranges-odr-test.cc)
 add_fmt_test(scan-test)
 add_fmt_test(std-test)
+add_fmt_test(ranges-std-test)
```
vitaut, line 81: Please merge this into `std-test`. Having a separate binary for a single test case is an overkill and including `fmt/ranges.h` from std-test is fine.

include/fmt/std.h
```diff
@@ -69,6 +70,33 @@ struct formatter<std::filesystem::path, Char>
         basic_string_view<Char>(quoted.data(), quoted.size()), ctx);
   }
 };
```
vitaut, line 72: I don't think this conditional branch is needed because character types other than char and wchar_t won't work anyway.

vitaut (changes requested):
Thanks for the PR.

include/fmt/std.h
phprus: Are you writing about a generic implementation?
Why?
``path::string<Char>()`` (and path formatter) supports ``char``, ``wchar_t``, ``char8_t``, ``char16_t``, and ``char32_t`` types.

vitaut: Ah, you are right. `string<Char>()` does the magic. In this case I suggest keeping just the generic branch and not introducing a workaround for broken msvc 2017. If anyone wants to format a path on a broken compiler version they can always use `.string()`.

include/fmt/core.h
Dani-Hub: I suggested the current implementation, because it is the same one that I committed for libstdc++ years ago and that worked for all known compilers regardless of completeness of variadic templates and minor rule changes in variadic templates. That being said, I guess the suggested alternative will work for all supported compilers. While adjusting this, we also can add a minor compile-conditional regarding `__cpp_lib_logical_traits` and take advantage of the STL variant in this case.

test/CMakeLists.txt
Dani-Hub: Are you really sure that you want that? This means that your preexisting test does not actually test anymore that the inclusion of stl.h suffices to get the wanted effects, it only tests that the combined inclusion has these effects. I really recommend to keep the separate test translation unit, because this introduces exactly the errornous situation.

include/fmt/core.h
vitaut: > we also can add a minor compile-conditional regarding __cpp_lib_logical_traits and take advantage of the STL variant in this case.

I suggest keeping it simple and not introducing conditional compilation.

include/fmt/std.h
Dani-Hub: At least according to `std::format` specification only `char` and `wchar_t` are supported for `charT`, that's the reason for my suggested workaround in this case. According to my experience, MSVC 2017 is still heavily in use. If you still prefer giving up 2017 support, may I suggest then a revised solution that still detects the problematic MSVC version and does *not* provide the partial specialization in this case?

include/fmt/core.h
Dani-Hub: OK.

test/CMakeLists.txt
vitaut: > Are you really sure that you want that?

Yes. Just make sure to keep `fmt/std.h` as the first include. I'm not concerned about other interactions between headers.

include/fmt/std.h
vitaut: > If you still prefer giving up 2017 support, may I suggest then a revised solution that still detects the problematic MSVC version and does not provide the partial specialization in this case?

Yes, I think it's better to disable `path` support on an older broken compiler completely rather than enable it partially for some character types. Detecting the problematic MSVC version sounds reasonable but maybe move it to a separate PR and leave this one to fixing 2019+?

Dani-Hub: Since I now need an extra conditional for the unit test anyway to *exclude* testing this scenario for Visual Studio 2017, which value would it have to keep the partial specialization without conditional?

phprus: Why disable path support if ranges are broken in MSVC 2017?

If I understand correctly, then the problem will be with formatter for any type that has member functions ``begin()`` and ``end()``. Or not?

Dani-Hub: I'm not suggesting to disable path support unconditionally. I'm suggesting to *enable* it conditionally for all cases that we know, so effectively *not* providing any specialization if `FMT_MSC_VERSION && FMT_MSC_VERSION < 1920`. 
The same conditional will need to be added to the ranges-std test anyway, so why providing functionality that we decide to not test, because it does not work? Does that make sense to you?

vitaut: > which value would it have to keep the partial specialization without conditional?

It will keep the implementation simple if I understood your question correctly. We generally try to avoid conditional compilation unless absolutely necessary and limiting it to test significantly reduces the scope.

Dani-Hub: > it's better to disable path support on an older broken compiler completely rather than enable it partially for some character types

My revised solution wasn't suggesting to enable it partially for some character types.

vitaut: The revised solution seems OK to me.

Dani-Hub:
Can someone explain to me the red cross that occurs now at my revised change in `core.h`?

test/CMakeLists.txt
Dani-Hub: Done.

include/fmt/std.h
Dani-Hub: Done

vitaut (approved):

vitaut:
Merged, thanks!

Dani-Hub:
Thanks for the review and guidance!

PR #2962, opened 2022-07-03
Add CMake option to configure the static thousands separator definition

jwillikers:
This PR just adds a CMake option to allow configuring the project to use a static thousands separator via a CMake cache variable.
This makes the functionality more discoverable.
This change will also make it easier to configure the option via Conan.

Note that some of the tests fail to link when configuring with this option.


vitaut:
Thanks for the PR but we generally don't provide CMake options for macros like this that can be passed by other means.

PR #2974, opened 2022-07-06
Exclude recursive ranges from the formatter specialization for ranges

Dani-Hub:


include/fmt/ranges.h
```diff
@@ -390,22 +390,36 @@ using range_formatter_type = conditional_t<
 template <typename R>
 using maybe_const_range =
     conditional_t<has_const_begin_end<R>::value, const R, R>;
+
+template <typename R>
+struct is_not_recursive_range : bool_constant<
```
vitaut, line 395: is_not_recursive_range -> is_nonrecursive_range

include/fmt/ranges.h
```diff
@@ -390,22 +390,36 @@ using range_formatter_type = conditional_t<
...
+
+template <typename R, typename Char>
+struct is_formattable_delayed : is_formattable<
+    detail::uncvref_type<detail::maybe_const_range<R>>, Char> {
+};
+
+template <typename R, typename Char>
+struct has_fallback_formatter_delayed : detail::has_fallback_formatter<
+    detail::uncvref_type<detail::maybe_const_range<R>>, Char> {
+};
```
vitaut, line 406: What are these for? I'd suggest keeping them inlined as before.

include/fmt/ranges.h
```diff
@@ -390,22 +390,36 @@ using range_formatter_type = conditional_t<
 template <typename R>
 using maybe_const_range =
     conditional_t<has_const_begin_end<R>::value, const R, R>;
+
+template <typename R>
+struct is_not_recursive_range : bool_constant<
+    !std::is_same<detail::uncvref_type<R>, R>::value> {};
```
vitaut, line 396: Contrary to the name of this trait it only checks recursiveness but not whether R is a range. I suggest adding is_range check here.

vitaut (changes requested):
Thanks for the PR.

include/fmt/ranges.h
Dani-Hub: Sure.

include/fmt/ranges.h
Dani-Hub: The whole point of this P/R is to ensure that the individual parts of the condition are evaluated in short-circuit manner. In C++ without language concepts this is only possible by defining type predicates that get only evaluated if the corresponding predicate template class definition becomes instantiated. Due to `std::conjunction` and friends we therefore need individual class templates. The trailing `delayed` is intended to emphasize their job.

include/fmt/ranges.h
Dani-Hub: Again, we want to prevent instantiation if the is_recursive predicate *unless* the is_range predicate has evaluated to true. See above. We could rename it to something different, such as `is_not_recursive`, if you prefer and add a comment that says that we assume here that is_range is evaluated to true.

vitaut: Yes, please add a comment.

include/fmt/ranges.h
vitaut: Could you clarify this in a comment?

brevzin: I'd suggest either adding a generic thing to delay instantiation, like so:

```cpp
template <template <class...> class F, typename... Args> struct delayed : F<Args...> { };
```
to be used as:
```cpp
disjunction<
    delayed<is_formattable, detail::uncvref_type<detail::maybe_const_range<R>>, Char>,
    delayed<detail::has_fallback_formatter, detail::uncvref_type<detail::maybe_const_range<R>>, Char>
    >
```
or, if we're going to manually do our own thing, do all of it together:
```cpp
template <class T, class Char>
struct delayed_fallback_formattable
    : disjunction<
        is_formattable<T, Char>,
        detail::has_fallback_formatter<T, Char>>
{ };
```
to be used as:
```cpp
conjunction<
    A,
    B,
    delayed_fallback_formattable<detail::uncvref_type<detail::maybe_const_range<R>>, Char>
    >
```

Put differently - instead of writing two bespoke delayed traits, write either one generic delayer or one bespoke delayed trait.

vitaut: This is a good suggestion but note that `has_fallback_formatter` will soon go away so I suggest keeping two bespoke traits. Keeping both will simplify removing `has_fallback_formatter_delayed` later.

brevzin: Oh. Ok that makes sense then. When you say soon, how soon?

include/fmt/ranges.h
Dani-Hub: Renamed as suggested.

include/fmt/ranges.h
Dani-Hub: I merged now has_fallback_formatter_delayed into is_formattable_delayed and got rid of the former.

include/fmt/ranges.h
Dani-Hub: Comment added, also relating to [P2286](https://wg21.link/p2286).

include/fmt/ranges.h
vitaut: > how soon?

In the next major release about a year from now.

vitaut (approved):

vitaut:
Merged, thanks!

PR #2979, opened 2022-07-08
Fix formatting of types with custom addressof operator

PeterFeicht:
I have a type with a custom `operator&` that I want to format. That currently doesn't work, because `fmt` uses `operator&` to get a `void*` to a formatted object.
This changes `fmt::detail::value` to use `std::addressof` instead, so types with an overloaded `operator&` will work.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -13,6 +13,7 @@
 #include <cstring>  // std::strlen
 #include <iterator>
 #include <limits>
+#include <memory>
 #include <string>
 #include <type_traits>
 
@@ -1270,7 +1271,7 @@ template <typename Context> class value {
 
   template <typename T> FMT_CONSTEXPR FMT_INLINE value(T& val) {
     using value_type = remove_cvref_t<T>;
-    custom.value = const_cast<value_type*>(&val);
+    custom.value = const_cast<value_type*>(std::addressof(val));
     // Get the formatter type through the context to allow different contexts
     // have different extension points, e.g. `formatter<T>` for `format` and
     // `printf_formatter<T>` for `printf`.


include/fmt/core.h
```diff
@@ -13,6 +13,7 @@
 #include <cstring>  // std::strlen
 #include <iterator>
 #include <limits>
+#include <memory>
```
vitaut, line 16: `fmt/core.h` is designed to have minimal includes and `<memory>` is not free (https://artificial-mind.net/projects/compile-health/). Can we avoid the include by using a cast (see e.g. https://en.cppreference.com/w/cpp/memory/addressof)?

vitaut (changes requested):
Thanks for the PR.

include/fmt/core.h
PeterFeicht: Sure, I'll add a function to the `detail` namespace. It won't be `constexpr` though, but since it's not used for parsing I guess that won't matter.

phprus: The function that calls ``addressof`` is ``constexpr`` for C++14+.

PeterFeicht: Dammit, didn't notice that. I'm pretty sure that constexpr `addressof` requires compiler support...
In Boost they detect the compiler and use a builtin function, would that be an option here too?

phprus: But ``std::addressof`` is ``constexpr`` only in C++17...

vitaut: Good point. This means we cannot use `std::addressof` or our own replacement. One can always use a wrapper type to format the one with overloaded & though which doesn't require any changes to {fmt}. So I'm going to close this PR.

PeterFeicht: Would it be an option to use `__builtin_addressof` with C++14 as long as it's supported in all supported compilers? I haven't done the research if it's available in C++14 for all compilers, but I wouldn't be surprised if it is.

vitaut: I think this would introduce a bunch of nonportable code for a very narrow use case. A potentially better approach is to extend `format_as` to work with classes and not just enums which will allow implementing a formatter for types with overloaded `operator&`.

vitaut: BTW what does overloaded operator& do in your case and what type does it return?

PeterFeicht: I have a handle type that references some resource, and the address of the handle itself has no meaning. So I overload `operator&` to return a wrapped copy of the handle, that results in the original handle when dereferenced.

I'll have a look at `format_as`. I'm not familiar at all with the code base, so any hints you could give me would be much appreciated.

vitaut: > May I create a PR for removing that constraint, or do you see any problems with that @vitaut?

Please do, I don't see any problems extending it to class types.

thomthom: @PeterFeicht did you find a way to format user types with custom `operator&`? (I also have a wrapper type, running into the same scenario.)

PeterFeicht: Oh, I completely forgot about that. We didn't yet have fmt v9 at that point so there was no `format_as`. I'll do a PR when I have some time, then `format_as` should work just fine for that purpose.

thomthom: Thank you Peter. Can you tag me in the PR if you remember?

Btw, when I look at the latest source it appear that this conditional change to be `is_enum` or `is_class`:
https://github.com/fmtlib/fmt/blob/552c43aba9d2f4fcc5309d49e586ddc90c90348c/include/fmt/core.h#L1322

Compared to 9.10 that was `is_enum` and `is_integral`.
https://github.com/fmtlib/fmt/blob/a33701196adfad74917046096bf5a2aa0ab0bb50/include/fmt/core.h#L1335

I updated to 10.0 yesterday, that includes that change. But that wasn't enough to allow a wrapper class with custom address-or. From this thread it appear that `std::addressof` cannot be used due to compatibility?

(I'm curious to what the fix would be.)

PeterFeicht: Oh nice, didn't notice that. Now it should be enough to provide an overload of `format_as` that maps to a wrapper type for formatting (which you would then provide a `formatter` specialization for).

thomthom: Having another look at this. Does this mean there is no need for any changes to fmt, but that one can already implement formatting for wrapper types? I tried to add `format_as` that return a string (for my string wrapping type) and then I tried to add a `fmt::formatter` specialisation. But I got `is not an entity that can be explicitly specialized`. (I'm not sure I fully understand how to correctly implement a specialisation.)

PR #2981, opened 2022-07-09
The n specifier for ranges.

brevzin:
Figure I'd slowly continue this work. As per P2286, the `n` specifier formats without brackets.

### DIFF
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -415,6 +415,7 @@ struct formatter<
       detail::range_formatter_type<Char, detail::uncvref_type<range_type>>;
   formatter_type underlying_;
   bool custom_specs_ = false;
+  bool use_brackets_ = true;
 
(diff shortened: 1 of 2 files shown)


include/fmt/ranges.h
```diff
@@ -415,15 +415,21 @@ struct formatter<
       detail::range_formatter_type<Char, detail::uncvref_type<range_type>>;
   formatter_type underlying_;
   bool custom_specs_ = false;
+  bool use_brackets_ = true;
```
vitaut, line 418: I suggest inverting this, i.e. replacing use_brackets_ with no_brackets_ for consistency with the specifier and to make the default false.

brevzin: I'm usually not a fan of double negatives, but it makes sense here.

vitaut:
Thanks!

PR #2983, opened 2022-07-12
Range formatter

brevzin:
More complete implementation of [P2286](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2286r8.html) and [P2585](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2585r0.html) (renaming a couple things). For now this subsumes #2981.

This is still just the `n` specifier (no `s`, `?s`, or `m` yet) but at least it's the full structure now.

### DIFF
--- a/include/fmt/ostream.h
+++ b/include/fmt/ostream.h
@@ -129,22 +129,16 @@ template <typename T> struct streamed_view { const T& value; };
 
 // Formats an object of type T that has an overloaded ostream operator<<.
 template <typename Char>
-struct basic_ostream_formatter {
-private:
-  formatter<basic_string_view<Char>, Char> underlying_;
-
-public:
-  template <typename ParseContext>
-  FMT_CONSTEXPR auto parse(ParseContext& ctx) -> decltype(ctx.begin()) {
-    return underlying_.parse(ctx);
-  }
+struct basic_ostream_formatter : formatter<basic_string_view<Char>, Char> {
+  void set_debug_format() = delete;
 
   template <typename T, typename OutputIt>
   auto format(const T& value, basic_format_context<OutputIt, Char>& ctx) const
       -> OutputIt {
     auto buffer = basic_memory_buffer<Char>();
     format_value(buffer, value, ctx.locale());
-    return underlying_.format({buffer.data(), buffer.size()}, ctx);
+    return formatter<basic_string_view<Char>, Char>::format(
+        {buffer.data(), buffer.size()}, ctx);
   }
 };
 
--- a/include/fmt/ostream.h
+++ b/include/fmt/ostream.h
@@ -129,7 +129,10 @@ template <typename T> struct streamed_view { const T& value; };
 
 // Formats an object of type T that has an overloaded ostream operator<<.
 template <typename Char>
-struct basic_ostream_formatter : formatter<basic_string_view<Char>, Char> {
+struct basic_ostream_formatter
+    : private formatter<basic_string_view<Char>, Char> {
+  using basic_ostream_formatter::formatter::parse;
+
   template <typename T, typename OutputIt>
   auto format(const T& value, basic_format_context<OutputIt, Char>& ctx) const
       -> OutputIt {
--- a/include/fmt/ostream.h
+++ b/include/fmt/ostream.h
@@ -129,17 +129,22 @@ template <typename T> struct streamed_view { const T& value; };
 
 // Formats an object of type T that has an overloaded ostream operator<<.
 template <typename Char>
-struct basic_ostream_formatter
-    : private formatter<basic_string_view<Char>, Char> {
-  using basic_ostream_formatter::formatter::parse;
+struct basic_ostream_formatter {
+private:
+  formatter<basic_string_view<Char>, Char> underlying_;
+
+public:
+  template <typename ParseContext>
+  FMT_CONSTEXPR auto parse(ParseContext& ctx) -> decltype(ctx.begin()) {
+    return underlying_.parse(ctx);
+  }
 
   template <typename T, typename OutputIt>
   auto format(const T& value, basic_format_context<OutputIt, Char>& ctx) const
       -> OutputIt {
     auto buffer = basic_memory_buffer<Char>();
     format_value(buffer, value, ctx.locale());
-    return formatter<basic_string_view<Char>, Char>::format(
-        {buffer.data(), buffer.size()}, ctx);
+    return underlying_.format({buffer.data(), buffer.size()}, ctx);
   }
 };
 
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -326,6 +326,9 @@ template <typename CharT, CharT... C> struct string_literal {
   }
 };
 
+template <typename CharT, CharT... C>
+constexpr CharT string_literal<CharT, C...>::value[];
+
 }  // namespace detail
 
 template <typename T> struct is_tuple_like {
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -356,7 +356,7 @@ struct formatter<TupleT, Char,
   };
 
  public:
-  FMT_CONSTEXPR formatter() { }
+  FMT_CONSTEXPR formatter() {}
 
   FMT_CONSTEXPR void set_separator(basic_string_view<Char> sep) {
     separator_ = sep;
@@ -453,7 +453,7 @@ struct range_formatter<
   }
 
  public:
-  FMT_CONSTEXPR range_formatter() { }
+  FMT_CONSTEXPR range_formatter() {}
 
   FMT_CONSTEXPR auto underlying() -> detail::range_formatter_type<Char, T>& {
     return underlying_;
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -352,7 +352,7 @@ struct formatter<TupleT, Char,
   // C++11 generic lambda for format().
   template <typename FormatContext> struct format_each {
     template <typename T> void operator()(const T& v) {
-      if (i > 0) out = detail::copy_str_range<Char>(separator, out);
+      if (i > 0) out = detail::copy_str<Char>(separator, out);
       out = detail::write_range_entry<Char>(out, v);
       ++i;
     }
@@ -383,9 +383,9 @@ struct formatter<TupleT, Char,
   auto format(const TupleT& values, FormatContext& ctx) const
       -> decltype(ctx.out()) {
     auto out = ctx.out();
-    out = detail::copy_str_range<Char>(opening_bracket_, out);
+    out = detail::copy_str<Char>(opening_bracket_, out);
     detail::for_each(values, format_each<FormatContext>{0, out, separator_});
-    out = detail::copy_str_range<Char>(closing_bracket_, out);
+    out = detail::copy_str<Char>(closing_bracket_, out);
     return out;
   }
 };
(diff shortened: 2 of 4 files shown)


brevzin:
I'll be honest I don't understand the gcc 4.8 error:

```
/home/runner/work/fmt/fmt/include/fmt/ranges.h:324:43: error: conversion from ‘const char*’ to non-scalar type ‘fmt::v9::basic_string_view<char>’ requested
```

What's wrong with [this constructor](https://github.com/fmtlib/fmt/blob/0db43cf7feb89f221f642c11b66bcc27043b2b5b/include/fmt/core.h#L450-L457)?

mwinterb:
Looks like it's just an "ancient GCC with new features" bug. If you explicitly provide a default constructor for `formatter`, it seems like it'll compile. https://gcc.godbolt.org/z/Wcosaqa6P. `= default` won't work. Also, if you move the initialization of `separator_`, etc to that constructor, it'll work, too.

mwinterb:
Also, to support the "esoteric" code unit types, I don't know if it'd be better to follow the `time_point` formatter model for specifying those three literals: https://github.com/fmtlib/fmt/blob/0db43cf7feb89f221f642c11b66bcc27043b2b5b/include/fmt/chrono.h#L2020-L2026
https://gcc.godbolt.org/z/q3KoTK9fK

phprus:
``default_specs[]`` without defined array size breaks Intel compiler (and other with EDG frontend).
See fix for this bug: https://github.com/fmtlib/fmt/pull/2982

mwinterb:
I was in a rush last night so I probably wasn't as clear as I should have been, but only `formatter() {}` would work.
However, you can also use brace initialization instead of assignment in the member initialization and that seems to be fine for 4.8: https://gcc.godbolt.org/z/E4boe4Gvr

vitaut (changes requested):
Thanks for the PR. Please address build failures.

brevzin:
> Thanks for the PR. Please address build failures.

Took me a while to figure this one out. Turns out this:

```
../include/fmt/core.h:2959:51: error: ‘constexpr decltype (ctx.begin()) fmt::v9::detail::parse_format_specs(ParseContext&) [with T = std::vector<int>; ParseContext = fmt::v9::detail::compile_parse_context<char, fmt::v9::detail::error_handler>; decltype (ctx.begin()) = const char*]’ called in a constant expression
 2959 |     return id >= 0 && id < num_args ? parse_funcs_[id](context_) : begin;
      |                                       ~~~~~~~~~~~~^
```

is a _slightly_ insufficient error diagnostic for "oh, you added a default constructor that you forgot to mark constexpr".

phprus:
@brevzin what do you think about comment https://github.com/fmtlib/fmt/pull/2983#issuecomment-1181291453?

The ``FMT_STATICALLY_WIDEN`` macro is limiting the ``xchar`` support, including the standard ``char16_t`` and ``char32_t`` types.

vitaut:
I think conversion is a better approach than having redundant literals for exotic code unit types.

brevzin:
> I think conversion is a better approach than having redundant constants for exotic code unit types.

Converting what, where?

vitaut:
I mean that instead of introducing a generic literal as suggested in https://github.com/fmtlib/fmt/pull/2983#issuecomment-1181291453 we could have a normal string literal and convert it into wide and other strings when needed.

phprus:
The implementation of the ``set_separator`` and ``set_brackets`` requires that string literals must immediately be of the correct character type.

include/fmt/ostream.h
```diff
@@ -129,14 +129,22 @@ template <typename T> struct streamed_view { const T& value; };
 
 // Formats an object of type T that has an overloaded ostream operator<<.
 template <typename Char>
-struct basic_ostream_formatter : formatter<basic_string_view<Char>, Char> {
+struct basic_ostream_formatter {
+private:
+  formatter<basic_string_view<Char>, Char> underlying_;
+
```
phprus, line 135: Why is this change needed?
Inheritance from formatter is allowed in the documentation (https://fmt.dev/latest/api.html):

> You can also reuse existing formatters via inheritance or composition, for example:
> ...
> ``template <> struct fmt::formatter<color>: formatter<string_view> {``
> ...
> Since parse is inherited from formatter<string_view> it will recognize all string format specifications, for example

brevzin: To avoid inheriting `set_debug_format()`

include/fmt/ostream.h
```diff
@@ -129,14 +129,22 @@ template <typename T> struct streamed_view { const T& value; };
...
+  }
+
   template <typename T, typename OutputIt>
   auto format(const T& value, basic_format_context<OutputIt, Char>& ctx) const
       -> OutputIt {
     auto buffer = basic_memory_buffer<Char>();
     format_value(buffer, value, ctx.locale());
-    return formatter<basic_string_view<Char>, Char>::format(
-        {buffer.data(), buffer.size()}, ctx);
+    return underlying_.format({buffer.data(), buffer.size()}, ctx);
```
vitaut, line 147: This change looks unnecessary.

include/fmt/ranges.h
```diff
@@ -391,56 +420,74 @@ template <typename R>
...
+        std::is_same<T, remove_cvref_t<T>>,
+        disjunction<is_formattable<T, Char>,
+                    detail::has_fallback_formatter<T, Char>>>::value>> {
+ private:
+  detail::range_formatter_type<Char, T> underlying_;
   bool custom_specs_ = false;
-  bool no_brackets_ = false;
+  basic_string_view<Char> separator_ = FMT_STATICALLY_WIDEN(Char, ", ");
+  basic_string_view<Char> opening_bracket_ = FMT_STATICALLY_WIDEN(Char, "[");
+  basic_string_view<Char> closing_bracket_ = FMT_STATICALLY_WIDEN(Char, "]");
```
vitaut, line 440: Let's do what was suggested in https://github.com/fmtlib/fmt/pull/2983#issuecomment-1181291453 and remove `FMT_STATICALLY_WIDEN`.

include/fmt/ranges.h
```diff
@@ -391,56 +420,74 @@ template <typename R>
...
+
+  template <class U>
+  FMT_CONSTEXPR static void maybe_set_debug_format(U&, long) {}
+
+  FMT_CONSTEXPR void maybe_set_debug_format() {
+    maybe_set_debug_format(underlying_, 0);
+  }
+
+ public:
+  FMT_CONSTEXPR range_formatter() { }
```
vitaut, line 456: Please apply clang-format.

include/fmt/ranges.h
```diff
@@ -391,56 +420,74 @@ template <typename R>
...
+  basic_string_view<Char> closing_bracket_ = FMT_STATICALLY_WIDEN(Char, "]");
+
+  template <class U>
+  FMT_CONSTEXPR static auto maybe_set_debug_format(U& u, int)
+      -> decltype(u.set_debug_format()) {
+    u.set_debug_format();
+  }
+
+  template <class U>
+  FMT_CONSTEXPR static void maybe_set_debug_format(U&, long) {}
```
vitaut, line 449: sneaky, we normally use ... for fallbacks like this.

vitaut (changes requested):

include/fmt/ostream.h
vitaut: Oh, I missed your earlier comment. Since we support all string specs here I don't see any reason to disable set_debug_format, please revert.

brevzin: The reason I did it was... let's say we have this:
```cpp
struct Int {
    int i;
    friend std:ostream& operator<<(std::ostream& os, Int i) { return os << i.i; }
};

std::vector<Int> ints = {Int{1}, Int{2}, Int{3}};
fmt::print("ints={}\n", ints);
```

If you keep the debug format, this'll print as `["1", "2", "3"]` (because the library thinks it's formatting string_views).

include/fmt/ranges.h
```diff
@@ -450,75 +497,104 @@ struct formatter<
...
+template <typename T> struct range_format_kind_ {
+  static constexpr auto value = std::is_same<range_reference_type<T>, T>::value
+                                    ? range_format::disabled
+                                : is_map<T>::value ? range_format::map
+                                : is_set<T>::value ? range_format::set
+                                                   : range_format::sequence;
+};
+
+template <range_format K, typename R, typename Char, typename Enable = void>
+struct range_default_formatter;
```
vitaut, line 532: Why is it called `range_default_formatter` and not `default_range_formatter`? =)

include/fmt/ostream.h
vitaut: Good point but can we disable `set_debug_format` more explicitly e.g. keep the inheritance but delete this function in ostream formatter?

include/fmt/ranges.h
brevzin: I originally called it `default_range_formatter` (since in the paper I originally called it _`default-range-formatter`_) and then I got asked to change that one to _`range-default-formatter`_ so I did that... and now this just maps to what's in the paper.

include/fmt/ostream.h
brevzin: When I did private inheritance, gcc 4.8 still picked it up anyway, but I didn't try `= delete;` which I agree is a more direct way of doing this. Will try and report back (or rather, if it works, I'll just change it. If it doesn't, then... I won't).

include/fmt/ranges.h
vitaut: OK, let's keep it consistent with the paper.

include/fmt/ranges.h
brevzin: Done, I think.

include/fmt/ostream.h
brevzin: Looks like that worked.

include/fmt/ranges.h
```diff
@@ -309,6 +309,16 @@ OutputIt write_range_entry(OutputIt out, const Arg& v) {
   return write<Char>(out, v);
 }
 
+template <typename CharT, CharT... C> struct string_literal {
+  static constexpr CharT value[] = {C...};
```
phprus, line 313: Please add array size for EDG based compilers. Details in PR https://github.com/fmtlib/fmt/pull/2982

```c++
static constexpr CharT value[sizeof...(C)] = {C...};
```

include/fmt/ranges.h
```diff
@@ -391,56 +419,76 @@ template <typename R>
...
+      detail::string_literal<Char, ']'>{};
+
+  template <class U>
+  FMT_CONSTEXPR static auto maybe_set_debug_format(U& u, int)
+      -> decltype(u.set_debug_format()) {
+    u.set_debug_format();
+  }
+
+  template <class U>
+  FMT_CONSTEXPR static void maybe_set_debug_format(U&, long) {}
```
vitaut, line 450: long -> ... per earlier comment

brevzin: Oh I didn't realize you actually wanted me to change it. Sure.

include/fmt/ranges.h
brevzin: Done.

vitaut (approved):

vitaut:
Thanks!

roman-orekhov:
Is this intentionally undocumented in [format syntax](https://fmt.dev/latest/syntax.html) until other specifiers are implemented as well?

PR #2984, opened 2022-07-12
Yet another simplification of Dragonbox

jk-jeon:
https://github.com/jk-jeon/dragonbox/issues/31 pointed out that I was doing something silly. I corrected that, which simplifies the source code a little bit (and the generated assembly as well), while the flow of the logic is not affected. Also, I replaced several logical operators into bitwise operators which seems to improve the code generation a little bit.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -1337,7 +1337,7 @@ template <typename T> decimal_fp<T> to_decimal(T x) noexcept {
 
   if (r < deltai) {
     // Exclude the right endpoint if necessary.
-    if (r == 0 && z_mul.is_integer && !include_right_endpoint) {
+    if (r == 0 && (z_mul.is_integer & !include_right_endpoint)) {
       --ret_value.significand;
       r = float_info<T>::big_divisor;
       goto small_divisor_case_label;
@@ -1389,7 +1389,7 @@ small_divisor_case_label:
   // or equivalently, when y is an integer.
   if (y_mul.parity != approx_y_parity)
     --ret_value.significand;
-  else if (y_mul.is_integer && ret_value.significand % 2 != 0)
+  else if (y_mul.is_integer & (ret_value.significand % 2 != 0))
     --ret_value.significand;
   return ret_value;
 }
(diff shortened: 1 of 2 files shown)


include/fmt/format-inl.h
```diff
@@ -1337,7 +1337,7 @@ template <typename T> decimal_fp<T> to_decimal(T x) noexcept {
 
   if (r < deltai) {
     // Exclude the right endpoint if necessary.
-    if (r == 0 && z_mul.is_integer && !include_right_endpoint) {
+    if (r == 0 && (z_mul.is_integer & !include_right_endpoint)) {
```
vitaut, line 1340: Why switch to bitwise AND? Better codegen?

include/fmt/format.h
```diff
@@ -1276,8 +1276,6 @@ template <> struct float_info<float> {
   static const int small_divisor = 10;
   static const int min_k = -31;
   static const int max_k = 46;
-  static const int divisibility_check_by_5_threshold = 39;
-  static const int case_fc_pm_half_lower_threshold = -1;
```
vitaut, line 1280: Should these be removed from `float_info<double>` as well?

vitaut (commented):
Nice simplification!

include/fmt/format-inl.h
jk-jeon: Yes. With logical AND, it generates three `test` instructions and three conditional branches. With bitwise AND, it generates two `test` instructions and two conditional branches. I thought compilers are aware of how to optimize this when all the things are just booleans variables (thus no short-circuit is really involved), but apparently they don't bother doing so. I am not sure why...

include/fmt/format.h
jk-jeon: Ha. Another careless mistake. Sorry about that, it's fixed now.

vitaut:
Merged, thanks!

include/fmt/format-inl.h
jk-jeon: I guess, compilers are not sure if all boolean variables involved are really just 0 or 1. Like, if one is `2` and another is `1`, they are both `true` so `&&` should give `true`, but `&` will give `0`. It is wrong from the first place to use `bool` for things with more than two states, but maybe the current standard allows `bool` variables to hold nonsensical values without UB.

PR #2985, opened 2022-07-14
Fixes IBM XLC behavior with uint128 fallback

federico-busato:
IBM XLC compiler evaluates `#if FMT_HAS_BUILTIN(__builtin_addcll)` to true but `__builtin_addcll` is not supported and the compilation fails.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -391,14 +391,11 @@ class uint128_fallback {
       hi_ += (lo_ < n ? 1 : 0);
       return *this;
     }
-#if FMT_XLC_VERSION && defined(_ARCH_PWR9)
-    lo_ = __addex(lo_, n, 0, &carry);
-    hi_ += carry;
-#elif FMT_HAS_BUILTIN(__builtin_addcll) && !FMT_XLC_VERSION
+#if FMT_HAS_BUILTIN(__builtin_addcll) && !defined(__ibmxl__)
     unsigned long long carry;
     lo_ = __builtin_addcll(lo_, n, 0, &carry);
     hi_ += carry;
-#elif FMT_HAS_BUILTIN(__builtin_ia32_addcarryx_u64) && !FMT_XLC_VERSION
+#elif FMT_HAS_BUILTIN(__builtin_ia32_addcarryx_u64) && !defined(__ibmxl__)
     unsigned long long result;
     auto carry = __builtin_ia32_addcarryx_u64(0, lo_, n, &result);
     lo_ = result;
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -391,7 +391,7 @@ class uint128_fallback {
       hi_ += (lo_ < n ? 1 : 0);
       return *this;
     }
-#if FMT_XLC_VERSION && defined(_ARCH_PWR9) && defined(_ARCH_PPC64)
+#if FMT_XLC_VERSION && defined(_ARCH_PWR9)
     lo_ = __addex(lo_, n, 0, &carry);
     hi_ += carry;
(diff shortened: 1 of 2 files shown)


vitaut:
Does it mean that `__has_builtin` is broken on IBM XLC? Please report to the compiler vendor.

As for the workaround, why make `FMT_HAS_BUILTIN` return false on this compiler if it doesn't work?

federico-busato:
> Does it mean that `__has_builtin` is broken on IBM XLC? Please report to the compiler vendor.

yes, I need further investigation to understand why `__has_builtin` does not work here. Just for completeness, I'm using xlc v16.1.1 compiler. 

> As for the workaround, why make `FMT_HAS_BUILTIN` return false on this compiler if it doesn't work?

`FMT_HAS_BUILTIN(__builtin_addcll)` return `true` and it enables the following line
```
__builtin_addcll(lo_, n, 0, &carry);
```
but `__builtin_addcll` is not supported by the compiler and this triggers the compile error.

please hold on. I would like to perform a couple of experiments to see if there are other solutions.

federico-busato:
IBM xlc has two versions, the legacy one, and the clang-based. The problem occurs in the clang-based version. The reason is that `FMT_HAS_BUILTIN(__builtin_addcll)` is evaluated to true because actually clang supports it. However, the intrinsic function is not recognized/translated by the xlc backend. If needed, this is the message
```
fmt/include/fmt/format.h:397:11: error: 1540-2993 The builtin "__builtin_addcll" is not supported.
    lo_ = __builtin_addcll(lo_, n, 0, &carry);
```

include/fmt/format.h
```diff
@@ -391,11 +391,14 @@ class uint128_fallback {
       hi_ += (lo_ < n ? 1 : 0);
       return *this;
     }
-#if FMT_HAS_BUILTIN(__builtin_addcll)
+#if FMT_XLC_VERSION && defined(_ARCH_PWR9)
+    lo_ = __addex(lo_, n, 0, &carry);
+    hi_ += carry;
```
vitaut, line 396: Do we need this optimization? I suggest not introducing it unless not having it is causing some real perf problems. The fallback normally generates identical code: https://godbolt.org/z/xMGE5aE4T.

include/fmt/format.h
```diff
@@ -391,11 +391,14 @@ class uint128_fallback {
...
+#elif FMT_HAS_BUILTIN(__builtin_addcll) && !FMT_XLC_VERSION
     unsigned long long carry;
     lo_ = __builtin_addcll(lo_, n, 0, &carry);
     hi_ += carry;
-#elif FMT_HAS_BUILTIN(__builtin_ia32_addcarryx_u64)
+#elif FMT_HAS_BUILTIN(__builtin_ia32_addcarryx_u64) && !FMT_XLC_VERSION
```
vitaut, line 401: Does `__builtin_ia32_addcarryx_u64` need to be excluded?

vitaut (changes requested):

include/fmt/format.h
federico-busato: __addex works only for IBM Power9 platforms. Anyway, it is not critical

include/fmt/format.h
federico-busato: as you removed the line
```
FMT_XLC_VERSION && defined(_ARCH_PWR9)
```
and in particular `defined(_ARCH_PWR9)` you don't need ` && !FMT_XLC_VERSION` anymore

federico-busato: edit: I misunderstand the issue. `__builtin_ia32_addcarryx_u64` must be skipped to avoid the same problem of `__builtin_addcll`

vitaut (approved):

vitaut:
Thanks!

PR #2991, opened 2022-07-19
Add additional search paths for doxygen on Windows

hwhsu1231:
`find_program(DOXYGEN doxygen)` didn't find the executable automatically even if I installed Doxygen on Windows. Therefore, I added some additional search paths.


vitaut:
Thanks

PR #2995, opened 2022-07-20
Reduce filesize of the tests on MinGW

dimztimz:
This patch removes the workaround applied here 1acfd07f1ed9d4055833a70d9195e57aee7bbe77. MinGW is not tested on Appveyor anymore. If its added in the future it should be added in Github Actions.

### DIFF
--- a/support/appveyor-build.py
+++ b/support/appveyor-build.py
@@ -11,32 +11,23 @@ path = os.environ['PATH']
 image = os.environ['APPVEYOR_BUILD_WORKER_IMAGE']
 jobid = os.environ['APPVEYOR_JOB_ID']
 cmake_command = ['cmake', '-DFMT_PEDANTIC=ON', '-DCMAKE_BUILD_TYPE=' + config, '..']
-if build == 'mingw':
-    cmake_command.append('-GMinGW Makefiles')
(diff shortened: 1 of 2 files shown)


support/appveyor-build.py
```diff
@@ -11,15 +11,7 @@
...
-if build == 'mingw':
-    cmake_command.append('-GMinGW Makefiles')
-    build_command = ['mingw32-make', '-j4']
-    test_command = ['mingw32-make', 'test']
-    # Remove the path to Git bin directory from $PATH because it breaks
-    # MinGW config.
-    path = path.replace(r'C:\Program Files (x86)\Git\bin', '')
-    os.environ['PATH'] = r'C:\MinGW\bin;' + path
-else:
+if build == 'msvc':
```
vitaut, line 14: Let's remove this `if` altogether because the condition is always true.

vitaut (changes requested):
Thanks for the PR. LGMT with one minor suggestion inline.

support/appveyor-build.py
dimztimz: fixed

vitaut (approved):

vitaut:
Thanks!

PR #3001, opened 2022-07-26
Fix Unicode handling for ostream under Windows with libc++.

dimztimz:
This PR fixes the Unicode handling for `std::print(ostream&, ...)` under Windows with libc++.

Also replaces the SFIANE tricks applied in ce7ecdb7afb207dadf55c8ae457e14aa44e7f537 with conditional compilation. The code was too complicated along with the other trick to access private data members.

### DIFF
--- a/include/fmt/ostream.h
+++ b/include/fmt/ostream.h
@@ -13,6 +13,8 @@
 #if defined(_WIN32) && defined(__GLIBCXX__)
 #  include <ext/stdio_filebuf.h>
 #  include <ext/stdio_sync_filebuf.h>
+#elif defined(_WIN32) && defined(_LIBCPP_VERSION)
+#  include <__std_stream>
 #endif
 
 #include "format.h"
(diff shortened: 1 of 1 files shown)


include/fmt/ostream.h
```diff
@@ -55,35 +57,25 @@ struct is_streamable<
...
 // type in an anonymous namespace.
 namespace {
-struct filebuf_access_tag {};
+struct priv_access_tag {};
 }  // namespace
-template <typename Tag, typename FileMemberPtr, FileMemberPtr file>
-class filebuf_access {
-  friend FILE* get_file(filebuf_type& buf) { return buf.*file; }
+template <class Tag, class BufType, FILE* BufType::*FileMemberPtr>
+class priv_access {
```
vitaut, line 66: I think the old naming was more descriptive. The only issue is that it no longer applies to just `filebuf` so let's rename it to `file_access` since it is used to access a `FILE*` member. And similarly `file_access_tag`.

vitaut (changes requested):
Thanks for the PR. Overall looks good, just a minor naming suggestion inline.

include/fmt/ostream.h
dimztimz: Fixed.

vitaut (approved):

vitaut:
Thanks

PR #3007, opened 2022-07-28
Improve CI on Windows, deprecate AppVeyor

### LINKED ISSUE #3004: Replace AppVeyor with Github actions
AppVeyor is usually slow so it is good idea to replace it. Now it is kept because it provides Visual Studio 2015 and GitHub actions does not, it has VS2019.

But you can get the toolchain of VS2015 inside VS2019. VS2019 provides three toolchains:

- v140 (from VS2015)
- v141 (from VS2017)
- v142, the default.

According to the docs, all three are [already installed](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#workloads-components-and-extensions). I even noticed there Clang as toolchain inside VS.

You can use the CMake -T CLI switch or set the appropriate variable [CMAKE_GENERATOR_TOOLSET](https://cmake.org/cmake/help/latest/variable/CMAKE_GENERATOR_TOOLSET.html) to select toolchain.

dimztimz:
Improve CI on Windows, deprecate AppVeyor. Fixes #3004.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -15,24 +15,30 @@ jobs:
         # https://github.com/actions/virtual-environments.
         os: [windows-2019]
         platform: [Win32, x64]
+        toolset: [v140, v141, v142]
+        standard: [14, 17, 20]
+        shared: ["", -DBUILD_SHARED_LIBS=ON]
         build_type: [Debug, Release]
-        standard: [11, 17, 20]
+        exclude:
+          - { toolset: v140, standard: 17 }
+          - { toolset: v140, standard: 20 }
+          - { toolset: v141, standard: 14 }
+          - { toolset: v141, standard: 20 }
+          - { toolset: v142, standard: 14 }
+          - { platform: Win32, toolset: v140 }
+          - { platform: Win32, toolset: v141 }
+          - { platform: Win32, standard: 14 }
+          - { platform: Win32, standard: 20 }
+          - { platform: x64, toolset: v140, shared: -DBUILD_SHARED_LIBS=ON }
+          - { platform: x64, toolset: v141, shared: -DBUILD_SHARED_LIBS=ON }
+          - { platform: x64, standard: 14, shared: -DBUILD_SHARED_LIBS=ON }
+          - { platform: x64, standard: 20, shared: -DBUILD_SHARED_LIBS=ON }
         include:
-          - os: windows-2019
-            platform: Win32
-            build_type: Debug
-            shared: -DBUILD_SHARED_LIBS=ON
           - os: windows-2022
             platform: x64
+            toolset: v143
             build_type: Debug
             standard: 20
-        exclude:
-          - os: windows-2019
-            standard: 11
-            platform: Win32
-          - os: windows-2019
-            standard: 20
-            platform: Win32
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


dimztimz:
This is ready. After this open the AppVeyor website and remove this repo from there. Hopefully, that should automatically remove the webhook, if not remove it manually from Github settings.

vitaut (changes requested):
Thanks for the PR. Looks great but please keep the old include/exclude rules and add one more include rule for the new config to keep them manageable.

dimztimz:
Just so you know in the old rules the one entry in include with `shared` 

https://github.com/fmtlib/fmt/blob/0b2862a1e4cf24797424501a3d13e3e563a90528/.github/workflows/windows.yml#L20-L24

was not adding a new job but was overriding an existing job generated by the build matrix above.

How about this: test toolset v140 (VS2015) only with standard C++14, v141 (VS2017) only with standard C++17 and toolset v142 with 2 standards, 17 and 20.

Win32 will get only 4 entries: (static, shared) x (debug, release) with the latest v142 toolchain.
on x64 shared will get only 2 entries with the latest toolchain with standard 17.

dimztimz:
The build matrix that I pushed now  generates 15 entries on Windows which is very much acceptable and manageable. There might be simpler rules that generate the same entries but I don't see them right now. Please review then in view for Github Actions.

It even detected a build failure. Maybe with some recent changes, because I also rebased this on top of master.

vitaut:
The failure seems related to #2957. cc @Dani-Hub

phprus:
@dimztimz, @vitaut 
Workaround commit: https://github.com/phprus/fmt/commit/af72c41eb3ded8134c4f793fd3633263b943fc00 (Branch: https://github.com/phprus/fmt/tree/ranges-msvc-1).

dimztimz:
I will cherry-puck that and put it in this PR as separated commit. Also I will clean my commits.

dimztimz:
I think this PR is in better shape now, it has two commits. I removed the commit that removes AppVeyor files to get the green check mark. You can remove them after you disable it from the settings.

vitaut (approved):

vitaut:
Thanks @dimztimz and @phprus!

vitaut:
@dimztimz I just noticed that v140 build still reports MSVC 2019, not 2015:

```
-- Building for: Visual Studio 16 2019
-- CMake version: 3.23.2
-- The CXX compiler identification is MSVC 1[9](https://github.com/fmtlib/fmt/runs/7632834167?check_suite_focus=true#step:4:10).0.24245.0
```
(https://github.com/fmtlib/fmt/runs/7632834167?check_suite_focus=true)

Do you know why?

vitaut:
What a versioning mess =). Where did you find the info that 19.0.24245.0 corresponds to 2015?

phprus:
See ``_MSC_VER`` value on page https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170

dimztimz:
https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B#Internal_version_numbering

It is total mess. There are like four different versions. One is year, one is IDE version, one is toolset version and the last is compiler version. Roughly: 
- toolset v14.0x = compiler v19.0x.
- toolset v14.1x = compiler v19.1x.
- toolset v14.2x = compiler v19.2x.

PR #3009, opened 2022-07-29
Add fmt:: namespace to doc

jcelerier:
Otherwise as-is the example does not compile on Visual Studio due to the conflict with std::format_to: 
https://gcc.godbolt.org/z/qe4jEvvqY


vitaut:
Thank you!

PR #3010, opened 2022-07-30
Simplify and improve chrono formatting

phprus:
1) Reused ``detail::string_literal``
2) Removed unneeded code.
3) Improve performance (prevented double code execution, because formatter<>::parse() is always called and for empty format spec too).


vitaut:
Thank you!

PR #3021, opened 2022-08-05
Suppress unused typedef warning

h-friederich:
On some compilers a (buggy) `-Wunused-local-typedef` warning is emitted.


vitaut:
Thanks!

PR #3026, opened 2022-08-06
Constexpr formatted_size

marksantaniello:
Minimal changes to make fmt::formatted_size constexpr in C++20.   Adjust the existing unit test accordingly.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -28,18 +28,9 @@ jobs:
             std: 17
             install: sudo apt install g++-8
             os: ubuntu-18.04
-          - cxx: g++-9
-            build_type: Debug
-            std: 20
-            install: sudo apt install g++-9
-            os: ubuntu-18.04
           - cxx: g++-10
             build_type: Debug
             std: 17
-            os: ubuntu-18.04
-          - cxx: g++-10
-            build_type: Debug
-            std: 20
             os: ubuntu-18.04
           - cxx: g++-11
             build_type: Debug
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -28,9 +28,18 @@ jobs:
             std: 17
             install: sudo apt install g++-8
             os: ubuntu-18.04
+          - cxx: g++-9
+            build_type: Debug
+            std: 20
+            install: sudo apt install g++-9
+            os: ubuntu-18.04
           - cxx: g++-10
             build_type: Debug
             std: 17
+            os: ubuntu-18.04
+          - cxx: g++-10
+            build_type: Debug
+            std: 20
             os: ubuntu-18.04
           - cxx: g++-11
             build_type: Debug
--- a/test/compile-test.cc
+++ b/test/compile-test.cc
@@ -227,18 +227,14 @@ TEST(compile_test, format_to_n) {
   EXPECT_STREQ("2a", buffer);
 }
 
-TEST(compile_test, formatted_size) {
 #ifdef __cpp_lib_bit_cast
-  FMT_CONSTEXPR20 
-#endif 
-  size_t s1 = fmt::formatted_size(FMT_COMPILE("{0}"), 42);
+TEST(compile_test, constexpr_formatted_size) {
+  FMT_CONSTEXPR20 size_t s1 = fmt::formatted_size(FMT_COMPILE("{0}"), 42);
   EXPECT_EQ(2, s1);
-#ifdef __cpp_lib_bit_cast
-  FMT_CONSTEXPR20 
-#endif 
-size_t s2 = fmt::formatted_size(FMT_COMPILE("{0:<4.2f}"), 42.0);
+  FMT_CONSTEXPR20 size_t s2 = fmt::formatted_size(FMT_COMPILE("{0:<4.2f}"), 42.0);
   EXPECT_EQ(5, s2);
 }
+#endif
 
 TEST(compile_test, text_and_arg) {
   EXPECT_EQ(">>>42<<<", fmt::format(FMT_COMPILE(">>>{}<<<"), 42));
--- a/test/compile-test.cc
+++ b/test/compile-test.cc
@@ -228,9 +228,15 @@ TEST(compile_test, format_to_n) {
 }
 
 TEST(compile_test, formatted_size) {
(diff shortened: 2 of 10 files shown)


test/compile-test.cc
```diff
@@ -228,8 +228,10 @@ TEST(compile_test, format_to_n) {
 }
 
 TEST(compile_test, formatted_size) {
-  EXPECT_EQ(2, fmt::formatted_size(FMT_COMPILE("{0}"), 42));
-  EXPECT_EQ(5, fmt::formatted_size(FMT_COMPILE("{0:<4.2f}"), 42.0));
+  static FMT_CONSTEXPR20 size_t s1 = fmt::formatted_size(FMT_COMPILE("{0}"), 42);
+  EXPECT_EQ(2, s1);
+  static FMT_CONSTEXPR20 size_t s2 = fmt::formatted_size(FMT_COMPILE("{0:<4.2f}"), 42.0);
+  EXPECT_EQ(5, s2);
```
vitaut, line 234: static seems unnecessary, please remove.

vitaut (commented):
Thanks for the PR! One minor comment inline, otherwise LGTM.

vitaut:
CI is failing: https://github.com/fmtlib/fmt/runs/7734651761?check_suite_focus=true

marksantaniello:
Seems Clang-11 has some issue with memcpy not being constexpr.  Trying to repro but for me clang 11 actually crashes...

marksantaniello:
OK got a repro.  I was missing that I needed to use libc++.  Will investigate.

marksantaniello:
Seems the problem is here:

https://github.com/fmtlib/fmt/blob/master/include/fmt/format.h#L295-L305

I think Clang 11 supports -std=c++20, but the included libc++ doesn't have __cpp_lib_bit_cast, so we end up trying to use memcpy in a constexpr function.

marksantaniello:
Maybe I can work around this by using __builtin_bit_cast

phprus:
For C++20 constexpr fmt needs a ``bit_cast``.
``bit_cast`` requires (https://en.cppreference.com/w/cpp/compiler_support):
GCC libstdc++: 11 (``_GLIBCXX_RELEASE >= 11``)
Clang libc++: 14 (``_LIBCPP_VERSION >= 14000``)
MSVC STL:  19.27

I suggest checking these conditions or ``__builtin_bit_cast`` when defining ``FMT_CONSTEXPR20``.
cc @vitaut 


@marksantaniello, You can enable Github Actions for your fmt fork

marksantaniello:
@phprus 
> You can enable Github Actions for your fmt fork

Are you saying that there's a way to get all the same test workflows running on my fork?

phprus:
@marksantaniello
Yes.

marksantaniello:
Thanks.  I'll have to figure do that if I make PRs in the future.  The test matrix is pretty large.

Anyway, it looks like the current version using __builtin_bit_cast is passing the checks.

phprus:
Current test matrix does not check:
gcc-10 + C++20 (defined ``FMT_CONSTEXPR20``, but not support ``__builtin_bit_cast``)
Clang libc++ < 14 + C++20

marksantaniello:
Ah, OK.  Maybe I should figure out the workflows now then.  I enabled "Actions" but I'm not sure how to clone the checks from here.  And then I guess I'd need to add at least two more?

phprus:
Strategies are defined in the .github/workflows/linux.yml file:

https://github.com/fmtlib/fmt/blob/8bd02e93b2d38ce3d69f117623e64e28c6f331cd/.github/workflows/linux.yml#L11-L61

I think for the test you need to add:
gcc-9,gcc-10 + C++20

vitaut:
`FMT_CONSTEXPR20` should definitely not be changed. I suggest making the test conditional on whether `__cpp_lib_bit_cast` is defined, removing the `bit_cast` changes and new CI configs (we already have way too many).
Then everything that used to work will continue to work and if `bit_cast` is `constexpr` and there is C++20 `constexpr` support then `formatted_size` will usable in `constexpr`.

test/compile-test.cc
```diff
@@ -228,8 +228,16 @@ TEST(compile_test, format_to_n) {
...
+  FMT_CONSTEXPR20 
+#endif 
+  size_t s1 = fmt::formatted_size(FMT_COMPILE("{0}"), 42);
+  EXPECT_EQ(2, s1);
+#ifdef __cpp_lib_bit_cast
+  FMT_CONSTEXPR20 
+#endif 
+size_t s2 = fmt::formatted_size(FMT_COMPILE("{0:<4.2f}"), 42.0);
+  EXPECT_EQ(5, s2);
 }
```
vitaut, line 241: Let's conditionally compile the test case itself because the whole purpose of it is to test constexpr formatted_size. Also formatted_size -> contexpr_formatted_size to make it clear what we are testing here.

```
#ifdef __cpp_lib_bit_cast
TEST(compile_test, constexpr_formatted_size) {
  FMT_CONSTEXPR20 size_t s1 = fmt::formatted_size(FMT_COMPILE("{0}"), 42);
  EXPECT_EQ(2, s1);
  FMT_CONSTEXPR20 size_t s2 = fmt::formatted_size(FMT_COMPILE("{0:<4.2f}"), 42.0);
  EXPECT_EQ(5, s2);
}
#endif
```

.github/workflows/linux.yml
```diff
@@ -28,10 +28,19 @@ jobs:
...
+            install: sudo apt install g++-9
+            os: ubuntu-18.04
           - cxx: g++-10
             build_type: Debug
             std: 17
             os: ubuntu-18.04
+          - cxx: g++-10
+            build_type: Debug
+            std: 20
+            os: ubuntu-18.04
```
vitaut, line 43: As commented earlier, let's not add new CI configs.

vitaut (changes requested):

.github/workflows/linux.yml
marksantaniello: Sorry, yes -- I'm just crap at git.  Will try again.

vitaut (approved):

vitaut:
Looks good but there are some merge conflicts, please rebase.

vitaut:
Thanks!

PR #3033, opened 2022-08-10
Update core.h

Altanb21:
C++ 20 build error fix


vitaut:
Please provide more context. I guess this is a workaround for buggy MSVC IntelliSense but what errors do you get?

Issue #3028, opened 2022-08-08
Tests fail using NVIDIA HPC compilers (PGI compilers)


olupton:
The `fmt` test suite does not pass using the 22.3 NVIDIA HPC compilers. I tested this with today's master (682e097bee9aaa88e9c9826e72d87c2ec1b1226b) and confirmed I could reproduce using NVIDIA's own Docker images that ship their compilers:
```
$ docker run -it nvcr.io/nvidia/nvhpc:22.3-devel-cuda11.6-ubuntu20.04
root@411ece11247c:~# git clone https://github.com/fmtlib/fmt.git
Cloning into 'fmt'...
...
root@411ece11247c:~# cd fmt/
root@411ece11247c:~/fmt# mkdir build
root@411ece11247c:~/fmt# cd build
root@411ece11247c:~/fmt/build# cmake .. -DCMAKE_CXX_FLAGS="-tp haswell"
-- CMake version: 3.16.3
-- The CXX compiler identification is PGI 22.3.0
-- Check for working CXX compiler: /opt/nvidia/hpc_sdk/Linux_x86_64/22.3/compilers/bin/nvc++
-- Check for working CXX compiler: /opt/nvidia/hpc_sdk/Linux_x86_64/22.3/compilers/bin/nvc++ — works
…
root@411ece11247c:~/fmt/build# make
Scanning dependencies of target fmt
[  1%] Building CXX object CMakeFiles/fmt.dir/src/format.cc.o
…
(there are some warnings here)
root@411ece11247c:~/fmt/build# make test
Running tests...
Test project /root/fmt/build
      Start  1: args-test
 1/20 Test  #1: args-test ........................   Passed    0.01 sec
…
80% tests passed, 4 tests failed out of 20

Total Test time (real) =   0.22 sec

The following tests FAILED:
	  3 - chrono-test (Failed)
	  7 - format-test (Failed)
	 12 - printf-test (Failed)
	 17 - xchar-test (Failed)
Errors while running CTest
make: *** [Makefile:97: test] Error 8
```
(I had to add `-DCMAKE_CXX_FLAGS="-tp haswell"` because these compilers use the equivalent of `-march=native` by default, but they seem to mis-detect the AVX512 capabilities of my Ice Lake laptop CPU and produce binaries that fail with illegal instruction errors. This might not be needed on other systems.)

[...]

PR #3043, opened 2022-08-18
nvhpc/22.3: workaround for c++17 mode

olupton:
https://github.com/fmtlib/fmt/issues/3028 discusses several issues with using fmt and the NVIDIA HPC compilers.

One issue leads to assertions failures and wrong results, even in simple examples. That issue seems to be a compiler but affecting version 22.3 of the compilers, specifically in C++17 mode. This workaround avoids the issue.

There are still other test failures with this patch, see https://github.com/fmtlib/fmt/issues/3028 for more information, but this seems sufficient to fix basic usage with 22.3 and C++17.

I assume it would also be fine to remove all of the macros and use the `*(foo + index)` style for all compilers (with an appropriate comment about nvhpc 22.3), if you prefer.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -1557,7 +1557,10 @@ FMT_CONSTEXPR inline fp get_cached_power(int min_exponent,
   const int dec_exp_step = 8;
   index = (index - first_dec_exp - 1) / dec_exp_step + 1;
   pow10_exponent = first_dec_exp + index * dec_exp_step;
-  return {data::pow10_significands[index], data::pow10_exponents[index]};
+  // Using *(x + index) instead of x[index] avoids an issue with some compilers
+  // using the EDG frontend (e.g. nvhpc/22.3 in C++17 mode).
+  return {*(data::pow10_significands + index),
+          *(data::pow10_exponents + index)};
 }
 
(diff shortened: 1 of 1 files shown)


phprus:
This error is also present in other compilers with EDG frontend. Not only in NVIDIA HPC compilers.

include/fmt/format.h
```diff
@@ -1557,7 +1557,10 @@ FMT_CONSTEXPR inline fp get_cached_power(int min_exponent,
   const int dec_exp_step = 8;
   index = (index - first_dec_exp - 1) / dec_exp_step + 1;
   pow10_exponent = first_dec_exp + index * dec_exp_step;
-  return {data::pow10_significands[index], data::pow10_exponents[index]};
+  // Using *(x + index) instead of x[index] avoids an issue with some compilers
+  // using the EDG frontend (e.g. nvhpc/22.3 in C++17 mode)
```
vitaut, line 1561: nit: please use proper punctuation in comments that are full sentences, i.e. please add `.` at the end.

olupton: Done.

vitaut:
Thank you.

Issue #3038, opened 2022-08-16
write_range_entry fails writing non-utf8 string


stevenzzzz:
there seems to be a bug in "Inline auto write_range_entry(OutputIt out, const std::T& t)" would cause ASAN error if the string is a invalid-utf8 string. 

using example: "\xf0\x28", it will cause an ASAN error.

stevenzzzz:
hmm, not sure if godbolt vm supports the asan sanitizer. 

Could you try this? 

```
#include <fmt/core.h>
#include <fmt/format.h>
#include <unordered_map>

int main(){

std::string invalid_utf8="\xf0\x28";

fmt::print("Foo Map: {}", invalid_utf8);

}
```

I actually got here from Envoy, which uses spdlog --> libfmt. 

I have synced to 9.0.0, and "bazel test" on logging the afore mentioned string would end up with: 

external/com_github_fmtlib_fmt/include/fmt/core.h:426: assertion failed: negative valuelibc++abi: terminating                                                                     [2022-08-16 17:57:43.537][16][critical][backtrace] [./source/server/backtrace.h:104] Caught Aborted, suspect faulting address 0xfffe00000010                                      
[2022-08-16 17:57:43.538][16][critical][backtrace] [./source/server/backtrace.h:91] Backtrace (use tools/stack_decode.py to get line numbers):                                    
[2022-08-16 17:57:43.538][16][critical][backtrace] [./source/server/backtrace.h:92] Envoy version: 0/1.24.0-dev/test/DEBUG/BoringSSL
[2022-08-16 17:57:43.551][16][critical][backtrace] [./source/server/backtrace.h:96] #0: Envoy::SignalAction::sigHandler() [0x1517d9c]                                             
[2022-08-16 17:57:43.551][16][critical][backtrace] [./source/server/backtrace.h:96] #1: __restore_rt [0x7f6475ecc420]                                                             
[2022-08-16 17:57:43.563][16][critical][backtrace] [./source/server/backtrace.h:96] #2: fmt::v8::detail::to_unsigned<>() [0xf4dbd4]                                               
[2022-08-16 17:57:43.573][16][critical][backtrace] [./source/server/backtrace.h:96] #3: fmt::v8::detail::find_escape() [0xf6f2a7]                                                 
[2022-08-16 17:57:43.584][16][critical][backtrace] [./source/server/backtrace.h:96] #4: fmt::v8::detail::write_range_entry<>() [0xf6e757]
[...]

PR #3044, opened 2022-08-18
Fix decoder on broken utf8 sequences.

phprus:
Issue: #3038

``utf8_decode`` did not check if the first byte is correct.
``for_each_codepoint`` did not check for overflow on broken codepoint.

cc @stevenzzzz @dyfrgi

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -602,6 +602,7 @@ FMT_CONSTEXPR FMT_NOINLINE auto copy_str_noinline(InputIt begin, InputIt end,
  */
 FMT_CONSTEXPR inline auto utf8_decode(const char* s, uint32_t* c, int* e)
     -> const char* {
+  constexpr const int prefix_masks[] = {0x00, 0x80, 0xe0, 0xf0, 0xf8};
   constexpr const int masks[] = {0x00, 0x7f, 0x1f, 0x0f, 0x07};
   constexpr const uint32_t mins[] = {4194304, 0, 128, 2048, 65536};
   constexpr const int shiftc[] = {0, 18, 12, 6, 0};
@@ -628,6 +629,8 @@ FMT_CONSTEXPR inline auto utf8_decode(const char* s, uint32_t* c, int* e)
   *e |= uchar(s[3]) >> 6;
   *e ^= 0x2a;  // top two bits of each tail byte correct?
   *e >>= shifte[len];
+  *e |= ((uchar(s[0]) & prefix_masks[len]) !=
+         uchar((prefix_masks[len] << 1) & 0xFF));  // first byte correct?
 
   return next;
 }
@@ -643,8 +646,8 @@ FMT_CONSTEXPR void for_each_codepoint(string_view s, F f) {
     auto error = 0;
     auto end = utf8_decode(buf_ptr, &cp, &error);
     bool result = f(error ? invalid_code_point : cp,
-                    string_view(ptr, to_unsigned(end - buf_ptr)));
-    return result ? end : nullptr;
+                    string_view(ptr, error ? 1 : to_unsigned(end - buf_ptr)));
+    return result ? (error ? buf_ptr + 1 : end) : nullptr;
   };
   auto p = s.data();
   const size_t block_size = 4;  // utf8_decode always reads blocks of 4 chars.
--- a/test/ranges-test.cc
+++ b/test/ranges-test.cc
@@ -380,8 +380,15 @@ TEST(ranges_test, escape_string) {
     EXPECT_EQ(fmt::format("{}", vec{"\xcd\xb8"}), "[\"\\u0378\"]");
     // Unassigned Unicode code points.
     EXPECT_EQ(fmt::format("{}", vec{"\xf0\xaa\x9b\x9e"}), "[\"\\U0002a6de\"]");
+    // Broken utf-8.
     EXPECT_EQ(fmt::format("{}", vec{"\xf4\x8f\xbf\xc0"}),
               "[\"\\xf4\\x8f\\xbf\\xc0\"]");
+    EXPECT_EQ(fmt::format("{}", vec{"\xf0\x28"}), "[\"\\xf0(\"]");
+    EXPECT_EQ(fmt::format("{}", vec{"\xe1\x28"}), "[\"\\xe1(\"]");
+    EXPECT_EQ(fmt::format("{}", vec{std::string("\xf0\x28\0\0anything", 12)}),
+              "[\"\\xf0(\\x00\\x00anything\"]");
+
+    // Correct utf-8.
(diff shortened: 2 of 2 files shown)


include/fmt/format.h
```diff
@@ -628,6 +629,8 @@ FMT_CONSTEXPR inline auto utf8_decode(const char* s, uint32_t* c, int* e)
   *e |= uchar(s[3]) >> 6;
   *e ^= 0x2a;  // top two bits of each tail byte correct?
   *e >>= shifte[len];
+  *e |= ((uchar(s[0]) & prefix_masks[len]) !=
+         uchar((prefix_masks[len] << 1) & 0xFF));  // first byte correct?
```
vitaut, line 633: I don't think this is needed because `utf8_decode` already gives an error for "\xf0\x28...".

vitaut (changes requested):
Thanks for the PR.

include/fmt/format.h
phprus: Without this check failed at sequence ``\xf4\x8f\xbf\xc0``.
```
/Users/phprus/Devel/fmt/test/ranges-test.cc:393: Failure
  fmt::format("{}", vec{"\xf4\x8f\xbf\xc0"})
    Which is: "[\"\\xf4\\x0f\xBF\\xc0\"]"
  "[\"\\xf4\\x8f\\xbf\\xc0\"]"
```

``utf8_decode`` return zero in ``e``.

vitaut: Could you add a test case then?

vitaut: Nevermind, I missed the code snippet which shows it's already there.

phprus: This test already present:
https://github.com/fmtlib/fmt/blob/541cd218381493728241fcc690e8c362434a37ea/test/ranges-test.cc#L383-L384

This patch fix two error.
First error fixed by add condition ``error ? 1 : to_unsigned(end - buf_ptr)``
Second - by add first byte check.

The first error masked the second on the sequence ``\xf4\x8f\xbf\xc0``

vitaut: AFAIC the current version of `utf8_decode` already returns an error for "\xf4\x8f\xbf\xc0" too: https://godbolt.org/z/E194qvshK

phprus: Step 1:
``"\xf4\x8f\xbf\xc0"`` - is not valid utf-8, print ``\xf4``, remove ``\xf4`` from sequence.
Step 2:
``"\x8f\xbf\xc0\0"`` - is not valid utf-8, but utf8_decode return no error and invalid codepoint (15, not 143).

See: https://godbolt.org/z/EbeEex4Gf

include/fmt/format.h
```diff
@@ -643,7 +646,7 @@ FMT_CONSTEXPR void for_each_codepoint(string_view s, F f) {
     auto error = 0;
     auto end = utf8_decode(buf_ptr, &cp, &error);
     bool result = f(error ? invalid_code_point : cp,
-                    string_view(ptr, to_unsigned(end - buf_ptr)));
+                    string_view(ptr, error ? 1 : to_unsigned(end - buf_ptr)));
```
stevenzzzz, line 649: we saw error here "end - buf_ptr" is < 0, but I dont follow how that could happen with invalid code point.

include/fmt/format.h
```diff
@@ -643,7 +646,7 @@ FMT_CONSTEXPR void for_each_codepoint(string_view s, F f) {
...
     return result ? end : nullptr;
```
stevenzzzz, line 650: could you help me to understand in the example of "\xF0\x28", is this end still buf_ptr+4 even tho there is an error?

phprus: @stevenzzzz 
No. On invalid utf-8 sequence, processing is one byte at a time.

Please describe your testing process.

My environment:
macOS 12.5.1
Apple clang version 13.1.6 (clang-1316.0.21.2.5)

Test steps:
```
CXXFLAGS="-fsanitize=address" cmake -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_STANDARD=20 -DFMT_PEDANTIC=ON -DFMT_WERROR=ON /path/to/patched/fmt
make
./bin/ranges-test
```
The test shows no errors.

String "\xF0\x28" present in updated test.


Without this PR:
```
ERROR: AddressSanitizer: negative-size-param: (size=-2)
```

stevenzzzz: I might have overlooked something, but for "\xF0\x28\0\0anything" the utf8_decode_new/decode would return 4 (as expected), when the error is set to 42.  Is that "4" gonna skip the "\x28\0\0"?

same test:  https://godbolt.org/z/EbeEex4Gf

stevenzzzz: thanks for working on this!

We got the ASAN error by printing a protobuf debugstring() which is arbitrary binary bytes, unfortunately I can't show the protobuf to you. 


I am a little bit lost on how the printing works, could you help me to understand where is the "skip one byte on error" pls?  per this piece of code: 
```
    for (auto end = p + s.size() - block_size + 1; p < end;) {
      p = decode(p, p);
      if (!p) return;
    }
```
it seems p would advance 4 bytes when encode(p, p) is called on "\xof\x28".

phprus: @stevenzzzz 
Please replace
```c++
  auto decode = [f](const char* buf_ptr, const char* ptr) {
    auto cp = uint32_t();
    auto error = 0;
    auto end = utf8_decode(buf_ptr, &cp, &error);
    bool result = f(error ? invalid_code_point : cp,
                    string_view(ptr, error ? 1 : to_unsigned(end - buf_ptr)));
    return result ? end : nullptr;
  };
```

with:
```c++
  auto decode = [f](const char* buf_ptr, const char* ptr) {
    auto cp = uint32_t();
    auto error = 0;
    auto end = utf8_decode(buf_ptr, &cp, &error);
    bool result = f(error ? invalid_code_point : cp,
                    string_view(ptr, error ? 1 : to_unsigned(end - buf_ptr)));
    return result ? (error ? buf_ptr + 1 : end) : nullptr;
  };
```

and run test.

stevenzzzz: I was wondering why you didn't replace that in this PR.

phprus: @stevenzzzz 
Did this replacement fix the bug? If yes, I will update the PR.

I didn't see this bug before.

include/fmt/format.h
vitaut: Makes sense, thanks.

vitaut (approved):

vitaut:
Merged, thanks! @phprus, could you report this to https://github.com/skeeto/branchless-utf8 where this implementation originates from?

phprus:
@vitaut : https://github.com/skeeto/branchless-utf8/issues/8

PR #3053, opened 2022-08-22
Disable non-type template args for nvhpc

bernhardmgruber:
Hi!

When I use `fmt` in a project with `nvhpc` as compiler in C++20 mode, I get the following error:
```bash
"/usr/local/share/vcpkg/installed/x64-linux/include/fmt/format.h", line 4138: error: a literal operator template must have a template parameter list equivalent to "<char ...>"
  template <detail_exported::fixed_string Str> constexpr auto operator""_a() {
                                                              ^
```

In spirit of #742, the use of non-type template paramters could also be disabled for nvhpc. This is what this PR proposes.

Workaround for now: I can specify `-DFMT_USE_NONTYPE_TEMPLATE_ARGS=0` in my project.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -276,7 +276,8 @@
 #ifndef FMT_USE_NONTYPE_TEMPLATE_ARGS
 #  if defined(__cpp_nontype_template_args) &&                  \
       ((FMT_GCC_VERSION >= 903 && FMT_CPLUSPLUS >= 201709L) || \
-       __cpp_nontype_template_args >= 201911L)
+       __cpp_nontype_template_args >= 201911L) &&              \
+      !defined(__NVCOMPILER)
 #    define FMT_USE_NONTYPE_TEMPLATE_ARGS 1
 #  else
 #    define FMT_USE_NONTYPE_TEMPLATE_ARGS 0


include/fmt/core.h
```diff
@@ -276,7 +276,8 @@
 #ifndef FMT_USE_NONTYPE_TEMPLATE_ARGS
 #  if defined(__cpp_nontype_template_args) &&                  \
       ((FMT_GCC_VERSION >= 903 && FMT_CPLUSPLUS >= 201709L) || \
-       __cpp_nontype_template_args >= 201911L)
+       __cpp_nontype_template_args >= 201911L) &&              \
+      !defined(_NVCOMPILER)
```
vitaut, line 280: Why do we need to exclude `_NVCOMPILER`, i.e. why does it pass the above condition if nontype template parameters are not supported?

phprus: ``__NVCOMPILER``
https://docs.nvidia.com/hpc-sdk/archive/22.1/compilers/hpc-compilers-ref-guide/index.html#unique_803747402

vitaut: Yeah, but does it lie about `__cpp_nontype_template_args` or incorrectly masquerades as gcc?

phprus: This may not be a nvhpc issue.

NTTP support - is EDG-frontend feature.

See:
    https://www.edg.com/c/features
    https://docs.google.com/spreadsheets/d/1H-aqjzVI2a-XQKGtw0xaS0tyjD0FcoQP8ttJI9JZQTc/edit#gid=0
    https://www.edg.com/docs/edg_cpp.pdf

Non-Type Template Parameters: Class Types in Non-Type Template Parameters - Release 6.2 (partial gcc-9 compatible).

We need to check the ``__EDG_VERSION__`` value, but I can't find the versioning scheme.

nvhpc 22.7 defined ``__EDG_VERSION__`` as ``603``.

I assume that need to check the condition:
```c++
(!defined(__EDG_VERSION__) || __EDG_VERSION__ >= 620)
```

I will try to clarify this question with the compiler developers.

bernhardmgruber: So `nvc++` (nvhpc version 22.7) on my system (Ubuntu 22.04), compiling with `-std=gnu++20` defines:
```c++
#define __cpp_nontype_template_args 201911L
#define __cplusplus 202002L
#define __GNUC__ 11
#define __GNUC_MINOR__ 2
```
fmt recognizes that it is not gcc and defines:
```c++
#define FMT_GCC_VERSION 0
```
However, since nvhpc defines `__cpp_nontype_template_args 201911L`, it passes the `#if` and sets `FMT_USE_NONTYPE_TEMPLATE_ARGS 1`, which later causes the error mentioned in the OP.

Feel free to suggest a better fix! :)

And @phprus thank you for catching my typo!

bernhardmgruber: I verified locally again that my diff fixes compilation.

vitaut:
Thank you

bernhardmgruber:
> Thank you

Well, thank you for providing this amazing library! :)

PR #3056, opened 2022-08-24
Fix bugs in utf8 decoder

phprus:
Info:
https://github.com/skeeto/branchless-utf8/issues/8


phprus:
cc @vitaut

vitaut:
Thank you!

PR #3057, opened 2022-08-25
Added workarounds for compilers with EDG frontend (LCC and Nvidia)

phprus:
Added workarounds for compilers with EDG frontend (LCC and Nvidia) that emulate GCC.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -755,8 +755,11 @@ class compile_parse_context
   using base::check_arg_id;
 
   FMT_CONSTEXPR void check_dynamic_spec(int arg_id) {
+    detail::ignore_unused(arg_id);
+#if !defined(__LCC__)
     if (arg_id < num_args_ && types_ && !is_integral_type(types_[arg_id]))
       this->on_error("width/precision is not integer");
+#endif
   }
 };
 FMT_END_DETAIL_NAMESPACE
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -286,7 +286,7 @@
 
 // Enable minimal optimizations for more compact code in debug mode.
 FMT_GCC_PRAGMA("GCC push_options")
-#if !defined(__OPTIMIZE__) && !defined(__NVCOMPILER)
+#if !defined(__OPTIMIZE__) && !defined(__NVCOMPILER) && !defined(__LCC__)
 FMT_GCC_PRAGMA("GCC optimize(\"Og\")")
 #endif
 
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -277,7 +277,7 @@
 #  if defined(__cpp_nontype_template_args) &&                  \
       ((FMT_GCC_VERSION >= 903 && FMT_CPLUSPLUS >= 201709L) || \
        __cpp_nontype_template_args >= 201911L) &&              \
-      !defined(__NVCOMPILER)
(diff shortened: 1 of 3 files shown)


vitaut:
Thanks for the PR! Could you post errors that are being fixed for reference?

phprus:
I have error messages in Russian. The error messages from the Nvidia compiler are the same.

1) (LCC only)
```
lcc: ".../include/fmt/core.h", строка 289: предупреждение #3317: неизвестная опция "Og" директивы #pragma "GCC optimize" [-Wpragmas].
```

2) (LCC and Nvidia)
```
lcc: ".../test/format-test.cc", строка 2154: ошибка fmtlib#28: выражение должно иметь константное значение

      static_assert(test_error<__VA_ARGS__>(fmt, error), "")
                    ^
 в раскрытие макроса "EXPECT_ERROR" на строке 2182
    EXPECT_ERROR("{:{}}", "width/precision is not integer", int, double);
    ^
lcc: ".../include/fmt/core.h", строка 759: примечание #2700: попытка доступа к просроченному хранилищу
      if (arg_id < num_args_ && types_ && !is_integral_type(types_[arg_id]))
                                                                  ^

```

3) (workaround for Nvidia in PR #3053)
```
lcc: ".../include/fmt/format.h", строка 4159: ошибка fmtlib#2500: шаблон литерального оператора должен иметь список параметров шаблона, эквивалентный "<char ...>"

  template <detail_exported::fixed_string Str> constexpr auto operator""_a() {
                                                              ^
```

4) Inconsistency FMT_CONSTEXPR20 and test
```
lcc: ".../test/compile-test.cc", строка 327: ошибка fmtlib#2933:
          call to consteval функция
          "test_format<max_string_length,Char,Args...,<auto-1>>(auto, const Args &...) [где max_string_length=6UL, Char=char, Args=<bool>, <auto-1>=FMT_COMPILE_STRING]"
          did not produce a valid constant expression
    EXPECT_EQ("true ", test_format<6>(FMT_COMPILE("{:5}"), true));
                                     ^
 в раскрытие макроса "EXPECT_EQ" на строке 327
```

phprus:
Update:
LCC - Successful build and tests (Debug and Release, C++14,17,20)
Nvidia compilers - Successful build, but some tests failed (float point formatting).

phprus:
Nvidia compilers error message for bug 2:
https://github.com/fmtlib/fmt/issues/3028#issuecomment-1208303478

include/fmt/core.h
```diff
@@ -755,8 +755,11 @@ class compile_parse_context
   using base::check_arg_id;
 
   FMT_CONSTEXPR void check_dynamic_spec(int arg_id) {
+    detail::ignore_unused(arg_id);
+#if !defined(__LCC__)
```
vitaut, line 759: Do we need to exclude `__NVCOMPILER` here too?

phprus: I don't know.
My tests have shown that this set of ``ifdef``s is sufficient for Nvidia, but I do not use the Nvidia compiler in production and do not know its specifics.

LCC supports ``is_constant_evaluated`` but Nvidia does not. Therefore, for Nvidia, disabling the test is enough.

This ``ifdef`` disables code for LCC, as there is an LCC specific bug (Internal compiler error).

LCC and Elbrus is my hobby, not production usage, but I have access to it for testing.

vitaut:
Thank you

Issue #2977, opened 2022-07-07
Support std::exception


fekir:
Exceptions are not supported, as they are not formattable:

````
int main() try {
  std::vector <int> vec;
  vec.at(42);
} catch(const std::out_of_range& ex){
    fmt::print("{}", ex);
}
````

https://godbolt.org/z/ejTMEYr3b

In practice, the expected/desired behavior would be to print the error message.
Ie if they would behave as-if

````
int main() try {
  std::vector<int> vec;
  vec.at(42);
} catch(const std::out_of_range& ex){
  fmt::print("{}", ex.what());
}
````

This can be accomplished by defining a custom formatter for `std::exception`, but as `fmt` already provides formatters for many/most types from the standard library, support for std::exception should be part of it.

Issue #3012, opened 2022-07-31
Questions about formatters for `std::` entities


alexezeder:
I was thinking about implementing https://github.com/fmtlib/fmt/issues/2977, it seems pretty straightforward, right? Just add a simple `fmt::formatter` with `write(exception.what())` into `fmt/std.h` and call it a day. But there is also the formatter for `std::variant`, it outputs the underlying value as `variant(42)` with escaping string and char values. The other formatter there, for `std::filesystem::path`, outputs `"/home"` without any `path` prefix, which just mimics `operator<<` of course.

So, such a long introduction to ask a few questions:
1. Should a formatter for `std::exception` have any prefix or escape the _what_ string?
2. Since I also want to add a formatter for `std::optional`: should it have `optional` prefix and also escape the string and char values? Just like the formatter for `std::variant` does.
3. Is there any point to add a formatter for `std::source_location`? Since it doesn't have `operator<<` provided and it consists of 4 components that can be combined in many different ways. For example, it can output using some _default_ format, but what would it be? For me, it's `source.cpp:line:column`, but who cares what is preferable for me.
4. Should we add an option to the `std::variant` formatter to skip `variant` prefix and an escaping of string and char values? The same decision should also apply to the `std::optional` formatter.
5. Should we add an option to the `std::filesystem` formatter to get a non-escaped path from it?
6. Should we add a formatter for `std::stacktrace` (and its entry) if it's currently not testable on CI?

And here comes the crazy question:

  7. Should we add format specifiers for underlying types of `std::optional` (easy) and `std::variant` (dizzy)? It's like with a range formatter IIRC, treat all specifiers after another `:` as underlying specifiers, so the format string `{::x}` with `std::optional{42}` would produce `2a` (considering that there is no `optional` prefix). This idea works fine with `std::optional`, but for `std::variant` it works only if the format string has underlying specs for the current type of variant.

vitaut:
> I was thinking about implementing https://github.com/fmtlib/fmt/issues/2977, it seems pretty straightforward, right? Just add a simple fmt::formatter with write(exception.what()) into fmt/std.h and call it a day.

Right.

> Should a formatter for std::exception have any prefix or escape the what string?

I don't think we should need to format anything other than the message by default but we should provide an escaped option similarly to strings.

> Since I also want to add a formatter for std::optional: should it have optional prefix and also escape the string and char values?

I would recommend checking what other languages do.

[...]

PR #3062, opened 2022-08-27
Add formatter for std::exception

zach2good:
Adds a single simple formatter for `std::exception` and accompanying test.

Reference:
https://github.com/fmtlib/fmt/issues/2977
https://github.com/fmtlib/fmt/issues/3012

Closes https://github.com/fmtlib/fmt/issues/2977

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -8,6 +8,7 @@
 #ifndef FMT_STD_H_
 #define FMT_STD_H_
 
+#include <exception>
 #include <thread>
 #include <type_traits>
 #include <utility>
@@ -166,6 +167,20 @@ struct formatter<
   }
 };
 FMT_END_NAMESPACE
-#endif
+#endif  // __cpp_lib_variant
+
+FMT_BEGIN_NAMESPACE
+template <typename T, typename Char>
+struct formatter<
+    T, Char,
+    typename std::enable_if<std::is_base_of<std::exception, T>::value>::type>
+    : formatter<string_view> {
+  template <typename FormatContext>
+  auto format(const std::exception& ex, FormatContext& ctx) const ->
+      typename FormatContext::iterator {
+    return fmt::formatter<string_view>::format(ex.what(), ctx);
+  }
+};
+FMT_END_NAMESPACE
 
 #endif  // FMT_STD_H_
--- a/test/std-test.cc
+++ b/test/std-test.cc
@@ -6,11 +6,11 @@
 // For the license information refer to format.h.
 
 #include "fmt/std.h"
-#include "fmt/ranges.h"
 
 #include <string>
 #include <vector>
 
+#include "fmt/ranges.h"
 #include "gtest/gtest.h"
 
 TEST(std_test, path) {
@@ -76,4 +76,23 @@ TEST(std_test, variant) {
   EXPECT_EQ(fmt::format("{}", v4), "variant(monostate)");
   EXPECT_EQ(fmt::format("{}", v5), "variant(\"yes, this is variant\")");
 #endif
+}
+
+TEST(std_test, exception) {
+  std::string str("Test Exception");
+  std::string escstr = fmt::format("\"{}\"", str);
+
+  try {
+    throw std::runtime_error(str);
+  } catch (const std::exception& ex) {
+    EXPECT_EQ(fmt::format("{}", ex), str);
+    EXPECT_EQ(fmt::format("{:?}", ex), escstr);
+  }
+
+  try {
+    throw std::runtime_error(str);
+  } catch (const std::runtime_error& ex) {
+    EXPECT_EQ(fmt::format("{}", ex), str);
+    EXPECT_EQ(fmt::format("{:?}", ex), escstr);
+  }
 }


zach2good:
Will come back and fix those failing CI jobs in a couple of days 👍

include/fmt/std.h
```diff
@@ -166,6 +167,21 @@ struct formatter<
...
-#endif
+
+#endif  // __cpp_lib_variant
+
+FMT_BEGIN_NAMESPACE
+template <typename T, typename Char>
+struct formatter<
+    T, Char,
+    typename std::enable_if<std::is_base_of<std::exception, T>::value>::type>
+    : formatter<std::string> {
```
phprus, line 178: Suggest:
Change ``std::string`` with ``string_view``.

Like example from documentation:
https://fmt.dev/latest/api.html#:~:text=template%20%3C%3E%20struct%20fmt%3A%3Aformatter%3Ccolor%3E%3A%20formatter%3Cstring_view%3E%20%7B

test/std-test.cc
```diff
@@ -77,3 +77,34 @@ TEST(std_test, variant) {
...
+#  if _MSC_VER > 1920
+  std::string str("invalid vector subscript");
+#  else
+   std::string str("invalid vector<T> subscript");
+#  endif
+#elif not defined(NDEBUG) && (defined(__clang__) || defined(__GNUC__)) || defined (__APPLE__)
+  std::string str("vector::_M_range_check: __n (which is 42) >= this->size() (which is 0)");
+#else
+    std::string str("vector");
+#endif
```
alexezeder, line 92: Maybe it's easier to create a custom exception type with some predefined message instead of trying to support all compilers?

zach2good: Yeah, that's a much better idea than going down this rabbit hole!

phprus: We can use standard exceptions.
For example (https://godbolt.org/z/qrGd455Wq):
```c++
#include <stdexcept>
#include <iostream>

int main() {
    try {
        throw std::runtime_error("My exception");
    } catch(const std::exception &e) {
        std::cerr << e.what() << std::endl; // Print: My exception
    }
    try {
        throw std::runtime_error("My exception");
    } catch(const std::runtime_error &e) {
        std::cerr << e.what() << std::endl; // Print: My exception
    }
    return 0;
}
```

alexezeder: I mean, it's not only easier, but it's also the only portable way to test it IMHO

Baardi:
Sorry for a stupid, uneducated question.
But does it make sense to include the exception type as well, when formatting?

E.g. a std::runtime_error("My exception") would then be formatted as "std::runtime_error: My exception", as it typically is formatted in C#, by using typeid(xxx).name()? 
Or is that what you previously discussed, and decided against it due to portability issues?

zach2good:
Not necessarily previously discussed (@vitaut's suggestions in https://github.com/fmtlib/fmt/issues/3012#issuecomment-1206515140), but you're right in that it isn't super portable to do that:

Using:
```cpp
return fmt::formatter<string_view>::format(fmt::format("{}: {}", typeid(T).name(), ex.what()), ctx);
```

clang (name mangling):
```
/mnt/c/dev/fmt/test/std-test.cc:88: Failure
  fmt::format("{}", ex)
    Which is: "St9exception: Test Exception"
  str
    Which is: "Test Exception"
/mnt/c/dev/fmt/test/std-test.cc:89: Failure
  fmt::format("{:?}", ex)
    Which is: "\"St9exception: Test Exception\""
  escstr
    Which is: "\"Test Exception\""
/mnt/c/dev/fmt/test/std-test.cc:95: Failure
  fmt::format("{}", ex)
    Which is: "St13runtime_error: Test Exception"
  str
    Which is: "Test Exception"
/mnt/c/dev/fmt/test/std-test.cc:96: Failure
  fmt::format("{:?}", ex)
    Which is: "\"St13runtime_error: Test Exception\""
  escstr
    Which is: "\"Test Exception\""
```

MSVC (addition of "class"):
```
C:\dev\fmt\test\std-test.cc(88): error: Expected equality of these values:
  fmt::format("{}", ex)
    Which is: "class std::exception: Test Exception"
  str
    Which is: "Test Exception"
C:\dev\fmt\test\std-test.cc(89): error: Expected equality of these values:
  fmt::format("{:?}", ex)
    Which is: "\"class std::exception: Test Exception\""
  escstr
    Which is: "\"Test Exception\""
C:\dev\fmt\test\std-test.cc(95): error: Expected equality of these values:
  fmt::format("{}", ex)
    Which is: "class std::runtime_error: Test Exception"
  str
    Which is: "Test Exception"
C:\dev\fmt\test\std-test.cc(96): error: Expected equality of these values:
  fmt::format("{:?}", ex)
    Which is: "\"class std::runtime_error: Test Exception\""
  escstr
    Which is: "\"Test Exception\""
```

bamo-volue:
[For gcc you can demangle the name](https://gcc.gnu.org/onlinedocs/libstdc++/manual/ext_demangling.html), and for msvc you can strip the class I guess. Not sure about clang.

[Searching online](https://stackoverflow.com/questions/281818/unmangling-the-result-of-stdtype-infoname), it generally seems like a doable thing, although it would introduce a lot of complexity that might not not be wanted in fmtlib.

Baardi:
[For gcc you can demangle the name](https://gcc.gnu.org/onlinedocs/libstdc++/manual/ext_demangling.html), and for msvc you can strip the class I guess. Not sure about clang.

[Searching online](https://stackoverflow.com/questions/281818/unmangling-the-result-of-stdtype-infoname), it generally seems like a doable thing, at least for the 3 main compilers, although it would introduce a lot of complexity that might not not be wanted in fmtlib.

vitaut:
Thank you!

PR #3066, opened 2022-08-29
Fix compilation error with gcc-7.2.0

gsjaardema:
Without the added line, the gcc-7.2.0 compiler will give the following error:
```
/opt/compiler-explorer/libs/fmt/trunk/include/fmt/format.h:1240:8: error: uninitialized variable 'buffer' in 'constexpr' function
   Char buffer[digits10<UInt>() + 1];
        ^~~~~~
```
See https://godbolt.org/z/fh7TMs9qs


vitaut:
Thank you

Issue #3068, opened 2022-09-01
Compile error when formatting volatile values before gcc11


NewbieOrange:
```
In file included from /opt/compiler-explorer/libs/fmt/trunk/include/fmt/std.h:26:0,
                 from <source>:1:
/opt/compiler-explorer/gcc-7.5.0/include/c++/7.5.0/variant: In instantiation of 'struct std::variant_size<volatile bool>':
/opt/compiler-explorer/libs/fmt/trunk/include/fmt/std.h:135:31:   required from 'constexpr const bool fmt::v9::is_variant_like<volatile bool>::value'
/opt/compiler-explorer/gcc-7.5.0/include/c++/7.5.0/type_traits:143:12:   required from 'struct std::__and_<fmt::v9::is_variant_like<volatile bool>, fmt::v9::is_variant_formattable<volatile bool, char> >'
/opt/compiler-explorer/gcc-7.5.0/include/c++/7.5.0/type_traits:162:12:   required from 'struct std::conjunction<fmt::v9::is_variant_like<volatile bool>, fmt::v9::is_variant_formattable<volatile bool, char> >'
/opt/compiler-explorer/gcc-7.5.0/include/c++/7.5.0/type_traits:177:27:   required from 'constexpr const bool std::conjunction_v<fmt::v9::is_variant_like<volatile bool>, fmt::v9::is_variant_formattable<volatile bool, char> >'
/opt/compiler-explorer/libs/fmt/trunk/include/fmt/std.h:146:22:   [ skipping 9 instantiation contexts, use -ftemplate-backtrace-limit=0 to disable ]
[...]

PR #3071, opened 2022-09-01
Fix for issue #3068

phprus:
Fix for issue #3068


phprus:
I think PR #3072 offers a more better solution.

vitaut:
> I think PR #3072 offers a more better solution.

Closing in favor of that PR then.

PR #3076, opened 2022-09-02
Add class name output to formatter for std::exception

phprus:
Discussion:
https://github.com/fmtlib/fmt/pull/3062#issuecomment-1235315255

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -8,9 +8,12 @@
 #ifndef FMT_STD_H_
 #define FMT_STD_H_
 
+#include <cstdlib>
 #include <exception>
+#include <memory>
 #include <thread>
 #include <type_traits>
+#include <typeinfo>
 #include <utility>
 
 #include "ostream.h"
@@ -25,6 +28,16 @@
 #  endif
 #  if FMT_HAS_INCLUDE(<variant>)
 #    include <variant>
+#  endif
+#endif
+
+// GCC 4 does not support FMT_HAS_INCLUDE.
+#if FMT_HAS_INCLUDE(<cxxabi.h>) || defined(__GLIBCXX__)
+#  include <cxxabi.h>
+// Android NDK with gabi++ library on some archtectures does not implement
+// abi::__cxa_demangle().
+#  ifndef __GABIXX_CXXABI_H__
+#    define FMT_HAS_ABI_CXA_DEMANGLE
 #  endif
 #endif
 
@@ -170,12 +183,56 @@ FMT_BEGIN_NAMESPACE
 template <typename T, typename Char>
 struct formatter<
     T, Char,
-    typename std::enable_if<std::is_base_of<std::exception, T>::value>::type>
-    : formatter<string_view> {
-  template <typename FormatContext>
-  auto format(const std::exception& ex, FormatContext& ctx) const ->
-      typename FormatContext::iterator {
-    return fmt::formatter<string_view>::format(ex.what(), ctx);
+    typename std::enable_if<std::is_base_of<std::exception, T>::value>::type> {
+ private:
+  bool with_typename_ = false;
+
+ public:
+  FMT_CONSTEXPR auto parse(basic_format_parse_context<Char>& ctx)
+      -> decltype(ctx.begin()) {
+    auto it = ctx.begin();
+    auto end = ctx.end();
+    if (it == end || *it == '}') return it;
+    if (*it == 't') {
+      ++it;
+      with_typename_ = true;
+    }
+    return it;
+  }
+
+  template <typename OutputIt>
+  auto format(const std::exception& ex,
+              basic_format_context<OutputIt, Char>& ctx) const -> OutputIt {
+    basic_format_specs<Char> spec;
+    auto out = ctx.out();
+    if (!with_typename_)
+      return detail::write_bytes(out, string_view(ex.what()), spec);
+
+    const std::type_info& ti = typeid(ex);
+#ifdef FMT_HAS_ABI_CXA_DEMANGLE
+    int status = 0;
+    std::size_t size = 0;
+    std::unique_ptr<char, decltype(&std::free)> demangled_name_ptr(
+        abi::__cxa_demangle(ti.name(), nullptr, &size, &status), &std::free);
+    out = detail::write_bytes(
+        out,
+        string_view(demangled_name_ptr ? demangled_name_ptr.get() : ti.name()),
+        spec);
+#elif FMT_MSC_VERSION
+    string_view demangled_name_view(ti.name());
+    if (demangled_name_view.starts_with("class "))
+      demangled_name_view.remove_prefix(6);
+    else if (demangled_name_view.starts_with("struct "))
+      demangled_name_view.remove_prefix(7);
+    out = detail::write_bytes(out, demangled_name_view, spec);
+#else
+    out = detail::write_bytes(out, string_view(ti.name()), spec);
+#endif
+    out = detail::write<Char>(out, Char(':'));
+    out = detail::write<Char>(out, Char(' '));
+    out = detail::write_bytes(out, string_view(ex.what()), spec);
+
+    return out;
   }
 };
 FMT_END_NAMESPACE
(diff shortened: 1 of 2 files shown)


phprus:
Rebased and fixed bug related to GCC < 5.

phprus:
cc @Baardi, @zach2good

Baardi:
I think this type of formatting looks good for exceptions, considering how important the exception type is, e.g. when logging errors.
It also looks consistent with the c# way of formatting exceptions, except for the callstack, which wouldn't be possible to include anways. Not that c++/fmtlib necessarily have to follow other languages on everything.

bilde
c# with string.Format on the left, c++ with fmtlib to the right

I guess the main issues would be that it locks support to the main compilers + and that the output isn't completely consistent, which I guess would violate the point below.

ref: 
bilde

include/fmt/core.h
```diff
@@ -486,6 +486,18 @@ template <typename Char> class basic_string_view {
...
+      basic_string_view<Char> sv) const noexcept {
+    return size_ >= sv.size_ &&
+           std::char_traits<Char>::compare(data_, sv.data_, sv.size_) == 0;
+  }
+  FMT_CONSTEXPR_CHAR_TRAITS bool starts_with(Char c) const noexcept {
+    return size_ >= 1 && std::char_traits<Char>::eq(*data_, c);
+  }
+  FMT_CONSTEXPR_CHAR_TRAITS bool starts_with(const Char* s) const {
+    return starts_with(basic_string_view<Char>(s));
+  }
```
vitaut, line 499: Could you move this to a separate PR and add unit tests?

vitaut (changes requested):
Thanks for the PR!

vitaut:
> I guess the main issues would be that it locks support to the main compilers + and that the output isn't completely consistent, which I guess would violate the point below.

Hmm, that's a good point.

phprus:
@vitaut 
Visual Studio ABI and Itanium ABI will show human-readable names.
Other ABIs will show compiler-specific (mangled) names. But other ABIs are rare.

include/fmt/core.h
phprus: PR #3080

phprus: @vitaut 
Rebased after merge #3080

vitaut:
I think we should keep the default format independent of ABI. Let's make type output an opt-in via a separate specifier (maybe `t`?)

phprus:
Like range format specifications:
Spec: ``{:t:OtherStringSpec}``, Out: ``exception_class: exception message``
or
Spec: ``{::OtherStringSpec}`` or ``{::}`` or ``{}``, Out: ``exception message``
?

phprus:
@vitaut 
Implemented.

Baardi:
> I think we should keep the default format independent of ABI. Let's make type output an opt-in via a separate specifier (maybe `t`?)

I like the idea. It makes sense to format the content of an item by default, which I guess would be the .what() on exceptions. 
And for exceptions it also makes sense to provide and easy way to format the type, considering how important it is.
While I think in general it would be better to provide the type by default, it makes sense to drop it, as there's no standardized way to format the type.

Btw @phprus, with the current PR, it still formats the whole typeinfo.name(), even if it's not one of the "big 3" compilers.
What do you think about avoiding to specify the whole formatter if it's not a "supported ABI"? Then the user would be free to specify a better formatter himself

vitaut:
Please don't use `:` in format specs because it will interfere with ranges. I think `t` should be a presentation type specifier like `s` for strings. We don't really need to combine it with other presentation specifiers.

phprus:
@Baardi 
I think the formatter should always be defined.
``typeinfo.name()`` will return the mangled name for any ABI. I think that if the rules for demangling the name are not known, then it is better to output it unchanged.

Baardi:
> @Baardi I think the formatter should always be defined. `typeinfo.name()` will return the mangled name for any ABI. I think that if the rules for demangling the name are not known, then it is better to output it unchanged.

Good point. There's argument both ways. I guess that not specifying the formatter could lead to unexpected exceptions etc,, when trying to write cross-platform code, which would be a good argument for always speciyfing it. And I guess "important" platforms could always create PR's adding proper formatting, if they're not happy with the current way it is formatted. 

Anyways, regardless of whether this gets merged or not.
Thanks @vitaut for creating an excellent formatting library, and thanks @phprus for implementing my idea, it would've taken significantly more time for me to create a PR.

phprus:
@vitaut
Implemented.

include/fmt/std.h
```diff
@@ -28,6 +31,16 @@
 #  endif
 #endif
 
+#if FMT_HAS_INCLUDE(<cxxabi.h>) || defined(__GLIBCXX__)
```
vitaut, line 34: Why do we need `__GLIBCXX__` check? Isn't `FMT_HAS_INCLUDE` enough?

include/fmt/std.h
```diff
@@ -28,6 +31,16 @@
...
+#  include <cxxabi.h>
+#  ifndef __GABIXX_CXXABI_H__
```
vitaut, line 36: What is `__GABIXX_CXXABI_H__` check for?

include/fmt/std.h
```diff
@@ -170,12 +183,56 @@ FMT_BEGIN_NAMESPACE
...
     T, Char,
-    typename std::enable_if<std::is_base_of<std::exception, T>::value>::type>
-    : formatter<string_view> {
-  template <typename FormatContext>
-  auto format(const std::exception& ex, FormatContext& ctx) const ->
-      typename FormatContext::iterator {
-    return fmt::formatter<string_view>::format(ex.what(), ctx);
+    typename std::enable_if<std::is_base_of<std::exception, T>::value>::type> {
+ private:
+  bool print_typename_{false};
```
vitaut, line 188: nit: we normally use `= false` initializer style

include/fmt/std.h
```diff
@@ -170,12 +183,56 @@ FMT_BEGIN_NAMESPACE
...
+    }
+    return it;
+  }
+
+  template <typename OutputIt>
+  auto format(const std::exception& ex,
+              basic_format_context<OutputIt, Char>& ctx) const -> OutputIt {
+    basic_format_specs<Char> spec;
+    auto out = ctx.out();
+    if (print_typename_ == false)
```
vitaut, line 208:
```
if (!print_typename_)
```

vitaut (commented):
Mostly looks good. A few comments/questions inline.

include/fmt/std.h
phprus: GCC 4 does not support FMT_HAS_INCLUDE, but it does support demangling.
It is possible to check for the presence of libstdc++ instead of gcc version, since all libstdc++ support demangling.

include/fmt/std.h
phprus: For Android.
See comment from Boost:

> // For some archtectures (mips, mips64, x86, x86_64) cxxabi.h in Android NDK is implemented by gabi++ library
> // (https://android.googlesource.com/platform/ndk/+/master/sources/cxx-stl/gabi++/), which does not implement
> // abi::__cxa_demangle(). We detect this implementation by checking the include guard here.

https://github.com/boostorg/core/blob/162a4e1d24d1cab4f24cbca9489203d891f352e6/include/boost/core/demangle.hpp#L33-L35

include/fmt/std.h
vitaut: Could you add a comment that the `__GLIBCXX__` is for gcc 4 here then?

include/fmt/std.h
vitaut: Could you add a comment here too?

phprus: Done.
And I renamed ``FMT_HAS_CXXABI_H`` to ``FMT_HAS_ABI_CXA_DEMANGLE``.

include/fmt/std.h
phprus: Done

vitaut (approved):

vitaut:
Thank you!

PR #3083, opened 2022-09-06
Workaround to make Intel LLVM 16 happy

andreagen0r:
The compiler interprets sign as a bit-field and throws an error in the lambda capture by reference.
Just remove the variable and use the parameter to solve the problem

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -2415,8 +2415,7 @@ FMT_CONSTEXPR20 auto do_write_float(OutputIt out, const DecimalFP& f,
   auto significand = f.significand;
   int significand_size = get_significand_size(f);
   const Char zero = static_cast<Char>('0');
-  auto sign = fspecs.sign;
-  size_t size = to_unsigned(significand_size) + (sign ? 1 : 0);
+  size_t size = to_unsigned(significand_size) + (fspecs.sign ? 1 : 0);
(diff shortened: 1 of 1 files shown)


include/fmt/format.h
```diff
@@ -2435,8 +2435,7 @@ FMT_CONSTEXPR20 auto do_write_float(OutputIt out, const DecimalFP& f,
   auto significand = f.significand;
   int significand_size = get_significand_size(f);
   const Char zero = static_cast<Char>('0');
-  auto sign = fspecs.sign;
```
vitaut, line 2438: Please keep `sign` definition here, we don't want to capture all specs. Perhaps changing `auto` to `sign_t` will workaround the intel compiler bug?

vitaut (changes requested):

include/fmt/format.h
andreagen0r: sign_t doesn’t work too, the alternative is to replace the lambda capture by value, but maybe we can cancel this PR and just wait, KornevNikita said they fixed it already, just need to upstream to the SYCL branch

vitaut:
> maybe we can cancel this PR and just wait, KornevNikita said they fixed it already, just need to upstream to the SYCL branch

Let's do that. Closing for now.

Issue #3084, opened 2022-09-06
formatted_size misses locale overload


gerboengels:
`formatted_size` doesn't have an overload which takes a `locale` (overloads 3 and 4 on [cpp reference](https://en.cppreference.com/w/cpp/utility/format/formatted_size)).

Godbolt link (using MSVC to compare with its std implementation, which does support this overload): https://godbolt.org/z/94dsPhTMb

vitaut:
It's just a matter of adding an overload similar to https://github.com/fmtlib/fmt/blob/1feb430faaac6bd8094e996861d6025f9903d34e/include/fmt/core.h#L3284-L3291

but taking and forwarding the locale. This overload should be added here: https://github.com/fmtlib/fmt/blob/1feb430faaac6bd8094e996861d6025f9903d34e/include/fmt/format.h#L4272

A PR is welcome.

PR #3087, opened 2022-09-07
Add locale overload for formatted_size (#3084)

gerboengels:
Fix for issue #3084. Adds an overload to `formatted_size` which takes a `locale`, as [`std::formatted_size`](https://en.cppreference.com/w/cpp/utility/format/formatted_size) does.

The unit test is more of a "does it compile" test than actually testing locale-dependent behaviour. Is that ok? My thought was that `vformat_to` is already tested, so no need to repeat that here...


vitaut:
Thank you!

PR #3091, opened 2022-09-10
Fix compilation with FMT_ENFORCE_COMPILE_STRING and FMT_WERROR

asmaloney:
With `Apple clang version 13.1.6 (clang-1316.0.21.2.5)`, turning on `FMT_ENFORCE_COMPILE_STRING` and `FMT_WERROR` fails to compile fmt itself.

This PR adds `FMT_STRING` around several strings in os.cc to fix it.

It also clang-formatted another line (172) which I left in this PR.


vitaut:
Thank you

Issue #3092, opened 2022-09-10
GCC-9 build error


phprus:
Commit with added GCC-9 CI: https://github.com/phprus/fmt/commit/4e386069c9733ebe8e42c5c01745fa1e695df303
Full build log: https://github.com/phprus/fmt/runs/8287290178?check_suite_focus=true

Error:

```
In file included from /home/runner/work/fmt/fmt/test/xchar-test.cc:8:
/home/runner/work/fmt/fmt/include/fmt/xchar.h: In instantiation of ‘bool fmt::v9::detail::write_loc(OutputIt, fmt::v9::basic_format_arg<fmt::v9::basic_format_context<std::back_insert_iterator<fmt::v9::detail::buffer<wchar_t> >, wchar_t> >, const fmt::v9::basic_format_specs<wchar_t>&, fmt::v9::detail::locale_ref) [with OutputIt = wchar_t*]’:
/home/runner/work/fmt/fmt/include/fmt/format.h:2200:16:   required from ‘constexpr OutputIt fmt::v9::detail::write(OutputIt, T, const fmt::v9::basic_format_specs<Char>&, fmt::v9::detail::locale_ref) [with Char = wchar_t; OutputIt = wchar_t*; T = int; typename std::enable_if<((fmt::v9::detail::is_integral<T>::value && (! std::is_same<T, bool>::value)) && (! std::is_same<OutputIt, typename std::conditional<std::is_same<Char, char>::value, fmt::v9::appender, std::back_insert_iterator<fmt::v9::detail::buffer<T> > >::type>::value)), int>::type <anonymous> = 0]’
/home/runner/work/fmt/fmt/include/fmt/format.h:3789:31:   required from ‘constexpr decltype (ctx.out()) fmt::v9::formatter<T, Char, typename std::enable_if<(fmt::v9::detail::type_constant<T, Char>::value != fmt::v9::detail::type::custom_type), void>::type>::format(const T&, FormatContext&) const [with FormatContext = fmt::v9::basic_format_context<wchar_t*, wchar_t>; T = int; Char = wchar_t; typename std::enable_if<(fmt::v9::detail::type_constant<T, Char>::value != fmt::v9::detail::type::custom_type), void>::type = void; decltype (ctx.out()) = wchar_t*]’
[...]

PR #3093, opened 2022-09-10
Fix build error on GCC-9

phprus:
Issue: #3092


phprus:
@vitaut
I think it is necessary to add a GCC-9 compiler to CI for testing fmt on all current versions of GCC.

vitaut:
LGTM but please rebase since there were some changes to `write_loc`.

phprus:
@vitaut 
Rebased

vitaut:
Thank you!

PR #3095, opened 2022-09-11
Fix warnings

HazardyKnusperkeks:
Fix some warnings.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -2585,14 +2585,14 @@ template <typename T, FMT_ENABLE_IF(std::is_floating_point<T>::value&&
 FMT_CONSTEXPR20 bool isfinite(T value) {
   constexpr T inf = T(std::numeric_limits<double>::infinity());
   if (is_constant_evaluated())
-    return !detail::isnan(value) && value != inf && value != -inf;
+    return !detail::isnan(value) && value < inf && value > -inf;
   return std::isfinite(value);
 }
 template <typename T, FMT_ENABLE_IF(!has_isfinite<T>::value)>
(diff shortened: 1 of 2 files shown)


vitaut (changes requested):
Thank for the PR. In general looks good except for the `-Wsuggest-attribute=format` part which is a bit messy. Considering that this code is going away I suggest not trying to suppress it.

phprus:
@vitaut 
Tell me please, sprintf is only used to print floating point numbers in hex format? Or not?

vitaut:
> Tell me please, sprintf is only used to print floating point numbers in hex format?

Yes, `s(n)printf` is only used for hexfloat.

HazardyKnusperkeks:
> Thank for the PR. In general looks good except for the `-Wsuggest-attribute=format` part which is a bit messy. Considering that this code is going away I suggest not trying to suppress it.

Is there an ETA on when it's going away?

vitaut (approved):

vitaut:
Merged, thanks!

> Is there an ETA on when it's going away?

it will happen in the next (major) release some time next year. The actual code might be removed earlier in preparation for the release.

HazardyKnusperkeks:
> it will happen in the next (major) release some time next year. The actual code might be removed earlier in preparation for the release.

We are hitting this warning in our tests (-Werror). When the code is going away, couldn't we just add the suppression, as it also will go away? I'd rather not change the our submodule to a fork.

vitaut:
You can suppress the warnings with `FMT_SYSTEM` or other method. I don't think it makes sense to add a suppression that won't be included in any release.

PR #3097, opened 2022-09-11
fix gcc <= 7.1 compile errors

### LINKED ISSUE #3096: Compile errors with GCC 7.1 and below
`scan-test.cc` fails to compile with GCC 7.1 and older issuing the following errors:
```console
<source>:172:21: error: enclosing class of constexpr non-static member function 'int fmt::v9::detail::scan_handler::on_arg_id()' is not a literal type
   FMT_CONSTEXPR int on_arg_id() { return on_arg_id(next_arg_id_++); }
                     ^~~~~~~~~
<source>:125:8: note: 'fmt::v9::detail::scan_handler' is not literal because:
 struct scan_handler : error_handler {
        ^~~~~~~~~~~~
<source>:125:8: note:   'fmt::v9::detail::scan_handler' is not an aggregate, does not have a trivial default constructor, and has no constexpr constructor that is not a copy or move constructor
<source>:173:21: error: enclosing class of constexpr non-static member function 'int fmt::v9::detail::scan_handler::on_arg_id(int)' is not a literal type
   FMT_CONSTEXPR int on_arg_id(int id) {
                     ^~~~~~~~~
<source>:178:21: error: enclosing class of constexpr non-static member function 'int fmt::v9::detail::scan_handler::on_arg_id(fmt::v9::string_view)' is not a literal type
   FMT_CONSTEXPR int on_arg_id(string_view id) {
                     ^~~~~~~~~
Compiler returned: 1
````
Reproducer: https://godbolt.org/z/jcxnx7MeT. A workaround is to remove `FMT_CONSTEXPR` from failing methods.

sergiud:
Fixes #3096

### DIFF
--- a/test/scan.h
+++ b/test/scan.h
@@ -42,7 +42,7 @@ struct scan_context {
  public:
   using iterator = const char*;
 
-  explicit scan_context(string_view input) : input_(input) {}
+  explicit FMT_CONSTEXPR scan_context(string_view input) : input_(input) {}
 
   iterator begin() const { return input_.data(); }
   iterator end() const { return begin() + input_.size(); }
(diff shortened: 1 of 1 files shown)


mwinterb:
Adding `FMT_CONSTEXPR` to the constructors for `scan_handler`, `scan_context`, `scan_args`, and `scan_arg` (and initializing `int_value` to `nullptr` for the default) also seemed to fix compilation going back to the first version that FMT_USE_CONSTEXPR is defined for, GCC 6.1. That seems cleaner. https://godbolt.org/z/oncnxz9nc

sergiud:
Good point, thanks. I can confirm that this works on my end too. I've updated the PR.

vitaut:
Thank you!

Issue #3098, opened 2022-09-12
std::chrono::utc_timestamp is not formattable via fmt/chrono.h


patrickroocks:
With C++20 we can format all the `std::chrono` types using `std::format`, e.g., the following program 

```
#include <iostream>
#include <chrono>
#include <format>
int main()
{
	std::chrono::utc_time<std::chrono::nanoseconds> ts = std::chrono::utc_clock::now();
	std::cout << std::format("{}", ts);
}
```

compiles and gives the output "2022-09-12 12:57:14.844838800".

So I would expect that this works also for `fmt::format`, at least if I include fmt/chrono.h or maybe also fmt/std.h. But the following program

```
#include <iostream>
#include <chrono>
#include <fmt/format.h>
#include <fmt/chrono.h>
#include <fmt/std.h>
int main()
{
	std::chrono::utc_time<std::chrono::nanoseconds> ts = std::chrono::utc_clock::now();
	std::cout << fmt::format("{}", ts);
}
```

doesn't compile with the current fmt master ([94ceb38](https://github.com/fmtlib/fmt/commit/94ceb38a0943c50ea9939050a7f4e8e4c9da43d2)). We get `error C2338: static_assert failed: 'Cannot format an argument. To make type T formattable provide a formatter<T> specialization: https://fmt.dev/latest/api.html#udt'`.

I'm using the following work-around now to get the above program compiling:

```
template<>
struct fmt::formatter<std::chrono::utc_time<std::chrono::nanoseconds>> : public formatter<std::string> {
	template<typename FormatContext>
	auto format(const std::chrono::utc_time<std::chrono::nanoseconds>& ts, FormatContext& ctx) const
	{
		return formatter<std::string>::format(std::format("{}", ts), ctx);
	}
};
```

I.e., I provide a custom formatter which does nothing else than forwarding the UTC timestamp to `std::format`. This works fine, but somehow feels unsatisfying. We doesn't provide fmt a formatter for a std::chrono UTC timestamp, even it is supported by `std::format`? Is this really intended or was this forgotten?

vitaut:
A PR to add a formatter for `utc_clock` is welcome.

patrickroocks:
[...]

PR #3110, opened 2022-09-22
Support formatting of time_point with utc_clock

patrickroocks:
According to the discussion in https://github.com/fmtlib/fmt/issues/3098 in this PR a formatter for `std::chrono::time_point<std::chrono::utc_clock>` is added. Also added a small unit test. fmtlib must be compiled with C++20 or later to use this, as `std::chrono::utc_clock` is a C++20 feature.

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -22,6 +22,15 @@
 
 FMT_BEGIN_NAMESPACE
 
+// Check if std::chrono::utc_timestamp is available.
+#ifndef FMT_USE_UTC_TIME
+#  ifdef __cpp_lib_chrono
+#    define FMT_USE_UTC_TIME (__cpp_lib_chrono >= 201907L)
+#  else
+#    define FMT_USE_UTC_TIME 0
+#  endif
+#endif
+
 // Enable tzset.
 #ifndef FMT_USE_TZSET
 // UWP doesn't provide _tzset.
@@ -2013,6 +2022,25 @@ struct formatter<std::chrono::time_point<std::chrono::system_clock, Duration>,
     return formatter<std::tm, Char>::format(localtime(val), ctx);
   }
 };
+
+#if FMT_USE_UTC_TIME
+template <typename Char, typename Duration>
+struct formatter<std::chrono::time_point<std::chrono::utc_clock, Duration>,
+                 Char> : formatter<std::tm, Char> {
+  FMT_CONSTEXPR formatter() {
+    basic_string_view<Char> default_specs =
+        detail::string_literal<Char, '%', 'F', ' ', '%', 'T'>{};
+    this->do_parse(default_specs.begin(), default_specs.end());
+  }
+
+  template <typename FormatContext>
+  auto format(std::chrono::time_point<std::chrono::utc_clock> val,
+              FormatContext& ctx) const -> decltype(ctx.out()) {
+    return formatter<std::tm, Char>::format(
+        localtime(std::chrono::utc_clock::to_sys(val)), ctx);
+  }
+};
+#endif
 
 template <typename Char> struct formatter<std::tm, Char> {
  private:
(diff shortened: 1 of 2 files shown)


include/fmt/chrono.h
```diff
@@ -22,6 +22,13 @@
 
 FMT_BEGIN_NAMESPACE
 
+// check if std::chrono::utc_timestamp is available
```
vitaut, line 25: nit: we normally use complete sentences in comments

```
// Check if std::chrono::utc_timestamp is available.
```

include/fmt/chrono.h
```diff
@@ -22,6 +22,13 @@
...
+#ifndef FMT_USE_UTC_TIME
+#  ifdef __cpp_lib_chrono
```
vitaut, line 27: I think this can be simplified:

```
#if !defined(FMT_USE_UTC_TIME) && defined(__cpp_lib_chrono)
```

test/chrono-test.cc
```diff
@@ -641,3 +641,12 @@ TEST(chrono_test, cpp20_duration_subsecond_support) {
 }
 
 #endif  // FMT_STATIC_THOUSANDS_SEPARATOR
+
+#if (FMT_USE_UTC_TIME)
```
vitaut, line 645: Let's remove the redundant parentheses around FMT_USE_UTC_TIME for consistency with the rest of the code.

vitaut (changes requested):
Thanks for the PR! Overall looks good, just a few nits inline.

include/fmt/chrono.h
patrickroocks: But I forgot `define FMT_USE_UTC_TIME 0` in the else-branch of `ifdef __cpp_lib_chrono`

patrickroocks:
Thanks for the review! Pushed the fixes.

vitaut (approved):

vitaut:
Looks good but there are some CI failures.

patrickroocks:
The chrono test fails with

`unknown file: error: C++ exception with description "The specified module could not be found." thrown in the test body.`
(https://github.com/fmtlib/fmt/actions/runs/3125553362/jobs/5078549521#step:6:134)

As this happens only for MSC version 1929 but not for version 1933 (the newest one), I check for this version now.

mwinterb:
It's probably not related to the MSVC version. Windows doesn't "[historically](https://archives.miloush.net/michkap/archive/2015/05/04/8770668856267196406.html)" use the IANA time zone names, and so to support them for `tzdb`, Microsoft's STL uses [ICU](https://github.com/microsoft/STL/blob/35078f3921e43aaee440033c56520338a97c5a7e/stl/src/tzdb.cpp#L66) to handle that translation. icu.dll is only included in [newer versions of Windows](https://devblogs.microsoft.com/oldnewthing/20210527-00/?p=105255). Windows Server 2019 is based on Windows 1809, so it is too old to have just "icu.dll".

`to_sys` for `utc_time` uses tzdb to handle [leap seconds](https://github.com/microsoft/STL/blob/35078f3921e43aaee440033c56520338a97c5a7e/stl/inc/chrono#L2661-L2663).

phprus:
@patrickroocks, @vitaut,
Maybe disable this test for Windows?
Like here:

https://github.com/fmtlib/fmt/blob/48f525d025cadbedce0b2288ff8e19b6877341e4/test/chrono-test.cc#L153-L156

vitaut (changes requested):
> Maybe disable this test for Windows?

Fine with me.

Also if @mwinterb is correct then the current MSVC check will only mask the issue and should be removed.

patrickroocks:
Removed the MSVC check, disabled the `utc_clock` test for windows. Squashed the commits.

vitaut (approved):

vitaut:
Thank you!

PR #3111, opened 2022-09-22
Add basic_format_string::get()

huangqinjin:


huangqinjin:
To conform with standard.

https://en.cppreference.com/w/cpp/utility/format/basic_format_string

vitaut:
Thanks!

phprus:
@vitaut 
``std::basic_format_string`` does not contain ``operator basic_string_view<Char>()``.
Maybe remove it and replace it with ``get()`` calls?

vitaut:
{fmt} provides a lot of functionality not available in `std::format`, this is just one of them. I don't think we need to remove anything.

PR #3114, opened 2022-09-25
Rename `uint` alias to prevent name collision with Qt

Bruflot:
Qt defines some [global typedefs](https://doc.qt.io/qt-6/qtglobal.html#uint-typedef) - one of which being `uint` - which causes a name collision when using Qt alongside fmtlib. Although this isn't fmtlib's fault, I think it's a common enough problem that it warrants a fix.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -1405,8 +1405,8 @@ template <typename Float> constexpr int num_significand_bits() {
 template <typename Float>
 constexpr auto exponent_mask() ->
     typename dragonbox::float_info<Float>::carrier_uint {
-  using uint = typename dragonbox::float_info<Float>::carrier_uint;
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR but typedefs shouldn't cause collisions with locally defined symbols.

Bruflot:
MSVC has a warning for precisely that:

https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4459?view=msvc-170

It's enabled by default when you compile with `/W4`.

vitaut:
You can suppress a warning with `FMT_SYSTEM` or other means.

Pokechu22:
Note: an equivalent PR was merged: #3137

PR #3115, opened 2022-09-26
Support formatting of subseconds

patrickroocks:
The formatting symbol `{%S}` should also print the subseconds, e.g., `fmt::format("{:%S}", std::chrono::microseconds(1234567))` should print "01.234567".

Didn't implement it for localized output as this localized output looks like black magic to me... may be we can omit it for the localized formatting for the first time?

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -982,7 +982,7 @@ inline Int to_nonnegative_int(T value, Int upper) {
   return static_cast<Int>(value);
 }
 
-template <class Rep, class Period,
+template <typename Rep, typename Period,
           FMT_ENABLE_IF(std::numeric_limits<Rep>::is_signed)>
 constexpr std::chrono::duration<Rep, Period> abs(
     std::chrono::duration<Rep, Period> d) {
@@ -995,7 +995,7 @@ constexpr std::chrono::duration<Rep, Period> abs(
   return d.count() >= d.zero().count() ? d : -d;
 }
 
-template <class Rep, class Period,
+template <typename Rep, typename Period,
           FMT_ENABLE_IF(!std::numeric_limits<Rep>::is_signed)>
 constexpr std::chrono::duration<Rep, Period> abs(
     std::chrono::duration<Rep, Period> d) {
@@ -1025,7 +1025,7 @@ struct count_fractional_digits<Num, Den, N, false> {
 
 // Format subseconds which are given as an integer type with an appropriate
 // number of digits.
-template <typename char_type, typename OutputIt, typename Duration>
+template <typename Char, typename OutputIt, typename Duration>
 void write_fractional_seconds(OutputIt& out, Duration d) {
   FMT_ASSERT(!std::is_floating_point<typename Duration::rep>::value, "");
   constexpr auto num_fractional_digits =
@@ -1052,26 +1052,33 @@ void write_fractional_seconds(OutputIt& out, Duration d) {
     int num_digits = detail::count_digits(n);
     if (num_fractional_digits > num_digits)
       out = std::fill_n(out, num_fractional_digits - num_digits, '0');
-    out = format_decimal<char_type>(out, n, num_digits).end;
+    out = format_decimal<Char>(out, n, num_digits).end;
   }
 }
 
 // Format subseconds which are given as a floating point type with an appropiate
 // number of digits. We cannot pass the Duration here, as we explicitly need to
 // pass the Rep value in the chrono_formatter.
-template <typename Rep, long long Num, long long Den>
-void format_floating_seconds(memory_buffer& buf, Rep val) {
-  FMT_ASSERT(std::is_floating_point<Rep>::value, "");
-  auto num_fractional_digits = count_fractional_digits<Num, Den>::value;
+template <typename Duration>
+void write_floating_seconds(memory_buffer& buf, Duration duration) {
+  FMT_ASSERT(std::is_floating_point<typename Duration::rep>::value, "");
+  auto num_fractional_digits =
+      count_fractional_digits<Duration::period::num,
+                              Duration::period::den>::value;
   // For non-integer values, we ensure at least 6 digits to get microsecond
   // precision.
-  if (num_fractional_digits < 6 && static_cast<Rep>(std::round(val)) != val)
+  auto val = duration.count();
+  if (num_fractional_digits < 6 &&
+      static_cast<typename Duration::rep>(std::round(val)) != val)
     num_fractional_digits = 6;
 
-  format_to(std::back_inserter(buf), runtime("{:.{}f}"),
-            std::fmod(val * static_cast<Rep>(Num) / static_cast<Rep>(Den),
-                      static_cast<Rep>(60)),
-            num_fractional_digits);
+  format_to(
+      std::back_inserter(buf), runtime("{:.{}f}"),
+      std::fmod(val *
+                    static_cast<typename Duration::rep>(Duration::period::num) /
+                    static_cast<typename Duration::rep>(Duration::period::den),
+                static_cast<typename Duration::rep>(60)),
+      num_fractional_digits);
 }
 
 template <typename OutputIt, typename Char,
@@ -1457,9 +1464,7 @@ class tm_writer {
       if (subsecs_) {
         if (std::is_floating_point<typename Duration::rep>::value) {
           auto buf = memory_buffer();
-          format_floating_seconds<typename Duration::rep, Duration::period::num,
-                                  Duration::period::den>(buf,
-                                                         subsecs_->count());
+          write_floating_seconds(buf, *subsecs_);
           if (buf.size() > 1) {
             // Remove the leading "0", write something like ".123".
             out_ = std::copy(buf.begin() + 1, buf.end(), out_);
@@ -1826,7 +1831,7 @@ struct chrono_formatter {
     if (ns == numeric_system::standard) {
       if (std::is_floating_point<rep>::value) {
         auto buf = memory_buffer();
-        format_floating_seconds<rep, Period::num, Period::den>(buf, val);
+        write_floating_seconds(buf, std::chrono::duration<rep, Period>(val));
         if (negative) *out++ = '-';
         if (buf.size() < 2 || buf[1] == '.') *out++ = '0';
         out = std::copy(buf.begin(), buf.end(), out);
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -1023,9 +1023,10 @@ struct count_fractional_digits<Num, Den, N, false> {
   static constexpr int value = (Num % Den == 0) ? N : 6;
 };
 
-// Format non-fractional subseconds with an appriate number of digits.
-template <typename OutputIt, typename Duration, typename char_type>
-void write_sub_sec(OutputIt& out, Duration d) {
+// Format subseconds which are given as an integer type with an appropriate
+// number of digits.
+template <typename char_type, typename OutputIt, typename Duration>
+void write_fractional_seconds(OutputIt& out, Duration d) {
   FMT_ASSERT(!std::is_floating_point<typename Duration::rep>::value, "");
   constexpr auto num_fractional_digits =
       count_fractional_digits<Duration::period::num,
@@ -1055,24 +1056,26 @@ void write_sub_sec(OutputIt& out, Duration d) {
   }
 }
 
-// Format fractional seconds with an appropiate number of digits.
+// Format subseconds which are given as a floating point type with an appropiate
+// number of digits. We cannot pass the Duration here, as we explicitly need to
+// pass the Rep value in the chrono_formatter.
 template <typename Rep, long long Num, long long Den>
-void format_sec_fractional(basic_memory_buffer<char>& buf, Rep val) {
+void format_floating_seconds(memory_buffer& buf, Rep val) {
   FMT_ASSERT(std::is_floating_point<Rep>::value, "");
   auto num_fractional_digits = count_fractional_digits<Num, Den>::value;
   // For non-integer values, we ensure at least 6 digits to get microsecond
   // precision.
-  if (num_fractional_digits < 6 && static_cast<Rep>(std::round(val)) != val) {
+  if (num_fractional_digits < 6 && static_cast<Rep>(std::round(val)) != val)
     num_fractional_digits = 6;
-  }
+
   format_to(std::back_inserter(buf), runtime("{:.{}f}"),
             std::fmod(val * static_cast<Rep>(Num) / static_cast<Rep>(Den),
                       static_cast<Rep>(60)),
             num_fractional_digits);
 }
 
-template <typename OutputIt, typename Char, typename Rep = int,
-          typename Period = std::ratio<1, 1>>
+template <typename OutputIt, typename Char,
+          typename Duration = std::chrono::seconds>
 class tm_writer {
  private:
   static constexpr int days_per_week = 7;
@@ -1080,8 +1083,7 @@ class tm_writer {
   const std::locale& loc_;
   const bool is_classic_;
   OutputIt out_;
-  std::chrono::duration<Rep, Period> subsecs_;
-  bool has_subsecs_;
+  const Duration* subsecs_;
   const std::tm& tm_;
 
   auto tm_sec() const noexcept -> int {
@@ -1245,19 +1247,11 @@ class tm_writer {
 
  public:
   tm_writer(const std::locale& loc, OutputIt out, const std::tm& tm,
-            std::chrono::duration<Rep, Period> subsecs)
+            const Duration* subsecs = nullptr)
       : loc_(loc),
         is_classic_(loc_ == get_classic_locale()),
         out_(out),
         subsecs_(subsecs),
-        has_subsecs_(true),
-        tm_(tm) {}
-
-  tm_writer(const std::locale& loc, OutputIt out, const std::tm& tm)
-      : loc_(loc),
-        is_classic_(loc_ == get_classic_locale()),
-        out_(out),
-        has_subsecs_(false),
         tm_(tm) {}
 
   OutputIt out() const { return out_; }
@@ -1460,18 +1454,18 @@ class tm_writer {
   void on_second(numeric_system ns) {
     if (is_classic_ || ns == numeric_system::standard) {
       write2(tm_sec());
-      if (has_subsecs_) {
-        if (std::is_floating_point<Rep>::value) {
+      if (subsecs_) {
+        if (std::is_floating_point<typename Duration::rep>::value) {
           auto buf = memory_buffer();
-          format_sec_fractional<Rep, Period::num, Period::den>(
-              buf, subsecs_.count());
+          format_floating_seconds<typename Duration::rep, Duration::period::num,
+                                  Duration::period::den>(buf,
+                                                         subsecs_->count());
           if (buf.size() > 1) {
             // Remove the leading "0", write something like ".123".
             out_ = std::copy(buf.begin() + 1, buf.end(), out_);
           }
         } else {
-          write_sub_sec<OutputIt, std::chrono::duration<Rep, Period>, Char>(
-              out_, subsecs_);
+          write_fractional_seconds<Char>(out_, *subsecs_);
         }
       }
     } else {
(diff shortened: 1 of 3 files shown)


patrickroocks:
Rebased on master.

include/fmt/chrono.h
```diff
@@ -966,13 +966,33 @@ inline void tzset_once() {
...
+  tm_ss tm_ss_;
+  empty_tm_ss() { tm_ss_.subsec_wdt = 0; }
+
+ public:
+  static empty_tm_ss& instance() {
+    static empty_tm_ss inst;
+    return inst;
+  }
+  const tm_ss& get_tm_ss() { return tm_ss_; }
+};
```
vitaut, line 986: I don't think we need these. Just pass subseconds as an additional argument to `tm_writer`.

include/fmt/chrono.h
```diff
@@ -966,13 +966,33 @@ inline void tzset_once() {
...
+
+template <typename OutputIt, typename Char> class tm_ss_writer {
```
vitaut, line 988: Let's keep the old name since formatting of subseconds is just a minor implementation detail.

include/fmt/chrono.h
```diff
@@ -966,13 +966,33 @@ inline void tzset_once() {
...
  private:
   static constexpr int days_per_week = 7;
 
   const std::locale& loc_;
   const bool is_classic_;
   OutputIt out_;
+  const tm_ss& tm_ss_;
```
vitaut, line 995: Let's replace this with a variable holding subseconds to avoid duplication.

include/fmt/chrono.h
```diff
@@ -1135,10 +1155,18 @@ template <typename OutputIt, typename Char> class tm_writer {
...
         out_(out),
+        tm_ss_(tm_ss_val),
+        tm_(tm_ss_val.tm) {}
+
+  tm_ss_writer(const std::locale& loc, OutputIt out, const std::tm& tm)
+      : loc_(loc),
+        is_classic_(loc_ == get_classic_locale()),
+        out_(out),
+        tm_ss_(empty_tm_ss::instance().get_tm_ss()),
         tm_(tm) {}
```
vitaut, line 1170: Similarly, there should be one ctor that takes an additional subseconds argument defaulted to 0.

include/fmt/chrono.h
```diff
@@ -2016,28 +2057,52 @@ struct formatter<std::chrono::time_point<std::chrono::system_clock, Duration>,
...
+         Period::den == 1e13 || Period::den == 1e14 || Period::den == 1e15);
+
+    if (reasonable_subsec_available) {
+      const auto epoch = val.time_since_epoch();
+      const auto subsecs =
+          std::chrono::duration_cast<std::chrono::duration<Rep, Period>>(
+              epoch - std::chrono::duration_cast<std::chrono::seconds>(epoch));
+
+      tm_ss_val.subsec = static_cast<uint64_t>(subsecs.count());
+      tm_ss_val.subsec_wdt =
```
vitaut, line 2084: What does wdt stand for?

include/fmt/chrono.h
```diff
@@ -2016,28 +2057,52 @@ struct formatter<std::chrono::time_point<std::chrono::system_clock, Duration>,
...
+    tm_ss_val.tm =
+        localtime(std::chrono::time_point_cast<std::chrono::seconds>(val));
+
+    constexpr bool reasonable_subsec_available =
+        Period::num == 1 &&
+        (Period::den == 1e1 || Period::den == 1e2 || Period::den == 1e3 ||
+         Period::den == 1e4 || Period::den == 1e5 || Period::den == 1e6 ||
+         Period::den == 1e7 || Period::den == 1e8 || Period::den == 1e9 ||
+         Period::den == 1e10 || Period::den == 1e11 || Period::den == 1e12 ||
+         Period::den == 1e13 || Period::den == 1e14 || Period::den == 1e15);
```
vitaut, line 2075: Can we support arbitrary fractions and not just 1/power-of-10?

include/fmt/chrono.h
```diff
@@ -2075,12 +2140,15 @@ template <typename Char> struct formatter<std::tm, Char> {
     return end;
   }
 
-  template <typename FormatContext>
-  auto format(const std::tm& tm, FormatContext& ctx) const
+  template <typename TimeType, typename FormatContext,
+            FMT_ENABLE_IF(std::is_same<TimeType, std::tm>::value ||
+                          std::is_same<TimeType, detail::tm_ss>::value)>
+  auto format(const TimeType& tm, FormatContext& ctx) const
```
vitaut, line 2146: Let's keep the first argument as `std::tm` and pass subseconds as an additional argument.

test/chrono-test.cc
```diff
@@ -652,3 +655,36 @@ TEST(chrono_test, utc_clock) {
...
+  EXPECT_EQ(fmt::format("{:%S}", t3), "00.123456789");
+
+  const auto t4 = std::chrono::time_point_cast<std::chrono::nanoseconds>(
+      std::chrono::system_clock::now());
+  const auto t4_sec = std::chrono::time_point_cast<std::chrono::seconds>(t4);
+
+  auto t4_sub_sec_part = fmt::format("{0:09}", (t4 - t4_sec).count());
+  EXPECT_EQ(fmt::format("{}.{}", strftime_full(t4_sec), t4_sub_sec_part),
+            fmt::format("{:%Y-%m-%d %H:%M:%S}", t4));
+}
```
vitaut, line 690: nit: missing newline

vitaut (changes requested):
Thanks for the PR.

include/fmt/chrono.h
patrickroocks: Removed and made subseconds an argumenbt.

include/fmt/chrono.h
patrickroocks: Kept old name.

include/fmt/chrono.h
patrickroocks: Done.

include/fmt/chrono.h
patrickroocks: I don't think it will work with an additional argument defaulted to 0. There is a difference between a timestamp based on `std::chrono::seconds(1)` (printed as "....01") and `std::chrono::milliseconds(1000)` (printed as "01.000" - the MS implementation of C++20 std::format does the same). Cf. the last unit test with `EXPECT_EQ(fmt::format("{:%S}", t10), "02.000");`

So I made two constructors and one sets `hasSubsecs` to true.

include/fmt/chrono.h
patrickroocks: Implicitly done.

patrickroocks:
I reworked everything again.

- The work regarding subseconds was already done in the `chrono_formatter` for formatting durations. I pulled the code out of that and put in helper functions `write_sub_sec` and `format_sec_fractional` used in the `tm_writer` and the `chrono_formatter`.
- This implicitly supports arbitrary ratios for the `Period`.
- I found a bug for printing durations with floating point types. The unit test 
`EXPECT_EQ(fmt::format("{:%S}", std::chrono::duration<double>{1.5}), "01.500000");`
didn't succeed, instead "01" was given as output. I changed the number of digits to at least 6, if the `Rep` type is fractional _and_ round(value) != value.

include/fmt/chrono.h
patrickroocks: width - removed anyway

include/fmt/chrono.h
patrickroocks: done

include/fmt/chrono.h
```diff
@@ -966,13 +966,123 @@ inline void tzset_once() {
...
+    int num_digits = detail::count_digits(n);
+    if (num_fractional_digits > num_digits)
+      out = std::fill_n(out, num_fractional_digits - num_digits, '0');
+    out = format_decimal<char_type>(out, n, num_digits).end;
+  }
+}
+
+// Format fractional seconds with an appropiate number of digits.
+template <typename Rep, long long Num, long long Den>
+basic_memory_buffer<char> format_sec_fractional(Rep val) {
```
phprus, line 1060: Moving ``basic_memory_buffer`` may use copy (for buffers shorter than 500 chars).
See implementation:

https://github.com/fmtlib/fmt/blob/662adf4f33346ba9aba8b072194e319869ede54a/include/fmt/format.h#L864-L893

I think it's better to pass basic_memory_buffer by reference, like this:

https://github.com/fmtlib/fmt/blob/662adf4f33346ba9aba8b072194e319869ede54a/include/fmt/std.h#L50-L53

include/fmt/chrono.h
```diff
@@ -966,13 +966,123 @@ inline void tzset_once() {
...
+          typename Period = std::ratio<1, 1>>
+class tm_writer {
  private:
   static constexpr int days_per_week = 7;
 
   const std::locale& loc_;
   const bool is_classic_;
   OutputIt out_;
+  std::chrono::duration<Rep, Period> subsecs_;
+  bool hasSubsecs_;
```
phprus, line 1085: Suggestion:
``has_subsecs_``

include/fmt/chrono.h
patrickroocks: Ok thanks, didn't know that. Fixed.

include/fmt/chrono.h
patrickroocks: Fixed.

include/fmt/chrono.h
```diff
@@ -966,13 +966,122 @@ inline void tzset_once() {
...
+};
+
+// Base case that doesn't instantiate any more templates
+// in order to avoid overflow.
+template <long long Num, long long Den, int N>
+struct count_fractional_digits<Num, Den, N, false> {
+  static constexpr int value = (Num % Den == 0) ? N : 6;
+};
+
+// Format non-fractional subseconds with an appriate number of digits.
```
vitaut, line 1026: appriate -> appropriate

Also please keep the old name, `write_fractional_seconds`, and remove or update misleading comment.

include/fmt/chrono.h
```diff
@@ -966,13 +966,122 @@ inline void tzset_once() {
...
+template <typename OutputIt, typename Duration, typename char_type>
```
vitaut, line 1027: char_type -> Char for consistency and make it the first argument so that the other ones are deduced.

include/fmt/chrono.h
```diff
@@ -966,13 +966,122 @@ inline void tzset_once() {
...
+    uint32_or_64_or_128_t<long long> n =
+        to_unsigned(to_nonnegative_int(subseconds, max_value<long long>()));
+    int num_digits = detail::count_digits(n);
+    if (num_fractional_digits > num_digits)
+      out = std::fill_n(out, num_fractional_digits - num_digits, '0');
+    out = format_decimal<char_type>(out, n, num_digits).end;
+  }
+}
+
+// Format fractional seconds with an appropiate number of digits.
```
vitaut, line 1058: Looks like the difference from the function above is that this one works with floating-point representation type rather than fractionality. Please update the comment accordingly.

include/fmt/chrono.h
```diff
@@ -966,13 +966,122 @@ inline void tzset_once() {
...
+template <typename Rep, long long Num, long long Den>
+void format_sec_fractional(basic_memory_buffer<char>& buf, Rep val) {
```
vitaut, line 1060: basic_memory_buffer<char> -> memory_buffer

Also pass duration directly instead of splitting it into components.

```c++
template <typename Rep, typename Period>
void format_sec_fractional(memory_buffer& buf, std::chrono::duration<Rep, Period> val) {
```

include/fmt/chrono.h
```diff
@@ -966,13 +966,122 @@ inline void tzset_once() {
...
+  FMT_ASSERT(std::is_floating_point<Rep>::value, "");
+  auto num_fractional_digits = count_fractional_digits<Num, Den>::value;
+  // For non-integer values, we ensure at least 6 digits to get microsecond
+  // precision.
+  if (num_fractional_digits < 6 && static_cast<Rep>(std::round(val)) != val) {
+    num_fractional_digits = 6;
+  }
```
vitaut, line 1067: Remove braces around the if body for consistency with the rest of the library.

include/fmt/chrono.h
```diff
@@ -966,13 +966,122 @@ inline void tzset_once() {
...
+  format_to(std::back_inserter(buf), runtime("{:.{}f}"),
+            std::fmod(val * static_cast<Rep>(Num) / static_cast<Rep>(Den),
+                      static_cast<Rep>(60)),
+            num_fractional_digits);
+}
+
+template <typename OutputIt, typename Char, typename Rep = int,
+          typename Period = std::ratio<1, 1>>
```
vitaut, line 1075: Let's replace Rep and Period with Duration for simplicity

include/fmt/chrono.h
```diff
@@ -1135,10 +1244,20 @@ template <typename OutputIt, typename Char> class tm_writer {
   }
 
  public:
+  tm_writer(const std::locale& loc, OutputIt out, const std::tm& tm,
+            std::chrono::duration<Rep, Period> subsecs)
```
vitaut, line 1248: Let's pass subsecs as a pointer with null indicating "no subseconds" and merge the two ctors into one to reduce copy pasta.

include/fmt/chrono.h
```diff
@@ -1337,9 +1456,28 @@ template <typename OutputIt, typename Char> class tm_writer {
...
   void on_second(numeric_system ns) {
-    if (is_classic_ || ns == numeric_system::standard) return write2(tm_sec());
-    format_localized('S', 'O');
+    if (is_classic_ || ns == numeric_system::standard) {
+      write2(tm_sec());
+      if (has_subsecs_) {
+        if (std::is_floating_point<Rep>::value) {
+          auto buf = memory_buffer();
+          format_sec_fractional<Rep, Period::num, Period::den>(
+              buf, subsecs_.count());
```
vitaut, line 1467: subsecs should be passed as duration and template arguments deduced.

include/fmt/chrono.h
```diff
@@ -1337,9 +1456,28 @@ template <typename OutputIt, typename Char> class tm_writer {
...
+          if (buf.size() > 1) {
+            // Remove the leading "0", write something like ".123".
+            out_ = std::copy(buf.begin() + 1, buf.end(), out_);
+          }
+        } else {
+          write_sub_sec<OutputIt, std::chrono::duration<Rep, Period>, Char>(
```
vitaut, line 1473: Only Char should be passed explicitly, the rest should be deduced once the argument order is changed per earlier comment.

include/fmt/chrono.h
```diff
@@ -2016,28 +2062,42 @@ struct formatter<std::chrono::time_point<std::chrono::system_clock, Duration>,
...
+      const auto epoch = val.time_since_epoch();
+      const auto subsecs =
+          std::chrono::duration_cast<std::chrono::duration<Rep, Period>>(
+              epoch - std::chrono::duration_cast<std::chrono::seconds>(epoch));
+
+      return formatter<std::tm, Char>::format(
+          localtime(std::chrono::time_point_cast<std::chrono::seconds>(val)),
+          ctx, subsecs);
+
+    } else {
```
vitaut, line 2081: else is redundant, please remove

vitaut (changes requested):

include/fmt/chrono.h
patrickroocks: done

include/fmt/chrono.h
patrickroocks: done

include/fmt/chrono.h
patrickroocks: done

include/fmt/chrono.h
patrickroocks: Multiline-if without braces feels dangerous... I saw problematic bugs caused to this already sometimes.

But, of course, consistency first :)

include/fmt/chrono.h
patrickroocks: ok, also replaced in
```
template <typename FormatContext, typename Duration>
  auto format(const std::tm& tm, FormatContext& ctx,
              const Duration& subsecs) const
```

include/fmt/chrono.h
patrickroocks: done

include/fmt/chrono.h
patrickroocks: done

include/fmt/chrono.h
patrickroocks: done

include/fmt/chrono.h
patrickroocks: Used pointer now

include/fmt/chrono.h
vitaut: This is better but I suggest changing to `write_floating_seconds` because we use `write` for consistency with `write_fractional_seconds`.

include/fmt/chrono.h
vitaut: I'm pretty sure it should work as you can pass any rep type as part of duration type. This is already done in

```
        write_fractional_seconds<char_type>(
            out, std::chrono::duration<Rep, Period>(val));
```
so we should do the same here.

include/fmt/chrono.h
```diff
@@ -966,13 +966,124 @@ inline void tzset_once() {
...
+  return static_cast<Int>(value);
+}
+template <typename T, typename Int, FMT_ENABLE_IF(!std::is_integral<T>::value)>
+inline Int to_nonnegative_int(T value, Int upper) {
+  if (value < 0 || value > static_cast<T>(upper))
+    FMT_THROW(format_error("invalid value"));
+  return static_cast<Int>(value);
+}
+
+template <class Rep, class Period,
```
vitaut, line 985: nit: class -> typename for consistency

include/fmt/chrono.h
```diff
@@ -966,13 +966,124 @@ inline void tzset_once() {
...
+// Base case that doesn't instantiate any more templates
+// in order to avoid overflow.
+template <long long Num, long long Den, int N>
+struct count_fractional_digits<Num, Den, N, false> {
+  static constexpr int value = (Num % Den == 0) ? N : 6;
+};
+
+// Format subseconds which are given as an integer type with an appropriate
+// number of digits.
+template <typename char_type, typename OutputIt, typename Duration>
```
vitaut, line 1028: char_type -> Char to follow naming conventions for template parameters

vitaut (changes requested):
Mostly looks good, just a few more minor comments.

include/fmt/chrono.h
patrickroocks: done

include/fmt/chrono.h
patrickroocks: done

include/fmt/chrono.h
patrickroocks: Ah, now I got it. I did exactly this, but used the `Rep` parameter and the tests failed for negative values... using `rep` as type it works fine.

include/fmt/chrono.h
patrickroocks: done

vitaut (approved):

vitaut:
Thank you!

vitaut:
Fuzzer discovered an issue with this change. The following code

```c++
#include <fmt/chrono.h>

int main() {
  fmt::format("{:%S}", std::chrono::duration<char, std::ratio<1, 100> >(0x80));
}
```
now results in an assertion failure:
```
    frame #7: 0x000000010000dddb a.out`fmt::v9::detail::assert_fail(char const*, int, char const*) + 59
    frame #8: 0x000000010000ac3e a.out`long long fmt::v9::detail::to_nonnegative_int<long long, long long, 0>(long long, long long) + 94
    frame #9: 0x000000010000a8dd a.out`void fmt::v9::detail::write_fractional_seconds<char, std::__1::back_insert_iterator<fmt::v9::basic_memory_buffer<char, 500ul, std::__1::allocator<char>>>, std::__1::chrono::duration<char, std::__1::ratio<1l, 100l>>>(std::__1::back_insert_iterator<fmt::v9::basic_memory_buffer<char, 500ul, std::__1::allocator<char>>>&, std::__1::chrono::duration<char, std::__1::ratio<1l, 100l>>) + 173
```

phprus:
@vitaut 
Fix: #3143

patrickroocks:
Thanks for the fix and sorry that I have overseen this!

vitaut:
Here's a link to the oss-fuzz report for future reference: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=52380.

PR #3119, opened 2022-09-27
Normalization of stdlib inline namespace names

phprus:
``libc++`` inline namespace:
```
std::__1::system_error                        ->  std::system_error
std::__1::__fs::filesystem::filesystem_error  ->  std::filesystem::filesystem_error
```

``libstdc++`` inline namespace:
```
std::__cxx11::system_error                 ->  std::system_error
std::filesystem::__cxx11::filesystem_error ->  std::filesystem::filesystem_error
```

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -214,10 +214,34 @@ struct formatter<
     std::size_t size = 0;
     std::unique_ptr<char, decltype(&std::free)> demangled_name_ptr(
         abi::__cxa_demangle(ti.name(), nullptr, &size, &status), &std::free);
-    out = detail::write_bytes(
-        out,
-        string_view(demangled_name_ptr ? demangled_name_ptr.get() : ti.name()),
-        spec);
+    string_view demangled_name_view(
+        demangled_name_ptr ? demangled_name_ptr.get() : ti.name());
+
+    // Normalization of stdlib inline namespace names.
+    if (demangled_name_view.starts_with("std::")) {
+      // Separation of user's and system exception classes.
+      demangled_name_view.remove_prefix(5);
+      out = detail::write_bytes(out, string_view("std::"), spec);
+
+      // libc++ inline namespaces.
+      // std::__1::*       -> std::*
+      // std::__1::__fs::* -> std::*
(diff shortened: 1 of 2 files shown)


vitaut (commented):
Thanks for the PR. I wonder if it would be better to remove all namespace components that start with `__` instead of hardcoding specific cases?

phprus:
I think you are right.
Since the ``__cxa_demangle`` allocates memory, the replacement can be making in place.

I will make changes in PR.

phprus:
@vitaut 
Implemented

include/fmt/std.h
```diff
@@ -214,10 +214,41 @@ struct formatter<
...
     std::unique_ptr<char, decltype(&std::free)> demangled_name_ptr(
         abi::__cxa_demangle(ti.name(), nullptr, &size, &status), &std::free);
-    out = detail::write_bytes(
-        out,
-        string_view(demangled_name_ptr ? demangled_name_ptr.get() : ti.name()),
-        spec);
+
+    string_view demangled_name_view;
+    if (demangled_name_ptr) {
+      demangled_name_view = string_view(demangled_name_ptr.get());
```
vitaut, line 220: I don't think you need to explicitly invoke the `string_view` ctor here.

include/fmt/std.h
```diff
@@ -214,10 +214,41 @@ struct formatter<
...
+            while (next < end && *next != ':') next++;
+            if (next[0] == ':' && next[1] == ':') {
+              from = next + 2;
+              continue;
+            }
+          }
+          *to++ = *from++;
+        }
+        demangled_name_view =
+            string_view(begin, detail::to_unsigned(to - begin));
```
vitaut, line 246: same here

vitaut (commented):
LGTM, just a few nits.

include/fmt/std.h
phprus: ``string_view`` does not contain assignment operators/functions for ``char *`` and ``{ char *, size_t }`` pair.
See: https://en.cppreference.com/w/cpp/string/basic_string_view

vitaut:
Thanks!

PR #3122, opened 2022-09-28
Add precision modifier for seconds in chrono format

SappyJoy:
Add precision modifier for seconds in chrono format and test for it.

It's became important for custom datetime types that store date, time and subseconds. I want to use chrono format for those types, but I have no way to control subsecond precision.

Use case:
```
EXPECT_EQ(fmt::format("{:%Y-%m-%d %H:%M:%.6S %Z}", datetime), "2022-09-28 12:14:25.123456 MSK");
```

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -593,6 +593,27 @@ template <typename Period> FMT_CONSTEXPR inline const char* get_units() {
   return nullptr;
 }
 
+template <typename Char>
+struct non_dynamic_precision_handler {
+  int precision = -1;
+
+  void on_error(const char* msg) { FMT_THROW(format_error(msg)); }
+  FMT_CONSTEXPR void on_precision(int _precision) {
+    precision = _precision;
+  }
+  template <typename Id> FMT_CONSTEXPR void on_dynamic_precision(Id  /*_precision*/) {
+    on_error("dynamic precision is not allowed here");
+  }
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR. Is this feature part of the standard?

vitaut (changes requested):
Thanks for the PR. Adding precision support to durations sounds reasonable but for consistency with the standard it should go at the beginning, not in `%S` (https://eel.is/c++draft/time.format), e.g.

```
fmt::print("{:.6%S}", std::chrono::nanoseconds{1234});
```

SappyJoy:
`{:.6%S}` looks like precision for _argument_. There is test for this:

```
  EXPECT_EQ("1.2 ms   ", fmt::format("{:7.1%Q %q}", dms(1.234)));
```

It is not obvious for user to use `{:.6%S}` and expect that this precision modifier belong to `%S`. But it can be done this way if the specifier has a higher priority on getting the modifier than the argument.

vitaut:
> But it can be done this way if the specifier has a higher priority on getting the modifier than the argument.

We should make existing precision apply to `%S` and not introduce a new one to avoid diverging too much from the standard.

SappyJoy:
Oh, I squashed last commits to resolve conflict. It led to close this PR. Please **reopen** it, I made changes. Now it should work.

vitaut:
I am not able to reopen this diff for some reason. Please submit a new one.

Issue #3123, opened 2022-09-29
Proposal: Move formatter<std::error_code> from fmt/os.h to fmt/std.h


phprus:
Move ``formatter<std::error_code>`` from ``fmt/os.h`` to ``fmt/std.h`` for consistency.

vitaut:
Seems reasonable, a PR is welcome!

PR #3125, opened 2022-10-01
Move formatter<std::error_code> from fmt/os.h to fmt/std.h

phprus:
Issue: #3123

### DIFF
--- a/test/std-test.cc
+++ b/test/std-test.cc
@@ -10,6 +10,7 @@
 #include <string>
 #include <vector>
 
+#include "fmt/os.h"  // fmt::system_category
 #include "fmt/ranges.h"
 #include "gtest-extra.h"  // StartsWith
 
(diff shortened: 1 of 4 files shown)


test/std-test.cc
```diff
@@ -10,6 +10,7 @@
 #include <string>
 #include <vector>
 
+#include "fmt/os.h"  // fmt::system_category()
```
vitaut, line 13: nit: we normally put symbol names without parentheses in comments like this

phprus: Fixed.

vitaut:
Thank you!

tupaschoal:
Should this PR also have touched documentation for `std.h`? It currently does not say that it provides formatting for `error_code`, just:
```
fmt/std.h provides formatters for:

[std::filesystem::path](https://en.cppreference.com/w/cpp/filesystem/path)

[std::thread::id](https://en.cppreference.com/w/cpp/thread/thread/id)

[std::monostate](https://en.cppreference.com/w/cpp/utility/variant/monostate)

[std::variant](https://en.cppreference.com/w/cpp/utility/variant/variant)

[std::optional](https://en.cppreference.com/w/cpp/utility/optional)
```

Issue #3128, opened 2022-10-06
Custom formatter can trigger compiler bug for GCC < v12.0 (constexpr downcast)


peterazmanov:
Test case (https://godbolt.org/z/1KvqKjf4n):
```C++
#include <fmt/core.h>

#include <string_view>

struct X
{
    double d = 0.1;
};

template <>
struct fmt::formatter<X> : fmt::formatter<double>
{
   constexpr auto parse(fmt::basic_format_parse_context<char> & ctx) -> decltype(ctx.begin())
   {
       return fmt::formatter<double>::parse(ctx);
   }

   template <class FormatContext>
   constexpr auto format(X x, FormatContext & ctx) const
   {
       return fmt::formatter<double>::format(x.d, ctx);
   }
};

int main()
{
    using namespace std::literals;
    return fmt::format("{:<{}}"sv, X{}, 30).size();
}
```
causes compilation error due to GCC compiler bug (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=103879#c6).
Error message:
`error: accessing value of '<anonymous>.fmt::v9::detail::format_string_checker<char, fmt::v9::detail::error_handler, X, int>::context_.fmt::v9::detail::compile_parse_context<char, fmt::v9::detail::error_handler>::<anonymous>' through a 'fmt::v9::detail::compile_parse_context<char, fmt::v9::detail::error_handler>' glvalue in a constant expression`

This can be workarounded with parse context template parameter:

```C++
   template <typename ParseContext>
   constexpr auto parse(ParseContext & ctx) -> decltype(ctx.begin())
```

Perhaps it's a good idea to add a workaround in fmt code, somehow.
If not, this issue can be closed, hope it'll help someone who stumbles upon this obscure bug.

vitaut:
Worked around in https://github.com/fmtlib/fmt/commit/491c32cbd9ba0cc4b36f7576afcaf6a0a37b0633. Thanks for reporting.

PR #3129, opened 2022-10-06
Fix broken condition

phprus:
``FMT_GCC_VERSION >= 1200`` equal to false for all non GCC compilers.
But this condition was added as a workaround for the GCC bug: https://github.com/fmtlib/fmt/issues/3128.


vitaut:
Good catch, thanks. Let's keep it a normal condition instead of `#ifdef` though.

phprus:
@vitaut 
Fixed.

mwinterb:
Should this condition be wrapped with const_check?

vitaut:
> Should this condition be wrapped with const_check?

I don't think so.

vitaut:
Thank you!

PR #3132, opened 2022-10-11
Explicitly export buffer<char> for MSVC

Youw:
Some MSVC-specific behavior:

Class ostream by inheriting detail::buffer<char> implicitly exports it when fmt is built as a shared library. Unless os.h is included, the compiler assumes detail::buffer<char> is not externally exported and instantiets a local copy of it, which causes ODR violation.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -960,6 +960,14 @@ template <typename T> class buffer {
   }
 };
 
+#if defined(FMT_MSC_VERSION)
+#  ifdef FMT_EXPORT
+extern template class buffer<char>;
+#  else
+template class FMT_API buffer<char>;
+#  endif
+#endif
+
 struct buffer_traits {
   explicit buffer_traits(size_t) {}
   auto count() const -> size_t { return 0; }
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -960,7 +960,7 @@ template <typename T> class buffer {
   }
 };
 
-#if defined(FMT_MSC_VERSION)
+#if FMT_MSC_VERSION
 #  ifdef FMT_EXPORT
 extern template class buffer<char>;
 #  else
--- a/src/format.cc
+++ b/src/format.cc
@@ -23,9 +23,9 @@ template FMT_API auto locale_ref::get<std::locale>() const -> std::locale;
 // Explicit instantiations for char.
 
 #if FMT_MSC_VERSION
-// ostream does it implicitly, so we need make it explicitly
-// otherwise we get ODR violation when {fmt} is built as a shared library
-// and fmt/os.h is not included by the user
+// fmt::ostream does it implicitly, so we need make it explicitly.
+// Otherwise we get ODR violation when {fmt} is built as a shared library
+// and fmt/os.h is not included by the user.
 template class FMT_API buffer<char>;
 #endif
 
--- a/src/format.cc
+++ b/src/format.cc
@@ -22,6 +22,13 @@ template FMT_API auto locale_ref::get<std::locale>() const -> std::locale;
 
 // Explicit instantiations for char.
 
+#if defined(FMT_MSC_VERSION)
+// ostream does it implicitly, so we need make it explicitly
+// otherwise we get ODR violation when {fmt} is built as a shared library
+// and fmt/os.h is not included by the user
+template class FMT_API buffer<char>;
+#endif
+
 template FMT_API auto thousands_sep_impl(locale_ref)
     -> thousands_sep_result<char>;
 template FMT_API auto decimal_point_impl(locale_ref) -> char;
--- a/src/format.cc
+++ b/src/format.cc
@@ -22,7 +22,7 @@ template FMT_API auto locale_ref::get<std::locale>() const -> std::locale;
 
 // Explicit instantiations for char.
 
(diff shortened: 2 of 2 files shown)


Youw:
So the scenario to trigger the ODR is a bit untrivial and gets reproduced only by MSVC, but still:
```
fmt.dll: is built and used as a shared library
my_application:
  main.cpp <- uses: fmt::ostream out = fmt::output_file(...)
  utest_logger.cpp <- uses fmt/format.h only
```
C++17 is used on all platforms.
It compiles and runs just fine on macOS(clang), Linux(GCC) and Android (Clang/NDK).
On Windows (MSVC2019) I get a linker (ODR violation) error:
```
fmt.lib(fmt.dll) : error LNK2005: "protected: __cdecl fmt::v9::detail::buffer<char>::buffer<char>(char *,unsigned __int64,unsigned __int64)" (??0?$buffer@D@detail@v9@fmt@@IEAA@PEAD_K1@Z) already defined in utest_logger.cpp.obj
fmt.lib(fmt.dll) : error LNK2005: "protected: void __cdecl fmt::v9::detail::buffer<char>::set(char *,unsigned __int64)" (?set@?$buffer@D@detail@v9@fmt@@IEAAXPEAD_K@Z) already defined in utest_logger.cpp.obj
fmt.lib(fmt.dll) : error LNK2005: "public: char * __cdecl fmt::v9::detail::buffer<char>::end(void)" (?end@?$buffer@D@detail@v9@fmt@@QEAAPEADXZ) already defined in utest_logger.cpp.obj
fmt.lib(fmt.dll) : error LNK2005: "public: unsigned __int64 __cdecl fmt::v9::detail::buffer<char>::size(void)const " (?size@?$buffer@D@detail@v9@fmt@@QEBA_KXZ) already defined in utest_logger.cpp.obj
fmt.lib(fmt.dll) : error LNK2005: "public: unsigned __int64 __cdecl fmt::v9::detail::buffer<char>::capacity(void)const " (?capacity@?$buffer@D@detail@v9@fmt@@QEBA_KXZ) already defined in utest_logger.cpp.obj
fmt.lib(fmt.dll) : error LNK2005: "public: char * __cdecl fmt::v9::detail::buffer<char>::data(void)" (?data@?$buffer@D@detail@v9@fmt@@QEAAPEADXZ) already defined in utest_logger.cpp.obj
fmt.lib(fmt.dll) : error LNK2005: "public: void __cdecl fmt::v9::detail::buffer<char>::try_resize(unsigned __int64)" (?try_resize@?$buffer@D@detail@v9@fmt@@QEAAX_K@Z) already defined in utest_logger.cpp.obj
fmt.lib(fmt.dll) : error LNK2005: "public: void __cdecl fmt::v9::detail::buffer<char>::try_reserve(unsigned __int64)" (?try_reserve@?$buffer@D@detail@v9@fmt@@QEAAX_K@Z) already defined in utest_logger.cpp.obj
fmt.lib(fmt.dll) : error LNK2005: "public: void __cdecl fmt::v9::detail::buffer<char>::push_back(char const &)" (?push_back@?$buffer@D@detail@v9@fmt@@QEAAXAEBD@Z) already defined in utest_logger.cpp.obj
target\bin\utest_core.exe : fatal error LNK1169: one or more multiply defined symbols found
```

---

This is the result running on current release 9.1.0.
On latest master there is less "`already defined in`" errors, but overall the issue persists.

include/fmt/core.h
```diff
@@ -960,6 +960,14 @@ template <typename T> class buffer {
   }
 };
 
+#if defined(FMT_MSC_VERSION)
```
Youw, line 963:
```suggestion
#if FMT_MSC_VERSION
```

src/format.cc
```diff
@@ -22,6 +22,13 @@ template FMT_API auto locale_ref::get<std::locale>() const -> std::locale;
 
 // Explicit instantiations for char.
 
+#if defined(FMT_MSC_VERSION)
```
Youw, line 25:
```suggestion
#if FMT_MSC_VERSION
```

src/format.cc
```diff
@@ -22,6 +22,13 @@ template FMT_API auto locale_ref::get<std::locale>() const -> std::locale;
 
 // Explicit instantiations for char.
 
+#if FMT_MSC_VERSION
+// ostream does it implicitly, so we need make it explicitly
+// otherwise we get ODR violation when {fmt} is built as a shared library
+// and fmt/os.h is not included by the user
```
vitaut, line 28: ostream -> fmt::ostream to avoid ambiguity with std::ostream

Also let's make it proper sentences:

```
// fmt::ostream does it implicitly, so we need make it explicitly.
// Otherwise we get ODR violation when {fmt} is built as a shared library
// and fmt/os.h is not included by the user.
```

Youw: thanks

vitaut:
> Is the any automated tools/scripts to check the performance before/after the change?

Not automated but there are some benchmarks in https://github.com/fmtlib/format-benchmark. They might need a bit of an update.

Youw:
It just came to my mind, but I think the alternative to this fix would be to aggregate `detail::buffer<char>` by the `ostream` instead of inheriting it.
But I need to test this assumption.

vitaut:
#3139 looks like a better workaround, let's go with it.

PR #3134, opened 2022-10-11
Fix options for C++20 experimental module in CMake

laitingsheng:
This PR, as discussed previously under commit 3def950b8479cdd5769b4d4e7794d94648ebdbd6, is to fix a mistake introduced in #2432 which disabled option for building the module with the new C++20 module system completely (even for compatible compiler).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,10 +84,8 @@ option(FMT_MODULE "Build a module instead of a traditional library." OFF)
 option(FMT_SYSTEM_HEADERS "Expose headers with marking them as system." OFF)
 
 set(FMT_CAN_MODULE OFF)
-if (CMAKE_CXX_STANDARD GREATER 17 AND
-    # msvc 16.10-pre4
-    MSVC AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 19.29.30035)
-  set(FMT_CAN_MODULE OFF)
+if (CMAKE_CXX_STANDARD GREATER 17 AND NOT MSVC)
+  set(FMT_CAN_MODULE ON)
 endif ()
 if (NOT FMT_CAN_MODULE)
   set(FMT_MODULE OFF)
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -94,7 +94,7 @@ add_fmt_test(enforce-checks-test)
 target_compile_definitions(enforce-checks-test PRIVATE
                            -DFMT_ENFORCE_COMPILE_STRING)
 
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -63,6 +64,11 @@ if (FMT_MASTER_PROJECT AND NOT CMAKE_BUILD_TYPE)
               "CMAKE_C_FLAGS used) Debug Release RelWithDebInfo MinSizeRel.")
 endif ()
 
+set(FMT_CAN_MODULE OFF)
+if (CMAKE_CXX_STANDARD GREATER 17 AND NOT MSVC)
+  set(FMT_CAN_MODULE ON)
+endif ()
```
vitaut, line 70: Most compiler versions don't support modules so this should be disabled by default, not enabled by default.

CMakeLists.txt
```diff
@@ -80,19 +86,13 @@ option(FMT_TEST "Generate the test target." ${FMT_MASTER_PROJECT})
 option(FMT_FUZZ "Generate the fuzz target." OFF)
 option(FMT_CUDA_TEST "Generate the cuda-test target." OFF)
 option(FMT_OS "Include core requiring OS (Windows/Posix) " ON)
-option(FMT_MODULE "Build a module instead of a traditional library." OFF)
+cmake_dependent_option(FMT_MODULE "Build a module instead of a traditional library." OFF FMT_CAN_MODULE OFF)
```
vitaut, line 89: I don't think it should be a dependent option. Users should be able to override detection if needed.

vitaut (changes requested):
Thanks for the PR.

CMakeLists.txt
laitingsheng: This is just a guard for the dependent option which will hide the option if the new module system is not available. If this is `TRUE`, `FMT_MODULE` will be the default value which is `FALSE`, as in the following code segment. Or otherwise, the option will be hidden completely from the user.

CMakeLists.txt
laitingsheng: Alright, I've changed this to the simpler method, which is just excluding the MSVC, but still allowing the user to force overriding the `FMT_MODULE`.

laitingsheng:
> replace FMT_CAN_MODULE with FMT_MODULE in test to fix CI

Thanks for the hint. I actually found the same issue from the CI logs, but I was hesitating if it's valid to replace the variable.

vitaut (approved):

vitaut:
Thank you

Issue #2809, opened 2022-03-13
fmt local scope `uint` conflicts with OpenCV's globally scoped `uint`


diablodale:
Unfortunately, OpenCV has a globally scoped `uint` type. It causes a conflict with the locally scoped `uint` used by fmt in 5 places.
I request fmt tweak this local typename to avoid the conflict. In my experience, OpenCV will not change.

### Setup

* fmt wth `FMT_VERSION 80101`
* OpenCV 4.5.3
* Microsoft Windows [Version 10.0.19044.1586]
* VS2019 v16.11.10

### Repro

1. Create code that includes both fmt and OpenCV core, then calls one of the log outputters like `spdlog::error("asdasd");`
2. Build with MSVC's code analyzer

### Result

Build fails with errors indicating the the global OpenCV type `uint` is being hidden by fmt's local `uint`

```
[build] vcpkg_installed\x64-windows-static-md-v142-sdk10\include\fmt/format.h(2042): error C2220: the following warning is treated as an error
[build] vcpkg_installed\x64-windows-static-md-v142-sdk10\include\fmt/format.h(2042): warning C4459: declaration of 'uint' hides global declaration
[build] C:\repos-nobackup\opencv\.install\Release\include\opencv2/core/hal/interface.h(45): note: see declaration of 'uint'
[build] vcpkg_installed\x64-windows-static-md-v142-sdk10\include\fmt/format.h(2197): note: see reference to function template instantiation 'OutputIt fmt::v8::detail::write<Char,std::back_insert_iterator<fmt::v8::detail::buffer<T>>,double,0>(OutputIt,double)' being compiled
[build]         with
[build]         [
[build]             OutputIt=std::back_insert_iterator<fmt::v8::detail::buffer<wchar_t>>,
[build]             Char=wchar_t,
[build]             T=wchar_t
[build]         ]
[build] vcpkg_installed\x64-windows-static-md-v142-sdk10\include\fmt/core.h(1611): note: see reference to function template instantiation 'std::back_insert_iterator<fmt::v8::detail::buffer<T>> fmt::v8::detail::default_arg_formatter<Char>::operator ()<double>(double)' being compiled
[build]         with
[build]         [
[build]             T=wchar_t,
[build]             Char=wchar_t
[build]         ]
[...]

PR #3137, opened 2022-10-12
Avoid using `uint` as a type name

Youw:
Sometime `uint` is defined as a global type by the project's code directly or by some 3rdparty libraries (e.g. Qt or OpenCV). Some versions of MSVC (e.g. v16.11.15) gives a type shadowing warning:
```
3rdparty\fmtlib\fmt\include\fmt/format.h(3251): warning C4459: declaration of 'uint' hides global declaration
opencv2/core/hal/interface.h(45): note: see declaration of 'uint'
```
This also causes a compilation failure when `/WX` is used.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -1405,8 +1405,8 @@ template <typename Float> constexpr int num_significand_bits() {
 template <typename Float>
 constexpr auto exponent_mask() ->
     typename dragonbox::float_info<Float>::carrier_uint {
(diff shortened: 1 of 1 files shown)


vitaut:
In general we don't try to workaround "conflicts" with third-party libraries but I guess this is a pretty minor change so we can make an exception this time.

Youw:
Fixes: #2809

Issue #3140, opened 2022-10-14
Clang-CL warning : declaration of 'Duration' shadows template parameter [-Wmicrosoft-template-shadow]


Epixu:
I get `fmt\chrono.h(2093,46): warning : declaration of 'Duration' shadows template parameter [-Wmicrosoft-template-shadow]` when building with latest clang-cl - other compilers don't seem to bother (tested on GNU 11.1.0 and MSVC2022).
As of [this commit](https://github.com/fmtlib/fmt/commit/9254cfa6f0cc9d6fc9eec0c16181bfbba08e75ec).

vitaut:
Fixed in #3142.

PR #3142, opened 2022-10-14
Remove duplicate template parameter.

phprus:
Fix for issue #3140.

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -2061,17 +2061,15 @@ struct formatter<std::chrono::time_point<std::chrono::system_clock, Duration>,
     this->do_parse(default_specs.begin(), default_specs.end());
   }
 
-  template <typename FormatContext, typename Rep, typename Period>
-  auto format(std::chrono::time_point<std::chrono::system_clock,
-                                      std::chrono::duration<Rep, Period>>
-                  val,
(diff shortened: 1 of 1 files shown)


include/fmt/chrono.h
```diff
@@ -2052,16 +2052,17 @@ struct formatter<std::chrono::duration<Rep, Period>, Char> {
   }
 };
 
-template <typename Char, typename Duration>
-struct formatter<std::chrono::time_point<std::chrono::system_clock, Duration>,
+template <typename Rep, typename Period, typename Char>
```
vitaut, line 2055: Why change template parameters? I suggest keeping the old ones at least for consistency with utc formatter and change the `format` function instead.

vitaut (changes requested):
Thanks for the PR.

include/fmt/chrono.h
phprus: We need ``Rep`` and ``Period`` types for the condition on lines 2070-2071.

vitaut: Yeah but these are available as `Duration::rep` and `Duration::period`. I don't think we need them here.

vitaut (approved):

vitaut:
Merged, thanks!

PR #3143, opened 2022-10-16
Fix overflow error

phprus:
Fix for https://github.com/fmtlib/fmt/pull/3115#issuecomment-1279995823


vitaut:
Thank you for the fix!

PR #3145, opened 2022-10-16
Mention MariaDB amongst the projects that use fmt

federico-razzoli:
MariaDB 10.7+ has an SFORMAT() function that is a bridge to fmt. It's documented here:
https://mariadb.com/kb/en/sformat/


vitaut:
Awesome, thanks!

PR #3148, opened 2022-10-19
Add precision modifier for seconds in chrono format

SappyJoy:
Continuation of #3122

Add precision modifier for seconds in chrono format and test for it.

It's became important for custom datetime types that store date, time and subseconds. I want to use chrono format for those types, but I have no way to control subsecond precision.

Use case:

```
EXPECT_EQ(fmt::format("{:.6%Y-%m-%d %H:%M:%S %Z}", datetime), "2022-09-28 12:14:25.123456 MSK");
```

So precision can be applied not only for %Q spec, but for %S too.

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -1038,12 +1038,13 @@ void write_fractional_seconds(OutputIt& out, Duration d, int precision = -1) {
 
   const auto fractional =
       detail::abs(d) - std::chrono::duration_cast<std::chrono::seconds>(d);
-  uint32_or_64_or_128_t<long long> subseconds =
+  const auto& subseconds =
       std::chrono::treat_as_floating_point<
           typename subsecond_precision::rep>::value
           ? fractional.count()
           : std::chrono::duration_cast<subsecond_precision>(fractional).count();
-  const int num_digits = detail::count_digits(subseconds);
+  auto n = static_cast<uint32_or_64_or_128_t<long long>>(subseconds);
+  const int num_digits = detail::count_digits(n);
 
   int leading_zeroes = std::max(0, num_fractional_digits - num_digits);
   if (precision < 0) {
@@ -1052,7 +1053,7 @@ void write_fractional_seconds(OutputIt& out, Duration d, int precision = -1) {
                         std::chrono::seconds::period>::value) {
       *out++ = '.';
       out = std::fill_n(out, leading_zeroes, '0');
-      out = format_decimal<Char>(out, subseconds, num_digits).end;
+      out = format_decimal<Char>(out, n, num_digits).end;
     }
   } else {
     *out++ = '.';
@@ -1060,12 +1061,11 @@ void write_fractional_seconds(OutputIt& out, Duration d, int precision = -1) {
     out = std::fill_n(out, leading_zeroes, '0');
     int remaining = precision - leading_zeroes;
     if (remaining < num_digits) {
-      subseconds /=
-          to_unsigned(detail::pow10(to_unsigned(num_digits - remaining)));
-      out = format_decimal<Char>(out, subseconds, remaining).end;
+      n /= to_unsigned(detail::pow10(to_unsigned(num_digits - remaining)));
+      out = format_decimal<Char>(out, n, remaining).end;
       return;
     }
-    out = format_decimal<Char>(out, subseconds, num_digits).end;
+    out = format_decimal<Char>(out, n, num_digits).end;
     remaining -= num_digits;
     out = std::fill_n(out, remaining, '0');
   }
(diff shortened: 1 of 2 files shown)


vitaut:
Could you copy the summary (updating it if needed) from #3122?

include/fmt/chrono.h
```diff
@@ -1081,6 +1119,26 @@ void write_floating_seconds(memory_buffer& buf, Duration duration) {
...
 
+template <typename Duration>
+void write_floating_seconds(memory_buffer& buf, Duration duration,
+                            int precision) {
+  if (precision < 0) {
+    write_floating_seconds(buf, duration);
+    return;
+  }
+
+  FMT_ASSERT(std::is_floating_point<typename Duration::rep>::value, "");
```
vitaut, line 1130: Let's add an alias for `typename Duration::rep` and use it here and in several places below.

SappyJoy: Done

include/fmt/chrono.h
```diff
@@ -1056,31 +1056,91 @@ void write_fractional_seconds(OutputIt& out, Duration d) {
   }
 }
 
+template <typename Char, typename OutputIt, typename Duration>
+void write_fractional_seconds(OutputIt& out, Duration d, int precision) {
```
vitaut, line 1060: Let's merge this overload with the one above to reduce code duplication.

include/fmt/chrono.h
```diff
@@ -1056,31 +1056,91 @@ void write_fractional_seconds(OutputIt& out, Duration d) {
...
+    auto fractional =
+        detail::abs(d) - std::chrono::duration_cast<std::chrono::seconds>(d);
+    auto subseconds =
+        std::chrono::treat_as_floating_point<
+            typename subsecond_precision::rep>::value
+            ? fractional.count()
+            : std::chrono::duration_cast<subsecond_precision>(fractional)
+                  .count();
+    uint32_or_64_or_128_t<long long> n =
+        to_unsigned(to_nonnegative_int(subseconds, max_value<long long>()));
```
vitaut, line 1080: Can subseconds be negative?

include/fmt/chrono.h
```diff
@@ -1056,31 +1056,91 @@ void write_fractional_seconds(OutputIt& out, Duration d) {
...
+  }
+}
+
 // Format subseconds which are given as a floating point type with an appropiate
 // number of digits. We cannot pass the Duration here, as we explicitly need to
 // pass the Rep value in the chrono_formatter.
 template <typename Duration>
 void write_floating_seconds(memory_buffer& buf, Duration duration) {
-  FMT_ASSERT(std::is_floating_point<typename Duration::rep>::value, "");
+  using Rep =  typename Duration::rep;
```
vitaut, line 1102: nit: Rep -> rep to follow naming conventions

include/fmt/chrono.h
```diff
@@ -1056,31 +1056,91 @@ void write_fractional_seconds(OutputIt& out, Duration d) {
...
-                static_cast<typename Duration::rep>(60)),
+                    static_cast<Rep>(Duration::period::num) /
+                    static_cast<Rep>(Duration::period::den),
+                static_cast<Rep>(60)),
       num_fractional_digits);
 }
 
+template <typename Duration>
+void write_floating_seconds(memory_buffer& buf, Duration duration,
+                            int precision) {
```
vitaut, line 1125: Similarly, let's merge the two `write_floating_seconds` overloads.

include/fmt/chrono.h
```diff
@@ -1518,6 +1578,9 @@ class tm_writer {
 };
 
 struct chrono_format_checker : null_chrono_spec_handler<chrono_format_checker> {
+  bool is_floating_point = false;
+  bool has_precision = false;
```
vitaut, line 1582: We only need one flag for the condition `has_precision && !is_floating_point`, something like `has_precision_integral`.

vitaut (changes requested):

include/fmt/chrono.h
SappyJoy: It's nonnegative because `fractional` is nonnegative:
```const auto fractional =
      detail::abs(d) - std::chrono::duration_cast<std::chrono::seconds>(d);```

But subseconds can be signed or unsigned and `to_unsigned` raises error if argument is unsigned, so i'll make `static_cast`.

File: include/fmt/chrono.h [hunk 7]
```diff
@@ -1036,49 +1035,66 @@ void write_fractional_seconds(OutputIt& out, Duration d) {
       typename std::common_type<typename Duration::rep,
                                 std::chrono::seconds::rep>::type,
       std::ratio<1, detail::pow10(num_fractional_digits)>>;
-  if (std::ratio_less<typename subsecond_precision::period,
-                      std::chrono::seconds::period>::value) {
+
+  const auto fractional =
+      detail::abs(d) - std::chrono::duration_cast<std::chrono::seconds>(d);
+  const auto& subseconds =
```
vitaut, line 1041: nit: `&` is not needed here

vitaut (approved):

vitaut:
Thanks!

PR #3179, opened 2022-11-13
Replacing snprintf-based hex float formatter with internal implementation

phprus:
1) Replacing snprintf-based hex float formatter with internal implementation.
2) Add missing env CTEST_OUTPUT_ON_FAILURE.
3) Enable C++17 tests on macOS for simplify hex float tests.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -360,6 +360,10 @@ class uint128_fallback {
       -> uint128_fallback {
     return {lhs.hi_ & rhs.hi_, lhs.lo_ & rhs.lo_};
   }
+  friend constexpr auto operator~(const uint128_fallback& n)
+      -> uint128_fallback {
+    return {~n.hi_, ~n.lo_};
+  }
   friend auto operator+(const uint128_fallback& lhs,
                         const uint128_fallback& rhs) -> uint128_fallback {
     auto result = uint128_fallback(lhs);
@@ -397,6 +401,10 @@ class uint128_fallback {
     FMT_ASSERT(new_hi >= hi_, "");
     lo_ = new_lo;
     hi_ = new_hi;
+  }
+  FMT_CONSTEXPR void operator&=(uint128_fallback n) {
+    lo_ &= n.lo_;
+    hi_ &= n.hi_;
   }
 
   FMT_CONSTEXPR20 uint128_fallback& operator+=(uint64_t n) noexcept {
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -1626,62 +1626,6 @@ FMT_CONSTEXPR inline fp get_cached_power(int min_exponent,
           *(data::pow10_exponents + index)};
 }
 
-#ifndef _MSC_VER
-#  define FMT_SNPRINTF snprintf
-#else
-FMT_API auto fmt_snprintf(char* buf, size_t size, const char* fmt, ...) -> int;
-#  define FMT_SNPRINTF fmt_snprintf
-#endif  // _MSC_VER
-
-// Formats a floating-point number with snprintf using the hexfloat format.
-template <typename T>
-auto snprintf_float(T value, int precision, float_specs specs,
-                    buffer<char>& buf) -> int {
-  // Buffer capacity must be non-zero, otherwise MSVC's vsnprintf_s will fail.
-  FMT_ASSERT(buf.capacity() > buf.size(), "empty buffer");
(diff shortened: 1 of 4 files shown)


phprus:
Benchmark:
https://github.com/phprus/fmt-bench/tree/hexfloat-1

MacbookPro M1 Max:

- Default format string: without changes.
- Compiled format string: ~3 times faster.

include/fmt/format.h
```diff
@@ -3171,6 +3123,88 @@ FMT_CONSTEXPR20 inline void format_dragon(basic_fp<uint128_t> value,
   buf[num_digits - 1] = static_cast<char>('0' + digit);
 }
 
+// Formats a floating-point number with snprintf using the hexfloat format.
```
vitaut, line 3126: "with snprintf" part no longer applies, let's remove.

test/format-test.cc
```diff
@@ -1344,16 +1344,82 @@ TEST(format_test, format_double) {
...
+  }
+#endif
+
+  d = (std::numeric_limits<double>::min)();
+  safe_sprintf(buffer, "%a", d);
+  std::string d_min_outputs[] = {
+      "0x1p-1022",               // UNIX
+      "0x1.0000000000000p-1022"  // Windows
+  };
+  EXPECT_THAT(d_min_outputs, testing::Contains(buffer));
```
vitaut, line 1380: I don't think we need to test sprintf output. Let's remove this check here and in similar places below.

test/format-test.cc
```diff
@@ -1344,16 +1344,82 @@ TEST(format_test, format_double) {
...
+  EXPECT_EQ(fmt::format("{:a}", d), "0x1p-1022");
+  EXPECT_EQ(fmt::format("{:#a}", d), "0x1.p-1022");
```
vitaut, line 1382: This should be guarded by a check that double is IEEE754.

vitaut (changes requested):
Thanks for the PR! Overall looks great, a few minor comments inline. Also please move the build config changes into a separate PR.

phprus:
@vitaut CI PR: #3188
I will do rebase this PR after the merge PR #3188

include/fmt/format.h
```diff
@@ -398,6 +402,10 @@ class uint128_fallback {
     lo_ = new_lo;
     hi_ = new_hi;
   }
+  FMT_CONSTEXPR void operator&=(uint128_fallback n) {
+    lo_ = lo_ & n.lo_;
+    hi_ = hi_ & n.hi_;
```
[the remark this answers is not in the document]
phprus: Done

include/fmt/format.h
phprus: Done

test/format-test.cc
phprus: Done

phprus:
@vitaut, Rebased.

phprus:
Rebased onto master

include/fmt/format.h
```diff
@@ -3169,6 +3121,88 @@ FMT_CONSTEXPR20 inline void format_dragon(basic_fp<uint128_t> value,
...
+  f.e += num_float_significand_bits;
+  if (!has_implicit_bit<Float>()) --f.e;
+
+  constexpr auto num_fraction_bits =
+      num_float_significand_bits + (has_implicit_bit<Float>() ? 1 : 0);
+  constexpr auto num_xdigits = (num_fraction_bits + 3) / 4;
+
+  constexpr auto leading_shift = ((num_xdigits - 1) * 4);
+  const auto leading_mask = carrier_uint(0xF) << leading_shift;
+  const auto leading_v =
```
vitaut, line 3150: What does "v" stand for?

vitaut (approved):

vitaut:
Merged, thanks!

include/fmt/format.h
phprus: "value".

But, probably, it would be more correct name: "leading_[hex]digit" or "most_significant_[hex]digit".
Now I think ``leading_*`` are not good names.

Make a PR with renaming? If yes, which option should I choose?

vitaut: A PR sounds good. I think leading_hexdigit or leading_xdigit (for consistency with num_xdigits) is fine. most_significant_* might be more precise but longer and I think leading is clear enough.

phprus: Done: PR #3203

Issue #3178, opened 2022-11-13
Building in Cygwin fails due to undeclared strptime in time.h


thesmurph:
Building in Cygwin fails due to undeclared strptime in time.h. This function is not available when __GNU_VISIBLE is not defined. The strptime() function is defined in the X/Open Portability Guide Issue 4 ("XPG4"). Ultimately to enable strptime you can either define _GNU_SOURCE or set _XOPEN_VERSION to 4/500/600/700. I was able to build fmt after adding "add_definitions(-D_XOPEN_SOURCE=700)" to CMakeLists.txt.

CMake version: 3.23.2
The CXX compiler identification is GNU 11.3.0
commit d2e89c8b080394e996d449371267365c223ca76b

**Error Message**
[ 68%] Building CXX object test/CMakeFiles/scan-test.dir/scan-test.cc.o
/home/Matt/fmt/test/scan-test.cc: In instantiation of ‘typename ScanContext::iterator fmt::v9::scanner<tm>::scan(tm&, ScanContext&) [with ScanContext = fmt::v9::scan_context; typename ScanContext::iterator = const char*]’:
/home/Matt/fmt/test/scan.h:111:26:   required from ‘static void fmt::v9::detail::scan_arg::scan_custom_arg(void*, fmt::v9::scan_parse_context&, fmt::v9::scan_context&) [with T = tm]’
/home/Matt/fmt/test/scan.h:102:17:   required from ‘fmt::v9::detail::scan_arg::scan_arg(T&) [with T = tm]’
/home/Matt/fmt/test/scan.h:231:20:   required from ‘std::array<fmt::v9::detail::scan_arg, sizeof... (Args)> fmt::v9::make_scan_args(Args& ...) [with Args = {tm}]’
/home/Matt/fmt/test/scan.h:244:49:   required from ‘const char* fmt::v9::scan(fmt::v9::string_view, fmt::v9::string_view, Args& ...) [with Args = {tm}; fmt::v9::basic_string_view<char>::iterator = const char*; fmt::v9::string_view = fmt::v9::basic_string_view<char>]’
/home/Matt/fmt/test/scan-test.cc:96:12:   required from here
/home/Matt/fmt/test/scan-test.cc:86:27: error: ‘strptime’ was not declared in this scope; did you mean ‘strftime’?
   86 |     auto result = strptime(ctx.begin(), format.c_str(), &t);
[...]

PR #3184, opened 2022-11-15
This skips the scan-test if strptime isn't defined.

thesmurph:
This skips the scan-test if strptime isn't defined. Newlib is used in cygwin and embedded systems. By default newlib doesn't provide strptime in time.h because it was added in a later X/Open versions. Issue: #3178


vitaut:
Thank you

PR #3200, opened 2022-11-22
Rename check to fmt_check (fixes #2965) so that it works with UE

### LINKED ISSUE #2965: Visual studio 2019 compile error
Reporting, was using older fmt 5.x version fine.  Started working on a new project with latest master (fa2eb2d2e3ec5c21629f8ccd88ae05ec40b963fa), it has compile error (after removing module: in fmt.cc)

I'm able to put the old fmt 5.x in new project and compiles fine.

image

vampy:
Fixes https://github.com/fmtlib/fmt/issues/2965

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -1354,10 +1354,10 @@ inline auto format_as(std::byte b) -> unsigned char {
 template <typename T> struct has_format_as {
   template <typename U, typename V = decltype(format_as(U())),
             FMT_ENABLE_IF(std::is_enum<U>::value&& std::is_integral<V>::value)>
-  static auto check(U*) -> std::true_type;
-  static auto check(...) -> std::false_type;
+  static auto fmt_check(U*) -> std::true_type;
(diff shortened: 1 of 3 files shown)


vampy:
Happy to rename it to something else if this name is not appropriate 🙏

vitaut:
Thanks for the PR but `check` are properly namespaced/local names so there is no reason to rename them. If this is causing problems with some third-party macros please report to the project that defines them.

vampy:
This is for unreal engine https://www.unrealengine.com/en-US/
Chance of them to rename the [`check` macro](https://docs.unrealengine.com/4.27/en-US/ProgrammingAndScripting/ProgrammingWithCPP/Assertions/) to something else is zero.

This change will not break anything but will allow unreal engine users to use this library.
I know this is not ideal, but this is the pain of using macros.

vitaut:
Please report to UE. It is clearly a problematic API that should be fixed on their side by prefixing the macro.

PR #3207, opened 2022-11-27
Remove workaround for GTest bug

phprus:
Gtest bug (https://github.com/google/googletest/issues/705) already fixed in used GTest version:

Upstream:
- https://github.com/google/googletest/issues/705#issuecomment-235067917
- https://github.com/google/googletest/commit/b5c81098a8ccc25e313ffca56c911200b3591ea0

FMT commit:
- https://github.com/fmtlib/fmt/commit/b4f9a058945a25d8a4af589d8e9168c70195077e


vitaut:
Good catch (and fix), thanks!

PR #3210, opened 2022-11-28
Bump tested CMake version to 3.25

phprus:
Version update will allow new cmake policies to be used.


ChrisThrasher:
What's the purpose of specifying a max? Why not just specify the minimum version?

PR #3214, opened 2022-11-30
fix #3105 - Compile-time error when mixing named argument with automatic indexing

### LINKED ISSUE #3105: Compile-time error when mixing named argument with automatic indexing
Minimal reproducer ([godbolt](https://godbolt.org/z/bxYGbn9cE)):
```
fmt::print("{} {a}", 1, "a"_a = 2);
```

I believe this is supposed to compile, but it causes:
```
on_error("cannot switch from automatic to manual argument indexing");
```

Are the named arguments considered as "indexed", or is it a bug?

It works with `fmt::runtime`, so if nothing else, there is a problem with consistency between compile-time and runtime checking.

rbrich:
Fixes #3105

Looks like this is enough to fix the issue.

I compared the runtime checking with compile-time checking and adjusted the compile-time version to mirror the runtime one.

The runtime check is this (second function for named args):
```
    FMT_CONSTEXPR auto on_arg_id(int id) -> int {
      return parse_context.check_arg_id(id), id;
    }
    FMT_CONSTEXPR auto on_arg_id(basic_string_view<Char> id) -> int {
      int arg_id = context.arg_id(id);
      if (arg_id < 0) on_error("argument not found");
      return arg_id;
    }
```
It returns the arg index directly without running it through the `check_arg_id`, unlike the indexed arg check.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -2984,7 +2984,7 @@ class format_string_checker {
 #if FMT_USE_NONTYPE_TEMPLATE_ARGS
     auto index = get_arg_index_by_name<Args...>(id);
     if (index == invalid_arg_index) on_error("named argument is not found");
-    return context_.check_arg_id(index), index;
+    return index;
 #else
     (void)id;
     on_error("compile-time checks for named arguments require C++20 support");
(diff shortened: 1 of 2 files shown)


vitaut:
Thank you!

Issue #3220, opened 2022-12-08
Modifier for timezone offset result in compile error


dgnthr:
Hi,

I'm currently trying to print ISO8601 compatible time strings.

As per the specification I can print the timezone offset to UTC with the "%z" conversion specifier. This should for example print "+0100". Using the modifier E or O as in "%Ez" should print "+01:00". If I use either modifier I get a compile time exception FMT_THROW(format_error("invalid format"));

The example code I use is
```
#include <chrono>
#include <fmt/chrono.h>

int main() {
    auto time = std::time(0);
    auto LocalTime = std::localtime(&time);
    fmt::print("{:%z}\n", *LocalTime);
    fmt::print("{:%Oz}\n", *LocalTime);

    return 0;
}
```

Also see the example on compiler explorer https://godbolt.org/z/9cEvzKPn9

Am I doing something wrong? Thank you for your time!

dgnthr:
I just checked the source code and saw that the modifiers (E, O) are not implemented for %z. I'll just use it without the colon.

phprus:
Modifiers (``E``, ``O``) are not defined for ``%z`` format:
https://en.cppreference.com/w/cpp/chrono/c/strftime

dgnthr:
I was referencing this https://en.cppreference.com/w/cpp/chrono/duration/formatter (close to the bottom).

> The modified commands %Ez and %Oz insert a : between the hours and minutes (e.g., -04:30).

phprus:
@dgnthr 
PR: #3222

dgnthr:
Wow! Amazingly fast. Thank you so much!

PR #3222, opened 2022-12-09
Implement %Ez, %Oz for chrono formatter

phprus:
Fix: #3220 

Specification:
https://en.cppreference.com/w/cpp/chrono/duration/formatter


vitaut:
Thank you!

Issue #3199, opened 2022-11-21
Formatting a std::chrono::system_clock renders in local time, not in UTC


ned14:
For some time now we've had this in our cmake, so I figured time to upstream this:

```
  message(STATUS "NOTE: Patching '${VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/include/fmt/chrono.h' to fix a formatting bug ...")
  file(READ "${VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/include/fmt/chrono.h" fmt_chrono_contents)
  string(REPLACE "  auto format(std::chrono::time_point<std::chrono::system_clock> val," "  auto format(std::chrono::time_point<std::chrono::system_clock, Duration> val," fmt_chrono_contents "${fmt_chrono_contents}")
  string(REPLACE "    return formatter<std::tm, Char>::format(localtime(val), ctx);" "    return formatter<std::tm, Char>::format(gmtime(std::chrono::time_point_cast<std::chrono::system_clock::time_point::duration>(val)), ctx);" fmt_chrono_contents "${fmt_chrono_contents}")
  file(WRITE "${VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/include/fmt/chrono.h" "${fmt_chrono_contents}")
```

Two things are changed here:

1. The `Duration` template parameter isn't passed through, breaking formatting time points with non-default duration types.

2. Formatting system clock time points is to UTC, not to localtime, as system clock specifically represents UTC not local time (if you want a local time point, use `chrono::local_t`). Microsoft's implementation of `std::format` agrees with this formatting interpretation i.e. system clocks format to UTC strings, local clocks format to local strings.

vitaut:
The first change has already been applied. Could you by any chance submit the second one as a PR?

ned14:
Last day of work for this year, so had a few spare minutes to get this done finally.

ned14:
Apologies for the unprompted reply. Currently on holidays in the United States without any ability to write code, which is deliberate. Glad you figured out some way to merge it, will look into your remaining ask when I return to Europe in January.

Jan-Otto-Bakkland-Volue:
Converting to local time not working as specified:
auto localTime = std::chrono::localtime(first_time)
![image](https://github.com/fmtlib/fmt/assets/116543087/5d643f2b-25fe-4bed-b482-dfb866c714eb)

or this:
[...]

PR #3230, opened 2022-12-16
Formatting of system clocks ought to be to UTC, not to local time.

ned14:
This improves standards conformance of fmt.

#3199

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -22,6 +22,15 @@
 
 FMT_BEGIN_NAMESPACE
 
+// Check if std::chrono::local_t is available.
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR! Overall looks good but please fix the compilation error: https://github.com/fmtlib/fmt/actions/runs/3715176616/jobs/6310218797.

vitaut:
Merged with one fix in https://github.com/fmtlib/fmt/commit/115001a3b165483563822645bb93fd9558908c50. Note that `local_system_clock_time_point` is still broken on Windows and therefore temporarily disabled. Another PR to fix it would be welcome.

Issue #3219, opened 2022-12-08
formatting chrono with precision


brevzin:
```cpp
#include <chrono>
#include <fmt/chrono.h>

int main() {
    using namespace std::chrono;

    // this one prints 1.00s, as expected
    fmt::print("{:.2}\n", duration<double>(1.0));

    // this one prints 1s but shouldn't compile
    fmt::print("{:.2}\n", duration<int>(1));

    // these both print .2
    fmt::print("{:.2}\n", time_point<system_clock, duration<double>>(duration<double>(1.0)));
    fmt::print("{:.2}\n", time_point<system_clock, duration<int>>(duration<int>(1)));
}
```

If we look at the standard spec, it's clear that using _precision_ with a `duration` is only valid for floating point (https://eel.is/c++draft/time.format#1.sentence-4), so the second call there should be rejected. 

There's some ambiguity in the standard spec right now about what formatting a `time_point<system_clock, duration<int>>` with `{:.2}` actually means (valid or not? probably not?) but either way, printing `.2` in either case is definitely incorrect.

vitaut:
Good catch(es), thanks. As usual, PRs are welcome =).

phprus:
Why is printing ``.2`` incorrect?
I think, in ``chrono-specs``, the string ``.2` is a valid ``literal-char``.

jwakely:
No, the first character of a chrono-spec must be `%`

jwakely:
If you want to write literal characters before the first % put them before the replacement field: `".2{}\n"`

vitaut:
The time point issue is fixed in #3232 (thanks @ShawnZhong) but precision of durations with integral representation is useful and will continue to be supported.

ShawnZhong:
Thanks for clarifying. I'm good with precision for integral durations. 

I have some follow-up comments/questions (which I can open a new issue if needed): 

- I agree that precision with `%S` is useful and should continue to be supported:
  ```cpp
  // %S writes the second as a decimal number
  fmt::print("{:.8%S}\n", 1234ns); // prints 00.000001234
  ```

- I'm not sure what should be the expected behavior of just specifying the precision:
   ```cpp
   fmt::print("{:.8}\n", 1234ns); // prints 1234ns
   ```

- Precision with `%Q` throws `format_error`, which is reasonable
   ```cpp
   // %Q writes the count of ticks of the duration
   fmt::print("{:.8%Q}", 1234ns); // precision not allowed for this argument type
   ```

[...]

PR #3231, opened 2022-12-17
Disallow formatting duration with precision for non-floating-point rep type

ShawnZhong:
#3219


vitaut:
Thanks for the PR but `fmt::format("{:.6%S}", std::chrono::nanoseconds{1234})` looks useful and I think we should continue supporting it.

PR #3239, opened 2022-12-22
doc: adjust chrono_format_spec documentation to match library behavior

jengelh:
References: #3237

### DIFF
--- a/doc/syntax.rst
+++ b/doc/syntax.rst
@@ -304,11 +304,11 @@ The available presentation types for pointers are:
 Chrono Format Specifications
 ============================
 
-Format specifications for chrono duration and time point types as well as
-``std::tm`` have the following syntax:
+Format specifications for chrono time points as well as ``std::tm`` have the
+following syntax:
 
 .. productionlist:: sf
-   chrono_format_spec: [[`fill`]`align`][`width`]["." `precision`][`chrono_specs`]
+   chrono_format_spec: [`chrono_specs`]
    chrono_specs: [`chrono_specs`] `conversion_spec` | `chrono_specs` `literal_char`
    conversion_spec: "%" [`modifier`] `chrono_type`
    literal_char: <a character other than '{', '}' or '%'>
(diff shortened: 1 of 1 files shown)


vitaut (changes requested):
Thanks for the PR. We shouldn't remove the part of the grammar because it is valid for durations. I suggest clarifying that time points only support chrono_specs at the moment in a sentence below the grammar.

jengelh:
Since durations don't do conversion (`{:%d}`), I've changed the syntax for durations to not include conversion_spec anymore.

vitaut:
Durations do support conversion specifiers except calendaric so please revert that change. We already explain the calendaric part in the docs.

vitaut:
@jengelh do you still plan to update this PR per my earlier comment?

jengelh:
I don't know how to dismiss this "Changes requested" box - GH does not notice the new change and does not show a "Done" button or something that I could have sworn existed elsewhere. whatever

vitaut (dismissed):

doc/syntax.rst
```diff
@@ -318,8 +318,14 @@ Format specifications for chrono duration and time point types as well as
               : "q" | "Q" | "r" | "R" | "S" | "t" | "T" | "u" | "U" | "V" |
               : "w" | "W" | "x" | "X" | "y" | "Y" | "z" | "Z" | "%"
 
-Literal chars are copied unchanged to the output. Precision is valid only for
-``std::chrono::duration`` types with a floating-point representation type.
+Literal chars are copied unchanged to the output. For durations, the syntax is:
+
+.. productionlist:: sf
+   chrono_format_spec: [[`fill`]`align`][`width`]["." `precision`][`chrono_specs`]
+   chrono_specs: [`chrono_specs`] `literal_char`
```
vitaut, line 325: This is incorrect because durations support conversion specifiers. Let's correct it by removing this line which will also fix duplicate definition of `chrono_specs`.

vitaut:
This is no longer needed because time points are getting missing format specifier support in #3260 but thanks anyway.

PR #3240, opened 2022-12-22
Improve timezone tests

phprus:
Issues present after commit 115001a3b165483563822645bb93fd9558908c50

1) macOS has broken std::time_put:
```
[ RUN      ] chrono_test.system_clock_time_point
/Users/phprus/Devel/fmt/test/chrono-test.cc:287: Failure
  sys_output
    Which is: "+0500"
  fmt::format(fmt::runtime(fmt_spec), t1)
    Which is: "+0000"
/Users/phprus/Devel/fmt/test/chrono-test.cc:288: Failure
  sys_output
    Which is: "+0500"
  fmt::format(fmt::runtime(fmt_spec), tm)
    Which is: "+0000"
/Users/phprus/Devel/fmt/test/chrono-test.cc:299: Failure
  sys_output
    Which is: "+05:00"
  fmt::format("{:%Ez}", t1)
    Which is: "+00:00"
/Users/phprus/Devel/fmt/test/chrono-test.cc:300: Failure
  sys_output
    Which is: "+05:00"
  fmt::format("{:%Ez}", tm)
    Which is: "+00:00"
/Users/phprus/Devel/fmt/test/chrono-test.cc:302: Failure
  sys_output
    Which is: "+05:00"
  fmt::format("{:%Oz}", t1)
    Which is: "+00:00"
/Users/phprus/Devel/fmt/test/chrono-test.cc:303: Failure
  sys_output
    Which is: "+05:00"
  fmt::format("{:%Oz}", tm)
    Which is: "+00:00"
[  FAILED  ] chrono_test.system_clock_time_point (0 ms)
```

2) Windows has not normal support UTC timezone for ``std::tm``.
We need to use ``std::chrono`` to format time with timezone.

### DIFF
--- a/test/chrono-test.cc
+++ b/test/chrono-test.cc
@@ -264,15 +264,15 @@ TEST(chrono_test, system_clock_time_point) {
       "%OU", "%W",  "%OW", "%V",  "%OV", "%j",  "%d",  "%Od", "%e",
       "%Oe", "%a",  "%A",  "%w",  "%Ow", "%u",  "%Ou", "%H",  "%OH",
       "%I",  "%OI", "%M",  "%OM", "%S",  "%OS", "%x",  "%Ex", "%X",
-      "%EX", "%D",  "%F",  "%R",  "%T",  "%p",  "%z",  "%Z"};
+      "%EX", "%D",  "%F",  "%R",  "%T",  "%p"};
 #ifndef _WIN32
   // Disabled on Windows because these formats are not consistent among
   // platforms.
   spec_list.insert(spec_list.end(), {"%c", "%Ec", "%r"});
 #elif defined(__MINGW32__) && !defined(_UCRT)
   // Only C89 conversion specifiers when using MSVCRT instead of UCRT
-  spec_list = {"%%", "%Y", "%y", "%b", "%B", "%m", "%U", "%W", "%j", "%d", "%a",
-               "%A", "%w", "%H", "%I", "%M", "%S", "%x", "%X", "%p", "%Z"};
+  spec_list = {"%%", "%Y", "%y", "%b", "%B", "%m", "%U", "%W", "%j", "%d",
+               "%a", "%A", "%w", "%H", "%I", "%M", "%S", "%x", "%X", "%p"};
 #endif
   spec_list.push_back("%Y-%m-%d %H:%M:%S");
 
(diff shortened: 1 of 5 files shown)


phprus:
Demo:
See errors in CI for branch https://github.com/phprus/fmt/tree/chrono-tz-test-demo

phprus:
Rebased

vitaut (commented):
Thanks for the PR! I noticed the timezone issues too but didn't have time to look into them myself.

What is the purpose of setting the timezone in CI and can we avoid it to keep configs simpler?

test/chrono-test.cc
```diff
@@ -287,19 +287,48 @@ TEST(chrono_test, system_clock_time_point) {
     EXPECT_EQ(sys_output, fmt::format(fmt::runtime(fmt_spec), tm));
   }
 
-  if (std::find(spec_list.cbegin(), spec_list.cend(), "%z") !=
-      spec_list.cend()) {
+#if defined(__MINGW32__) && !defined(_UCRT)
+  spec_list = {"%Z"};
+#else
+  spec_list = {"%z", "%Z"};
+#endif
```
vitaut, line 294: I suggest moving this code to a separate function and reuse here and elsewhere in chrono-test.

vitaut: A comment clarifying why these are handled separately would be nice too.

phprus:
@vitaut 
By default GitHub Actions uses UTC and possible time zone related issues are not tested in CI.
I think this change makes sense for testing future versions of fmt.

test/chrono-test.cc
phprus: For loops are not equivalent.

First ``for`` loop tests ``std::chrono::time_point`` (``std::chrono::system_clock::now()``) and ``std::tm`` (in UTC).
Second ``for`` loop tests timezone formatters only for ``std::tm`` (in localtime).

I think we should not do this, since timezone support is a weak point of fmt and the standard library.

phprus: @vitaut 
Comments added.

vitaut:
> By default GitHub Actions uses UTC and possible time zone related issues are not tested in CI.
> I think this change makes sense for testing future versions of fmt.

Sounds reasonable but is it possible to move changing the timezone to code instead of CI?

phprus:
@vitaut 
> Sounds reasonable but is it possible to move changing the timezone to code instead of CI?

Unfortunately, I am not aware of a cross-platform and universal API for changing the timezone settings. So I changed it at the CI settings level. I think it's more reliable.

phprus:
I understand that this is not are best solution, but I have no other universal solution.
Chrono supports timezones only in C++20 and only in MSVC.

test/chrono-test.cc
vitaut: I didn't mean merging the for loops, just moving the selected block, namely

```c++
#if defined(__MINGW32__) && !defined(_UCRT)
  spec_list = {"%Z"};
#else
  spec_list = {"%z", "%Z"};
#endif
```

into a separate function (say, `get_timezone_specs`) to keep the system-specific conditional compilation in one place.

Other than that, all looks good.

phprus: @vitaut
I don't understand how this will simplify the code.
The condition ``defined(__MINGW32__) && !defined(_UCRT)`` is used 4 times in this test. Partially with precondition ``ifndef _WIN32``.

vitaut: It is not perfect but would be slightly cleaner because we could reuse this function in two of these places. Another option is to move the `defined(__MINGW32__) && !defined(_UCRT)` condition into one place, define our own macro (e.g. `FMT_HAS_C99_STRFTIME`) and reuse that in all places.

phprus: Done.
I added macro ``FMT_HAS_C99_STRFTIME``.

vitaut:
Thank you!

Issue #3241, opened 2022-12-22
fmt::to_string/compilation doesn't work with a null string_view


DenisYaroshevskiy:
At the moment you can't fmt a default constructed string_view: https://godbolt.org/z/ddzYqYE99

vitaut:
Specifically, this is about `fmt::to_string`.

DenisYaroshevskiy:
I don't think so https://godbolt.org/z/Y6e4oPdPx

DenisYaroshevskiy:
FYI: this is how libc++ does this: https://github.com/llvm/llvm-project/blob/ce5ef7d1d5dc632a1ff35110821a44978e004e5f/libcxx/include/__algorithm/copy.h#L60

PR #3244, opened 2022-12-23
Fix for issue #3241

phprus:
Possible fix for issue #3241


vitaut:
Thank you for the fix!

PR #3254, opened 2022-12-29
Fix build with MSVC C++20 modules

Orvid:
When using fmt with C++20 modules under MSVC, it can end up requiring certain things to have storage that would not otherwise have needed to. Since I didn't see anything that was already doing detection for `inline constexpr` variable support, I've just moved the entire thing into the only function where it's used. I've also swapped the function from `constexpr` to `FMT_CONSTEXPR` since it's no longer a simple C++11 constexpr function.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -395,9 +395,9 @@ FMT_CONSTEXPR auto to_unsigned(Int value) ->
   return static_cast<typename std::make_unsigned<Int>::type>(value);
 }
 
-FMT_MSC_WARNING(suppress : 4566) constexpr unsigned char section[] = "\u00A7";
+FMT_CONSTEXPR auto is_utf8() -> bool {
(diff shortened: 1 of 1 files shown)


include/fmt/core.h
```diff
@@ -395,9 +395,9 @@ FMT_CONSTEXPR auto to_unsigned(Int value) ->
   return static_cast<typename std::make_unsigned<Int>::type>(value);
 }
 
-FMT_MSC_WARNING(suppress : 4566) constexpr unsigned char section[] = "\u00A7";
+FMT_CONSTEXPR auto is_utf8() -> bool {
```
vitaut, line 398: You need `inline` here because `FMT_CONSTEXPR` expands to nothing on compilers that don't support relaxed constexpr.

vitaut:
Thanks

Issue #428, opened 2016-11-23
Why is `println` missing?


jiri:
`fmt::println` seems like a natural addition to the API. Is there a reason why it's missing?

foonathan:
What's wrong with `fmt::print("...\n")`?

vitaut:
I think that `println` is of limited use because you can always specify the newline explicitly as @foonathan showed. But fmt provides sufficient infrastructure to easily and efficiently build this and similar methods without making them the part of the core library. In particular println can be written in five lines of code:

```c++
template <typename... Args>
void print(fmt::CStringRef format, const Args&... args) {
  fmt::print(format, args...);
  std::putc('\n', stdout);
}
```

jiri:
Sure, and you could always use `stdout` explicitly, but it's the default file for convenience. One could as well argue "What's wrong with `fmt::print(stdout, "...\n")`.

Having to include an explicit newline everywhere is noisy and prone to forgetting them. There's a reason why many languages include `println` in their core libraries ;) Moreover, I thought that the point of a formatting library is eliminating the reason for all those non-newline terminated prints, since I could just format them into the strings, but nevertheless, you seem to be pretty set on not including it, so feel free to keep this closed.

Cheers.

vitaut:
The difference from `fmt::print` overload taking `stdout` is that the latter saves a bit of typing

```c++
fmt::print("foo");
fmt::print(stdout, "foo");
```
and directly corresponds to `std::printf`. But you are right, it's somewhat redundant too.

jiri:
It's true that `println` would save exactly zero characters. What's the approach to printing a foreign string? E.g., is something like

```cpp
void foo(std::string s) {
    fmt::print(s);
}
```

considered bad form or bad api usage? `println(s)` can make the intention clearer than `print("{}\n", s)` or god forbid `print(s + "\n")`.

vitaut:
> is something like ... considered bad form or bad api usage? 

Yes. As with `printf`, it's better to write

```c++
void foo(std::string s) {
    fmt::print("{}", s);
}
```
Otherwise `s` will be parsed as a format string and will need to be escaped.

slonm:
>What's wrong with `fmt::print("...\n")`?

For Windows it's: `fmt::print("...\r\n")`
Howto to hide this difference?
I'm vote for API extension

vitaut:
You can use `fmt::print("...\n")` on Windows as well.

foonathan:
[...]

PR #3267, opened 2023-01-11
Implement `println`

ShawnZhong:
Close https://github.com/fmtlib/fmt/issues/428 (which is already closed)

https://eel.is/c++draft/iostream.format#ostream.formatted.print-2
https://eel.is/c++draft/iostream.format#print.fun-3


vitaut:
Thank you

zencatalyst:
Thanks a ton! Recently I was thinking about why `println` was not part of the {fmt} but here we have it now!

Issue #3262, opened 2023-01-06
Try a new formatting algorithm for float/double with a small given precision


jk-jeon:
So this is a continuation from the discussion done here: https://github.com/fmtlib/fmt/pull/2750.

I'm trying to implement the plan discussed but currently a little bit confused.

1. To my current understanding, it is `format_float` which produces actual digits for the input floating-point number, and it does not print the decimal dot, and it does not attempt to choose between fixed / scientific form. It just print digits to a buffer, and later on in another function up in the call stack it is adjusted according to the precise spec. Is that the right understanding?
2. It seems that the parameter `precision` of `format_float` is different from the precision given by the input. What exactly is that number? Why, for example, we decide shortest roundtrip output if `precision` is negative? Ultimately, what we need is the number of digits to print, counted from the first nonzero digit. How can I get that number?
3. The case when the input is zero, either positive or negative, regardless of the formatting specs, returns early at the beginning of `format_float`, right? So can I assume that the input is not zero if it passed over the branch `if (value <= 0)`?

vitaut:
> To my current understanding, it is format_float which produces actual digits for the input floating-point number, and it does not print the decimal dot, and it does not attempt to choose between fixed / scientific form. It just print digits to a buffer, and later on in another function up in the call stack it is adjusted according to the precise spec. Is that the right understanding?

Yes.

> It seems that the parameter precision of format_float is different from the precision given by the input. What exactly is that number?

It is the number of digits to generate or -1 for the shortest number of digits that round trip. We need to adjust it from the user-specified precision for a number of reasons:

*  The exponent/scientific format (`e`) count precision differently from the general format (`g`): the former counts the number of digits after the decimal point while the latter counts the total number of digits (https://godbolt.org/z/cYas9ds79):

   ```c++
   fmt::print("{:g}\n", 1.23456789); // 1.23457
   fmt::print("{:e}\n", 1.23456789); // 1.234568e+00
   ```

* We need to generate at least one digit even if the user-specified precision is zero:

  ```c++
  fmt::print("{:.0}\n", 1.23456789); // 1
  ```

> Why, for example, we decide shortest roundtrip output if precision is negative?

-1 is just a special value that denotes the shortest roundtrip format to distinguish it from nonnegative values that specify "fixed" precision.

> Ultimately, what we need is the number of digits to print, counted from the first nonzero digit. How can I get that number?
  
If `precision` passed to `format_float` is nonnegative, it gives the number of digits to print.

[...]

PR #3269, opened 2023-01-12
Implement a new formatting algorithm for small given precision

jk-jeon:
Implement the algorithm discussed in https://github.com/fmtlib/fmt/issues/3262 and https://github.com/fmtlib/fmt/pull/2750.

- Didn't touch the `constexpr`-route; it still goes into the Grisu branch. We can eliminate this branch (and other Grisu-related stuffs) completely by `constexpr`-ifying several more functions after this PR.
- Moved several stuffs from `format-inl.h` into `format.h` to satisfy the compiler.
- Appended several entries into the Dragonbox table, 15 into the full table, and 1 into the compressed table. Modified `float_info<double>::max_k` accordingly, and adjusted the corresponding test case from `format-impl-test`.
- Added `uint64_t` version of `countl_zero`.
- The implementation of the algorithm is in the added `else` branch of `format_float`.
- Currently all tests seem to be passing, but I suggest to add some more test cases. I can do it in a separate PR if you think it's needed.

Contrary to what we have worried, the performance even seems a bit better than the original one. Here is a benchmark result I've done:

image

- **Red**: after this PR.
- **Green**: after this PR, with `#define FMT_USE_FULL_CACHE_DRAGONBOX 1`.
- **Blue**: original implementation.

Probably the main reason of this boost is that now digits are generated in pair rather than individually, so the number of multiplications in digit generation is halved (or maybe less than that). To be fair, we could do the same thing for the original implementation and maybe that will give an even better performance, though the rounding logic then might become a bit more complicated.

Also, it's worth mentioning that the new algorithm just gives up doing anything for the 19-digits case while the original implementation occasionally succeeds with Grisu for this case. That's probably the reason why we have the regression for this case.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -1125,7 +1125,7 @@ template <> struct cache_accessor<double> {
   }
 };
 
-inline uint128_fallback get_cached_power(int k) noexcept {
+FMT_FUNC uint128_fallback get_cached_power(int k) noexcept {
   return cache_accessor<double>::get_cached_power(k);
 }
 
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -174,58 +174,10 @@ FMT_CONSTEXPR inline uint64_t rotr(uint64_t n, uint32_t r) noexcept {
   return (n >> r) | (n << (64 - r));
 }
 
-// Computes 128-bit result of multiplication of two 64-bit unsigned integers.
-inline uint128_fallback umul128(uint64_t x, uint64_t y) noexcept {
-#if FMT_USE_INT128
-  auto p = static_cast<uint128_opt>(x) * static_cast<uint128_opt>(y);
-  return {static_cast<uint64_t>(p >> 64), static_cast<uint64_t>(p)};
-#elif defined(_MSC_VER) && defined(_M_X64)
-  auto result = uint128_fallback();
-  result.lo_ = _umul128(x, y, &result.hi_);
-  return result;
-#else
-  const uint64_t mask = static_cast<uint64_t>(max_value<uint32_t>());
 
-  uint64_t a = x >> 32;
-  uint64_t b = x & mask;
-  uint64_t c = y >> 32;
-  uint64_t d = y & mask;
-
-  uint64_t ac = a * c;
-  uint64_t bc = b * c;
-  uint64_t ad = a * d;
-  uint64_t bd = b * d;
-
-  uint64_t intermediate = (bd >> 32) + (ad & mask) + (bc & mask);
-
-  return {ac + (intermediate >> 32) + (ad >> 32) + (bc >> 32),
-          (intermediate << 32) + (bd & mask)};
-#endif
-}
 
 // Implementation of Dragonbox algorithm: https://github.com/jk-jeon/dragonbox.
 namespace dragonbox {
-// Computes upper 64 bits of multiplication of two 64-bit unsigned integers.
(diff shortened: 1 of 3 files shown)


include/fmt/format-inl.h
```diff
@@ -1169,6 +1125,10 @@ template <> struct cache_accessor<double> {
   }
 };
 
+FMT_FUNC uint128_fallback get_cached_power(int k) noexcept {
+  return cache_accessor<double>::get_cached_power(k);
+}
```
vitaut, line 1130: Do we need this function? It seems to be used only in one place and can be replaced with calling `cache_accessor<double>::get_cached_power` directly.

include/fmt/format.h
```diff
@@ -495,6 +495,16 @@ FMT_CONSTEXPR20 inline auto countl_zero(uint32_t n) -> int {
...
 }
 
+FMT_CONSTEXPR20 inline auto countl_zero(uint64_t n) -> int {
+#ifdef FMT_BUILTIN_CLZLL
+  if (!is_constant_evaluated()) return FMT_BUILTIN_CLZLL(n);
+#endif
+  int lz = 0;
+  constexpr uint64_t msb_mask = 1ull << (num_bits<uint64_t>() - 1);
+  for (; (n & msb_mask) == 0; n <<= 1) lz++;
+  return lz;
```
vitaut, line 505: I would suggest moving this logic into a separate function (e.g. `countl_zero_fallback`) and reusing here and in 32-bit overload.

vitaut (commented):
Thanks for the PR!

include/fmt/format-inl.h
jk-jeon: The reason I did that way was I didn't want to move the entire `cache_accessor<double>` class to `format.h`. I guess we will anyway need to move the the table into `format.h` to `constexpr`-ify the current implementation, but I considered that as a job of a later work. Or is there a simpler solution?

include/fmt/format.h
jk-jeon: You mean having a template `countl_zero_fallback` that is called from both of the overloads?

jk-jeon:
By the way, I think in previous discussions I said we will be able to reliably generate 18 digits with this new algorithm, but that's not the case. The number of digits that can be reliably generated without Dragon4 fallback is 17, and for the 18 digits case it depends on whether the multiplication by 10^k gives us 18 or 19 digits number. We have to have a one-digit margin for the rounding, so if we initially got a 19-digits number then we can print 18 digits out of it, and if we got an 18-digit number then we can print 17 digits out of it.

Actually, since we check (`has_more_segments`) if there are further digits other than those contained in the number we got from said multiplication, so if it turns out that there is no further digit (i.e., if `has_more_segments == false`), then we can print the last digit we spared for the rounding. But I considered it more harm than good. I think 17 digits should be enough for all the common use cases.

include/fmt/format.h
vitaut: Yes

include/fmt/format-inl.h
vitaut: I see. Let's keep `get_cached_power` for now.

include/fmt/format.h
vitaut: Anyway, it can be done separately so I'll merge this as is.

vitaut:
>  I think 17 digits should be enough for all the common use cases.

I agree. Precision of 17 can be used for roundtrip with `g` and `e`, larger values are pretty uncommon.

vitaut:
Merged, thanks!

PR #3275, opened 2023-01-16
Fix spec "{:}" for time point

ShawnZhong:
My PR https://github.com/fmtlib/fmt/pull/3260 results in a `format_error` when formatting a timepoint with spec `{:}` due to a typo in parsing. This PR fixes the issue.


vitaut:
Thanks for the quick fix.

Issue #3247, opened 2022-12-24
clang, exception raised or not depending on compiler options


m5k8:
Isolated problem - `test.cc`:
        
        #include <fmt/format.h>
        #include <iostream>

        int main()
        {
	        char buf[100];
	        char* iterator = buf;
	        auto ret = fmt::format_to_n(iterator, sizeof(buf), "Test {}", "OK");
		std::cout << ret.size << std::endl;
        
	        return 0;
        }

clang version 14.0.6. Then running compiled executable:
`clang++ -I fmt/include -std=c++20                      test.cc fmt/src/format.cc` - OK
`clang++ -I fmt/include -std=c++20 -fomit-frame-pointer test.cc fmt/src/format.cc` - OK
`clang++ -I fmt/include -std=c++2b                      test.cc fmt/src/format.cc` - OK
`clang++ -I fmt/include -std=c++2b -fomit-frame-pointer test.cc fmt/src/format.cc` - runtime exception:

    terminate called after throwing an instance of 'fmt::v9::format_error'
      what():  argument not found

`clang++ -I fmt/include -std=c++2b -fomit-frame-pointer -O2 test.cc fmt/src/format.cc` - OK again.

So the outcome depends on selected C++ standard and compiler options. c++2b and option for omitting frame pointer fails, but adding optimization O2 to this makes it work again.
I'm not competent enough to dig into this problem, not sure whether it's fmt library issue or compiler issue.
gcc is fine with all these variants. MSVC also (albeit options do not apply).

Please advise.

vitaut:
I wasn't able to repro this on clang++ 14: https://godbolt.org/z/99v59KEhb. Likely a compiler issue on the specific compiler version.

m5k8:
More tests: problem happens on current Arch Linux, when compiled with clang 14.0.6, using libstdc++.
When using -stdlib=libc++ it's fine. So switching C++ library solves the issue.
When compiled on Debian (still using clang 14.0.6) , it's fine with both libraries. The only difference is that Arch has libstdc++ 6.0.30, Debian has 6.0.25.

m5k8:
Does godbolt compile the library using the same options as main source file?
Because I've also tried using system-wide installed fmt library, version 9.1.0-2 - and then it works properly.

vitaut:
It doesn't but we can force compilation with the same options using `FMT_HEADER_ONLY` which also doesn't reproduce the issue: https://godbolt.org/z/Mvv3zfPWd

[...]

PR #3281, opened 2023-01-21
Workaround for incompatibility between libstdc++ consteval-based std::is_constant_evaluated() implementation and clang-14

phprus:
Fix for issue #3247

cc @m5k8

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -329,7 +329,15 @@ template <typename... T> FMT_CONSTEXPR void ignore_unused(const T&...) {}
 
 constexpr FMT_INLINE auto is_constant_evaluated(
     bool default_value = false) noexcept -> bool {
-#ifdef __cpp_lib_is_constant_evaluated
+// Workaround for incompatibility between libstdc++ consteval-based
+// std::is_constant_evaluated() implementation and clang-14.
+// https://github.com/fmtlib/fmt/issues/3247
+#if FMT_CPLUSPLUS >= 202002L && defined(_GLIBCXX_RELEASE) && \
+    _GLIBCXX_RELEASE >= 12 &&                                \
+    (FMT_CLANG_VERSION >= 1400 && FMT_CLANG_VERSION < 1500)
+  ignore_unused(default_value);
+  return __builtin_is_constant_evaluated();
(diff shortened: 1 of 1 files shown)


include/fmt/core.h
```diff
@@ -329,7 +329,14 @@ template <typename... T> FMT_CONSTEXPR void ignore_unused(const T&...) {}
 
 constexpr FMT_INLINE auto is_constant_evaluated(
     bool default_value = false) noexcept -> bool {
-#ifdef __cpp_lib_is_constant_evaluated
+// Workaround for incompatible between libstdc++ consteval based
```
vitaut, line 332: incompatible -> incompatibility
consteval based -> consteval-based

include/fmt/core.h
```diff
@@ -329,7 +329,14 @@ template <typename... T> FMT_CONSTEXPR void ignore_unused(const T&...) {}
...
+// std::is_constant_evaluated implementation and clang-14+.
+// https://github.com/fmtlib/fmt/issues/3247
+#if FMT_CPLUSPLUS >= 202002L && defined(_GLIBCXX_RELEASE) && \
+    _GLIBCXX_RELEASE >= 12 && FMT_CLANG_VERSION >= 14
```
vitaut, line 336: Should this be `FMT_CLANG_VERSION == 14` per https://github.com/fmtlib/fmt/issues/3247#issuecomment-1399299689?

vitaut (commented):
Thanks for the PR.

include/fmt/core.h
phprus: Should be
```
(FMT_CLANG_VERSION >= 1400 && FMT_CLANG_VERSION < 1500)
```
This is my mistake, I forgot about the minor version part (00):

https://github.com/fmtlib/fmt/blob/f89cd276f7dead38f11cebc73d1e91a1b1b38124/include/fmt/core.h#L23

Since bug is fixed in clang-15, I added a second condition.

include/fmt/core.h
phprus: Fixed

vitaut:
Thank you!

Issue #3287, opened 2023-01-25
Visual Studio 2022: fmt/format-inl.h(1145,60): warning C4310: cast truncates constant value


chronoxor:
Visual Studio 2022 generates the following warning in /W4 pedantic mode:
```
fmt/format-inl.h(1145,60): warning C4310: cast truncates constant value
```

On the following code fragment:
```
const uint32_t mod_inv_25 = uint32_t(uint64_t(mod_inv_5) * mod_inv_5);
```

chronoxor:
Thanks! Yes, I usually try to fix pedantic warnings by myself and provide a PR in order not to disturb too much.

PR #3288, opened 2023-01-25
Visual Studio 2022: fmt/format-inl.h(1145,60): warning C4310: cast truncates constant value #3287

chronoxor:


vitaut:
Thanks for the PR. MSVC is such a mess =).

PR #3290, opened 2023-01-26
Allowing formatting non-copyable ranges.

### LINKED ISSUE #3286: Can't format move-only ranges
[Example](https://godbolt.org/z/cE1P9YdG6):

```cpp
#include <concepts>
#include <vector>
#include <fmt/ranges.h>

template <bool Copyable>
struct Vector {
    std::vector<int> v;

    Vector(std::initializer_list<int> elems) : v(elems) { }

    Vector(Vector&&) = default;
    Vector& operator=(Vector&&) = default;

    Vector(Vector const&) requires Copyable = default;
    Vector& operator=(Vector const&) requires Copyable = default;

    auto begin() { return v.begin(); }
    auto end() { return v.end(); }
};

static_assert(std::movable<Vector<false>>);
static_assert(std::movable<Vector<true>>);
static_assert(!std::copyable<Vector<false>>);
static_assert(std::copyable<Vector<true>>);

int main() {
    fmt::print("{}\n", Vector<true>{1, 2, 3});  // ok [1, 2, 3]
    fmt::print("{}\n", Vector<false>{1, 2, 3}); // error
}
```

This is because the range check for non-const ranges requires copyability and should probably just be removed (h/t @timsong-cpp):

https://github.com/fmtlib/fmt/blob/a2c05a10ec875712bfee739a05ed99fc3915f4e9/include/fmt/ranges.h#L154-L159

brevzin:
Fixes #3286

### DIFF
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -155,6 +155,8 @@ template <typename T>
 struct has_mutable_begin_end<
     T, void_t<decltype(detail::range_begin(std::declval<T>())),
               decltype(detail::range_end(std::declval<T>())),
+              // the extra int here is because older versions of MSVC don't
+              // SFINAE properly unless there are distinct types
               int>>
     : std::true_type {};
 
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -154,7 +154,8 @@ struct has_const_begin_end<
 template <typename T>
 struct has_mutable_begin_end<
     T, void_t<decltype(detail::range_begin(std::declval<T>())),
-              decltype(detail::range_end(std::declval<T>()))>>
+              decltype(detail::range_end(std::declval<T>())),
+              int>>
     : std::true_type {};
 
 template <typename T>
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -154,8 +154,7 @@ struct has_const_begin_end<
 template <typename T>
 struct has_mutable_begin_end<
     T, void_t<decltype(detail::range_begin(std::declval<T>())),
-              decltype(detail::range_end(std::declval<T>())),
-              enable_if_t<std::is_copy_constructible<T>::value>>>
+              decltype(detail::range_end(std::declval<T>()))>>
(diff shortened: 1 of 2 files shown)


brevzin:
Some windows build is complaining about `noncopyable_range<int> &` being non-formattable, but I'm not sure why. Do you know what version "(windows-2019, x64, v140, 14)" corresponds to on compiler-explorer? Could try to repro something there.

mwinterb:
It's not an exact match, but the general "era" matches "x64 msvc v19.0 (WINE)".

brevzin:
wat: https://godbolt.org/z/b9TYxqh6s

```cpp
namespace my {
namespace detail {
    template <typename... > using void_t = void;
}
}

template <typename T>
struct has_mutable_begin_end<
    T, 
    #if 1
    fmt::detail::void_t<
    #else
    my::detail::void_t<
    #endif
        decltype(fmt::detail::range_begin(std::declval<T>()))
        , decltype(fmt::detail::range_end(std::declval<T>()))
        #if 0
        , int
        #endif
        >>
    : std::true_type {};

static_assert(has_mutable_begin_end<noncopyable_range>::value, "");
```

If I use `my::detail::void_t`, this works. 

If I use `fmt::detail::void_t`, which is ... the same thing as `my::detail::void_t`, it fails. Unless I add the extra `int` at the end, in which case it works?

include/fmt/ranges.h
```diff
@@ -155,7 +155,7 @@ template <typename T>
 struct has_mutable_begin_end<
     T, void_t<decltype(detail::range_begin(std::declval<T>())),
               decltype(detail::range_end(std::declval<T>())),
-              enable_if_t<std::is_copy_constructible<T>::value>>>
+              int>>
```
vitaut, line 158: Please add a comment briefly explaining why we have int here.

vitaut (commented):
Thanks for the PR.

include/fmt/ranges.h
brevzin: Sure.

PR #3293, opened 2023-02-06
change sopen_s to wsopen_s (fmtlib#3234)

### LINKED ISSUE #3234: _wfopen support
On Windows it is important to use UTF-16 file paths for opening files (at least until recent versions that allow using UTF-8 locale for WinAPI).

std::fstream has overload for std::wstring
for std::FILE* there is Windows-specific "_wfopen" method as a replacement for standard std::fopen.

Is it possible to provide a way to use fmt::output_file with UTF-16 encoded filepath?

Fros1er:
Fix #3234.

### DIFF
--- a/src/os.cc
+++ b/src/os.cc
@@ -213,22 +213,25 @@ int buffered_file::descriptor() const {
 }
 
 #if FMT_USE_FCNTL
-file::file(cstring_view path, int oflag) {
 #  ifdef _WIN32
-  using mode_t = int;
+using mode_t = int;
 #  endif
-  constexpr mode_t mode =
-      S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH;
+constexpr mode_t default_open_mode =
+    S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH;
+
+file::file(cstring_view path, int oflag) {
 #  if defined(_WIN32) && !defined(__MINGW32__)
   fd_ = -1;
   auto converted = detail::utf8_to_utf16(string_view(path.c_str()));
-  FMT_POSIX_CALL(wsopen_s(&fd_, converted.c_str(), oflag, _SH_DENYNO, mode));
+  auto err =
+      _wsopen_s(&fd_, converted.c_str(), oflag, _SH_DENYNO, default_open_mode);
 #  else
-  FMT_RETRY(fd_, FMT_POSIX_CALL(open(path.c_str(), oflag, mode)));
+  FMT_RETRY(fd_, FMT_POSIX_CALL(open(path.c_str(), oflag, default_open_mode)));
+  auto err = errno;
 #  endif
   if (fd_ == -1)
     FMT_THROW(
-        system_error(errno, FMT_STRING("cannot open file {}"), path.c_str()));
+        system_error(err, FMT_STRING("cannot open file {}"), path.c_str()));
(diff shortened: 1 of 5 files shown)


test/posix-mock.h
```diff
@@ -39,6 +39,8 @@ typedef unsigned size_t;
 typedef int ssize_t;
 errno_t sopen_s(int* pfh, const char* filename, int oflag, int shflag,
                 int pmode);
+errno_t wsopen_s(int* pfh, const wchar_t* filename, int oflag, int shflag,
+                int pmode);
```
vitaut, line 43: I don't think this is needed because there is no corresponding test in posix-mock-test. You'll need to remove `FMT_POSIX_CALL` around the call to `wsopen_s` as well.

src/os.cc
```diff
@@ -353,6 +354,20 @@ buffered_file file::fdopen(const char* mode) {
   return bf;
 }
 
+#  if defined(_WIN32) && !defined(__MINGW32__)
+file file::open_windows_file(wcstring_view path, int oflag) {
+  int fd_ = -1;
+  using mode_t = int;
+  constexpr mode_t mode =
+      S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH;
+  FMT_POSIX_CALL(wsopen_s(&fd_, path.c_str(), oflag, _SH_DENYNO, mode));
```
vitaut, line 363: Looks like wsopen_s returns an error code so I suggest using it instead of global errno.

src/os.cc
```diff
@@ -353,6 +354,20 @@ buffered_file file::fdopen(const char* mode) {
   return bf;
 }
 
+#  if defined(_WIN32) && !defined(__MINGW32__)
+file file::open_windows_file(wcstring_view path, int oflag) {
+  int fd_ = -1;
+  using mode_t = int;
+  constexpr mode_t mode =
+      S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH;
```
vitaut, line 362: Let's move to a constant, e.g. `default_open_mode` and reuse here and above to make sure modes are consistent.

vitaut (changes requested):
Thanks for the PR. Overall looks good, just a few minor comments inline.

Fros1er:
It seems windows doesn't return EINTR on _wsopen_s, so I removed `open_retry` in posix mock test.
And sorry for those force pushes.

src/os.cc
```diff
@@ -213,21 +213,25 @@ int buffered_file::descriptor() const {
...
+constexpr mode_t default_open_mode =
+    S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH;
+
+file::file(cstring_view path, int oflag) {
 #  if defined(_WIN32) && !defined(__MINGW32__)
   fd_ = -1;
-  FMT_POSIX_CALL(sopen_s(&fd_, path.c_str(), oflag, _SH_DENYNO, mode));
+  auto converted = detail::utf8_to_utf16(string_view(path.c_str()));
+  auto err =
+      _wsopen_s(&fd_, converted.c_str(), oflag, _SH_DENYNO, default_open_mode);
```
vitaut, line 227: I think we could avoid a bit of duplication by calling `open_windows_file` here and moving its result into `*this`.

vitaut (approved):

vitaut:
Thank you

PR #3294, opened 2023-02-07
Fix precision with non-ascii characters

### LINKED ISSUE #3284: Extra characters printed when formatting invalid UTF-8 with precision
Formatting using string precision fails when there are non-ascii characters.

Here is a code to reproduce (first line works as expected, second doesn't as there is an extra "diamond-shaped" character printed):
```
#include <fmt/format.h>
int main()
{
    fmt::print(">>{:.6}<<\n", "1234567");
    fmt::print(">>{:.6}<<\n", "123456\xad");
    return 0;
}
```

https://godbolt.org/z/zzY4Y4Pec

ShawnZhong:
Fix #3284

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -773,7 +773,8 @@ inline auto code_point_index(string_view s, size_t n) -> size_t {
   for (size_t i = 0, size = s.size(); i != size; ++i) {
     if ((data[i] & 0xc0) != 0x80 && ++num_code_points > n) return i;
   }
-  return s.size();
+  size_t size = s.size();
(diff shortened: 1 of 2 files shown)


vitaut (changes requested):
Thanks for the PR but unfortunately it won't work with multibyte code points, e.g.

```
  fmt::print(">>{:.6}<<\n", "123ж56\xad");
```
It should print `>>123ж56<<` but would print `>>123ж5<<` with this change.

The correct fix is to handle invalid bytes as having width 1.

PR #3296, opened 2023-02-07
[msvc] fix warning about non-inline variable

russelltg:
With MSVC 17.4.4, I'm getting a warning originating from fmt headers:

```
fmt\core.h(2909,15): warning C5260: the constant variable 'fmt::v9::detail::invalid_arg_index' has internal linkage in an included header file context, but external linkage in imported header unit context; consider declaring it 'inline' as well if it will be shared across translation units, or 'static' to express intent to use it local to this translation unit
fmt\core.h(2909,15): message : to simplify migration, consider the temporary use of /Wv:18 flag with the version of the compiler with which you used to build without warnings
fmt\core.h(413,58): warning C5260: the constant variable 'fmt::v9::detail::micro' has internal linkage in an included header file context, but external linkage in imported header unit context; consider declaring it 'inline' as well if it will be shared across translation units, or 'static' to express intent to use it local to this translation unit
fmt\core.h(413,58): message : to simplify migration, consider the temporary use of /Wv:18 flag with the version of the compiler with which you used to build without warnings
fmt\format.h(639,20): warning C5260: the constant variable 'fmt::v9::detail::invalid_code_point' has internal linkage in an included header file context, but external linkage in imported header unit context; consider declaring it 'inline' as well if it will be shared across translation units, or 'static' to express intent to use it local to this translation unit
fmt\format.h(639,20): message : to simplify migration, consider the temporary use of /Wv:18 flag with the version of the compiler with which you used to build without warnings
```

It seems to me `inline` is the proper linkage specifier here, as the definition won't vary from compilation unit to compilation unit. 

However, since it's only a C++17 feature, I've only enabled it when the feature is detected.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -281,6 +281,12 @@
 #  endif
 #endif
 
+#if defined __cpp_inline_variables && __cpp_inline_variables >=  201606L
+#    define FMT_INLINE_VARIABLE inline
+#else
+#    define FMT_INLINE_VARIABLE
+#endif
+
 // Enable minimal optimizations for more compact code in debug mode.
 FMT_GCC_PRAGMA("GCC push_options")
 #if !defined(__OPTIMIZE__) && !defined(__NVCOMPILER) && !defined(__LCC__)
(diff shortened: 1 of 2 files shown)


russelltg:
I tried to run clang-format, but it touched lines that I didn't touch. What is the suggested clang-format version so I can make sure I'm formatting the code correctly?

vitaut:
Thanks for the PR! Merged with a minor tweak in https://github.com/fmtlib/fmt/commit/6e6eb63770a8f69bba48d079fb0f43f036d6b543. Most of the code is formatted with clang-format 12.

PR #3259, opened 2023-01-01
Fix errors setting of FMT_USE_FLOAT128

Roman-Koshelev:
Target=x86_64-linux-gnu
```
ld.lld: error: undefined symbol: __extenddftf2
>>> referenced by format-test.cc:89 (.../contrib/libs/fmt/test/format-test.cc:89)
>>>               /home/romankoshelev/.ya/build/build_root/zb02/000035/contrib/libs/fmt/test/format-test/__/format-test.cc.o:(void check_isfinite<__float128>())
>>> referenced by format-test.cc:94 (.../contrib/libs/fmt/test/format-test.cc:94)
>>>               /home/romankoshelev/.ya/build/build_root/zb02/000035/contrib/libs/fmt/test/format-test/__/format-test.cc.o:(void check_isfinite<__float128>())
>>> referenced by format-test.cc:95 (.../contrib/libs/fmt/test/format-test.cc:95)
>>>               /home/romankoshelev/.ya/build/build_root/zb02/000035/contrib/libs/fmt/test/format-test/__/format-test.cc.o:(void check_isfinite<__float128>())
>>> referenced 8 more times

ld.lld: error: undefined symbol: __eqtf2
>>> referenced by format.h:2491 (.../contrib/libs/fmt/include/fmt/format.h:2491)
>>>               /home/romankoshelev/.ya/build/build_root/zb02/000035/contrib/libs/fmt/test/format-test/__/format-test.cc.o:(bool fmt::v9::detail::isfinite<__float128, 0>(__float128))

ld.lld: error: undefined symbol: __netf2
>>> referenced by format.h:2491 (.../contrib/libs/fmt/include/fmt/format.h:2491)
>>>               /home/romankoshelev/.ya/build/build_root/zb02/000035/contrib/libs/fmt/test/format-test/__/format-test.cc.o:(bool fmt::v9::detail::isfinite<__float128, 0>(__float128))

ld.lld: error: undefined symbol: __unordtf2
>>> referenced by format.h:2469 (.../contrib/libs/fmt/include/fmt/format.h:2469)
>>>               /home/romankoshelev/.ya/build/build_root/zb02/000035/contrib/libs/fmt/test/format-test/__/format-test.cc.o:(bool fmt::v9::detail::isnan<__float128>(__float128))

ld.lld: error: undefined symbol: __trunctfdf2
>>> referenced by format.h:2504 (.../contrib/libs/fmt/include/fmt/format.h:2504)
[...]

PR #3297, opened 2023-02-08
Disable some tests for a issue of clang on OpenBSD

ChihHao-Su:
This is a workaround of https://github.com/llvm/llvm-project/issues/56694.

It takes a long time for bug fix of that issue to actually take effect inside OpenBSD. So I done that in order to make fmt avaliable for OpenBSD users asap.


phprus:
@ChihHao-Su
Please check the solution from PR #3259

ChihHao-Su:
That works for me, it should be the right way to solve the issue.

ChihHao-Su:
Duplicate with https://github.com/fmtlib/fmt/pull/3259

Issue #3299, opened 2023-02-09
Add official Ubuntu mirrors


phprus:
To fix the error when installing Ubuntu packages, I suggest adding official mirrors, like in PR:
https://github.com/zlib-ng/zlib-ng/pull/1425

vitaut:
Good idea! Could you by any chance submit a PR?

phprus:
@vitaut 
PR #3302

PR #3302, opened 2023-02-10
Add ubuntu mirrors

phprus:
Github Actions caching proxy is at times unreliable

Fixes: #3299
More info: https://github.com/actions/runner-images/issues/7048


vitaut:
Thank you!

PR #3311, opened 2023-02-20
Improve fmt::print*

phprus:
*) Reduce the memory allocations
*) Reduce symbols (with ABI break)
*) Add ``println`` for compiled format string


vitaut:
Thanks for the PR but I don't think it's worth optimizing `println` and we should definitely not penalize the other APIs for this.

Issue #2811, opened 2022-03-15
Glibc extensions no longer supported in std::tm formatter


euclio:
In short, this assertion (using the `make_hour` function from the chrono tests) passes in 8.0.1 but fails in 8.1.0.

```cpp
EXPECT_EQ("8", fmt::format("{:%-I}", make_hour(8)));
```

According to `git bisect`, this regressed in 7911d8d3f56ec88f75e1f25663f4079ff2e9c67d.

Seems like this might be expected breakage given the goal of https://github.com/fmtlib/fmt/pull/2544, but [the documentation](https://fmt.dev/latest/syntax.html#chrono-specs) claims that the code uses the format specifiers from the system's `strftime`, which is no longer the case.

cc https://github.com/Alexays/Waybar/issues/1447

vitaut:
This is indeed expected, updated the docs in https://github.com/fmtlib/fmt/commit/1f3d44b859ba55d6e71194e563e7ee477882c92e. Thanks for reporting.

PR #3314, opened 2023-02-23
Support %P (lowercase am/pm) glibc extension

chris-morgan:
Fun fact: in glibc you can also spell %P %#p, # meaning “swap case”. I haven’t implemented glibc’s ^ or # flags (and don’t think there’s any particularly urgent need for it, so long as %P is around).

----

When #2544 broke this functionality for me and many other users of Waybar, I gathered from the response in #2811 that there was *actively* no interest in supporting glibc extensions. But now that #3271 has added the other important glibc extension (unpadded hours), I figured it might be worth making and trying submitting this patch after all.

The naming is arbitrary, and the code order for a quite possibly misguided form of consistency. I’m not fussed about it, feel free to amend it as you desire.

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -838,6 +838,9 @@ FMT_CONSTEXPR const Char* parse_chrono_format(const Char* begin,
     case 'p':
       handler.on_am_pm();
       break;
+    case 'P':
+      handler.on_am_pm_lower();
+      break;
     case 'Q':
       handler.on_duration_value();
       break;
@@ -976,6 +979,7 @@ template <typename Derived> struct null_chrono_spec_handler {
   FMT_CONSTEXPR void on_24_hour_time() { unsupported(); }
   FMT_CONSTEXPR void on_iso_time() { unsupported(); }
   FMT_CONSTEXPR void on_am_pm() { unsupported(); }
+  FMT_CONSTEXPR void on_am_pm_lower() { unsupported(); }
   FMT_CONSTEXPR void on_duration_value() { unsupported(); }
   FMT_CONSTEXPR void on_duration_unit() { unsupported(); }
   FMT_CONSTEXPR void on_utc_offset(numeric_system) { unsupported(); }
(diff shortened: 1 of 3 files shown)


include/fmt/chrono.h
```diff
@@ -835,6 +835,9 @@ FMT_CONSTEXPR const Char* parse_chrono_format(const Char* begin,
     case 'T':
       handler.on_iso_time();
       break;
+    case 'P':
+      handler.on_am_pm_lower();
+      break;
```
vitaut, line 840: nit: I'd suggest moving this after the 'p' case to keep the standard version first (and similarly for `am_pm_lower` functions below).

include/fmt/chrono.h
```diff
@@ -1627,6 +1632,15 @@ class tm_writer {
     out_ = copy_str<Char>(std::begin(buf), std::end(buf), out_);
   }
 
+  void on_am_pm_lower() {
+    if (is_classic_) {
+      *out_++ = tm_hour() < 12 ? 'a' : 'p';
+      *out_++ = 'm';
+    } else {
+      format_localized('P');
```
vitaut, line 1640: Please add a test case to ensure that it actually works with localized formatting.

vitaut (changes requested):
Thanks for the PR.

Looks good to me but please address the inline comments and update the docs (https://github.com/fmtlib/fmt/blob/master/doc/syntax.rst). Make sure to mention that `%P` is a non-standard extension.

include/fmt/chrono.h
chris-morgan: I don’t know how to do this. I wrote the code by blindly copying and modifying, not by deeply understanding what’s going on. 🙂

chris-morgan:
I’ve made the requested changes that I know how to, and request assistance on how to test the `format_localized('P')` branch.

> update the docs

I missed this because #3271 also missed this.

include/fmt/chrono.h
vitaut: Something along the lines of https://github.com/fmtlib/fmt/blob/050293646fc0337cfdb5084bf16f3d6efc84e9d8/test/chrono-test.cc#L756-L757
but with `%P` instead of `%a` and in a separate test case.

vitaut: Any updates? We'll need to have locale path tested before this can be merged.

vitaut:
Closing for now but feel free to resubmit with locale questions addressed.

Issue #3319, opened 2023-02-25
Double escaping on ranges of std::filesystem::path


phprus:
Code (https://godbolt.org/z/MGsqqnfb5):

```c++
#include <filesystem>
#include <vector>
#include <variant>
#include <fmt/ranges.h>
#include <fmt/std.h>

int main(int argc, char* argv[]) {
    std::vector<std::filesystem::path> files = {"path/file.ext", "path2/file2.txt"};
    fmt::println("{}", files);

    std::vector<std::variant<std::filesystem::path, std::string, int>> v = {std::filesystem::path("path/file.ext"), std::string("string"), 1};
    fmt::println("{}", v);

    return 0;
}
```

Output:

```
["\"path/file.ext\"", "\"path2/file2.txt\""]
[variant("path/file.ext"), variant("string"), variant(1)]
```

Expected output:

```
["path/file.ext", "path2/file2.txt"]
[variant("path/file.ext"), variant("string"), variant(1)]

```

PR #3321, opened 2023-02-26
Fix for issue #3319.

phprus:
Fix for issue #3319.
Disable debug format because path is already escaped.

### DIFF
--- a/test/std-test.cc
+++ b/test/std-test.cc
@@ -21,6 +21,8 @@ TEST(std_test, path) {
   EXPECT_EQ(fmt::format("{:8}", std::filesystem::path("foo")), "\"foo\"   ");
   EXPECT_EQ(fmt::format("{}", std::filesystem::path("foo\"bar.txt")),
             "\"foo\\\"bar.txt\"");
+  EXPECT_EQ(fmt::format("{:?}", std::filesystem::path("foo\"bar.txt")),
+            "\"foo\\\"bar.txt\"");
 
 #  ifdef _WIN32
   // File.txt in Russian.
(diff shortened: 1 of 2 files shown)


phprus:
Depends on #3322

phprus:
@vitaut 
Rebased and ready for review.

test/std-test.cc
```diff
@@ -46,6 +48,25 @@ TEST(ranges_std_test, format_vector_path) {
 #endif
 }
 
+TEST(ranges_std_test, format_quote_path) {
+  // Test quote problem described in #3319.
```
vitaut, line 52: Since the issue is very simple I suggest summarizing it instead of linking elsewhere, e.g. "Test that path is not escaped twice in the debug mode."

test/std-test.cc
```diff
@@ -46,6 +48,25 @@ TEST(ranges_std_test, format_vector_path) {
...
+            "[\"path1/file1.txt\", \"path2/file2.txt\"]");
+#  ifdef __cpp_lib_optional
+  auto o = std::optional<std::filesystem::path>("path/file.txt");
+  EXPECT_EQ(fmt::format("{}", o), "optional(\"path/file.txt\")");
+  EXPECT_EQ(fmt::format("{:?}", o), "optional(\"path/file.txt\")");
+#  endif
+#  ifdef __cpp_lib_variant
+  auto v = std::variant<std::filesystem::path, int>("path/file.txt");
+  EXPECT_EQ(fmt::format("{}", v), "variant(\"path/file.txt\")");
+#  endif
```
vitaut, line 66: I don't think we need these two because optional and variant use the same escaping mechanism as ranges.

vitaut (changes requested):
Thanks for the PR! Overall looks good, just two minor comments inline.

test/std-test.cc
phprus: Done.

test/std-test.cc
phprus: Done.

phprus:
@vitaut 
PR updated.

vitaut (approved):

vitaut:
Thank you!

Issue #3322, opened 2023-02-26
Incompatibility of fmt/ranges.h and std::filesystem::path on Windows


phprus:
Code (https://godbolt.org/z/qYPafnW9a):
```c++
#include <iostream>
#include <filesystem>
#include <variant>
#include <fmt/std.h>
#include <fmt/ranges.h>

int main(int argc, char* argv[]) {
    auto v = std::variant<std::filesystem::path, int>("path/file.txt");
    std::cout << fmt::format("{}", v) << std::endl;
    return 0;
}
```

Errror:
```
example.cpp
C:\data\libraries\installed\x64-windows\include\fmt\core.h(1855): error C2968: 'is_formattable<std::filesystem::path,char>': recursive alias declaration
C:\data\libraries\installed\x64-windows\include\fmt\core.h(320): note: see reference to class template instantiation 'fmt::v9::detail::is_formattable_delayed<std::filesystem::path,char>' being compiled
C:\data\libraries\installed\x64-windows\include\fmt\core.h(323): note: see reference to class template instantiation 'fmt::v9::conjunction<fmt::v9::detail::is_formattable_delayed<std::filesystem::path,char>>' being compiled
C:/data/msvc/14.29.30145/include\type_traits(593): note: see reference to class template instantiation 'fmt::v9::conjunction<std::integral_constant<bool,true>,fmt::v9::detail::is_formattable_delayed<std::filesystem::path,char>>' being compiled
C:/data/libraries/installed/x64-windows/include\fmt/std.h(114): note: see reference to class template instantiation 'std::is_constructible<fmt::v9::formatter<std::filesystem::path,char,void>>' being compiled
C:/data/libraries/installed/x64-windows/include\fmt/std.h(114): note: see reference to alias template instantiation 'fmt::v9::is_formattable<std::filesystem::path,char>' being compiled
C:/data/libraries/installed/x64-windows/include\fmt/std.h(140): note: see reference to class template instantiation 'fmt::v9::detail::is_variant_formattable_<T,C>' being compiled
        with
        [
            T=std::variant<std::filesystem::path,int>,
            C=char
        ]
C:/data/msvc/14.29.30145/include\type_traits(593): note: see reference to class template instantiation 'fmt::v9::is_variant_formattable<std::variant<std::filesystem::path,int>,char>' being compiled
[...]

PR #3323, opened 2023-02-26
std::filesystem::path is not format as a range

phprus:
Fix: #3322

### DIFF
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -59,6 +59,17 @@ template <typename T> class is_std_string_like {
       !std::is_void<decltype(check<T>(nullptr))>::value;
 };
 
+// Returns true if T has a std::filesystem::path-like interface.
+template <typename T> class is_std_filesystem_path_like {
+  template <typename U>
+  static auto check(U* p) -> decltype((void)U::preferred_separator, int());
+  template <typename> static void check(...);
+
+ public:
+  static constexpr const bool value =
+      !std::is_void<decltype(check<T>(nullptr))>::value;
+};
+
 template <typename Char>
 struct is_std_string_like<fmt::basic_string_view<Char>> : std::true_type {};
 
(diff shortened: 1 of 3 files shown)


vitaut:
`std::filesystem::path` should be excluded by the recursive range check in https://github.com/fmtlib/fmt/blob/73b7cee7fb0d46487607f95a7fc67229a3aea17c/include/fmt/ranges.h#L509

See e.g. http://eel.is/c++draft/format#range.fmtkind-2. I wonder why it doesn't work on Windows?

phprus:
@vitaut 

https://github.com/fmtlib/fmt/blob/73b7cee7fb0d46487607f95a7fc67229a3aea17c/include/fmt/ranges.h#L560-L564

On windows ``std::filesystem::path`` uses ``wchar_t``.
``is_range`` exclude ``std::filesystem::path`` on unix because ``std::filesystem::path`` is convertible to ``std::basic_string<char>``:
https://github.com/fmtlib/fmt/blob/73b7cee7fb0d46487607f95a7fc67229a3aea17c/include/fmt/ranges.h#L382-L387

On Windows ``std::filesystem::path`` does not convertible to ``std::basic_string<char>``.
Demo: https://godbolt.org/z/K1f1jPxjv

Original test for ``char`` (UNIX): https://github.com/fmtlib/fmt/blob/c644c753d7124b0161ad87003ce340ebfb0526c8/test/ranges-test.cc#L182-L191

phprus:
@vitaut, what do you think about this PR?

vitaut:
I think it's OK for `is_range` to return true for `path` on Windows and the root cause is broken recursion check (it didn't strip cvref). This should be fixed in https://github.com/fmtlib/fmt/commit/e0748e61ddc0cbef840596bf9748199a2197e613. So I'm closing this PR but thanks anyway, especially for the test case which has been added in the above commit.

PR #3054, opened 2022-08-23
Disable bogus -Wstringop-overflow on GCC 11

phprus:
``gcc version 11.3.0 (SUSE Linux)``

Warning:

```
[ 43%] Building CXX object test/CMakeFiles/unicode-test.dir/test-main.cc.o
cd /home/phprus/devel/tmp/fmt/fmt-git/build/cxx20r-asan/test && /usr/bin/g++-11 -DFMT_HEADER_ONLY=1 -DGTEST_HAS_STD_WSTRING=1 -D_SILENCE_TR1_NAMESPACE_DEPRECATION_WARNING=1 -I/home/phprus/devel/tmp/fmt/fmt-git/include -isystem /home/phprus/devel/tmp/fmt/fmt-git/test/gtest/. -fsanitize=address -O3 -DNDEBUG -fvisibility=hidden -fvisibility-inlines-hidden -pedantic-errors -Wall -Wextra -pedantic -Wold-style-cast -Wundef -Wredundant-decls -Wwrite-strings -Wpointer-arith -Wcast-qual -Wformat=2 -Wmissing-include-dirs -Wcast-align -Wctor-dtor-privacy -Wdisabled-optimization -Winvalid-pch -Woverloaded-virtual -Wconversion -Wno-ctor-dtor-privacy -Wno-format-nonliteral -Wno-dangling-else -Wno-unused-local-typedefs -Wdouble-promotion -Wtrampolines -Wzero-as-null-pointer-constant -Wuseless-cast -Wvector-operation-performance -Wsized-deallocation -Wshadow -Wshift-overflow=2 -Wnull-dereference -Wduplicated-cond -Werror -fno-delete-null-pointer-checks -std=gnu++20 -MD -MT test/CMakeFiles/unicode-test.dir/test-main.cc.o -MF CMakeFiles/unicode-test.dir/test-main.cc.o.d -o CMakeFiles/unicode-test.dir/test-main.cc.o -c /home/phprus/devel/tmp/fmt/fmt-git/test/test-main.cc
In file included from /home/phprus/devel/tmp/fmt/fmt-git/include/fmt/format.h:48,
                 from /home/phprus/devel/tmp/fmt/fmt-git/test/format-impl-test.cc:15:
In member function ‘constexpr void fmt::v9::detail::fill_t<Char>::operator=(fmt::v9::basic_string_view<Char>) [with Char = char]’,
[...]

Issue #2989, opened 2022-07-16
-Wstringop-overflow in GCC 12.1 at -O3 (not reproducible in GCC trunk)


hgkjshegfskef:
fmtlib: trunk
Repro: https://godbolt.org/z/7YM9oY4cj

Perhaps this is just to add to the collection of bogus `-Wstringop-overflow` warnings. Unless I'm using the library incorrectly. Seems to be fixed in GCC trunk (future version 13?). Note that the warning does not appear at -O2 and lower.

vitaut:
This looks like a false positive which is fixed in the latest gcc so I don't think we should do anything about it. You can suppress the warning with `FMT_SYSTEM_HEADERS` or some other method.

vitaut:
Worked around in https://github.com/fmtlib/fmt/commit/d9bc5f1320332db9a4bf7e103b0813b94e369304.

PR #3333, opened 2023-03-03
Fix code causing spurious Wstringop-overflow warning

codeinred:
Despite attempts to disable `-Wstringop-overflow` for tests, building the library from source still results in spurious Wstringop-overflow warnings. This occurs both for downstream consumers (eg, those using CMake Package Manager), and for people building the library from scratch.

Rather than attempting to disable the warning, I've modified the code such that the compiler is able to "figure out" that `fmt::detail::code_point_length` returns a value between 1 and 4, inclusive. The change also allows using bitshifts to calculate the codepoint length, which avoids a second array access.

While the `Wstringop-overflow` warning *was* spurious in this case, the new code should be more performant, and downstream consumers will no longer have the headache of dealing with a spurious warning when compiling their code. 

See Issue #2989, as well as #3054.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -2204,20 +2204,32 @@ constexpr auto to_ascii(Char c) -> char {
   return c <= 0xff ? static_cast<char>(c) : '\0';
 }
 
+// Returns the length of a codepoint. Returns 0 for invalid codepoints.
 FMT_CONSTEXPR inline auto code_point_length_impl(char c) -> int {
   return "\1\1\1\1\1\1\1\1\1\1\1\1\1\1\1\1\0\0\0\0\0\0\0\0\2\2\2\2\3\3\4"
       [static_cast<unsigned char>(c) >> 3];
+}
+// Returns the length of a codepoint. Returns 1 for invalid codepoints.
+// This is equivalent to
+//
+// int len = code_point_length_impl(c);
+// return len + !len;
+//
+// This is useful because it allows the compiler to check that the
+// length is within the range [1, 4]
+FMT_CONSTEXPR inline auto code_point_length_impl_2(char c) -> int {
+  return ((0x3a55000000000000ull >> (2 * (static_cast<unsigned char>(c) >> 3))) & 0x3) + 1;
 }
 
 template <typename Char>
 FMT_CONSTEXPR auto code_point_length(const Char* begin) -> int {
   if (const_check(sizeof(Char) != 1)) return 1;
-  int len = code_point_length_impl(static_cast<char>(*begin));
+  int len = code_point_length_impl_2(static_cast<char>(*begin));
 
   // Compute the pointer to the next character early so that the next
   // iteration can start working on the next character. Neither Clang
   // nor GCC figure out this reordering on their own.
-  return len + !len;
+  return len;
 }
 
 // Return the result via the out param to workaround gcc bug 77539.
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -2218,7 +2218,7 @@ FMT_CONSTEXPR inline auto code_point_length_impl(char c) -> int {
 // This is useful because it allows the compiler to check that the
 // length is within the range [1, 4]
 FMT_CONSTEXPR inline auto code_point_length_impl_2(char c) -> int {
-  return ((0x3a55000000000000ull >> (2 * (static_cast<unsigned char>(c) >> 3))) & 0x3) + 1;
+  return static_cast<int>((0x3a55000000000000ull >> (2 * (static_cast<unsigned char>(c) >> 3))) & 0x3) + 1;
 }
 
 template <typename Char>
--- a/test/core-test.cc
+++ b/test/core-test.cc
@@ -898,3 +898,18 @@ TEST(core_test, has_const_formatter) {
 TEST(core_test, format_nonconst) {
   EXPECT_EQ(fmt::format("{}", nonconst_formattable()), "test");
 }
+
+TEST(core_test, code_point_length_impl) {
+  // code_point_length_impl_2 is a bit-shifted version of code_point_length_impl
+  // that returns 1 for invalid codepoints, so that length is always in [1..4]
+  int min = CHAR_MIN;
+  int max = CHAR_MAX;
+
+  for(int ch = min; ch <= max; ch++) {
+    char c = static_cast<char>(ch);
+    int len1 = fmt::detail::code_point_length_impl(c);
+    int len2 = fmt::detail::code_point_length_impl_2(c);
+
+    ASSERT_EQ(len1 + !len1, len2);
+  }
+}


vitaut:
Please provide a godbolt repro that demonstrates the warning. The link in the referenced issue doesn't actually show any warnings.

codeinred:
[Here's a reproduction that demonstrates the warning](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:2,endLineNumber:7,positionColumn:1,positionLineNumber:1,selectionStartColumn:2,selectionStartLineNumber:7,startColumn:1,startLineNumber:1),source:'%23include+%3Cfmt/core.h%3E%0A%23include+%3Cfmt/ranges.h%3E%0A%23include+%3Cvector%3E%0A%0Avoid+print_vec(std::vector%3Cint%3E+nums)+%7B%0A++++fmt::print(%22%7B%7D%5Cn%22,+nums)%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:41.172877214416616,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x64+msvc+v19.latest',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+12.2+(Compiler+%231)',t:'0'),(h:compiler,i:(compiler:g122,deviceViewOpen:'1',filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'1',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!((name:fmt,ver:'910')),options:'-O3',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+12.2+(Editor+%231)',t:'0')),k:58.827122785583384,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

include/fmt/core.h
```diff
@@ -2204,20 +2204,32 @@ constexpr auto to_ascii(Char c) -> char {
...
+// This is equivalent to
+//
+// int len = code_point_length_impl(c);
+// return len + !len;
+//
+// This is useful because it allows the compiler to check that the
+// length is within the range [1, 4]
+FMT_CONSTEXPR inline auto code_point_length_impl_2(char c) -> int {
+  return static_cast<int>((0x3a55000000000000ull >> (2 * (static_cast<unsigned char>(c) >> 3))) & 0x3) + 1;
+}
```
vitaut, line 2222: I don't think we need both `code_point_length_impl` and `code_point_length_impl_2`, let's merge this logic into `code_point_length_impl` and make the latter take `unsigned char` to have one case instead of 3.

include/fmt/core.h
```diff
@@ -2204,20 +2204,32 @@ constexpr auto to_ascii(Char c) -> char {
...
+// Returns the length of a codepoint. Returns 1 for invalid codepoints.
+// This is equivalent to
+//
+// int len = code_point_length_impl(c);
+// return len + !len;
+//
+// This is useful because it allows the compiler to check that the
+// length is within the range [1, 4]
+FMT_CONSTEXPR inline auto code_point_length_impl_2(char c) -> int {
+  return static_cast<int>((0x3a55000000000000ull >> (2 * (static_cast<unsigned char>(c) >> 3))) & 0x3) + 1;
```
vitaut, line 2221: I think this logic deserves some explanation.

include/fmt/core.h
vitaut: On a second thought, this logic can be moved directly into `code_point_length` and both impl functions removed.

phprus: The return value of ``code_point_length`` is used as ``len`` and as ``len + !len``.

First case (first and last lines):
https://github.com/fmtlib/fmt/blob/3daf33837c98f95a7763459e06ce443abb883726/include/fmt/format.h#L654-L664

Second case:
https://github.com/fmtlib/fmt/blob/3daf33837c98f95a7763459e06ce443abb883726/include/fmt/core.h#L2215-L2223

vitaut: Ah, right. Then I think we should merge `code_point_length_impl` into `utf8_decode` and `code_point_length_impl_2` into `code_point_length`. Unless, of course, there is a similar false positive in `utf8_decode`.

include/fmt/core.h
codeinred: Hi! Sorry for the delay.

code_point_length_impl is initially implemented like this, which gets the length of a code point based on the first character in that codepoint.

```cpp
MT_CONSTEXPR inline auto code_point_length_impl(char c) -> int {
  return "\1\1\1\1\1\1\1\1\1\1\1\1\1\1\1\1\0\0\0\0\0\0\0\0\2\2\2\2\3\3\4"
      [static_cast<unsigned char>(c) >> 3];
}
```

For valid codepoints, we return the length of the codepoint, and for invalid codepoints we return a length of 0.

When used in `code_point_length`, the length of invalid codepoints gets converted to 1, so we don't actually need to distinguish between valid and invalid codepoints for `code_point_length`.

As a result, we could write it like this:

```cpp
MT_CONSTEXPR inline auto code_point_length_impl_2(char c) -> int {
  return "\1\1\1\1\1\1\1\1\1\1\1\1\1\1\1\1\1\1\1\1\1\1\1\1\2\2\2\2\3\3\4"
      [static_cast<unsigned char>(c) >> 3];
}
```

Now we have 4 possible return values: [1, 2, 3, 4]. If we instead use [0, 1, 2, 3] and just add one, it looks like this now:

```cpp
MT_CONSTEXPR inline auto code_point_length_impl_2(char c) -> int {
  return ("\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\1\1\1\1\2\2\3"
      [static_cast<unsigned char>(c) >> 3]) + 1;
}
```

Because this array only contains values 0..3, each value can be represented in 2 bits. This means that the entire array fits in a 64 bit integer.

We can get that integer like this:

```py
arr = [0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,1,1,1,1,2,2,3]
result = 0
pow = 1

for x in arr:
    result += pow * x
    pow *= 4

print(f'0x{result:x}')
```
This gives us `0x3a55000000000000`.

That logic basically indexes into this integer, treating it as an array of 2-bit values, resulting in the final implementation:

```cpp
FMT_CONSTEXPR inline auto code_point_length_impl_2(char c) -> int {
  return static_cast<int>((0x3a55000000000000ull >> (2 * (static_cast<unsigned char>(c) >> 3))) & 0x3) + 1;
}
```

This has two benefits:

- The compiler recognizes that the return value is always between 1 and 4, so it gets rid of the stringop-overflow warning
- We no longer need to access memory when computing the length of a codepoint

That being said the code could definitely be less ugly.

I'll apply your recommendations with regard to merging `code_point_length_impl` into utf8_decode, and merging `code_point_length_impl_2` into `code_point_length`. I haven't noticed any similar false positives in utf8_decode.

vitaut: Makes sense, thanks for the explanation.

vitaut:
Merged with some tweaks in https://github.com/fmtlib/fmt/commit/d9bc5f1320332db9a4bf7e103b0813b94e369304. Thanks for the PR.

PR #3338, opened 2023-03-10
Cleanup dead variable

ShawnZhong:
`custom_specs_` is introduced in https://github.com/fmtlib/fmt/pull/2673/files#diff-6ac6e767e83158d3347137b8f2ca90a37cb361e79c27fc3ba6a935074c26a342R646

The variable is unused after the removal of `if (custom_specs_)` in https://github.com/fmtlib/fmt/pull/2983/files#diff-6ac6e767e83158d3347137b8f2ca90a37cb361e79c27fc3ba6a935074c26a342L466


vitaut:
Good catch, thanks!

PR #3340, opened 2023-03-12
Fix GCC6 compilation error in `count_digits`

### LINKED ISSUE #3339: v9.1.0 gcc 6 compilation error: use of 'fmt::v9::detail::count_digits` before deduction of 'auto'
Error message:

```
include/fmt/format.h:1116:4: error: use of 'fmt::v9::detail::count_digits(UInt)::<lambda(UInt)> [with int BITS = 4; UInt = long long unsigned int]' before deduction of 'auto'
   return [](UInt m) {
          ~~~~~~~~~~~~
     int num_digits = 0;
     ~~~~~~~~~~~~~~~~~~~
     do {
     ~~~~
       ++num_digits;
       ~~~~~~~~~~~~~
     } while ((m >>= BITS) != 0);
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     return num_digits;
     ~~~~~~~~~~~~~~~~~~
   }(n);
   ~^~~
```

CI job: https://app.circleci.com/pipelines/github/diasurgical/devilutionX/17077/workflows/5baa3d10-eb3f-48d1-89f0-b8297057feea/jobs/108883 (from https://github.com/diasurgical/devilutionX/pull/5875)

Not quite sure what this error means.

glebm:
GCC6 appears to need the explicit return type there.

Otherwise, it fails with:

> error: use of 'fmt::v9::detail::count_digits(UInt)::<lambda(UInt)> [with int BITS = 4; UInt = long long unsigned int]' before deduction of 'auto'

Fixes #3339

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -1107,7 +1107,7 @@ FMT_CONSTEXPR auto count_digits(UInt n) -> int {
     return (FMT_BUILTIN_CLZ(static_cast<uint32_t>(n) | 1) ^ 31) / BITS + 1;
 #endif
   // Lambda avoids unreachable code warnings from NVHPC.
-  return [](UInt m) {
+  return [](UInt m) -> int {
     int num_digits = 0;
     do {
       ++num_digits;


vitaut:
Could you provide a godbolt repro?

glebm:
Do you have a godbolt template with libfmt?

glebm:
Note that I encountered this using a somewhat esoteric fork of GCC (Amiga GCC).
The full compilation error is available in the linked issue.

vitaut:
You could use this as a starting point: https://godbolt.org/z/b4TEMh5s4.

glebm:
There does not appear to be any version of GCC 6.5.x on godbolt

glebm:
Note that the error happens during compilation of fmt itself

glebm:
Godbolt: https://godbolt.org/z/8Wx1PxPss

I think the issue is simply a symptom of #3341. It is not reproducible on master, where #3341 is fixed by #3066.

It only results in a much shorter error message when #3341 is present. Please close the PR if that doesn't seem worth it.

vitaut:
Closing then but thanks for investigation.

sergiud:
I observe the issue again using fmt 10.1.1 and GCC 6.4.0. fmt 10.1.0 worked just fine. Regression?

vitaut:
Could you provide a godbolt repro? I wasn't able to reproduce the issue in https://godbolt.org/z/bbs165G9f.

sergiud:
Sure: https://godbolt.org/z/9c8c895Wo. Important bit is `-DFMT_HEADER_ONLY=1` which triggers the errors also in your repro.

vitaut:
Looks like a gcc 6 bug. Worked around in https://github.com/fmtlib/fmt/commit/84e666151755850ba8dde010ec8dd0ce02bb4ac7.

KaganCanSit:
https://github.com/gabime/spdlog/issues/3050 
In the issue above, I get the same error for the master and release version. The release and gcc tests in spdlog are as attached.

vitaut:
@KaganCanSit if it happens on master please open a separate issue and provide a self-contained repro.

KaganCanSit:
Hello @vitaut,
Today, I took the fmt library from the master tree and compiled it in the following gcc versions. I haven't received any errors so far.

GCC Versions:
7.5.0
7.1
6.5.0
5.5.0
5.1
4.9
4.8.5
4.8.4

When I compiled the spdlog by updating the header files in the spdlog, I received an error. Looks like a little more time is needed here. A cleaner operation may occur after Fmt is released and spdlog implements this version.

I wanted to give information. See you later.

KaganCanSit:
Hello again,
After checking the spdlog update, I got this error in g++/gcc 4.8.1. The compiler does not fully support lambda expressions. I have made various issue reports and tests regarding this situation. 

After deciding on this situation, it may be necessary to change the following statement in the documentation.

"The library is highly portable and requires only a minimal subset of C++11 features which are available in GCC 4.8, Clang 3.4, MSVC 19.0 (2015) and later. Newer compiler and standard library features are used if available, and enable additional functionality."

https://github.com/gabime/spdlog/issues/3050

I would like to help with this issue.

vitaut:
The docs have already been updated:

> GCC 4.9, Clang 3.4, MSVC 19.10 (2017) and later

https://fmt.dev/dev/

KaganCanSit:
@vitaut Ohh, I missed. I am focusing readme and repository URL.

image
image


Maybe we update this  partitions.

vitaut:
We don't retroactively changed published docs.

PR #3347, opened 2023-03-15
fix case of variant which is valueless by exception

TheOmegaCarrot:
[This](https://github.com/TheOmegaCarrot/breaking-fmt) tiny repository contains a small bit of code which forces an `std::variant` to be valueless by exception, and demonstrates that upstream fmtlib does not handle that case.

My small change simply handles this case.

### DIFF
--- a/test/std-test.cc
+++ b/test/std-test.cc
@@ -96,24 +96,19 @@ TEST(std_test, optional) {
 #endif
 }
 
-// this struct exists only to force a variant to be
-// valueless by exception
-// NOLINTNEXTLINE
-struct throws_on_move_t {
-  throws_on_move_t() = default;
+struct throws_on_move {
+  throws_on_move() = default;
 
-  // NOLINTNEXTLINE
-  [[noreturn]] throws_on_move_t(throws_on_move_t&&) {
-    throw std::runtime_error{"Thrown by throws_on_move_t"};
+  [[noreturn]] throws_on_move(throws_on_move&&) {
+    throw std::runtime_error("Thrown by throws_on_move_t");
   }
 
-  throws_on_move_t(const throws_on_move_t&) = default;
+  throws_on_move(const throws_on_move&) = default;
 };
 
-template <> struct fmt::formatter<throws_on_move_t> : formatter<string_view> {
-  template <typename FormatContext>
-  auto format(const throws_on_move_t&, FormatContext& ctx) const {
-    string_view str{"<throws_on_move_t>"};
+template <> struct fmt::formatter<throws_on_move> : formatter<string_view> {
+  auto format(const throws_on_move&, format_context& ctx) const {
+    string_view str("<throws_on_move_t>");
     return formatter<string_view>::format(str, ctx);
   }
 };
@@ -150,13 +145,11 @@ TEST(std_test, variant) {
   volatile int i = 42;  // Test compile error before GCC 11 described in #3068.
   EXPECT_EQ(fmt::format("{}", i), "42");
 
-  using V2 = std::variant<std::monostate, throws_on_move_t>;
-
-  V2 v6{};
+  std::variant<std::monostate, throws_on_move> v6;
 
   try {
-    throws_on_move_t thrower{};
-    v6.emplace<throws_on_move_t>(std::move(thrower));
+    throws_on_move thrower;
+    v6.emplace<throws_on_move>(std::move(thrower));
   } catch (const std::runtime_error&) {
   }
   // v6 is now valueless by exception
--- a/test/std-test.cc
+++ b/test/std-test.cc
@@ -107,7 +107,8 @@ struct throws_on_move {
 };
 
 template <> struct fmt::formatter<throws_on_move> : formatter<string_view> {
-  auto format(const throws_on_move&, format_context& ctx) const {
+  auto format(const throws_on_move&, format_context& ctx) const
+      -> decltype(ctx.out()) {
     string_view str("<throws_on_move>");
     return formatter<string_view>::format(str, ctx);
   }
--- a/test/std-test.cc
+++ b/test/std-test.cc
@@ -7,6 +7,7 @@
 
 #include "fmt/std.h"
 
+#include <stdexcept>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


vitaut:
Thanks for the PR. Please add a test case to https://github.com/fmtlib/fmt/blob/master/test/std-test.cc.

test/std-test.cc
```diff
@@ -95,6 +96,28 @@ TEST(std_test, optional) {
...
+    throw std::runtime_error{"Thrown by throws_on_move_t"};
+  }
+
+  throws_on_move_t(const throws_on_move_t&) = default;
+};
+
+template <> struct fmt::formatter<throws_on_move_t> : formatter<string_view> {
+  template <typename FormatContext>
+  auto format(const throws_on_move_t&, FormatContext& ctx) const {
+    string_view str{"<throws_on_move_t>"};
```
vitaut, line 116: We don't use brace initialization except for initializer lists, let's replace with normal initialization.

test/std-test.cc
```diff
@@ -95,6 +96,28 @@ TEST(std_test, optional) {
...
+  [[noreturn]] throws_on_move_t(throws_on_move_t&&) {
+    throw std::runtime_error{"Thrown by throws_on_move_t"};
+  }
+
+  throws_on_move_t(const throws_on_move_t&) = default;
+};
+
+template <> struct fmt::formatter<throws_on_move_t> : formatter<string_view> {
+  template <typename FormatContext>
+  auto format(const throws_on_move_t&, FormatContext& ctx) const {
```
vitaut, line 115: This doesn't need to be a template, you can use `format_context` instead.

test/std-test.cc
```diff
@@ -126,6 +149,20 @@ TEST(std_test, variant) {
 
   volatile int i = 42;  // Test compile error before GCC 11 described in #3068.
   EXPECT_EQ(fmt::format("{}", i), "42");
+
+  using V2 = std::variant<std::monostate, throws_on_move_t>;
```
vitaut, line 153: This alias is not needed, just use the variant type below.

test/std-test.cc
```diff
@@ -126,6 +149,20 @@ TEST(std_test, variant) {
...
+
+  V2 v6{};
```
vitaut, line 155: `{}` is redundant, please remove.

test/std-test.cc
```diff
@@ -95,6 +96,28 @@ TEST(std_test, optional) {
 #endif
 }
 
+// this struct exists only to force a variant to be
+// valueless by exception
+// NOLINTNEXTLINE
+struct throws_on_move_t {
```
vitaut, line 102: `_t` suffix is normally used for aliases (typedefs), please remove.

test/std-test.cc
```diff
@@ -126,6 +149,20 @@ TEST(std_test, variant) {
 
   volatile int i = 42;  // Test compile error before GCC 11 described in #3068.
   EXPECT_EQ(fmt::format("{}", i), "42");
+
+  using V2 = std::variant<std::monostate, throws_on_move_t>;
+
+  V2 v6{};
+
+  try {
+    throws_on_move_t thrower{};
```
vitaut, line 158: ditto

vitaut (changes requested):

test/std-test.cc
```diff
@@ -95,6 +96,28 @@ TEST(std_test, optional) {
 #endif
 }
 
+// this struct exists only to force a variant to be
+// valueless by exception
+// NOLINTNEXTLINE
```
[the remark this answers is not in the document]
TheOmegaCarrot: The purpose of the `NOLINTNEXTLINE` is to suppress clang-tidy warnings concerning the `throws_on_move` struct which is guaranteed to throw on move construction. I have removed it.

test/std-test.cc
TheOmegaCarrot: My apologies, fixed.

test/std-test.cc
TheOmegaCarrot: Fixed

test/std-test.cc
TheOmegaCarrot: Fixed

test/std-test.cc
TheOmegaCarrot: Fixed

test/std-test.cc
TheOmegaCarrot: Fixed

test/std-test.cc
TheOmegaCarrot: Fixed

test/std-test.cc
```diff
@@ -95,6 +96,23 @@ TEST(std_test, optional) {
...
+
+  [[noreturn]] throws_on_move(throws_on_move&&) {
+    throw std::runtime_error("Thrown by throws_on_move");
+  }
+
+  throws_on_move(const throws_on_move&) = default;
+};
+
+template <> struct fmt::formatter<throws_on_move> : formatter<string_view> {
+  auto format(const throws_on_move&, format_context& ctx) const {
```
vitaut, line 110: This needs a return type for compatibility with older compilers.

TheOmegaCarrot: My apologies once again. I've fixed this.

vitaut:
Almost there. Just one more issue on older compilers: https://www.internalfb.com/code/fbsource/[ae2a1f4ad0986b7c6686fa486466a9f6f47e9c05]/fbcode/thrift/compiler/test/fixtures/constants/gen-cpp2/module_constants.h?lines=140.

TheOmegaCarrot:
> Almost there. Just one more issue on older compilers: [https://www.internalfb.com/code/fbsource/[ae2a1f4ad0986b7c6686fa486466a9f6f47e9c05]/fbcode/thrift/compiler/test/fixtures/constants/gen-cpp2/module_constants.h?lines=140](https://www.internalfb.com/code/fbsource/%5Bae2a1f4ad0986b7c6686fa486466a9f6f47e9c05%5D/fbcode/thrift/compiler/test/fixtures/constants/gen-cpp2/module_constants.h?lines=140).

I am unable to view this link.

phprus:
@TheOmegaCarrot 
https://github.com/fmtlib/fmt/actions/runs/4450676746/jobs/7817044471?pr=3347

TheOmegaCarrot:
That ought to fix it.

vitaut (approved):

vitaut:
Thank you!

PR #3349, opened 2023-03-16
%T is %H:%M:%S

brevzin:
`%S` was fixed to print sub-second resolutions, but `%T` still only formats seconds.


vitaut:
Thank you!

PR #3356, opened 2023-03-17
Create a Security Policy

### LINKED ISSUE #3355: Create a Security Policy
A Security Policy is a GitHub standard document (`SECURITY.md`) that can be seen in the “Security Tab”  to instruct users about how to report vulnerability in the safest and most efficient way possible.

image

It is a [Scorecard Recommendation](https://github.com/ossf/scorecard/blob/main/docs/checks.md#security-policy) (being one check of medium priority) and a [Github Recommendation](https://docs.github.com/en/code-security/getting-started/securing-your-repository#setting-a-security-policy).

Together with this issue I’ll submit one suggestion of Security Policy, feel free to edit it directly or ask me for editions until it is in compliance with how fmt would best handle with vulnerabilities reports.

joycebrum:
Closes #3355 

I’ve created the `SECURITY.md` file considering the [report vulnerability through security advisory](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability), which is a new github feature still in beta and that has to be enabled.

If you rather not enabling it there is also the possibility to receive the vulnerability report through an email, in this case just let me know which email it would be and I’ll submit the change.

Besides that feel free to edit or suggest any changes to this document, it is supposed to reflect the amount of effort the team can offer to handle vulnerabilities.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,5 @@
+# Security Policy
+
+To report a security issue, please disclose it at [security advisory](https://github.com/fmtlib/fmt/security/advisories/new).
+
+We will respond within 7 working days of your submission. If the issue is confirmed as a vulnerability, we will open a Security Advisory and acknowledge your contributions as part of it. This project follows a 90 day disclosure timeline.


vitaut:
Thanks for the PR. Considering that we have limited resources, everything is a best effort. We will still prioritize security-related issues but I'd rather not create any expectations that we cannot meet. I enabled private vulnerability reporting which I think is sufficient for now.

joycebrum:
@vitaut I've updated the document to be a best effort bases. The Security Policy file is more about explaining the user what they should do when finding a vulnerability and what they should expect. Since the expectation is best effort bases, there is no problem to disclose it on the SECURITY.md (this is actually very common among open sources).

vitaut:
This looks reasonable but I don't think we need a top-level file for this. I suggest adding a section to the readme.

PR #3364, opened 2023-03-25
Optional Win9x support

glebm:
Adds optional support for Windows 9x/ME and the original Xbox (nxdk).

Most of libfmt is available except functions that depend on `fileno`, such as file descriptor duplication.

For now, support must be enabled explicitly by defining `FMT_WIN9X=1`.

/cc @AJenbo

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -46,8 +46,8 @@ FMT_BEGIN_NAMESPACE
 #  if FMT_HAS_INCLUDE("winapifamily.h")
 #    include <winapifamily.h>
 #  endif
-#  if defined(_WIN32) && (!defined(WINAPI_FAMILY) || \
-                          (WINAPI_FAMILY == WINAPI_FAMILY_DESKTOP_APP))
+#  if defined(_WIN32) && !FMT_WIN9X && \
+     (!defined(WINAPI_FAMILY) || (WINAPI_FAMILY == WINAPI_FAMILY_DESKTOP_APP))
 #    define FMT_USE_TZSET 1
 #  else
 #    define FMT_USE_TZSET 0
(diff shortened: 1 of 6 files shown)


vitaut:
Thanks for the PR but considering that even extended support for Windows 98/ME ended back in 2006, I don't think it's worth complicating our build configs for these systems.

include/fmt/chrono.h
```diff
@@ -46,8 +46,8 @@ FMT_BEGIN_NAMESPACE
 #  if FMT_HAS_INCLUDE("winapifamily.h")
 #    include <winapifamily.h>
 #  endif
-#  if defined(_WIN32) && (!defined(WINAPI_FAMILY) || \
-                          (WINAPI_FAMILY == WINAPI_FAMILY_DESKTOP_APP))
+#  if defined(_WIN32) && !FMT_WIN9X && \
```
ax3l, line 49: Typo here and below:
```suggestion

#  if defined(_WIN32) && !FMT_WIN98X && \
```

Clearly needs CI coverage 😁

glebm: Good catch, thanks, fixed! (`FMT_WIN9X` now used throughout)

glebm:
For context, we want to port DevilutionX to Windows 98, which the original Diablo was compatible with (https://github.com/diasurgical/devilutionX/issues/301).
Also, April 1st is coming and releasing a version with experimental Windows 98 support is good fun :grin:

include/fmt/chrono.h
ax3l: You are most welcome.

vitaut:
The usual approach to dealing with esoteric platform is adding compatibility workarounds to the core part of the library only (`core.h`, `format.h`, `format-inl.h`) and require the consumer to do define the build config for a subset of the library themselves. So it would be acceptable if the change was limited to these files and, possibly, `chrono.h` if the latter is not too intrusive.

PR #3369, opened 2023-04-01
fix compilation for MS-DOS

Mixaill:
It least Hello World works fine
DJGPP was used as a toolchain

image


vitaut:
image

jwt27:
Nice to see that this got merged, but you'll still need some way to work around missing `std::wstring` to build the static libary, no?  (ref: #2820)

Also see 910adb5132f3abfc6bd2dbea6c37d27b0be8bf5b for an implementation of `getpagesize()`.

PR #3375, opened 2023-04-09
Make fmt::println support text style

Nukoooo:
image
Supports all text style


vitaut:
Thanks for the PR but since `println` is redundant we only provide overloads required by the standard.

Issue #3377, opened 2023-04-10
[MSVC][c++latest] error C2668: 'std::format_to': ambiguous call to overloaded function


Zhaojun-Liu:
Hi all,
I tried to build Fmt with VS2022 on Windows, but it failed due to error C2668. Could anyone look at this issue? Thanks in advance.

**Environment:**
VS2022 + Windows Server

**Repro steps:**
1. Open a VS 2022 x64 Native Tools command.
2. git clone https://github.com/fmtlib/fmt F:\gitP\fmtlib\fmt
3. git -C "F:\gitP\fmtlib\fmt" reset --hard 4a4a2a2
4. set _CL_= /std:c++latest
5. mkdir F:\gitP\fmtlib\fmt\build_amd64 and cd F:\gitP\fmtlib\fmt\build_amd64
6. cmake -G "Visual Studio 17 2022" -A x64 -DCMAKE_SYSTEM_VERSION=10.0.22621.0 -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_STANDARD=20 ..
7. cmake --build . --config Release --verbose -- /maxcpucount

**Expected result:**
Build successfully.

**Actual result:**
```
"F:\gitP\fmtlib\fmt\build_amd64\ALL_BUILD.vcxproj" (default target) (1) ->
       "F:\gitP\fmtlib\fmt\build_amd64\test\format-impl-test.vcxproj" (default target) (19) ->
       (ClCompile target) -> 
         F:\gitP\fmtlib\fmt\test\format-impl-test.cc(304,5): error C2668: 'fmt::v9::format_to': ambiguous call to overloaded function [F:\gitP\fmtlib\fmt\build_amd64\test\format-impl-test.vcxproj]


       "F:\gitP\fmtlib\fmt\build_amd64\ALL_BUILD.vcxproj" (default target) (1) ->
       "F:\gitP\fmtlib\fmt\build_amd64\test\args-test.vcxproj" (default target) (3) ->
         F:\gitP\fmtlib\fmt\test\args-test.cc(47,12): error C2668: 'std::format_to': ambiguous call to overloaded function [F:\gitP\fmtlib\fmt\build_amd64\test\args-test.vcxproj]


       "F:\gitP\fmtlib\fmt\build_amd64\ALL_BUILD.vcxproj" (default target) (1) ->
       "F:\gitP\fmtlib\fmt\build_amd64\test\format-test.vcxproj" (default target) (18) ->
[...]

PR #3378, opened 2023-04-11
Fix error C2668 on msvc

Zhaojun-Liu:
Fmt built failed due to `error C2668: 'std::format_to': ambiguous call to overloaded function ` on MSVC, add namespace to qualify the call to format_to to solve this issue.

The related issue is: #3377


vitaut:
thanks

PR #3380, opened 2023-04-11
cmake: detect C++ standard automatically

Mixaill:
By default `CMAKE_CXX_STANDARD` is not defined, which causes module support to be disabled.

In this PR we try to use the latest standard supported by the compiler, if it has not been explicitly specified by `CMAKE_CXX_STANDARD`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,6 +157,16 @@ option(FMT_OS "Include core requiring OS (Windows/Posix) " ON)
 option(FMT_MODULE "Build a module instead of a traditional library." OFF)
 option(FMT_SYSTEM_HEADERS "Expose headers with marking them as system." OFF)
 
+#  Set the highest supported standard if it has not been externally defined
+if(NOT DEFINED CMAKE_CXX_STANDARD)
+  foreach(std_ver 23 20 17 14 11)
+    if(CMAKE_CXX_COMPILE_FEATURES MATCHES "cxx_std_${std_ver}")
+      set(CMAKE_CXX_STANDARD ${std_ver})
+      break()
+    endif()
+  endforeach()
+endif()
+
 set(FMT_CAN_MODULE OFF)
 if (CMAKE_CXX_STANDARD GREATER 17 AND NOT MSVC)
   set(FMT_CAN_MODULE ON)


flagarde:
@Mixaill This should force all subproject to have the same CMAKE_CXX_STANDARD. Maybe after the check the standard could be added as [target_compile_features](https://cmake.org/cmake/help/latest/manual/cmake-compile-features.7.html#requiring-language-standards) ?

Mixaill:
> @Mixaill This should force all subproject to have the same CMAKE_CXX_STANDARD. Maybe after the check the standard could be added as [target_compile_features](https://cmake.org/cmake/help/latest/manual/cmake-compile-features.7.html#requiring-language-standards) ?

This should force only fmt subprojects, right now it's only `gtest`, so it's no big deal. 

However, I also don't like polluting the current scope, so I switched to `FMT_CXX_STANDARD` and `set_target_properties`.

vitaut:
Thanks for the PR. I think making `FMT_MODULE=ON` not require specifying a standard is a good idea. However, I don't think we should change handling of the standard in other cases. We generally prefer keeping the default configuration as close to the compiler default as possible to prevent compatibility issues.

.github/workflows/linux.yml
```diff
@@ -94,7 +94,7 @@ jobs:
         CXXFLAGS: ${{matrix.cxxflags}}
       run: |
         cmake -DCMAKE_BUILD_TYPE=${{matrix.build_type}} ${{matrix.fuzz}} ${{matrix.shared}} \
-              -DCMAKE_CXX_STANDARD=${{matrix.std}} -DFMT_DOC=OFF \
+              -DFMT_CXX_STANDARD=${{matrix.std}} -DFMT_DOC=OFF \
```
vitaut, line 97: We should use the standard mechanism of specifying standard version.

Mixaill: In this case we can just drop the second commit

Mixaill:
@vitaut I've drop the second commit and rebase the branch. Should be much simpler.

Mixaill:
@vitaut do you want to change the standard only in the case of `FMT_MODULE` or are you satisfied with the current approach (use the newest version if the user does not explicitly specify a version)? ?

vitaut:
> do you want to change the standard only in the case of FMT_MODULE or are you satisfied with the current approach (use the newest version if the user does not explicitly specify a version or parent project specify something)?

I think we should only bump the standard for `FMT_MODULE`. It is mostly done already via https://github.com/fmtlib/fmt/blob/75f3b1c09485986427b8a46272f6b9570366e1aa/CMakeLists.txt#L70
so the only thing that remains is relaxing the check here:
https://github.com/fmtlib/fmt/blob/75f3b1c09485986427b8a46272f6b9570366e1aa/CMakeLists.txt#L59-L67

Mixaill:
Looks like modules project is a better place to address this issue https://github.com/vitaut/modules/pull/2

vitaut:
> Looks like modules project is a better place to address this issue

Agreed. Closing this one.

PR #3386, opened 2023-04-16
fix and improve module

DanielaE:
* export public documented API
* don't export `namespace detail`
* add `std.h` into module
* add missing namespace qualification in `xchar.h`
* fix call to `detail::get_iterator` in `xchar.h`
* fix ambiguous overload of `detail::isfinite` in `chrono.h`

### DIFF
--- a/include/fmt/args.h
+++ b/include/fmt/args.h
@@ -59,6 +59,7 @@ class dynamic_arg_list {
   }
 };
 }  // namespace detail
+FMT_BEGIN_EXPORT
 
 /**
   \rst
@@ -229,6 +230,7 @@ class dynamic_format_arg_store
   }
 };
 
+FMT_END_EXPORT
 FMT_END_NAMESPACE
 
 #endif  // FMT_ARGS_H_
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -1666,7 +1666,7 @@ struct chrono_format_checker : null_chrono_spec_handler<chrono_format_checker> {
   FMT_CONSTEXPR void on_duration_unit() {}
 };
 
-template <typename T, FMT_ENABLE_IF(std::is_integral<T>::value)>
+template <typename T, FMT_ENABLE_IF(std::is_integral<T>::value && has_isfinite<T>::value)>
 inline bool isfinite(T) {
   return true;
 }
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -394,6 +394,7 @@ FMT_CONSTEXPR inline auto is_utf8() -> bool {
   compiled with a different ``-std`` option than the client code (which is not
   recommended).
  */
+FMT_MODULE_EXPORT
 template <typename Char> class basic_string_view {
  private:
   const Char* data_;
@@ -496,9 +497,11 @@ template <typename Char> class basic_string_view {
   }
 };
 
+FMT_MODULE_EXPORT
 using string_view = basic_string_view<char>;
 
 /** Specifies if ``T`` is a character type. Can be specialized by users. */
+FMT_MODULE_EXPORT
 template <typename T> struct is_char : std::false_type {};
 template <> struct is_char<char> : std::true_type {};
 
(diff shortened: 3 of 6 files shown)


include/fmt/chrono.h
```diff
@@ -1666,7 +1666,7 @@ struct chrono_format_checker : null_chrono_spec_handler<chrono_format_checker> {
   FMT_CONSTEXPR void on_duration_unit() {}
 };
 
-template <typename T, FMT_ENABLE_IF(std::is_integral<T>::value)>
+template <typename T, FMT_ENABLE_IF(std::is_integral<T>::value && has_isfinite<T>::value)>
```
DanielaE, line 1669: Without the additional predicate, compilation will fail due to a clash with `format.h` line 2756 when compiling
```
fmt::format("{}", std::chrono::seconds{42});
fmt::format(L"{}", std::chrono::seconds{42});
```

include/fmt/core.h
```diff
@@ -258,6 +258,7 @@ FMT_GCC_PRAGMA("GCC optimize(\"Og\")")
 #endif
 
 FMT_BEGIN_NAMESPACE
+FMT_BEGIN_EXPORT
```
vitaut, line 261: The symbols below are undocumented and shouldn't be exported.

include/fmt/core.h
```diff
@@ -502,6 +505,7 @@ using string_view = basic_string_view<char>;
 template <typename T> struct is_char : std::false_type {};
 template <> struct is_char<char> : std::true_type {};
```
vitaut, line 506: These are also undocumented and semi-internal so let's just export `basic_string_view` via `FMT_MODULE_EXPORT` (we don't need the block here).

include/fmt/core.h
```diff
@@ -2024,6 +2036,7 @@ template <typename Char> struct fill_t {
   }
 };
 }  // namespace detail
+FMT_BEGIN_EXPORT
```
vitaut, line 2039: These are also undocumented APIs that we don't need to export yet.

include/fmt/std.h
```diff
@@ -73,6 +73,7 @@ inline void write_escaped_path<std::filesystem::path::value_type>(
 }
 
 }  // namespace detail
+FMT_BEGIN_EXPORT
```
vitaut, line 76: Here and a few other places: no need for a block, it's enough to mark the single definition here with `FMT_MODULE_EXPORT`.

vitaut (changes requested):
Thanks for the PR.

include/fmt/core.h
DanielaE: The primary template needs to be exported. It is documented to be [specializable by users](https://fmt.dev/latest/api.html#wchar-t-support). The (library-internal) specializations don't need to be exported as long as they are reachable (ehich they are).

vitaut: My bad. We should probably make it private but not in this PR.

include/fmt/args.h
```diff
@@ -59,6 +59,7 @@ class dynamic_arg_list {
   }
 };
 }  // namespace detail
+FMT_BEGIN_EXPORT
```
vitaut, line 62: We only have one class template here so I suggest applying FMT_MODULE_EXPORT to it.

include/fmt/core.h
```diff
@@ -775,11 +780,12 @@ FMT_CONSTEXPR void basic_format_parse_context<Char>::check_dynamic_spec(
   }
 }
 
-template <typename Context> class basic_format_arg;
-template <typename Context> class basic_format_args;
-template <typename Context> class dynamic_format_arg_store;
+FMT_MODULE_EXPORT template <typename Context> class basic_format_arg;
+FMT_MODULE_EXPORT template <typename Context> class basic_format_args;
+FMT_MODULE_EXPORT template <typename Context> class dynamic_format_arg_store;
```
vitaut, line 785: These are just forward declarations. Do we still need to export them if the main templates are exported?

DanielaE: Names must be declared `export` when **introduced** into a namespace [[module.interface]/6](http://eel.is/c++draft/module.interface#6). The compiler will then collect **all** semantic properties that pertain to that entity that is referenced by the name. This collection continues throughout the whole module purview up to the end of the TU or the start of the _private module fragment_, whichever comes first.

In this case, the (forward) _declarations_ introduce the names. So they need to be marked as exported at this very place in the source text.

This reminds me to remove the `export` declarations from the _definitions_ of the primary templates 🤦‍♂️

vitaut (approved):

vitaut:
Merged, thanks!

PR #3387, opened 2023-04-16
Feature/attach to gmf

DanielaE:


vitaut:
This overlaps with #3386. Did you mean to include only the first commit in this PR?

DanielaE:
I pushed only the second as a PR from the Github UI. I.e. the changes to `fmt.cc`.

vitaut:
Please rebase.

DanielaE:
Done.

vitaut:
Please provide some details why this change is needed.

DanielaE:
If you build {fmt} as a module, you're basically compiling it as a _static library_ that happens to export its interface in a binary form called BMI. _All_ declarations and definitions are _attached to to the module_, all signatures and _symbols carry an additional part_ that says so. In msvc this is a suffix `<!fmt>`, clang and gcc have something similar now that they both are following msvc's example and have switched over to the so-called _'strong module ownership' model_.

Suppose you have a program in transition: parts are using {fmt} in module form, the rest is still using it through #includes. This means you have _one set_ of declarations and symbols (all of them!) _attached to the module_, and another, _additional set_ (brought in by the #includes) attached to the so-called _global module_. Both can coexist in the same program because their signatures and symbols are different. But you pay for all those duplications.

If you define macro `FMT_ATTACH_TO_GLOBAL_MODULE`, _all declarations_ are subjected to _explicit language linkage specification_ `"extern C++"`. This _detaches_ them from the module, the additional symbol name parts don't exist and the module behaves like a static library build in the traditional sense. The private module partition provides all the definitions of entitites declared out-of-line.

DanielaE:
I have this feature in my Asio module to support exactly this kind of scenario.

vitaut:
Makes sense, thanks for the explanation. Please add a short comment clarifying this before the `FMT_ATTACH_TO_GLOBAL_MODULE` check.

src/fmt.cc
```diff
@@ -73,6 +73,10 @@ export module fmt;
   export {
 // All library-provided declarations and definitions must be in the module
 // purview to be exported.
```
vitaut, line 75: It looks like this comment applies to `FMT_ATTACH_TO_GLOBAL_MODULE` which is not right. Should it be moved some place else?

vitaut (changes requested):

vitaut (approved):

vitaut:
Merged, thanks!

PR #3391, opened 2023-04-18
names declared to be exported at the point of introduction into a nam…

DanielaE:
…espace *must not* be (re-)declared as exported later in the TU, e.g. when they are redeclared, defined, or specialized. [module.export]/6


DanielaE:
This needs more work

PR #3395, opened 2023-04-20
Update tests to use recommended MOCK_METHOD

kevinhwang:
`MOCK_METHODn()` has been deprecated.

See https://google.github.io/googletest/gmock_cook_book.html#old-style-mock_methodn-macros.


vitaut:
Thank you!

PR #3397, opened 2023-04-20
towards better module testing

DanielaE:
* use the standard `test-main.cc` component instead of injected test infrastructure sources
 * undo now obsolete commit `00235d8a` from July 2021
 * test current API: remove obsolete tests, add tests for new API functions


vitaut:
Now that module support is getting better, I think we should kill a separate module test and make normal tests compatible with modules. This will avoid duplication and improve coverage.

DanielaE:
I agree. All the tests for internal, non-exported entities need to be excluded from the test suites when the API is brought in through imports rather than includes. As far as I can see, this requires

- the ability to build a {fmt} module, preferably also consumable by users of the library
- for all compilers that have reasonable support for C++ modules (i.e. clang and msvc these days)
- with all the necessary workarounds applied (this affects both clang and msvc)
- subjected to all test suites on CI test runners.

For the time being I have this working only in the module testing branch of my fork of {fmt} that focuses on the API surface, the development of work arounds, and the deployment of CI. This builds on the existing, old test approach and the related infrastructure in CMakeTests.txt.

vitaut:
We don't need to modularize all tests at once. Even modularizing core-test alone would be better than the current state.

test/gtest-extra.h
```diff
@@ -12,12 +12,7 @@
 
 #include <string>
 
-#ifdef FMT_MODULE_TEST
-import fmt;
-#else
-#  include "fmt/os.h"
-#endif  // FMG_MODULE_TEST
-
+#include "fmt/os.h"
```
vitaut, line 15: Why change this?

test/module-test.cc
```diff
@@ -552,21 +505,27 @@ TEST(module_test, custom_context) {
...
+  EXPECT_EQ(R"(optional("42"))",
+            fmt::format("{}", std::optional<std::string>("42")));
+  EXPECT_EQ("none", fmt::format("{}", std::optional<std::string>()));
+  EXPECT_EQ("variant(42)",
+            fmt::format("{}", std::variant<int, std::string>(42)));
+  EXPECT_EQ("monostate", fmt::format("{}", std::monostate()));
+  EXPECT_EQ("system:0", fmt::format("{}", std::error_code()));
+  EXPECT_EQ("0", fmt::format("{}", std::thread::id()));
+  EXPECT_EQ("42", fmt::format("{}", std::runtime_error("42")));
+}
```
vitaut, line 531: I think we should stop adding new test cases to module-test.

vitaut (changes requested):

test/gtest-extra.h
DanielaE: With all compilers having converged to the 'strong module ownership' model we no longer need to fold module `fmt` into the test machinery for module tests. The module has *all* its symbols attached to the module (e.g. `void foo()` becomes `void foo@fmt()`), and the test machinery can have the same definitions attached to the global module **within the same program**.

Therefore, the test-main library can be reused as with all other tests. And all stuff conditionally compiled by `FMT_MODULE_TEST` can kicked out again. Both Clang and MSVC are fine with it. Strong ownership rulez!

test/module-test.cc
DanielaE: I can drop this for your repo and keep it in mine. I want tests **now**.

vitaut (approved):

vitaut:
Thank you!

PR #3398, opened 2023-04-20
Fix MSVC C4365 warning (signed/unsigned mismatch) on 32-bit Windows

louiswins:
Just needs a cast to `int` similar to the cast a few lines later.


vitaut:
thanks

PR #3399, opened 2023-04-23
modules missing pieces

DanielaE:
- don't export names from `detail`
- put more headers into the global module fragment
- support MSYS2 and Clang

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -1039,10 +1039,12 @@ template <typename T, size_t SIZE, typename Allocator>
 struct is_contiguous<basic_memory_buffer<T, SIZE, Allocator>> : std::true_type {
 };
 
+FMT_END_EXPORT
 namespace detail {
 FMT_API bool write_console(std::FILE* f, string_view text);
 FMT_API void print(std::FILE*, string_view);
 }  // namespace detail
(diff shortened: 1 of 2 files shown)


include/fmt/format.h
```diff
@@ -1039,10 +1039,10 @@ template <typename T, size_t SIZE, typename Allocator>
 struct is_contiguous<basic_memory_buffer<T, SIZE, Allocator>> : std::true_type {
 };
 
-namespace detail {
+FMT_BEGIN_DETAIL_NAMESPACE
 FMT_API bool write_console(std::FILE* f, string_view text);
 FMT_API void print(std::FILE*, string_view);
-}  // namespace detail
+FMT_END_DETAIL_NAMESPACE
```
vitaut, line 1045: I think we should use `FMT_{BEGIN,END}_EXPORT` instead of "magically" folding exports into namespace macros.

vitaut (commented):
Thanks for the PR! Mostly looks good, just one comment inline.

include/fmt/format.h
DanielaE: Ok, will do!

DanielaE: Done!

vitaut:
Thank you!

PR #3400, opened 2023-04-23
Name `vfprintf` clashes with the identically named declaration in 'st…

DanielaE:
…dio.h' if that happens to be #included into the same TU. Fix this by using qualified name lookup instead of unqualified lookup that also enables ADL.


vitaut:
It's probably the one in `<cstdio>` not `<stdio.h`> that is problematic. In any case the change seems reasonable, thank you!

PR #3407, opened 2023-04-27
Add spellcheck

Vertexwahn:
Add [codespell ](https://github.com/codespell-project/codespell)to avoid introduction of new spelling mistakes


vitaut:
Thanks for the PR but I think running spellcheck in CI is an overkill.

PR #3413, opened 2023-04-30
Add `FMT_STRING` for `format_to()` call with clang-cl

mogemimi:
I get the following error when compiling `os.cc` on clang-cl 15.

```cpp
1>..\fmt\src\os.cc(172,52): error : call to consteval function 'fmt::basic_format_string<char, const char *&, fmt::detail::utf16_to_utf8 &>::basic_format_string<char[7], 0>' is not a constant expression
1>..\fmt\include\fmt/core.h(3165,62): message : non-constexpr constructor 'basic_string_view' cannot be used in a constant expression
1>..\fmt\src\os.cc(172,52): message : in call to 'basic_format_string("{}: {}")'
1>..\fmt\include\fmt/core.h(451,3): message : declared here
```

This pull request adds a `FMT_STRING` wrapper to resolve a format-string type and fixes the error.

### DIFF
--- a/src/os.cc
+++ b/src/os.cc
@@ -169,8 +169,8 @@ void detail::format_windows_error(detail::buffer<char>& out, int error_code,
     if (msg) {
       auto utf8_message = utf16_to_utf8();
       if (utf8_message.convert(msg) == ERROR_SUCCESS) {
-        fmt::format_to(buffer_appender<char>(out), "{}: {}", message,
-                       string_view(utf8_message));
+        fmt::format_to(buffer_appender<char>(out), FMT_STRING("{}: {}"),
+                       message, string_view(utf8_message));
         return;
       }
     }


vitaut:
`FMT_STRING` shouldn't be needed here. Please provide a complete repro, ideally on godbolt.

mogemimi:
@vitaut
Thank you for replying. When I compiled `os.cc` with `/std:c++20` and [`-fmsc-version=1913`](https://releases.llvm.org/15.0.0/tools/clang/docs/CommandGuide/clang.html#cmdoption-fmsc-version) on clang-cl 15, the above error occurred. I had overlooked that I had specified `-fmsc-version=` option.

The reproduction procedure is as follows

1. Install [LLVM 15](https://github.com/llvm/llvm-project/releases/tag/llvmorg-15.0.7) on Windows. Or install clang-cl 15 from [Visual Studio 2022 installer](https://learn.microsoft.com/en-us/cpp/build/clang-support-msbuild?view=msvc-170).
    ```sh
    # "/c/Program Files/LLVM/bin/clang-cl.exe" --version
    
    clang version 15.0.7
    Target: x86_64-pc-windows-msvc
    Thread model: posix
    InstalledDir: C:\Program Files\LLVM\bin
    ```
    ```sh
    # "/c/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/Llvm/bin/clang-cl.exe" --version
    
    clang version 15.0.1
    Target: i686-pc-windows-msvc
    Thread model: posix
    InstalledDir: C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\Llvm\bin
    ```
2. Go to the `fmt` directory. 
    ```sh
    # cd path/to/fmt
    # ls src/os.cc
    ```
3. Run the following command to compile `src/os.cc` using clang-cl:
    ```sh
    # "/c/Program Files/LLVM/bin/clang-cl.exe" -Xclang -Iinclude /std:c++20 -fmsc-version=1913 /EHsc /diagnostics:column -Xclang -fcxx-exceptions src/os.cc
    
    src/os.cc(172,52): error: call to consteval function 'fmt::basic_format_string<char, const char *&, fmt::basic_string_view<char>>::basic_format_string<char[7], 0>' is not a constant expression
    include\fmt/core.h(2758,62): note: non-constexpr constructor 'basic_string_view' cannot be used in a constant expression
    src/os.cc(172,52): note: in call to 'basic_format_string("{}: {}")'
    include\fmt/core.h(421,3): note: declared here
    ```

In addition, when I specified `/std:c++14` or `/std:c++17` instead of `/std:c++20`, the error did not occur. Also, specifying `1914` or higher for `-fmsc-version=` like `-fmsc-version=1914` did not cause an error.

The reason seems to be that `FMT_CONSTEXPR_CHAR_TRAITS` is enabled when `FMT_MSC_VERSION` is `1914` or higher.
https://github.com/fmtlib/fmt/blob/02cae7e48ac9b1d90acb0ee07c76fbde2462a99e/include/fmt/core.h#L124-L126

vitaut:
I think it's OK as a workaround considering that we use `FMT_STRING` in similar cases elsewhere. Thanks.

PR #3414, opened 2023-05-01
Suppress most likely false positive with GCC 13

HazardyKnusperkeks:


gendalph (changes requested):
decltype(arg) -> decltype(val)

return {arg} -> return {arg_mapper<Context>().map(FMT_FORWARD(val))}

gendalph:
ref to arg_mapper().map(FMT_FORWARD(val)) some time is realy dangling

vitaut:
Closing as the warning appears to be legit. But thanks for the PR anyway.

Issue #3415, opened 2023-05-01
gcc13.1 warning: possibly dangling reference to a temporary w/ostream operator


mcharneyamp:
Hi.  Using current to of tree (02cae7e4),  I see this warning with aarch64 gcc 13.1 -Wall. On godbolt, I used "arm64 gcc trunk" to compile and got this error (They don't have an arm64 gcc 13.1 listed yet).  Oddly, I don't see the error with x86-64 gcc13.1.  No error with gcc 11.3 or 12.2 on aarch64.

```
/opt/compiler-explorer/libs/fmt/trunk/include/fmt/core.h:1674:15: warning: possibly dangling reference to a temporary [-Wdangling-reference]
 1674 |   const auto& arg = arg_mapper<Context>().map(FMT_FORWARD(val));
      |               ^~~
/opt/compiler-explorer/libs/fmt/trunk/include/fmt/core.h:1674:46: note: the temporary was destroyed at the end of the full expression 'fmt::v9::detail::arg_mapper<fmt::v9::basic_format_context<fmt::v9::appender, char> >().fmt::v9::detail::arg_mapper<fmt::v9::basic_format_context<fmt::v9::appender, char> >::map<foo&>((* & val))'
 1674 |   const auto& arg = arg_mapper<Context>().map(FMT_FORWARD(val));
      |                     ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
Compiler returned: 0
```

Here is a godbolt repro link:

https://godbolt.org/z/ajsGKe4nE

```
#include <cstdint>
#include <ostream>
#include <string>
#include <fmt/core.h>
#include <fmt/ranges.h>
#include <fmt/ostream.h>

class foo {
public:
    int a;
    std::string b;

    foo(int aa, std::string bb) : a(aa), b(bb) {}

    friend std::ostream& operator<< (std::ostream& os, const foo& obj);
};

template <> struct fmt::formatter<foo> : ostream_formatter {};
std::ostream& operator<< (std::ostream& os, const foo& obj) {
    os << obj.a << " " << obj.b;
    return os;
}



int main(int argc, char** argv) {
    foo f(5,"hello");
    fmt::print("{}\n", f);
}

```

pattop:
Same happens when a specialisation of fmt::formatter is added, e.g https://godbolt.org/z/r657Yhc9Y

```
enum class foo { a, b };

template<>
struct fmt::formatter<foo> {
[...]

PR #3420, opened 2023-05-05
Fix dangling reference

phprus:
Fix for issue: #3415

Before: https://godbolt.org/z/ajsGKe4nE
After: https://godbolt.org/z/bsaEYYWMW

Thanks to @gendalph for the suggestion for PR #3414.


vitaut:
Thanks for the PR. As commented on #3415 the warning is clearly wrong but I applied a (slightly simpler) workaround in https://github.com/fmtlib/fmt/commit/f61f15cc5b11582d50d02ba0514c5344f7b2600e.

PR #3421, opened 2023-05-08
Silence NVCC warning in has_isfinite

scramsby:
This avoids warnings like:
```
: warning: pointless comparison of unsigned integer with zero
          detected during:
            instantiation of "auto fmt::v9::detail::to_unsigned(Int)->std::make_unsigned<Int>::type [with Int=size_t]" 
```
It also has the benefit of being a little clearer in its intent.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -2802,7 +2802,7 @@ template <typename T, typename Enable = void>
 struct has_isfinite : std::false_type {};
 
 template <typename T>
-struct has_isfinite<T, enable_if_t<sizeof(std::isfinite(T())) != 0>>
+struct has_isfinite<T, void_t<decltype(std::isfinite(T()))>>
     : std::true_type {};
 
 template <typename T, FMT_ENABLE_IF(std::is_floating_point<T>::value&&


vitaut:
Thanks for the PR but I am not sure what `has_isfinite` has to do with the warning you quoted. Is it a wrong warning?

scramsby:
I didn't. The offending line is this one from the diff:
`enable_if_t<sizeof(std::isfinite(T())) != 0`
It's the `!= 0` comparison that the warning is complaining about.

mwinterb:
Is the warning in this case because sizeof produces a constant? i.e. it's "1u != 0"?

vitaut:
But there is no `to_unsigned` here. Please post the full warning with file names and line numbers.

scramsby:
Ahh, you're right. I apparently in my sleep-deprived state misread the error. So this doesn't actually fix a warning, but is marginally more clear, so you're free to take or leave it.

This is the original error which I didn't actually fix apparently:
```
<redacted>\fmt\include\fmt\core.h(409): warning: pointless comparison of unsigned integer with zero
          detected during:
            instantiation of "auto fmt::v9::detail::to_unsigned(Int)->std::make_unsigned<Int>::type [with Int=size_t]" 
<redacted>\fmt\include\fmt\format.h(570): here
            instantiation of "auto fmt::v9::detail::fill_n(T *, Size, char)->T * [with T=char, Size=size_t]" 
<redacted>\fmt\include\fmt\format.h(843): here
            instantiation of "fmt::v9::basic_memory_buffer<T, SIZE, Allocator>::basic_memory_buffer(const Allocator &) [with T=char, SIZE=500ULL, Allocator=std::allocator<char>]" 
<redacted>\fmt\include\fmt\ostream.h(192): here
```

vitaut:
Closing then but a PR to fix the actual warning is welcome.

PR #3422, opened 2023-05-08
Add some missing argument forwarding

scramsby:
These additional argument forwards don't appear to presently result in any functional changes that I can tell, but are more future proof, more indicative of intent, and more consistent with other code.

Note: It's not clear why `FMT_FORWARD()` is used in some places in the existing code and sometimes `std::forward<>()` is used directly instead. In this change I opted to use `FMT_FORWARD()` as it doesn't imply additional dependence on the `<utility>` header. I leave it as a later exercise to investigate whether one or the other should be used consistently.


vitaut:
Thanks for the PR but we don't need forwarding for argument capture.

PR #3261, opened 2023-01-04
Fix negative subsec for time_point

ShawnZhong:
Fix #3117

The following piece of code (https://godbolt.org/z/4qr5389ns) used to print `59.000 00.750 00.500 00.250 00.000 00.250 00.500 00.750 01.000`, which is wrong. After this PR, it now correctly prints `59.000 59.250 59.500 59.750 00.000 00.250 00.500 00.750 01.000`

```cpp
const auto epoch = std::chrono::time_point<std::chrono::system_clock,
                                           std::chrono::milliseconds>();
const auto d = std::chrono::milliseconds(250);
for (int i = -4; i <= 4; i++) {
  fmt::print("{:%S} ", epoch + i * d);
}
```

PR #3430, opened 2023-05-11
Fix time_point formatting for durations with certain ratios

patlkli:
#3261 unfortunately broke formatting for `std::chrono::time_point`s with durations with certain ratios, since `std::chrono::seconds` is not implicitly convertible to these durations. 

This PR fixes this by explicitly `duration_cast`-ing back to the `time_point`'s `Duration`.

Repro: The following worked in 9.1.0, but doesn't compile in 10.0.0:
```cpp
#include <fmt/chrono.h>

int main() {
    fmt::print("{}\n", std::chrono::time_point_cast<std::chrono::minutes>(std::chrono::system_clock::now()));
}
```


vitaut:
Thanks for the PR. Could you add a test case to chrono-test?

patlkli:
Sure, they're in now.

PR #3432, opened 2023-05-13
Fix parameter type of formatter<char*>

timsong-cpp:
Since this is macro expansion, `const Type&` expands to `const Char*&` instead of the desired `Char* const&`.

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -2693,7 +2693,7 @@ struct formatter<T, Char,
   template <typename Char>                                               \
   struct formatter<Type, Char> : formatter<Base, Char> {                 \
     template <typename FormatContext>                                    \
-    auto format(const Type& val, FormatContext& ctx) const               \
+    auto format(Type const& val, FormatContext& ctx) const               \
         -> decltype(ctx.out()) {                                         \
       return formatter<Base, Char>::format(static_cast<Base>(val), ctx); \
     }                                                                    \
--- a/test/format-test.cc
+++ b/test/format-test.cc
@@ -2251,3 +2251,19 @@ TEST(format_test, format_named_arg_with_locale) {
 }
 
 #endif  // FMT_STATIC_THOUSANDS_SEPARATOR
+
+struct convertible_to_nonconst_cstring {
+    operator char*() const {
+        static char c[]="bar";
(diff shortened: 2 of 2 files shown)


vitaut:
Thanks for the PR but this specialization should be there at all, removed in https://github.com/fmtlib/fmt/commit/aeedac588490ebeea17bd083f0696742e0d295ed.

timsong-cpp:
> this specialization should [not] be there at all

Hmm, are you sure? I think the result of this removal is that I can still format a `char*` but not a range of it (and `std::formatter` does provide this specialization.

timsong-cpp:
OK, `range_formatter` works because it uses the mapper, but not `fmt::join` for either ranges or tuples. I'm not sure why this inconsistency (within fmt and also with `std::formatter`) is desirable.

vitaut:
The `char*` formatter is redundant. The only reason it has a `formatter` in the standard is because there is no other ways to represent formattability there (yet).

timsong-cpp:
If I have a `my::Optional<T>` whose formatter wants to support all the specifiers supported by `T`, how should I write it so that it works for `my::Optional<char*>`?

Also as mentioned, this breaks `fmt::join` for both ranges and tuples: https://gcc.godbolt.org/z/45q54oe4T

brevzin:
A simple wrapping example [now breaks](https://gcc.godbolt.org/z/GoxT7nETx):

```cpp
#include <fmt/format.h>

template <typename T>
struct wrapped {
    T t;
};

template <typename T>
struct fmt::formatter<wrapped<T>> {
    fmt::formatter<T> underlying;

    constexpr auto parse(auto& ctx) {
        return underlying.parse(ctx);
    }

    auto format(wrapped<T> const& w, auto& ctx) const {
        return underlying.format(w.t, ctx);
    }
};

auto f(wrapped<char*> w) -> std::string {
    return fmt::format("{}", w); // error: formatter<char*>::formatter() is deleted
}
```

vitaut:
Good point, reverted in https://github.com/fmtlib/fmt/commit/616a49378652ebd7bb3f3dbd0b3f01f53f34b8cb.

vitaut:
merged, thanks

Issue #3433, opened 2023-05-13
Failing tests with 10.0.0 on alpine linux


a16bitsysop:
When running the tests for fmt 10.0.0 on alpine linux edge with gcc 12.2.1 some tests fail for x86_64 and x86, but all tests pass for ppc64le and s390x.

The failure is the same for both:
```
Test project /builds/a16bitsysop/aports/main/fmt/src/fmt-10.0.0
      Start  1: args-test
      Start  2: assert-test
      Start  3: color-test
      Start  4: core-test
      Start  5: gtest-extra-test
      Start  6: format-test
      Start  7: format-impl-test
      Start  8: ostream-test
      Start  9: compile-test
      Start 10: compile-fp-test
      Start 11: printf-test
      Start 12: ranges-test
      Start 13: scan-test
      Start 14: std-test
      Start 15: enforce-checks-test
      Start 16: posix-mock-test
      Start 17: os-test
 1/17 Test  #1: args-test ........................   Passed    0.02 sec
 2/17 Test  #2: assert-test ......................   Passed    0.02 sec
 3/17 Test  #3: color-test .......................   Passed    0.02 sec
 4/17 Test  #4: core-test ........................   Passed    0.02 sec
 5/17 Test  #5: gtest-extra-test .................   Passed    0.01 sec
 6/17 Test  #8: ostream-test .....................   Passed    0.01 sec
 7/17 Test  #9: compile-test .....................   Passed    0.01 sec
 8/17 Test #10: compile-fp-test ..................   Passed    0.01 sec
 9/17 Test #11: printf-test ......................   Passed    0.01 sec
10/17 Test #12: ranges-test ......................   Passed    0.01 sec
11/17 Test #13: scan-test ........................   Passed    0.01 sec
12/17 Test #14: std-test .........................   Passed    0.01 sec
13/17 Test #15: enforce-checks-test ..............   Passed    0.01 sec
14/17 Test  #6: format-test ......................***Failed    0.02 sec
[==========] Running 114 tests from 6 test suites.
[----------] Global test environment set-up.
[----------] 5 tests from uint128_test
[ RUN      ] uint128_test.ctor
[       OK ] uint128_test.ctor (0 ms)
[ RUN      ] uint128_test.shift
[       OK ] uint128_test.shift (0 ms)
[ RUN      ] uint128_test.minus
[       OK ] uint128_test.minus (0 ms)
[ RUN      ] uint128_test.plus_assign
[...]

PR #3434, opened 2023-05-13
Make hex float test more stable

phprus:
Fix for issue: #3433


vitaut:
Thank you for the quick fix!

PR #3439, opened 2023-05-15
Always assert in FMT_THROW

### LINKED ISSUE #3418: Error handling in an -fno-exceptions world
`error_handler::on_error` does this:

https://github.com/fmtlib/fmt/blob/d8973bf16be2af304065df1e210d2681a78cc82e/include/fmt/core.h#L635-L638

Which hides the throwing behavior behind a macro:

https://github.com/fmtlib/fmt/blob/d8973bf16be2af304065df1e210d2681a78cc82e/include/fmt/format-inl.h#L39-L41

Which conditionally either throws or asserts:

https://github.com/fmtlib/fmt/blob/d8973bf16be2af304065df1e210d2681a78cc82e/include/fmt/format.h#L105-L128

But asserting... might not assert:

https://github.com/fmtlib/fmt/blob/d8973bf16be2af304065df1e210d2681a78cc82e/include/fmt/core.h#L330-L341

What this means is that a program like:

```cpp
#include <fmt/format.h>

int main() {
    fmt::println(fmt::runtime("Hi {:d}"), "10");
}
```

could:

1. If exceptions are enabled, throw an exception for bad format specifier.
2. Otherwise, if `NDEBUG` isn't defined, terminate after printing an error about bad format specifier.
3. Otherwise (if `-fno-exceptions -DNDEBUG`), just... keep... going?

So the above program could just [print "Hi 10"](https://godbolt.org/z/YP1nh34xc). That's pretty weird. Also we're violating the `[[noreturn]]` on `throw_format_error` since in this case it actually returns.

Maybe we should introduce a `FMT_ALWAYS_ASSERT` that always does the `(condition) ? (void)0 : assert_fail(~)` bit, and have `FMT_THROW` fallback to `FMT_ALWAYS_ASSERT` instead of `FMT_ASSERT`?

brevzin:
Fixes #3418

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -114,10 +114,8 @@ FMT_END_NAMESPACE
 #      define FMT_THROW(x) throw x
 #    endif
 #  else
-#    define FMT_THROW(x)               \
-      do {                             \
-        FMT_ASSERT(false, (x).what()); \
-      } while (false)
+#    define FMT_THROW(x) \
+      ::fmt::detail::assert_fail(__FILE__, __LINE__, (x).what())
 #  endif
 #endif
 


vitaut:
Thank you!

PR #3448, opened 2023-05-17
remove code duplication

florimond-collette:


include/fmt/format-inl.h
```diff
@@ -1162,25 +1162,10 @@ FMT_INLINE int remove_trailing_zeros(uint64_t& n) noexcept {
...
   // Is n is divisible by 10^8?
   if ((nm.high() & ((1ull << (90 - 64)) - 1)) == 0 && nm.low() < magic_number) {
-    // If yes, work with the quotient.
+    // If yes, work with the quotient...
     auto n32 = static_cast<uint32_t>(nm.high() >> (90 - 64));
-
-    const uint32_t mod_inv_5 = 0xcccccccd;
-    const uint32_t mod_inv_25 = mod_inv_5 * mod_inv_5;
-
-    int s = 8;
```
vitaut, line 1171: Looks like 8 here gets lost.

florimond-collette: totally right, I changed the signature of `remove_trailing_zeros(uint32_t& n, int s = 0)` in order to allow passing the s

vitaut:
Thank you!

jk-jeon:
How about to revert the change for the 32-bit version and just add 8 to the result in the 64-bit version? In my opinion that's more readable and the impact is more local.

vitaut:
> How about to revert the change for the 32-bit version and just add 8 to the result in the 64-bit version? In my opinion that's more readable and the impact is more local.

I was thinking about the same. A PR is welcome!

Issue #3163, opened 2022-11-03
warning C4307: “*” on VS2017


soncfe:
build in spdlog 1.11.0 header only,

spdlog\fmt\bundled\format-inl.h(1154): warning C4307: “*”

// Remove trailing zeros from n and return the number of zeros removed (float)
FMT_INLINE int remove_trailing_zeros(uint32_t& n) noexcept {
FMT_ASSERT(n != 0, "");
const uint32_t mod_inv_5 = 0xcccccccd;
const uint32_t mod_inv_25 = mod_inv_5 * mod_inv_5; //// ?????

vitaut:
I wasn't able to reproduce this in https://godbolt.org/z/baaGbjebz. Please provide more repro details.

vitaut:
Closing for now but feel free to reopen with a repro (ideally on godbolt).

Qujamlee:
> I wasn't able to reproduce this in https://godbolt.org/z/baaGbjebz. Please provide more repro details.

There is indeed this problem, and I have reproduced it on the website you provided, but you need to choose x64msvc19.22 and earlier versions of the compiler.I copied the warning message:

x64 msvc v19.22 /std:c++latest /W2
Compiler stdout
example.cpp
<source>(5): warning C4307: '*': integral constant overflow
Program returned: 0

But using newer versions of the compiler does not have this warning. Why? Did the compiler do the optimization?
I'm doing the math, 
const uint32_t mod_inv_5 = 0xcccccccd = 3_435_973_837 (decimal), 
the square of it is 11_805_916_208_548_502_569, 
It does exceed the maximum range of uint32_t (4_294_967_295) ,
so the warning message seems reasonable, is it a bug?

vitaut:
@jk-jeon, what is the logic behind https://github.com/fmtlib/fmt/blob/115ca96e0e8755524592beb99f1f0669335b8fdf/include/fmt/format-inl.h#L1183-L1184?

Would it make sense to avoid overflow there?

jk-jeon:
`0xcccccccd` is the modular inverse of 5 (mod $2^{32}$ ), so the square of it is the modular inverse of 25. This is indeed an intended overflow. There are several ways I can think of to silence this warning:
1. Disable the warning. Preferably locally, if possible.
[...]

PR #3450, opened 2023-05-17
revert non-working workaround for issue #3163

florimond-collette:
revert non-working workaround for issue #3163

This fix for warning C4307: '*': integral constant overflow leads to warning C4309: 'static_cast': truncation of constant value, which is not better (both are level 3 VS warnings). Revert to the original code that is clearer and also present further in this source code.

Alternatively, to avoid any VS warning, one could follow the third suggestion of https://github.com/fmtlib/fmt/issues/3163#issuecomment-1332608091 and write the explicit value of mod_inv_25 which is 0xc28f5c29 and replace the lines with:
`const uint32_t mod_inv_25 = 0xc28f5c29 // = mod_inv_5 * mod_inv_5;`

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -1131,11 +1131,8 @@ bool is_left_endpoint_integer_shorter_interval(int exponent) noexcept {
 FMT_INLINE int remove_trailing_zeros(uint32_t& n, int s = 0) noexcept {
   FMT_ASSERT(n != 0, "");
   // Modular inverse of 5 (mod 2^32): (mod_inv_5 * 5) mod 2^32 = 1.
-  // See https://github.com/fmtlib/fmt/issues/3163 for more details.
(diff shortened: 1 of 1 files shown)


vitaut:
Could you apply the alternative suggestion to suppress the warning? i.e. 

```
const uint32_t mod_inv_25 = 0xc28f5c29 // = mod_inv_5 * mod_inv_5;
```

florimond-collette:
The value of mod_inv_25 is now explicit in 32 and 64 bit:
`const uint32_t mod_inv_25 = 0xc28f5c29; // = mod_inv_5 * mod_inv_5`
`const uint64_t mod_inv_25 = 0x8f5c28f5c28f5c29; // = mod_inv_5 * mod_inv_5`

vitaut:
Please rebase.

vitaut:
Thanks

Issue #3449, opened 2023-05-17
Hash pin github workflow dependencies


joycebrum:
## Description

I also would like to suggest a security practice recommended by the [OpenSSF Scorecard][scorecard-repo] which is to hash pin dependencies to prevent dependency-confusion, typosquatting and tag renaming attacks. 

The change would only be applied to GitHub workflows dependencies.

This means:
- Hash pinning GitHub Workflow actions.
- Using a package-lock.json on npm installs.

I can submit one PR for each type of change above to be easier to review if you prefer. Just let me know if that's the case.

Along with hash-pinning dependencies, I also recommend adopting dependabot or renovatebot to help keep the dependencies up to date. Both tools can update  hashes and associated semantic version comments. 

Let me know if you are open to evaluate those changes and I'll submit the PR ASAP. 

Any questions or concerns just let me know.
Thanks!

## Additional Context

A tag renaming attack is a type of attack whereby an attacker:
- Hijack an action.
- Upload a malicious version.
- Replace existing tags with malicious versions.

A [dependency-confusion][deps-confusion] attack occurs when an attacker:
- Find the name of a package that the victims wants to install
- Create an identically named package and publish it under the public or default registry.
- Assign the package with a higher version number to trick the package manager tool to download it from the public repo.

 
A [typosquatting attack][typosquatting] is a type of attack whereby an attacker:
- Create a malicious package
- Publish it with a similar name of a known package (example: numpi instead of numpy)

For more informations about the dependency-update tools:
- [Dependabot][dependabot]
- [Renovatebot][renovatebot]

[scorecard-repo]: https://github.com/ossf/scorecard
[deps-confusion]: https://www.websecuritylens.org/how-dependency-confusion-attack-works-and-how-to-prevent-it/ 
[typosquatting]: https://snyk.io/blog/typosquatting-attacks/ 
[dependabot]: https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/
[renovatebot]:https://www.mend.io/renovate/

vitaut:
Sounds like a good idea, PRs are welcome.

Frank-Buss:
That's an interesting point for the "actions" workflows, they are probably pretty safe and might be useful to get the latest versions, if problems with old versions are found. But malicious software could be still introduced to it. Dependabot looks like a good compromise. Looks like it creates PRs for new workflows, and then they could be reviewed.

[...]

PR #3451, opened 2023-05-19
Hash pin Github Workflows

joycebrum:
#3449 

## Changes
- Hash pin actions used on the github workflows. 
- cifuzz workflow not pinned since it only works on master (pinning it would cause too much notifications from dependabot/renovatebot without need)


vitaut:
Thank you

PR #3452, opened 2023-05-19
Create dependabot.yml

joycebrum:
#3449 

## Changes

I'm submiting a configuration file for dependabot (official from GitHub) to update only github workflows  and limited up to 3 PRs per week (let me know if you rather this to be greater of smaler)

I think it also has to enable `Dependabot version updated` at https://github/fmtlib/fmt/settings/security_analysis to actually works if merged.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions" # Necessary to update action hashs	
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    # Allow up to 3 opened pull requests for github-actions versions
+    open-pull-requests-limit: 3


.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions" # Necessary to update action hashs	
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    # Allow up to 3 opened pull requests for github-actions versions
+    open-pull-requests-limit: 3
```
vitaut, line 8: Do you have an example of such PR?

vitaut:
> https://github/fmtlib/fmt/settings/security_analysis

This link doesn't work.

florimond-collette:
I guess it should be [https://github/fmtlib/fmt/settings/security_analysis](https://github**.com**/fmtlib/fmt/settings/security_analysis)

.github/dependabot.yml
fredemmott: Here's one that's also using dependabot to update a pinned-by-hash github action: https://github.com/OpenKneeboard/OpenKneeboard/pull/353

vitaut:
Thank you!

PR #3453, opened 2023-05-21
CMake: Do not fail on unknown compiler features

glebm:
If CMake does not know much about a compiler, `target_compile_features` will fail. Issue a warning instead.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -297,7 +297,11 @@ if (FMT_PEDANTIC)
   target_compile_options(fmt PRIVATE ${PEDANTIC_COMPILE_FLAGS})
 endif ()
 
-target_compile_features(fmt PUBLIC cxx_std_11)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -297,7 +297,11 @@ if (FMT_PEDANTIC)
   target_compile_options(fmt PRIVATE ${PEDANTIC_COMPILE_FLAGS})
 endif ()
 
-target_compile_features(fmt PUBLIC cxx_std_11)
+if(cxx_std_11 IN_LIST CMAKE_CXX_COMPILE_FEATURES)
+  target_compile_features(fmt PUBLIC cxx_std_11)
+else()
+  message(WARNING "Feature cxx_std_11 is unknown for the CXX compiler")
+endif()
```
vitaut, line 304: nit: we use spaces after control flow CMake commands like if/else/endif, e.g.

```
if (...)
```

glebm: Done

vitaut:
Thanks

PR #3457, opened 2023-05-23
Pass correct Char to base format_as formatter

gix:
Returning a non-`char` `string_view` from `format_as` fails to compile since the base formatter always uses `char`.


vitaut:
Thank you

PR #3459, opened 2023-05-23
Make constexpr ceil

phprus:
Fix for https://github.com/fmtlib/fmt/commit/858e528abd2be97cf11f9a9c1c180d6ad1a4d255#r114625556

Broken on macOS 13.3.1
Apple clang version 14.0.3 (clang-1403.0.22.14.1)
C++20.
```
/.../fmt/test/compile-fp-test.cc:32: error: call to consteval function 'test_format<11UL, char, float, FMT_COMPILE_STRING>' is not a constant expression
/.../fmt/test/compile-fp-test.cc:32:27: error: call to consteval function 'test_format<11UL, char, float, FMT_COMPILE_STRING>' is not a constant expression
  EXPECT_EQ("392.500000", test_format<11>(FMT_COMPILE("{0:f}"), 392.5f));
                          ^
/.../fmt/include/fmt/format.h:3351:9: note: non-constexpr function 'ceil' cannot be used in a constant expression
        std::ceil((f.e + count_digits<1>(f.f) - 1) * inv_log2_10 - 1e-10));
```

``std::ceil`` is constexpr since C++23 (https://en.cppreference.com/w/cpp/numeric/math/ceil)

This is not a compiler bug. This is unimplemented feature in clang and msvs. Only gcc support it.
https://en.cppreference.com/w/cpp/compiler_support

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -3322,19 +3322,16 @@ template <typename Float> FMT_CONSTEXPR auto iceil(Float value) -> int {
   auto max = (std::numeric_limits<int>::max)();
   ignore_unused(min, max);
   FMT_ASSERT(value >= min && value <= max, "value not in int range");
-  if (is_constant_evaluated()) {
-    do {
-      auto mid = min + static_cast<int>((static_cast<unsigned>(max) -
-                                         static_cast<unsigned>(min)) /
-                                        2);
-      if (mid < value)
-        min = mid;
-      else
-        max = mid;
-    } while (min + 1 != max);
-    return max;
-  }
-  return static_cast<int>(std::ceil(value));
+  do {
+    auto mid = min + static_cast<int>((static_cast<unsigned>(max) -
+                                       static_cast<unsigned>(min)) /
+                                      2);
+    if (mid < value)
+      min = mid;
+    else
+      max = mid;
+  } while (min + 1 != max);
+  return max;
 }
 
 template <typename Float>
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -3317,6 +3317,26 @@ FMT_CONSTEXPR20 void format_hexfloat(Float value, int precision,
   format_hexfloat(static_cast<double>(value), precision, specs, buf);
 }
 
+template <typename Float> FMT_CONSTEXPR auto iceil(Float value) -> int {
+  auto min = (std::numeric_limits<int>::min)();
+  auto max = (std::numeric_limits<int>::max)();
+  ignore_unused(min, max);
+  FMT_ASSERT(value >= min && value <= max, "value not in int range");
+  if (is_constant_evaluated()) {
+    do {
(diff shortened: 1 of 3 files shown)


include/fmt/format.h
```diff
@@ -3317,6 +3317,32 @@ FMT_CONSTEXPR20 void format_hexfloat(Float value, int precision,
   format_hexfloat(static_cast<double>(value), precision, specs, buf);
 }
 
+template <typename Integral, typename Float>
+FMT_CONSTEXPR auto iceil(Float value) -> Integral {
+  static_assert(
+      std::is_integral<Integral>::value && !std::is_same<Integral, bool>::value,
+      "Return type is not integer");
```
vitaut, line 3324: Then this can go away.

include/fmt/format.h
```diff
@@ -3317,6 +3317,32 @@ FMT_CONSTEXPR20 void format_hexfloat(Float value, int precision,
...
+  auto min = (std::numeric_limits<Integral>::min)();
+  auto max = (std::numeric_limits<Integral>::max)();
+  ignore_unused(min, max);
+  FMT_ASSERT(min <= value && value <= max, "value not in Integral range");
+  if (is_constant_evaluated()) {
```
vitaut, line 3329: I think inverting the condition here will be more readable.

vitaut (changes requested):
Thanks for the PR!

include/fmt/format.h
```diff
@@ -3317,6 +3317,26 @@ FMT_CONSTEXPR20 void format_hexfloat(Float value, int precision,
   format_hexfloat(static_cast<double>(value), precision, specs, buf);
 }
 
+template <typename Float> FMT_CONSTEXPR auto iceil(Float value) -> int {
+  auto min = (std::numeric_limits<int>::min)();
+  auto max = (std::numeric_limits<int>::max)();
+  ignore_unused(min, max);
+  FMT_ASSERT(value >= min && value <= max, "value not in int range");
+  if (is_constant_evaluated()) {
```
vitaut, line 3325: On a second thought, I don't think we need this check at all because `iceil` is not called at runtime except in tests.

test/format-impl-test.cc
```diff
@@ -523,3 +523,37 @@ TEST(format_impl_test, to_utf8) {
...
+           0.2,
+           1.2,
+           4.0,
+           ((std::numeric_limits<int>::max)() - 0.5),
+       }) {
+    auto r = fmt::detail::iceil(v);
+    fmt::detail::ignore_unused(r);
+  }
+  return true;
+}
```
vitaut, line 542: Once the branching on constant evaluation is removed, we won't need this anymore.

include/fmt/format.h
phprus: Done

vitaut (approved):

vitaut:
Thanks!

PR #3467, opened 2023-05-30
Add VxWorks7 user space and kernel space support

wenshan1:
Add VxWorks7 user space and kernel space support.
All test cases are passed in Ubuntu Linux 22.04.

### DIFF
--- a/src/os.cc
+++ b/src/os.cc
@@ -18,6 +18,10 @@
 #  include <sys/stat.h>
 #  include <sys/types.h>
 
+#  ifdef _WRS_KERNEL   //added VxWorks7 kernel support
+#    include <ioLib.h> // header file for getpagesize()
+#  endif
+
 #  ifndef _WIN32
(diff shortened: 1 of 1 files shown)


src/os.cc
```diff
@@ -20,6 +20,9 @@
 
 #  ifndef _WIN32
 #    include <unistd.h>
+#    ifdef _WRS_KERNEL  //added VxWorks7 kernel support
+#        include <ioLib.h>
+#    endif
```
vitaut, line 25: I think this better be moved outside of `ifndef _WIN32` block because those are independent.
Also it would be nice to add a comment clarifying that the new include is for `getpagesize`, something like 

```c++
# include <ioLib.h>  // getpagesize
```

wenshan1: Thank you for your review. 
I have changed my code.

vitaut:
Thanks!

PR #3476, opened 2023-06-02
Use correct Char type in std::filesystem::path formatter

hmbj:
Fixed template specialization for std::filesystem::path formatter.

Godbolt repro:
https://godbolt.org/z/o4bjG6ddo


vitaut:
Thank you!

PR #3482, opened 2023-06-11
Use FMT_TRY and FMT_CATCH in std.h

Minty-Meeo:
This naked try-catch block prevents compilation when exceptions are disabled.


vitaut:
Thanks

PR #3485, opened 2023-06-12
Give basic_memory_buffer allocator [[no_unique_address]]

Minty-Meeo:
This allows stateless allocators to take up no space while still avoiding the empty base class optimization.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -79,6 +79,18 @@
 #  endif
 #endif
 
+#ifndef FMT_NO_UNIQUE_ADDRESS
+#  if FMT_CPLUSPLUS >= 202002L
+#    if FMT_HAS_CPP_ATTRIBUTE(no_unique_address)
+#      define FMT_NO_UNIQUE_ADDRESS [[no_unique_address]]
+#    elif FMT_MSC_VERSION >= 1929 // VS2019 v16.10 and later
+#      define FMT_NO_UNIQUE_ADDRESS [[msvc::no_unique_address]]
+#    endif
+#  else
+#    define FMT_NO_UNIQUE_ADDRESS
+#  endif
+#endif
+
 #if FMT_GCC_VERSION || defined(__clang__)
 #  define FMT_VISIBILITY(value) __attribute__((visibility(value)))
 #else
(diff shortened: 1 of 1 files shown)


Minty-Meeo:
Following the advice from https://devblogs.microsoft.com/cppblog/msvc-cpp20-and-the-std-cpp20-switch/ on how to use [[msvc::no_unique_address]].

include/fmt/core.h
```diff
@@ -154,6 +157,16 @@
...
+#ifndef FMT_NO_UNIQUE_ADDRESS
+#  if FMT_HAS_CPP20_ATTRIBUTE(no_unique_address)
+#    define FMT_NO_UNIQUE_ADDRESS [[no_unique_address]]
+#  elif FMT_MSC_VERSION >= 1929 && FMT_CPLUSPLUS >= 202002L // VS2019 v16.10 and later
+#    define FMT_NO_UNIQUE_ADDRESS [[msvc::no_unique_address]]
+#  else
+#    define FMT_NO_UNIQUE_ADDRESS
+#  endif
+#endif
+
```
vitaut, line 169: Please move to `format.h` where it is used.

include/fmt/core.h
```diff
@@ -87,6 +87,9 @@
 #define FMT_HAS_CPP17_ATTRIBUTE(attribute) \
   (FMT_CPLUSPLUS >= 201703L && FMT_HAS_CPP_ATTRIBUTE(attribute))
 
+#define FMT_HAS_CPP20_ATTRIBUTE(attribute) \
+  (FMT_CPLUSPLUS >= 202002L && FMT_HAS_CPP_ATTRIBUTE(attribute))
```
vitaut, line 91: I don't think we need this for a single check esp. since MSVC requires special handling. Let's just inline in the call site.

vitaut (changes requested):
Thanks for the PR

include/fmt/core.h
Minty-Meeo: Would it not be better to create this feature now, in the event that the other C++20 attributes (likely, unlikely) are desired in the future?

vitaut: There is little value in these variations of `FMT_HAS_CPP_ATTRIBUTE` so let's just inline.

vitaut (approved):

vitaut:
Thanks

Issue #3489, opened 2023-06-14
Cannot cmake --install since 10.0.0


petrmanek:
Ever since I upgraded to 10.0.0 I am experiencing issues running cmake install on my project. This appears in 10.0.0 as well as 8732ad877369172f863fd14e9600306ba8bb5da2 (latest master), so I thought I would be a good idea to bring to your attention. Full disclosure, I am uncertain whether this is a bug on my end or yours -- one reason that leads me to believe this is of concern upstream is that my usage of fmtlib is conventional, so there are not really that many issues where things could be going wrong

I include fmtlib in my project as a git submodule, using the following directory tree:

```
├── build
├── CMakeLists.txt
└── lib
    ├── CMakeLists.txt
    └── fmt            <---- this is a git submodule
```

Then wire it up in cmake like so:

1. My project's `CMakeLists.txt` includes `lib/CMakeLists.txt`
2. The `lib/CMakeLists.txt` sets `FMT_SHARED=ON`
3. Then `lib/CMakeLists.txt`  includes `lib/fmt/CMakeLists.txt`
4. fmtlib's `CMakeLists.txt` does its thing
5. After that,  `lib/CMakeLists.txt`  specifies `install(TARGETS fmt)`, so that the built version of fmtlib is shipped with the rest of my project.

At compilation- and runtime, everything works as expected. The issue appears when I attempt to install the project. I see the following message:

```bash
$ cmake --install .. --prefix $(pwd)                                                                                                                                                  
-- Install configuration: "Debug"
-- Installing: <path to repo redacted>/build/asd/lib/libfmtd.so.10.0.1
-- Installing: <path to repo redacted>/build/asd/lib/libfmtd.so.10
-- Set runtime path of "<path to repo redacted>/build/asd/lib/libfmtd.so.10.0.1" to "/usr/lib/<project name redacted>:/usr/lib64/<project name redacted>:$ORIGIN:$ORIGIN/../lib"
-- Installing: <path to repo redacted>/build/asd/lib/libfmtd.so
CMake Error at <path to repo redacted>/build/lib/cmake_install.cmake:99 (file):
  file INSTALL cannot find
  "<path to repo redacted>/lib/include/fmt/args.h": No such file
  or directory.
Call Stack (most recent call first):
  <path to repo redacted>/build/cmake_install.cmake:47 (include)
```

[...]

PR #3496, opened 2023-06-18
Revert ae25f79 to restore capability to `cmake --install` when included as a subproject

petrmanek:
Following the discussion in #3489, this PR reverts commit ae25f7968efca30a6483ad7802ccfa5419f6330a.

This commit introduces a relative path dependency to public header files, which does not hold when fmtlib is included as non-root project in more complex directory tree. At a later point during a call to `cmake --install` this relative path is resolved w.r.t. the path of the root project rather than the fmtlib project, causing a failure.

Since this PR does not add anything new, compliance with coding conventions is trivially satisfied. Furthermore, the submitter hereby agrees to fmtlib's license conditions.


vitaut:
Thank you

PR #3508, opened 2023-06-25
Fix typo in assert message.

TobiSchluter:
I noticed this one while looking at the compiler output in godbolt. I reviewed all strings in the file but this was the only typo I found.

oh, and sorry about the typo in the commit message :(


vitaut:
Thanks

PR #3515, opened 2023-06-29
README: Add link to merged clang-tidy check

mikecrowe:
A much-improved version of the main clang-tidy-fmt check for converting printf and fprintf has landed in upstream LLVM. It converts to std::print by default, but can be configured to convert to fmt::print instead. It makes more sense for the README to point to that version instead now.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -295,9 +295,12 @@ or the bloat test::
 Migrating code
 --------------
 
-`clang-tidy-fmt <https://github.com/mikecrowe/clang-tidy-fmt>`_ provides clang
-tidy checks for converting occurrences of ``printf`` and ``fprintf`` to
-``fmt::print``.
+`clang-tidy <https://clang.llvm.org/extra/clang-tidy/>`_ v17 (not yet
+released) provides the `modernize-use-std-print
+<https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-std-print.html>`_
+check that is capable of converting occurrences of ``printf`` and
+``fprintf`` to ``fmt::print`` if configured to do so. (By default it
+converts to ``std::print``.)
 
 Projects using this library
 ---------------------------


mikecrowe:
Since you're the world experts on the format specification syntax, it would
be lovely if you could also have a look at the test cases for the check at
https://github.com/llvm/llvm-project/blob/main/clang-tools-extra/test/clang-tidy/checkers/modernize/use-std-print.cpp
to see if there are any glaring omissions or mistakes.

Thanks!

vitaut:
Thank you!

In https://github.com/llvm/llvm-project/blob/fc37f717770acdfe5504bb9b969a01bb16a187f9/clang-tools-extra/test/clang-tidy/checkers/modernize/use-std-print.cpp#L339C34-L339C36
why is the format string `"{} {{[{][{]}}}}"` and not `"{} {{}}"?

Also note that the order of dynamic width/precision and the argument it applies to should be reversed in https://github.com/llvm/llvm-project/blob/fc37f717770acdfe5504bb9b969a01bb16a187f9/clang-tools-extra/test/clang-tidy/checkers/modernize/use-std-print.cpp#L1025 and elsewhere.

mikecrowe:
> In https://github.com/llvm/llvm-project/blob/fc37f717770acdfe5504bb9b969a01bb16a187f9/clang-tools-extra/test/clang-tidy/checkers/modernize/use-std-print.cpp#L339C34-L339C36
why is the format string "{} {{[{][{]}}}}" and not `"{} {{}}"?

LLVM FileCheckexpects anything in `{{` and `}}` to be a regex. So, to match literal `{{}}` I need to write a regex that matches it and enclose it in those characters. The [documentation for FileCheck](https://llvm.org/docs/CommandGuide/FileCheck.html#filecheck-regex-matching-syntax) says:
> In the rare case that you want to match double braces explicitly from the input, you can use something ugly like {{[}][}]}} as your pattern.

which is almost what I needed.

> Also note that the order of dynamic width/precision and the argument it applies to should be reversed in https://github.com/llvm/llvm-project/blob/fc37f717770acdfe5504bb9b969a01bb16a187f9/clang-tools-extra/test/clang-tidy/checkers/modernize/use-std-print.cpp#L1025 and elsewhere.

Thanks for the review. I shall try to implement that - if I can fathom out how!

mikecrowe:
> Also note that the order of dynamic width/precision and the argument it applies to should be reversed in https://github.com/llvm/llvm-project/blob/fc37f717770acdfe5504bb9b969a01bb16a187f9/clang-tools-extra/test/clang-tidy/checkers/modernize/use-std-print.cpp#L1025 and elsewhere.

Hopefully fixed by [D154283](https://reviews.llvm.org/D154283). Thanks for pointing out the flaw. It was easier to fix than I expected.

PR #3521, opened 2023-07-04
constexpr-ed a 'shift', of constexpr variables

TheMR-777:
Credits to, **Code Analyzer of Visual Studio 2022 (MSVC 17.6.4)**.


TheMR-777:
> Thanks for the PR but constexpr is not needed here since the constant will be optimized anyway.

Oh alright, welcome :)

Issue #3517, opened 2023-06-30
Install doesn't install header files


mborn-adi:
On Windows, install only installs everything except the header files. In CMakeLists.txt, the install is set up like this:
```
  # Install the library and headers.
  install(TARGETS ${INSTALL_TARGETS} EXPORT ${targets_export_name}
          LIBRARY DESTINATION ${FMT_LIB_DIR}
          ARCHIVE DESTINATION ${FMT_LIB_DIR}
          PUBLIC_HEADER DESTINATION "${FMT_INC_DIR}/fmt"
          RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
```
But nowhere else are the header files marked "PUBLIC_HEADER".

Adding
```
set_property(TARGET fmt PROPERTY PUBLIC_HEADER ${FMT_HEADERS})
```
in line 289 right after
```
add_library(fmt::fmt ALIAS fmt)
```
does just that. But cmake is not my strongest suit.

Best Regards,
Mathias

vitaut:
I was managed to repro this on macOS and it looks like a recent regression related to modules work.

rxguy:
I am getting this today with Fedora Linux.

vitaut:
Header installation was working on https://github.com/fmtlib/fmt/commit/75f3b1c09485986427b8a46272f6b9570366e1aa.

vitaut:
Bisected the regression to 13156e54bf91e44641ce3aac041d31f9a15a8042.

PR #3523, opened 2023-07-05
fix missing header files when using CMake install

setoye:
this pr fixes https://github.com/fmtlib/fmt/issues/3517#issue-1782395710


vitaut:
Thank you!

lanfeust69:
@vitaut : I'm a bit surprised this got merged. It surely fixes #3517, but, as it is, at the cost of re-introducing #3489, it seems. Maybe the [suggested patch](https://github.com/fmtlib/fmt/issues/3489#issuecomment-1595869444) could avoid playing ping-pong between the two issues ?

PR #3531, opened 2023-07-15
Add an option to disable file descriptors

glebm:
Systems such as Win98 and the original Xbox do not provide file numeric file descriptors.

Allows the user to selectively disable the functionality that relies on numeric file descriptors.

---

Not sure if this is the right way to go about it, advice appreciated.
I'm trying to extract the changes needed to build for NXDK and Win98 into separate options.

Here are all the changes needed for NXDK: https://github.com/diasurgical/fmt/commit/e421e854efcb21ee641349bab604d91b34af26b8

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -338,6 +338,13 @@ target_include_directories(fmt-header-only ${FMT_SYSTEM_HEADERS_ATTRIBUTE} INTER
   $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
   $<INSTALL_INTERFACE:${FMT_INC_DIR}>)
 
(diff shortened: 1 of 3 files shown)


vitaut:
Descriptor-based API are provided by the C runtime, not by the operating system API. Normally the latter provides `_fileno` on Windows. Is it documented somewhere that it is not available on Win98?

vitaut:
Anyway, I don't think we can disable descriptors in such way because it will break a lot of tests. You could create a custom build config with `os.h` excluded for this legacy system(s).

glebm:
Ah, I didn't realize the whole thing can be disabled entirely. This is great, thanks!

PR #3544, opened 2023-07-21
Turn off error-producing NVCC workaround when using c++20

alexeedm:
Turn off NVCC workaround for c++20 standard. It's not need and triggers multiple errors instead.


vitaut:
Thanks

PR #3550, opened 2023-07-25
Remove std::copy usage.

Ozomahtli:
Fixes missing `<algorithm>` include error.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -2896,7 +2896,7 @@ class bigint {
     auto size = other.bigits_.size();
     bigits_.resize(size);
     auto data = other.bigits_.data();
-    std::copy(data, data + size, bigits_.data());
+    std::uninitialized_copy(data, data + size, bigits_.data());
     exp_ = other.exp_;
   }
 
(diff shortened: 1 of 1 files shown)


include/fmt/format.h
```diff
@@ -2896,7 +2896,7 @@ class bigint {
     auto size = other.bigits_.size();
     bigits_.resize(size);
     auto data = other.bigits_.data();
-    std::copy(data, data + size, bigits_.data());
+    std::uninitialized_copy(data, data + size, bigits_.data());
```
vitaut, line 2899: Looks like `uninitialized_copy` is not `constexpr`. Let's replace it with

```c++
copy_str<bigit>(data, data + size, bigits_.data());
```

vitaut (changes requested):
Thanks for the PR!

include/fmt/format.h
Ozomahtli: Done!

vitaut (approved):

vitaut:
Thanks!

Issue #3529, opened 2023-07-14
no member named 'on_error' in 'fmt::basic_format_parse_context<char>'


mjerabek:
The [example for custom formatter](https://fmt.dev/latest/api.html#formatting-user-defined-types) specifies that in case of format parse error, one should call `ctx.on_error("invalid format")`. But this method got removed in d1745084e03cde6d055f3db293392197815b32d9 and the impl uses `detail::throw_format_error`, which should not be used from outsude.

What is the recommended way to report parse errors now?

Works in 9.1.0. Broken in 10.0.0. Still fails on master.

godbolt link: https://godbolt.org/z/dEWEKEsTc

mjerabek:
I worked this around by defining my own `throw_format_error` function and using that, but I think the lib should expose such a function. One can just `throw format_error(...)` directly, but then including `<fmt/core.h>` is not enough (and including `<fmt/format.h>` in a header is too heavy).

vitaut:
Duplicate of https://github.com/fmtlib/fmt/issues/3458.

mjerabek:
Ah, right, didn't find that.

The example is still wrong, though, as now including `<fmt/core.h>` is not enough. We could change the include in just the one example (as the rest doesn't need `<fmt/format.h>`).

But IMO having the throwing function available directly from the library (and having the includes consistent) would be less confusing.

I can create a PR with whatever variant you want to use if you want.

vitaut:
Good catch, thanks. Fixed the include in https://github.com/fmtlib/fmt/commit/8e87d3a8bead3f37eb513af093f58a4222f45038.

A PR to expose `throw_format_error` via the public API is welcome but note that for ABI compatibility reasons it should be done via an alias rather than moving the function to another namespace.

PR #3551, opened 2023-07-26
expose detail::throw_format_error

mjerabek:
This enables to only #include <fmt/core.h> when defining formatters.

Addresses https://github.com/fmtlib/fmt/issues/3529.


vitaut:
Thanks

PR #3552, opened 2023-07-27
Add missing inline specifier

razaqq:
Add missing inline specifier for `auto get_path_string<char>(const std::filesystem::path& p)`, causing duplicate symbol build errors on windows


vitaut:
Thanks!

PR #3561, opened 2023-07-30
fix ambiguous formatter lookup for flat_set

### LINKED ISSUE #3556: Ambiguous partial specialization when formatting boost `flat_set` since fmt v10.0.0
When trying to format a `flat_set` from boost, an ambiguous lookup for `formatter` happens.

Reproducer:
https://godbolt.org/z/dhqnMf755
boost version 1.82.0 (changing it to some random previous versions did not change something) 
Clang 16.0.0 with fmt trunk
gcc 13.2 with fmt trunk
gcc 13.2 with fmt 9.1.0 as a proof it worked with fmt 9.1.0

The ambiguous formatters are in lines
https://github.com/fmtlib/fmt/blob/757564f5cd2fa78b634dd698c63dbf069818e6fd/include/fmt/ranges.h#L560
and
https://github.com/fmtlib/fmt/blob/757564f5cd2fa78b634dd698c63dbf069818e6fd/include/fmt/ranges.h#L671

5chmidti:
Fixes #3556

The underlying issue was that the `flat_set` of boost (and in theory of std), could match `is_set` and `is_container_adaptor_like` at the same time. This resulted in an ambiguous lookup of the formatters for ranges and range adaptors.

I added the change that @vitaut suggested in the issue and tested the change with a minimal reproducer of the original issue (a range that matches `is_set` and `is_container_adaptor_like`).

### DIFF
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -668,8 +668,11 @@ template <typename Container> struct all {
 }  // namespace detail
 
 template <typename T, typename Char>
-struct formatter<T, Char,
-                 enable_if_t<detail::is_container_adaptor_like<T>::value>>
+struct formatter<
+    T, Char,
+    enable_if_t<conjunction<detail::is_container_adaptor_like<T>,
+                            bool_constant<range_format_kind<T, Char>::value ==
+                                          range_format::disabled>>::value>>
     : formatter<detail::all<typename T::container_type>, Char> {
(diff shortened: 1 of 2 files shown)


vitaut (changes requested):
Thanks for the PR. Overall looks good but please fix build failures.

vitaut (approved):

Issue #3560, opened 2023-07-29
Many packages fail to compile with libfmt-10


yurivict:
For example all "Failed" lines in the following log:
```

[00:39:58] [01] [00:31:00] Finished emulators/dolphin-emu | dolphin-emu-5.0.16795_5: Failed: build
[00:40:00] [01] [00:00:01] Building net/ceph14 | ceph14-14.2.22_11
[00:53:44] [03] [00:51:56] Finished graphics/openimageio | openimageio-2.4.13.0_1: Success
[00:53:49] [03] [00:00:02] Building graphics/py-openimageio@py39 | py39-openimageio-2.4.13.0_1
[01:10:18] [08] [00:32:16] Finished multimedia/kodi | kodi-20.2_1: Failed: build
[01:10:39] [08] [00:00:01] Building cad/openroad | openroad-2.0.2930_4,1
[01:19:22] [06] [01:15:37] Finished lang/solidity | solidity-0.8.21: Failed: build
[01:19:25] [06] [00:00:00] Building lang/lfortran | lfortran-0.19.0_2
[01:22:09] [02] [01:00:28] Finished net/ecal | ecal-5.11.5: Success
[01:22:12] [02] [00:00:00] Building net-p2p/litecoin-daemon | litecoin-daemon-0.21.2.1_5
[01:27:48] [01] [00:47:49] Finished net/ceph14 | ceph14-14.2.22_11: Failed: build
[01:27:56] [01] [00:00:00] Building science/openmc | openmc-0.13.3_1
[01:43:19] [07] [01:41:31] Finished games/0ad | 0ad-0.0.26_13: Failed: build
[01:43:30] [07] [00:00:00] Building net-im/telegram-desktop@qt5 | telegram-desktop-4.8.4_2
[01:49:13] [01] [00:21:17] Finished science/openmc | openmc-0.13.3_1: Success
[01:49:14] [01] [00:00:00] Building net/knxd | knxd-0.14.59
[02:00:16] [01] [00:11:02] Finished net/knxd | knxd-0.14.59: Failed: build
[02:00:17] [01] [00:00:00] Building audio/musicpd | musicpd-0.23.13
[02:01:55] [06] [00:42:30] Finished lang/lfortran | lfortran-0.19.0_2: Success
[02:01:59] [06] [00:00:00] Building devel/bear | bear-3.1.2_1
[02:08:47] [08] [00:58:09] Finished cad/openroad | openroad-2.0.2930_4,1: Failed: build
[...]

PR #3562, opened 2023-07-31
cmake: respect semantic versioning

abouvier:
`fmt` is [clearly incompatible between different major versions ](https://github.com/fmtlib/fmt/issues/3560).
This will allow `find_package(fmt 9 REQUIRED)` to reject version 10.


vitaut:
Thanks for the PR but compatibility should be specified by the caller and not here. It is easy to write code that is compatible between major versions and it is possible to write code that is incompatible between minor versions.

PR #3563, opened 2023-07-31
cmake: fix FMT_PKGCONFIG_DIR path

abouvier:
The `PATH` keyword will make the path absolute. Use `STRING` like in the other `FMT_*_DIR` variables.


vitaut:
Thank you

PR #3568, opened 2023-08-02
Add fmt::println without args

bugdea1er:
Hi there!

I added a println function with no arguments. This function allows you to output an empty string to the console, which can be useful for output formatting.

It is just a short way to write this:
```cpp
fmt::println("");
```

The Java language also has a println function with no arguments, which prints only a newline to the console. This println function in the fmt library will work similarly.


vitaut:
Thanks for the PR but I don't think it's worth adding a new API for this specific case considering that it's trivial to do by other means.

PR #3570, opened 2023-08-04
Added formatter for bit_reference-like types

phprus:
Fix for issue #3567

Test is based on #3569

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -8,6 +8,7 @@
 #ifndef FMT_STD_H_
 #define FMT_STD_H_
 
+#include <bitset>
 #include <cstdlib>
 #include <exception>
 #include <memory>
@@ -15,6 +16,7 @@
 #include <type_traits>
 #include <typeinfo>
 #include <utility>
+#include <vector>
 
 #include "format.h"
 #include "ostream.h"
@@ -389,6 +391,50 @@ struct formatter<
 #endif
   }
 };
+
+namespace detail {
+
+template <typename T, typename Enable = void>
+struct has_flip : std::false_type {};
+
+template <typename T>
+struct has_flip<T, void_t<decltype(std::declval<T>().flip())>>
+    : std::true_type {};
+
+template <typename T> struct is_bit_reference_like {
+  static constexpr const bool value =
+      std::is_convertible<T, bool>::value &&
+      std::is_nothrow_assignable<T, bool>::value && has_flip<T>::value;
+};
+
+#ifdef _LIBCPP_VERSION
+
+// Workaround for libc++ incompatibility with C++ standard.
+// According to the Standard, `bitset::operator[] const` returns bool.
+template <typename C>
+struct is_bit_reference_like<std::__bit_const_reference<C>> {
+  static constexpr const bool value = true;
+};
+
+#endif
+
+}  // namespace detail
+
+// We can't use std::vector<bool, Allocator>::reference and
+// std::bitset<N>::reference because the compiler can't deduce Allocator and N
+// in partial specialization.
+FMT_EXPORT
+template <typename BitRef, typename Char>
+struct formatter<BitRef, Char,
+                 enable_if_t<detail::is_bit_reference_like<BitRef>::value>>
+    : formatter<bool, Char> {
+  template <typename FormatContext>
+  FMT_CONSTEXPR auto format(const BitRef& v, FormatContext& ctx) const
+      -> decltype(ctx.out()) {
+    return formatter<bool, Char>::format(v, ctx);
+  }
+};
+
 FMT_END_NAMESPACE
 
 #endif  // FMT_STD_H_
--- a/test/std-test.cc
+++ b/test/std-test.cc
@@ -7,6 +7,7 @@
 
 #include "fmt/std.h"
 
+#include <bitset>
 #include <stdexcept>
 #include <string>
 #include <vector>
@@ -222,7 +223,9 @@ TEST(std_test, exception) {
 #endif
 }
 
-TEST(std_test, format_vector_bool_specialization) {
+TEST(std_test, format_bit_reference) {
+  std::bitset<2> bs(1);
+  EXPECT_EQ(fmt::format("{} {}", bs[0], bs[1]), "true false");
   std::vector<bool> v = {true, false};
   EXPECT_EQ(fmt::format("{} {}", v[0], v[1]), "true false");
 }
--- a/test/std-test.cc
+++ b/test/std-test.cc
@@ -221,3 +221,8 @@ TEST(std_test, exception) {
   }
 #endif
 }
+
+TEST(std_test, format_vector_bool_specialization) {
+  std::vector<bool> v = {true, false};
+  EXPECT_EQ(fmt::format("{} {}", v[0], v[1]), "true false");
+}
--- a/test/std-test.cc
+++ b/test/std-test.cc
@@ -229,3 +229,10 @@ TEST(std_test, format_bit_reference) {
   std::vector<bool> v = {true, false};
   EXPECT_EQ(fmt::format("{} {}", v[0], v[1]), "true false");
 }
+
+TEST(std_test, format_const_bit_reference) {
+  const std::bitset<2> bs(1);
+  EXPECT_EQ(fmt::format("{} {}", bs[0], bs[1]), "true false");
+  const std::vector<bool> v = {true, false};
+  EXPECT_EQ(fmt::format("{} {}", v[0], v[1]), "true false");
+}
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -13,6 +13,7 @@
 #include <cstring>  // std::strlen
 #include <iterator>
 #include <limits>
+#include <memory>  // std::addressof
 #include <string>
 #include <type_traits>
 
@@ -1281,9 +1282,9 @@ template <typename Context> class value {
   FMT_INLINE value(const named_arg_info<char_type>* args, size_t size)
       : named_args{args, size} {}
 
-  template <typename T> FMT_CONSTEXPR FMT_INLINE value(T& val) {
+  template <typename T> FMT_CONSTEXPR20 FMT_INLINE value(T& val) {
     using value_type = remove_const_t<T>;
-    custom.value = const_cast<value_type*>(&val);
+    custom.value = const_cast<value_type*>(std::addressof(val));
     // Get the formatter type through the context to allow different contexts
     // have different extension points, e.g. `formatter<T>` for `format` and
     // `printf_formatter<T>` for `printf`.


phprus:
Final version.
Ready for review.

include/fmt/std.h
```diff
@@ -389,6 +390,16 @@ struct formatter<
 #endif
   }
 };
+
+FMT_EXPORT
+template <typename Char>
+struct formatter<std::vector<bool>::reference, Char> : formatter<bool, Char> {
```
wangzw, line 396: Maybe need to handle non default allocator

phprus: We can't deduce Allocator type in this context.

```
/Users/phprus/Devel/fmt/include/fmt/std.h:395:8: error: class template partial specialization contains a template parameter that cannot be deduced; this partial specialization will never be used [-Wunusable-partial-specialization]
struct formatter<typename std::vector<bool, Allocator>::reference, Char>
       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/phprus/Devel/fmt/include/fmt/std.h:394:20: note: non-deducible template parameter 'Allocator'
template <typename Allocator, typename Char>
                   ^
```

phprus:
@vitaut

Final-2 version :)
Ready for review.

include/fmt/std.h
```diff
@@ -389,6 +391,38 @@ struct formatter<
...
+template <typename T>
+struct has_flip<T, void_t<decltype(std::declval<T>().flip())>>
+    : std::true_type {};
+
+template <typename T> struct is_bit_reference_like {
+  static constexpr const bool value =
+      std::is_convertible<T, bool>::value &&
+      std::is_nothrow_assignable<T, bool>::value && has_flip<T>::value
+
+      ;
```
wangzw, line 409: formatter issue?

phprus: Unexpected clang-format behaviour.
Fixed.

test/std-test.cc
```diff
@@ -221,3 +222,10 @@ TEST(std_test, exception) {
   }
 #endif
 }
+
+TEST(std_test, format_bit_reference) {
+  std::bitset<2> bs(1);
+  EXPECT_EQ(fmt::format("{} {}", bs[0], bs[1]), "true false");
+  std::vector<bool> v = {true, false};
+  EXPECT_EQ(fmt::format("{} {}", v[0], v[1]), "true false");
```
wangzw, line 230: try this

```
  const std::vector<bool> cv = {true, false};
  EXPECT_EQ(fmt::format("{} {}", cv[0], cv[1]), "true false");
```

```
[ 90%] Building CXX object test/CMakeFiles/std-test.dir/std-test.cc.o
In file included from /Users/wangzw/workspace/fmt/test/std-test.cc:8:
In file included from /Users/wangzw/workspace/fmt/include/fmt/std.h:21:
In file included from /Users/wangzw/workspace/fmt/include/fmt/format.h:49:
/Users/wangzw/workspace/fmt/include/fmt/core.h:1621:63: error: implicit instantiation of undefined template 'fmt::detail::type_is_unformattable_for<std::__bit_const_reference<std::vector<bool>>, char>'
    type_is_unformattable_for<T, typename Context::char_type> _;
                                                              ^
/Users/wangzw/workspace/fmt/include/fmt/core.h:1853:23: note: in instantiation of function template specialization 'fmt::detail::make_arg<true, fmt::basic_format_context<fmt::appender, char>, std::__bit_const_reference<std::vector<bool>>, 0>' requested here
        data_{detail::make_arg<is_packed, Context>(args)...} {
                      ^
/Users/wangzw/workspace/fmt/include/fmt/core.h:1871:10: note: in instantiation of function template specialization 'fmt::format_arg_store<fmt::basic_format_context<fmt::appender, char>, std::__bit_const_reference<std::vector<bool>>, std::__bit_const_reference<std::vector<bool>>>::format_arg_store<std::__bit_const_reference<std::vector<bool>>, std::__bit_const_reference<std::vector<bool>>>' requested here
  return {args...};
         ^
/Users/wangzw/workspace/fmt/include/fmt/core.h:2833:28: note: in instantiation of function template specialization 'fmt::make_format_args<fmt::basic_format_context<fmt::appender, char>, std::__bit_const_reference<std::vector<bool>>, std::__bit_const_reference<std::vector<bool>>>' requested here
  return vformat(fmt, fmt::make_format_args(args...));
                           ^
/Users/wangzw/workspace/fmt/test/std-test.cc:231:18: note: in instantiation of function template specialization 'fmt::format<std::__bit_const_reference<std::vector<bool>>, std::__bit_const_reference<std::vector<bool>>>' requested here
  EXPECT_EQ(fmt::format("{} {}", cv[0], cv[1]), "true false");
                 ^
/Users/wangzw/workspace/fmt/include/fmt/core.h:1599:45: note: template is declared here
template <typename T, typename Char> struct type_is_unformattable_for;
                                            ^
/Users/wangzw/workspace/fmt/include/fmt/core.h:1624:3: error: static assertion failed due to requirement 'formattable': Cannot format an argument. To make type T formattable provide a formatter<T> specialization: https://fmt.dev/latest/api.html#udt
  static_assert(
  ^
```

phprus: Fixed

vitaut (changes requested):
Thanks for the PR.

This looks like a lot of effort and implementation-specific code for such a corner case. I suggest just using std::addressoff and replacing FMT_CONSTEXPR with FMT_COSTEXPR20 in the value ctor overload that uses it. It only affects compile-time formatting anyway that requires a recent standard.

phprus:
@vitaut 
Done.

wangzw:
Works well for me. Thanks

include/fmt/std.h
```diff
@@ -389,6 +391,52 @@ struct formatter<
...
+      std::is_convertible<T, bool>::value &&
+      std::is_nothrow_assignable<T, bool>::value && has_flip<T>::value;
+};
+
+}  // namespace detail
+
+FMT_EXPORT
+template <typename BitRef, typename Char>
+struct formatter<BitRef, Char,
+                 enable_if_t<detail::is_bit_reference_like<BitRef>::value>>
```
vitaut, line 415: I don't think we need SFINAE here. Let's use `std::vector<bool>::reference`.

include/fmt/std.h
```diff
@@ -389,6 +391,52 @@ struct formatter<
...
+  }
+};
+
+#ifdef _LIBCPP_VERSION
+
+// Workaround for libc++ incompatibility with C++ standard.
+// According to the Standard, `bitset::operator[] const` returns bool.
+FMT_EXPORT
+template <typename C, typename Char>
+struct formatter<std::__bit_const_reference<C>, Char> : formatter<bool, Char> {
```
vitaut, line 430: Similarly, this should be `bitset::reference` and remove ifdef and a comment.

phprus: See: https://github.com/fmtlib/fmt/pull/3570#discussion_r1285057473

vitaut: I suggest moving this to a separate PR and adding more context why this is needed.

include/fmt/std.h
phprus: 1) The compiler won't be deduce the allocator type from ``std::vector<bool, Allocator>::reference`` or ``N`` value from ``std::bitset<N>::reference``.

2) What is ``std::vector<bool>::reference`` or ``std::bitset<N>::reference``?
On libc++ is ``std::__bit_reference`` and ``std::__bit_const_reference``.
On gcc and msvs ``std::vector<bool>::reference`` and ``std::bitset<N>::reference`` - is different proxy classes.

We need separate implementation for libc++ and gcc/msvc. But in this case, we will not be able to implement output for ``std::bitset``.

See:
https://github.com/fmtlib/fmt/pull/3570#discussion_r1284995184 (Allocator)
https://github.com/fmtlib/fmt/pull/3570#discussion_r1285057473 (const)

Previous branch:
https://github.com/fmtlib/fmt/tree/85bc49b83d59070a8023f529ec80c38652c14910

Commit with formatters: https://github.com/fmtlib/fmt/commit/64c905e27868c246289fd5d5e13663cc7410cfce
Failed tests: https://github.com/fmtlib/fmt/actions/runs/5769810223/job/15642249348 , https://github.com/fmtlib/fmt/actions/runs/5769810222/job/15642249612

include/fmt/std.h
phprus: > I suggest moving this to a separate PR and adding more context why this is needed.

I do not understand why.
Without this workaround we can't format values from ``const`` containers on libc++.

include/fmt/std.h
vitaut: Why do we care about allocator? Reference type should normally be the same between vectors with different allocators: https://godbolt.org/z/5sKM5b6vb.

phprus: No. It is implementation specific behaviour:
libc++: https://godbolt.org/z/3xbxj89hn

vitaut: That's unfortunate.

include/fmt/std.h
```diff
@@ -389,6 +391,52 @@ struct formatter<
...
+
+FMT_EXPORT
+template <typename BitRef, typename Char>
+struct formatter<BitRef, Char,
+                 enable_if_t<detail::is_bit_reference_like<BitRef>::value>>
+    : formatter<bool, Char> {
+  template <typename FormatContext>
+  FMT_CONSTEXPR auto format(const BitRef& v, FormatContext& ctx) const
+      -> decltype(ctx.out()) {
+    return formatter<bool, Char>::format(static_cast<bool>(v), ctx);
```
vitaut, line 420: I don't think this cast is needed because the reference should be implicitly convertible to bool.

include/fmt/std.h
vitaut: It would make sense to move if we used concrete types but in view of the above discussion it seems impossible so ignore this comment.

include/fmt/std.h
```diff
@@ -389,6 +391,52 @@ struct formatter<
...
+struct has_flip : std::false_type {};
+
+template <typename T>
+struct has_flip<T, void_t<decltype(std::declval<T>().flip())>>
+    : std::true_type {};
+
+template <typename T> struct is_bit_reference_like {
+  static constexpr const bool value =
+      std::is_convertible<T, bool>::value &&
+      std::is_nothrow_assignable<T, bool>::value && has_flip<T>::value;
```
vitaut, line 407: Can we remove the workaround for libc++ if we drop the `is_nothrow_assignable` check?

include/fmt/std.h
vitaut: A short comment explaining why we can't use `vector<bool>::reference` would be nice though.

include/fmt/std.h
phprus: No. ``__bit_const_reference`` does not contain ``flip`` method.

``is_bit_reference_like`` - is standard compatibility reference.
``__bit_const_reference`` is libc++ non standard issue.

include/fmt/std.h
phprus: Done.

include/fmt/std.h
phprus: Done.

include/fmt/std.h
```diff
@@ -389,6 +391,52 @@ struct formatter<
...
+template <typename C, typename Char>
+struct formatter<std::__bit_const_reference<C>, Char> : formatter<bool, Char> {
+  template <typename FormatContext>
+  FMT_CONSTEXPR20 auto format(const std::__bit_const_reference<C>& v,
+                              FormatContext& ctx) const -> decltype(ctx.out()) {
+    return formatter<bool, Char>::format(v, ctx);
+  }
+};
+
+#endif
```
vitaut, line 441: I think this can be replaced by a specialization of `is_bit_reference_like` for `__bit_const_reference` which seems a bit cleaner (no duplicate formatter).

phprus: Done.

vitaut (approved):

vitaut:
Thank you!

PR #3574, opened 2023-08-06
Add formatter for std::atomic

wangzw:
There is no formatter for `std::atomic` and `std:errc` and inject `format_as` in `std` namespace is not a good idea.

This patch implement formatter for above types.

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -8,6 +8,7 @@
 #ifndef FMT_STD_H_
 #define FMT_STD_H_
 
+#include <atomic>
 #include <bitset>
 #include <cstdlib>
 #include <exception>
@@ -435,6 +436,17 @@ struct formatter<BitRef, Char,
   }
 };
 
-FMT_END_NAMESPACE
+FMT_EXPORT
+template <typename T, typename Char>
+struct formatter<std::atomic<T>, Char,
+                 enable_if_t<is_formattable<T, Char>::value>>
+    : formatter<T, Char> {
+  template <typename FormatContext>
+  auto format(const std::atomic<T>& v, FormatContext& ctx) const
+      -> decltype(ctx.out()) {
+    return formatter<T, Char>::format(v.load(), ctx);
+  }
+};
 
+FMT_END_NAMESPACE
 #endif  // FMT_STD_H_
--- a/test/std-test.cc
+++ b/test/std-test.cc
@@ -236,3 +236,11 @@ TEST(std_test, format_const_bit_reference) {
   const std::vector<bool> v = {true, false};
   EXPECT_EQ(fmt::format("{} {}", v[0], v[1]), "true false");
 }
+
+TEST(std_test, format_atomic) {
+  std::atomic<bool> b(false);
+  EXPECT_EQ(fmt::format("{}", b), "false");
+
+  const std::atomic<bool> cb(true);
+  EXPECT_EQ(fmt::format("{}", cb), "true");
+}


include/fmt/std.h
```diff
@@ -389,6 +389,29 @@ struct formatter<
 #endif
   }
 };
-FMT_END_NAMESPACE
 
+FMT_EXPORT
+template <typename T, typename Char>
+struct formatter<std::atomic<T>, Char> : formatter<T, Char> {
+  template <typename FormatContext>
+  FMT_CONSTEXPR20 auto format(const std::atomic<T>& v, FormatContext& ctx) const
```
phprus, line 397: ``std::atomic<T>::load`` is not constexpr:
https://en.cppreference.com/w/cpp/atomic/atomic/load

wangzw: fixed

vitaut:
You need to include `<atomic>` in `fmt/std.h`.

wangzw:
> You need to include `<atomic>` in `fmt/std.h`.

Fixed and passed all tests

include/fmt/std.h
```diff
@@ -435,6 +437,28 @@ struct formatter<BitRef, Char,
...
+      -> decltype(ctx.out()) {
+    return formatter<T, Char>::format(v.load(), ctx);
+  }
+};
 
+FMT_EXPORT
+template <typename Char>
+struct formatter<std::errc, Char>
+    : formatter<std::underlying_type<std::errc>::type, Char> {
+  typedef std::underlying_type<std::errc>::type underlying_type;
```
phprus, line 454: I suggest move ``underlying_type`` into ``format`` function and replace ``typedef`` with
```c++
using underlying_type = std::underlying_type<std::errc>::type;
```

wangzw: Fixed

include/fmt/std.h
```diff
@@ -435,6 +437,28 @@ struct formatter<BitRef, Char,
   }
 };
 
-FMT_END_NAMESPACE
+FMT_EXPORT
+template <typename T, typename Char>
+struct formatter<std::atomic<T>, Char> : formatter<T, Char> {
```
vitaut, line 442: This formatter should only be enabled if T is formattable.

include/fmt/std.h
```diff
@@ -435,6 +437,28 @@ struct formatter<BitRef, Char,
...
+struct formatter<std::errc, Char>
+    : formatter<std::underlying_type<std::errc>::type, Char> {
+  template <typename FormatContext>
+  FMT_CONSTEXPR auto format(std::errc v, FormatContext& ctx) const
+      -> decltype(ctx.out()) {
+    using underlying_type = std::underlying_type<std::errc>::type;
+    return formatter<underlying_type, Char>::format(
+        static_cast<underlying_type>(v), ctx);
+  }
+};
```
vitaut, line 461: I am not sure if providing a formatter for errc is a good idea. I think it should be up to the user to decide how they want to present it and if they want a number they could use to_underlying.

vitaut (changes requested):
Thanks for the PR.

include/fmt/std.h
wangzw: Fixed

include/fmt/std.h
wangzw: Is there any way to provide a default formatter if no other formatter exists?

vitaut (approved):

vitaut:
Thank you!

include/fmt/std.h
vitaut: Considering that it is easy to format `errc` as a number with `to_underlying`, I don't think it's worth exploring the default formatter option.

vitaut:
BTW could you add std::atomic to the list of formattable types in https://github.com/fmtlib/fmt/blob/master/doc/api.rst#standard-library-types-formatting?

include/fmt/std.h
```diff
@@ -435,6 +436,17 @@ struct formatter<BitRef, Char,
...
-FMT_END_NAMESPACE
+FMT_EXPORT
+template <typename T, typename Char>
+struct formatter<std::atomic<T>, Char,
+                 enable_if_t<is_formattable<T, Char>::value>>
+    : formatter<T, Char> {
+  template <typename FormatContext>
+  auto format(const std::atomic<T>& v, FormatContext& ctx) const
+      -> decltype(ctx.out()) {
+    return formatter<T, Char>::format(v.load(), ctx);
```
dyumin, line 447: Maybe its better to use load(std::memory_order_acquire) or even load(std::memory_order_relaxed)?

PR #3582, opened 2023-08-11
Issue 3545

pmqtt:
Hello folks,
I hope my changes meet your needs and is useful?!

It solves #3545 

Best regrads


vitaut:
Thanks for the PR but this is already addressed by #3575.

PR #3594, opened 2023-08-14
`atomic_flag` formatting

AlexGuteniev:
To have complete atomic support

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -451,8 +451,7 @@ struct formatter<std::atomic<T>, Char,
 #if defined(__cpp_lib_atomic_flag_test) && __cpp_lib_atomic_flag_test >= 201907L
 FMT_EXPORT
 template <typename Char>
-struct formatter<std::atomic_flag, Char,
-                 enable_if_t<is_formattable<bool, Char>::value>>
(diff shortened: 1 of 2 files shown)


include/fmt/std.h
```diff
@@ -448,5 +448,18 @@ struct formatter<std::atomic<T>, Char,
   }
 };
 
+#if defined(__cpp_lib_atomic_flag_test) && __cpp_lib_atomic_flag_test >= 201907L
```
AlexGuteniev, line 451: I observe the other feature checks don't check for the feature value, shouldn't check here too?

AlexGuteniev: No, I mean make like `#ifdef __cpp_lib_optional` above without comparing the value, as we're looking for the first and (currently) the only version of the feature

vitaut: I think it's enough to check if `__cpp_lib_atomic_flag_test` is defined unless this needs to rely on some specific change to `atomic_flag` which is unlikely.

vitaut (commented):
Thanks for the PR!

vitaut:
Merged, thanks.

Issue #3598, opened 2023-08-18
� cause C2001 and C1075 errors in VisualStudio 2022


qliang88:
https://github.com/fmtlib/fmt/blob/6c845f57e5db589208ff0c2808238587cafafa82/include/fmt/format.h#L1456C12-L1456C40

logs:
[build] .../fmt/format.h(1456,34): error C2001: newline in constant 
[build] .../fmt/format.h(1456,33): fatal  error C1075: '{': no matching token found

vitaut:
There is no newline there, mostly likely you are using some legacy encoding. Make sure to compile with `/utf-8`.

YuHuanTin:
> There is no newline there, mostly likely you are using some legacy encoding. Make sure to compile with `/utf-8`.

I think this is a little bit wrong, because most `msvc` never specify `utf-8` encoding by default, I may need to specify the whole project to `utf-8` encoding because of one character, and he will affect the standard output (`utf8` can be output normally but other encodings can not be displayed), and I have already encountered this problem, except for specifying `\x`style characters I can not think of a better way to increase the compatibility of the

vitaut:
What encoding do you use? BTW it looks like you comment was truncated.

YuHuanTin:
> I think this is a little bit wrong, because most `msvc` never specify `utf-8` encoding by default, I may need to specify the whole project to `utf-8` encoding because of one character, and he will affect the standard output (`utf8` can be output normally but other encodings can not be displayed), and I have already encountered this problem, except for specifying `\x`style characters I can not think of a better way to increase the compatibility of the

I'm using `GBK` encoding for the whole project now.
Maybe it's a translation problem, I retranslated it back and it seems to be fine.

vitaut:
@qliang88, could you confirm that #3599 fix those MSVC errors?

qliang88:
yes, no more errors after replace by 0xFFFD.

vitaut:
Merged the "fix" but please note that using legacy encodings is not recommended.

PR #3599, opened 2023-08-19
fix issues #3598

YuHuanTin:
fix issues #3598


vitaut:
Thanks for the PR but � is a valid Unicode character and I don't think we should replace it.

vitaut:
OK, let's merge it since it's a simple change that doesn't affect the common case of UTF-8.

AlexGuteniev:
44rmwd

PR #3600, opened 2023-08-20
Fix `FMT_NO_UNIQUE_ADDRESS` warning with clang-cl.

lukester1975:
Hi

clang-cl is generating a warning with the `no_unique_address` change from c86fe0b8d3aa6553160ad04aed608c2e694e9136: `warning : unknown attribute 'no_unique_address' ignored [-Wunknown-attributes]`

Thread on clang-cl and no_unique_address: https://reviews.llvm.org/D110485

Here's an additional check check for clang-cl (plus use `__has_cpp_attribute(msvc::no_unique_address)` to cope in case clang-cl adds support; cl returns true since 19.31).

Thanks

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -83,7 +83,8 @@
 #  if FMT_CPLUSPLUS >= 202002L
 #    if FMT_HAS_CPP_ATTRIBUTE(no_unique_address)
 #      define FMT_NO_UNIQUE_ADDRESS [[no_unique_address]]
-#    elif FMT_MSC_VERSION >= 1929  // VS2019 v16.10 and later
+// VS2019 v16.10 and later except clang-cl (https://reviews.llvm.org/D110485)
+#    elif (FMT_MSC_VERSION >= 1929) && !FMT_CLANG_VERSION
 #      define FMT_NO_UNIQUE_ADDRESS [[msvc::no_unique_address]]
 #    endif
 #  endif


include/fmt/format.h
```diff
@@ -83,7 +83,9 @@
 #  if FMT_CPLUSPLUS >= 202002L
 #    if FMT_HAS_CPP_ATTRIBUTE(no_unique_address)
 #      define FMT_NO_UNIQUE_ADDRESS [[no_unique_address]]
-#    elif FMT_MSC_VERSION >= 1929  // VS2019 v16.10 and later
+// VS2019 v16.10 and later except clang-cl (https://reviews.llvm.org/D110485)
+#    elif __has_cpp_attribute(msvc::no_unique_address) || \
```
vitaut, line 87: __has_cpp_attribute -> FMT_HAS_CPP_ATTRIBUTE

vitaut: Actually, I don't think we need an attribute check since the branch below only handles specific compiler (versions).

lukester1975: OK, `FMT_HAS_CPP_ATTRIBUTE` done, missed that.

The reason I put the `msvc::no_unique_address` attr check is to future proof if support gets added to clang-cl, rather than always not setting.

I suppose if clang-cl waits to enable the standard `no_unique_address` until after cl does and never bothers with the `msvc` attr it'll be moot, but this way at least it won't need revising either way... Or at least that was the intention!

vitaut: I don't think we should be adding support for some not-standard attribute that may or may not be supported in the future. Let's remove the attribute check.

lukester1975: OK sure, done.

Seems a shame to miss out on the size optimization if it does get added, though! Not that I've looked at where it's used in fmtlib 😀

vitaut:
Thanks

PR #3624, opened 2023-09-05
Remove type cast as mxe(mingw32) compiler complains about useless-cast

vinayyadav3016:
Remove type cast as mxe(mingw32) compiler complains about useless-cast
when FMT_PEDANTIC && FMT_WERROR options are enabled """
error: useless cast to type 'class fmt::v10::basic_format_args<fmt::v10::basic_format_context<fmt::v10::appender, char> >' [-Werror=useless-cast]
 1449 |                      basic_format_args<buffer_context<char>>(args));
"""


vitaut (changes requested):
Thanks for the PR

vinayyadav3016:
Hi Vitaut,

Changes done as per your suggestion.

vitaut (approved):

vitaut:
Thanks

Issue #3626, opened 2023-09-05
Reconsider forcibly exporting `format_error` symbol on GCC


feltech:
In e0fc0e8 the export attribute macro for the `format_error` exception in `format.h` was switched from the conditional `FMT_API` to the unconditional `FMT_VISIBILITY("default")`

This causes an ODR violation when fmt is bundled in a (shared) library, intended as a private internal dependency, but consumers of the library are using their own copy of fmt too.

Kinda tricky to show in a Compiler Explorer example. A useful `nm` command-line to check exported symbols in a library:
```shell
nm --demangle --dynamic --defined-only --extern-only mylibrary.so | grep fmt
```

After reverting to 9.1.0, `nm` reveals no errant public symbols.

vitaut:
We can't use `FMT_API` there because of windows but we could allow overriding `FMT_VISIBILITY` similarly to other macros. A PR is welcome.

feltech:
> We can't use FMT_API there because of windows

Could you elaborate on that a little?  From what I understand, the usage of `FMT_VISIBILITY` rather than `FMT_API` on Windows means that the `format_error` exception is not exported at all on Windows?  I'd be curious to understand why (I'm no Windows guru, but sadly still have to build for it).

Am I correct in my understanding that if we're using fmt header-only, then `FMT_API` expands to nothing (v9 and v10), so nothing is exported at all regardless of platform (except for `format_error` in v10 on GCC/Clang - i.e. this issue)?  So we don't have to worry about Windows shenanigans when header-only?

vitaut:
I don't remember the details of windows issues from the top of my head and it's irrelevant anyway because you need not to export something on windows but avoid changing symbol visibility on POSIX. And this should be achieved by means of changing the definition of `FMT_VISIBILITY` and not by switching to `FMT_API`.

vitaut:
So I think the correct solution is to define `FMT_VISIBILITY` to the default visibility if `defined(FMT_LIB_EXPORT) || defined(FMT_SHARED)` and to nothing otherwise.

feltech:
The fix in https://github.com/fmtlib/fmt/pull/3627 looks good to me. 

I'd be interested, just to plug a gap in my own knowledge, if anyone can explain why this special case for `format_error` is needed?

vitaut:
As commented on the PR we don't need a special case there and should change `FMT_VISIBILITY` directly.

PR #3627, opened 2023-09-08
Fix symbol leak

phprus:
Fix for issue #3626

@feltech, please test this PR.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -99,6 +99,13 @@
 #  define FMT_VISIBILITY(value)
 #endif
 
+#if !defined(FMT_HEADER_ONLY) && !defined(_WIN32) && \
+    (defined(FMT_LIB_EXPORT) || defined(FMT_SHARED))
+#  define FMT_INLINE_API FMT_VISIBILITY("default")
+#else
+#  define FMT_INLINE_API
+#endif
+
 #ifdef __has_builtin
 #  define FMT_HAS_BUILTIN(x) __has_builtin(x)
 #else
@@ -1046,7 +1053,7 @@ FMT_BEGIN_EXPORT
 #endif
 
 /** An error reported from a formatting function. */
-class FMT_VISIBILITY("default") format_error : public std::runtime_error {
+class FMT_INLINE_API format_error : public std::runtime_error {
  public:
   using std::runtime_error::runtime_error;
 };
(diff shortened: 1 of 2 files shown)


feltech:
Fix looks good to me.  I can confirm that, at least for our specific case of  GCC 9 and CMake target `fmt::fmt-header-only`:

Current `main` leaks symbols:
```
$ nm --demangle --dynamic --defined-only --extern-only libopenassetio.so | grep fmt
00000000000b55f0 D typeinfo for fmt::v10::format_error
000000000007e410 R typeinfo name for fmt::v10::format_error
$
```
and this PR does not ;)
```
$ nm --demangle --dynamic --defined-only --extern-only libopenassetio.so | grep fmt
$
```

I'm still personally curious why this special case for `format_error` exists (not that its wrong, I just genuinely want to understand), but I'll follow-up in the issue.

feltech (approved):

include/fmt/format.h
```diff
@@ -99,6 +99,13 @@
 #  define FMT_VISIBILITY(value)
 #endif
 
+#if !defined(FMT_HEADER_ONLY) && !defined(_WIN32) && \
+    (defined(FMT_LIB_EXPORT) || defined(FMT_SHARED))
+#  define FMT_INLINE_API FMT_VISIBILITY("default")
+#else
+#  define FMT_INLINE_API
+#endif
```
vitaut, line 107: I don't think we should introduce a new macro. Let's update `FMT_VISIBILITY` instead since we shouldn't be changing visibility in general when not compiling a shared library.

vitaut (changes requested):
Thanks for the PR.

include/fmt/format.h
phprus: @vitaut 
FMT_VISIBILITY is used unconditionally:
https://github.com/fmtlib/fmt/blob/a4b7b24b7b20d5462565e9cadbdfc7f101a4b843/include/fmt/format.h#L1885

Therefore, we need two macros.

vitaut (approved):

include/fmt/format.h
vitaut: OK, we can simplify this later.

vitaut:
Merged, thanks!

PR #3630, opened 2023-09-08
fix redundant redeclaration of ‘constexpr’ static data member

cyyever:

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -336,8 +336,6 @@ template <typename CodeUnit> struct codecvt_result {
   CodeUnit buf[max_size];
   CodeUnit* end;
 };
-template <typename CodeUnit>
-constexpr const size_t codecvt_result<CodeUnit>::max_size;
 
 template <typename CodeUnit>
 void write_codecvt(codecvt_result<CodeUnit>& out, string_view in_buf,


include/fmt/chrono.h
```diff
@@ -336,8 +336,6 @@ template <typename CodeUnit> struct codecvt_result {
   CodeUnit buf[max_size];
   CodeUnit* end;
 };
-template <typename CodeUnit>
-constexpr const size_t codecvt_result<CodeUnit>::max_size;
```
phprus, line 340: Suggest:
```c++
#if FMT_CPLUSPLUS < 201703L
template <typename CodeUnit>
constexpr const size_t codecvt_result<CodeUnit>::max_size;
#endif
```

vitaut: I'd prefer avoiding conditional compilation if possible.

phprus: @vitaut 
This is C++ :(

https://en.cppreference.com/w/cpp/language/static#Static_data_members

> If a const _non-inline (since C++17)_ static data member _or a constexpr static data member (since C++11)(until C++17)_ is odr-used, a definition at namespace scope is still required, but it cannot have an initializer.

vitaut: But it doesn't need to be ODR-used so it should be fine.

vitaut:
Thanks!

PR #3633, opened 2023-09-11
Fix compilation error if `PACKED` is defined before including FMT

### LINKED ISSUE #3632: Compilation error with Nintendo 3DS homebrew compiler
I tried updating fmt to v10.1.1 in DevilutionX (https://github.com/diasurgical/devilutionX/pull/6585) and got an error on one of the oddball platforms.

The platform is Nintendo 3DS homebrew, it uses a GCC 13.2.0-based compiler, and the error is this:

```
In file included from /root/repo/Source/utils/log.hpp:6,
                 from /root/repo/Source/platform/ctr/sockets.cpp:8:
/root/repo/build/_deps/libfmt-src/include/fmt/core.h:1557:68: error: template argument 1 is invalid
 1557 | template <bool PACKED, typename Context, typename T, FMT_ENABLE_IF(PACKED)>
      |                                                                    ^~~~~~
/root/repo/build/_deps/libfmt-src/include/fmt/core.h:286:48: note: in definition of macro 'FMT_ENABLE_IF'
  286 | #  define FMT_ENABLE_IF(...) fmt::enable_if_t<(__VA_ARGS__), int> = 0
      |                                                ^~~~~~~~~~~
/root/repo/build/_deps/libfmt-src/include/fmt/core.h:286:59: error: expected identifier before ')' token
  286 | #  define FMT_ENABLE_IF(...) fmt::enable_if_t<(__VA_ARGS__), int> = 0
      |                                                           ^
/root/repo/build/_deps/libfmt-src/include/fmt/core.h:1557:54: note: in expansion of macro 'FMT_ENABLE_IF'
 1557 | template <bool PACKED, typename Context, typename T, FMT_ENABLE_IF(PACKED)>
      |                                                      ^~~~~~~~~~~~~
/root/repo/build/_deps/libfmt-src/include/fmt/core.h:286:59: error: expected '>' before ')' token
  286 | #  define FMT_ENABLE_IF(...) fmt::enable_if_t<(__VA_ARGS__), int> = 0
      |                                                           ^
/root/repo/build/_deps/libfmt-src/include/fmt/core.h:1557:54: note: in expansion of macro 'FMT_ENABLE_IF'
 1557 | template <bool PACKED, typename Context, typename T, FMT_ENABLE_IF(PACKED)>
      |                                                      ^~~~~~~~~~~~~
/root/repo/build/_deps/libfmt-src/include/fmt/core.h:286:59: error: expected unqualified-id before ')' token
  286 | #  define FMT_ENABLE_IF(...) fmt::enable_if_t<(__VA_ARGS__), int> = 0
      |                                                           ^
/root/repo/build/_deps/libfmt-src/include/fmt/core.h:1557:54: note: in expansion of macro 'FMT_ENABLE_IF'
 1557 | template <bool PACKED, typename Context, typename T, FMT_ENABLE_IF(PACKED)>
      |                                                      ^~~~~~~~~~~~~
```

Failing build: https://app.circleci.com/pipelines/github/diasurgical/devilutionX/18895/workflows/2775b1cd-f0ad-4945-911c-530cfc7aca1d/jobs/116663

I cannot reproduce this error on godbolt. I also don't really understand the error message, the code seems fine to me.
I'd appreciate any thoughts on what this could be.

glebm:
Reduces the chance of conflicts between a macro declaration and the template argument name by using CamelCase (`PACKED` -> `IsPacked`)

This improves compatibility with oddball SDKs, such as the 3DS libctru.

Fixes #3632

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -1554,7 +1554,7 @@ constexpr auto encode_types() -> unsigned long long {
 template <typename T, typename Char> struct type_is_unformattable_for;
 #endif
 
-template <bool PACKED, typename Context, typename T, FMT_ENABLE_IF(PACKED)>
+template <bool IsPacked, typename Context, typename T, FMT_ENABLE_IF(IsPacked)>
 FMT_CONSTEXPR FMT_INLINE auto make_arg(T& val) -> value<Context> {
   using arg_type = remove_cvref_t<decltype(arg_mapper<Context>().map(val))>;
 
@@ -1591,7 +1591,7 @@ FMT_CONSTEXPR auto make_arg(T& val) -> basic_format_arg<Context> {
   return arg;
 }
 
-template <bool PACKED, typename Context, typename T, FMT_ENABLE_IF(!PACKED)>
+template <bool IsPacked, typename Context, typename T, FMT_ENABLE_IF(!IsPacked)>
 FMT_CONSTEXPR inline auto make_arg(T& val) -> basic_format_arg<Context> {
   return make_arg<Context>(val);
 }


vitaut:
Thanks for the PR but this should be fixed in libctru by prefixing the macro. We shouldn't be trying to avoid conflicts with poorly named macros from random third-party libraries. In the meantime you can undefined the problematic macro yourself after including the headers that define it.

glebm:
Philosophically correct but why name a template parameter in all caps? This style is usually used for C macros and enums and doesn't seem consistent with the rest of libfmt anyway

vitaut:
Non-type template parameters normally follow this style.

nelul-lut:
For what it's worth, my company also defines a `PACKED` macro. It's a fairly common alias for `__attribute__((PACKED))`, you see it a lot in embedded contexts for example.

Would you be open to changing this parameter name to `IS_PACKED`? There's already `IS_CONSTEXPR` in a few places, so it's still consistent.

You're right that it isn't this library's responsibility to ensure that your names don't conflict with other people's macros, but `PACKED` happens to be a fairly common macro people define, so it could save a lot of headaches for people.

lapointexavier:
I would like to second that, ‘PACKED’ is also a bit too common and it’s causing headaches trying to upgrade fmt. Would you reconsider @vitaut ? If not we’ll find a solution but it seems like a small lift if this PR is all it takes.

vitaut:
We don't try to workaround conflicts with poorly written macros unless they are part of system APIs but this particular one is no longer there for other reasons.

PR #3634, opened 2023-09-11
Fix deprecation warning in libc++ (LLVM-16)

phprus:
Warning:

```
'char_traits<custom_char>' is deprecated: char_traits<T> for T not equal to char, wchar_t, char8_t, char16_t or char32_t is non-standard and is provided for a temporary period. It will be removed in LLVM 18, so please migrate off of it. [-Werror,-Wdeprecated-declarations]
```

LLVM source:

https://github.com/llvm/llvm-project/blob/7cbf1a2591520c2491aa35339f227775f4d3adf6/libcxx/include/__string/char_traits.h#L74-L80

### DIFF
--- a/test/xchar-test.cc
+++ b/test/xchar-test.cc
@@ -102,10 +102,70 @@ struct custom_char {
   template <typename T>
   constexpr custom_char(T val) : value(static_cast<int>(val)) {}
 
-  operator char() const {
+  constexpr operator char() const {
     return value <= 0xff ? static_cast<char>(value) : '\0';
   }
+  constexpr bool operator<(custom_char c) const { return value < c.value; }
 };
+
+namespace std {
+
+template <> struct char_traits<custom_char> {
+  using char_type = custom_char;
+  using int_type = int;
+  using off_type = streamoff;
+  using pos_type = streampos;
+  using state_type = mbstate_t;
+
+  static constexpr void assign(char_type& r, const char_type& a) { r = a; }
+  static constexpr bool eq(char_type a, char_type b) { return a == b; }
+  static constexpr bool lt(char_type a, char_type b) { return a < b; }
+  static FMT_CONSTEXPR int compare(const char_type* s1, const char_type* s2,
+                                   size_t count) {
+    for (; count; count--, s1++, s2++) {
+      if (lt(*s1, *s2)) return -1;
+      if (lt(*s2, *s1)) return 1;
+    }
+    return 0;
+  }
+  static FMT_CONSTEXPR size_t length(const char_type* s) {
+    size_t count = 0;
+    while (!eq(*s++, custom_char(0))) count++;
+    return count;
+  }
+  static const char_type* find(const char_type*, size_t, const char_type&);
+  static FMT_CONSTEXPR char_type* move(char_type* dest, const char_type* src,
+                                       size_t count) {
+    if (count == 0) return dest;
+    char_type* ret = dest;
+    if (src < dest) {
+      dest += count;
+      src += count;
+      for (; count; count--) assign(*--dest, *--src);
+    } else if (src > dest)
+      copy(dest, src, count);
+    return ret;
+  }
+  static FMT_CONSTEXPR char_type* copy(char_type* dest, const char_type* src,
+                                       size_t count) {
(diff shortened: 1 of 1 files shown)


.github/workflows/linux.yml
```diff
@@ -54,6 +54,11 @@ jobs:
             std: 20
             cxxflags: -stdlib=libc++
             install: sudo apt install libc++-11-dev libc++abi-11-dev
+          - cxx: clang++-16
+            build_type: Debug
+            std: 20
+            cxxflags: -stdlib=libc++
+            install: sudo apt install g++-11 libstdc++-11-dev clang-16 libomp-16-dev libc++-16-dev libc++abi-16-dev
```
vitaut, line 61: We have too many CI configs already and shouldn't introduce yet another one for this.

test/xchar-test.cc
```diff
@@ -102,11 +102,62 @@ struct custom_char {
...
+  constexpr operator char() const {
     return value <= 0xff ? static_cast<char>(value) : '\0';
   }
+  constexpr bool operator<(custom_char c) const { return value < c.value; }
 };
 
+#  ifdef _LIBCPP_VERSION
+namespace std {
+
+template <> struct char_traits<custom_char> {
```
vitaut, line 114: Does the standard permit user specializations of `char_traits`?

vitaut (changes requested):
Thanks for the PR. A few comments inline.

.github/workflows/linux.yml
phprus: Maybe be better to remove the older ``clang``?
For example ``clang++-8``?

test/xchar-test.cc
phprus: https://en.cppreference.com/w/cpp/string/char_traits:

> The char_traits class template serves as a basis for explicit instantiations. The user can [provide a specialization](https://en.cppreference.com/w/cpp/language/extending_std) for any custom character types. Several explicit specializations are provided for the standard character types (see below), other specializations are not required to satisfy the requirements of [CharTraits](https://en.cppreference.com/w/cpp/named_req/CharTraits).

.github/workflows/linux.yml
vitaut: The old one is more likely to catch compatibility issues.

test/xchar-test.cc
```diff
@@ -102,11 +102,62 @@ struct custom_char {
...
   constexpr custom_char(T val) : value(static_cast<int>(val)) {}
 
-  operator char() const {
+  constexpr operator char() const {
     return value <= 0xff ? static_cast<char>(value) : '\0';
   }
+  constexpr bool operator<(custom_char c) const { return value < c.value; }
 };
 
+#  ifdef _LIBCPP_VERSION
```
vitaut, line 111: I think we should specialize the trait unconditionally here. Since relying on the standard one is discouraged we don't care whether it works or not with custom code unit types.

.github/workflows/linux.yml
phprus: But old don't catch compatibility issues with new libc++.
I think we need CI for the newest versions of the all compilers, which is updated periodically.
If you tell me to delete this CI (or update to lasted clang-17), I will do so.

test/xchar-test.cc
phprus: Done

.github/workflows/linux.yml
vitaut: We care about old compilers because of explicit compatibility guarantees which are easy to accidentally break and need to be tested against. The newest versions on the other hand rarely cause real problems because they usually ensure compatibility with existing code. We should care about commonly used and not latest versions. It is a bit hard to estimate which versions are more used but I think a reasonable rule of thumb is to have a year or two lag or check which versions ship with popular LTS Linux releases. I agree that we need to periodically bump versions (but not to the latest). The only exception is when we need to test against specific major features like modules or consteval. This particular case is just a warning in test which doesn't need extra coverage so please remove the new CI config.

phprus: Done.

vitaut (approved):

vitaut:
Thank you!

PR #3636, opened 2023-09-12
Add an option to avoid wchar APIs on Windows

### LINKED ISSUE #3631: Windows API systems without UTF-16 APIs
Currently the Windows build assumes that `WriteConsoleW` is available.

None of the Unicode stuff is actually available on Windows 98 (and similar API level systems such as the original Xbox).

Would you be OK with adding a build option to disable it? I can send a PR if so.
Is `FMT_WINDOWS_HAS_UNICODE_APIS` a good name for the build option?

Alternatively, perhaps we could add an option to disable the entire `print` subsystem (something like `set(FMT_PRINT OFF)`)

glebm:
With this, fmt can be used on Windows 98 and the Original Xbox with:

```cmake
set(FMT_OS OFF)
target_compile_definitions(fmt PUBLIC FMT_WINDOWS_NO_WCHAR)
```

Fixes #3631

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -18,7 +18,7 @@
 #  include <locale>
 #endif
 
-#ifdef _WIN32
+#if defined(_WIN32) && !defined(FMT_WINDOWS_NO_WCHAR)
 #  include <io.h>  // _isatty
 #endif
 
@@ -1426,7 +1426,7 @@ FMT_FUNC std::string vformat(string_view fmt, format_args args) {
 }
 
 namespace detail {
-#ifndef _WIN32
+#if !defined(_WIN32) || defined(FMT_WINDOWS_NO_WCHAR)
 FMT_FUNC bool write_console(std::FILE*, string_view) { return false; }
 #else
 using dword = conditional_t<sizeof(long) == 4, unsigned long, unsigned>;
@@ -1441,7 +1441,9 @@ FMT_FUNC bool write_console(std::FILE* f, string_view text) {
   return WriteConsoleW(reinterpret_cast<void*>(_get_osfhandle(fd)), u16.c_str(),
                        static_cast<uint32_t>(u16.size()), &written, nullptr) != 0;
 }
+#endif
 
+#ifdef _WIN32
 // Print assuming legacy (non-Unicode) encoding.
 FMT_FUNC void vprint_mojibake(std::FILE* f, string_view fmt, format_args args) {
   auto buffer = memory_buffer();
--- a/test/format-impl-test.cc
+++ b/test/format-impl-test.cc
@@ -351,7 +351,7 @@ TEST(format_impl_test, write_dragon_even) {
   if (!FMT_MSC_VERSION) EXPECT_EQ(s, "33554450");
 }
 
-#ifdef _WIN32
+#if defined(_WIN32) && !defined(FMT_WINDOWS_NO_WCHAR)
 #  include <windows.h>
 
 TEST(format_impl_test, write_console_signature) {


CMakeLists.txt
```diff
@@ -151,6 +151,7 @@ option(FMT_TEST "Generate the test target." ${FMT_MASTER_PROJECT})
 option(FMT_FUZZ "Generate the fuzz target." OFF)
 option(FMT_CUDA_TEST "Generate the cuda-test target." OFF)
 option(FMT_OS "Include core requiring OS (Windows/Posix) " ON)
+option(FMT_WINDOWS_NO_WCHAR "Do not use wide-char APIs, such as WriteConsoleW, on Windows" OFF)
```
vitaut, line 154: We shouldn't pollute options with such an esoteric thing. At most it could be configured with a macro. Please revert all changes to the CMake config.

include/fmt/format-inl.h
```diff
@@ -18,7 +18,7 @@
 #  include <locale>
 #endif
 
-#ifdef _WIN32
+#if defined(_WIN32) && !defined(FMT_WINDOWS_NO_WCHAR)
```
vitaut, line 21: Can this be detected automatically e.g. via some predefined macro or `__has_include(<io.h>)`?

vitaut (changes requested):
In general I'm not a fan of supporting dead platforms but will be OK with merging this if the changes are more contained. In particular don't touch the CMake config and try to detect it automatically if possible.

CMakeLists.txt
glebm: Done

include/fmt/format-inl.h
glebm: Done. It can't always be detected automatically but I've added the 2 cases where it can (missing `<io.h>` on xbox and defined WINVER on properly configured Windows projects). Even when it can't be detected automatically, one can always `add_definitions(FMT_WINDOWS_NO_WCHAR)` in the fmt subproject, so that's not a problem.

vitaut: I was hoping that it will be less messy =). It is particularly weird to have this tested in core.h which explicitly avoids `wchar_t`. Let's go back to just requiring the user to define `FMT_WINDOWS_NO_WCHAR` (and without the config option).

glebm: Done!

image

include/fmt/core.h
```diff
@@ -2675,7 +2675,7 @@ void vformat_to(buffer<Char>& buf, basic_string_view<Char> fmt,
                 typename vformat_args<Char>::type args, locale_ref loc = {});
 
 FMT_API void vprint_mojibake(std::FILE*, string_view, format_args);
-#ifndef _WIN32
+#if !defined(_WIN32) || defined(FMT_WINDOWS_NO_WCHAR)
```
vitaut, line 2678: Can we not change this? It's weird to disable overload that doesn't even use `wchar_t`.

glebm: If we don't have wchar, we don't have `vprint_mojibake` either, and the compiler complains about the lack of a definition.

glebm: Actually maybe I'm wrong, let me investigate

glebm: This only happens when `is_utf8()` returns `false` but this is the error:

```
fmt/core.h:2894: undefined reference to `fmt::v10::detail::vprint_mojibake(_iobuf*, fmt::v10::basic_string_view<char>, fmt::v10::basic_format_args<fmt::v10::basic_format_context<fmt::v10::appender, char> >)'
```

vitaut: This is the wrong fix though because this will just discard the output. The correct fix is to enable the `vprint_mojibake` definition in `format-inl.h`.

glebm: Thanks, done

vitaut (approved):

Issue #3645, opened 2023-09-18
Unexpected output from intel fmt::fprint


gsjaardema:
With the following program:
```
#include <fmt/printf.h>
int main()
{
    fmt::printf("%d:%d %f %s\n", 3, 1234, 12.345, "weekend");
}
```
I get the expected output on most compiler and lib::fmt version combinations:
```
3:1234 12.345000 weekend
```

However, if I use the intel compiler with either 10.0.0 or trunk, I get:
```
3:1234 12.345001 weekend
```
Notice the `1` at the end of the floating point output.  I'm not sure where this is coming from or how to get the same output on intel with 10.0.X as I get with intel and earlier lib::fmt versions or with all other compilers I have tried.

Here is a compiler explorer link:  https://godbolt.org/z/MzG7T9bfr

gsjaardema:
I added the c-library printf call:
```
    printf("%d:%d %f %s\n", 3, 1234, 12.345, "weekend");
```
And that gives me the "expected" output:
```
3:1234 12.345000 weekend
```

gsjaardema:
I also tried with the `%g` format specifier for the double parameter and get:
```
3:1234 12.3451 weekend
``` 
which is worse accuracy.

If I loop from 0.345 to 100.345:
```
include <fmt/printf.h>
int main()
{
   for (int i = 0; i < 100; i++) { 
        double di = (double)i;
        fmt::printf("%d:%d %g %s\n", 3, 1234, i+.345, "weekend");
    }
}
```
The extraneous `1` is added on for these values:
```
3:1234 7.345 weekend
3:1234 8.34501 weekend
3:1234 9.34501 weekend
3:1234 10.3451 weekend
3:1234 11.3451 weekend
3:1234 12.3451 weekend
3:1234 13.3451 weekend
3:1234 14.3451 weekend
3:1234 15.3451 weekend
3:1234 16.345 weekend
```

vitaut:
Interestingly, the issue doesn't happen with `print`: https://godbolt.org/z/1s3ebbv9K.

gsjaardema:
> Interestingly, the issue doesn't happen with `print`: https://godbolt.org/z/1s3ebbv9K.

Yes, I noticed that.  The assembly output is also much cleaner and smaller.

gsjaardema:
We found another strangeness:
```
int main()
{
    fmt::printf("%g %g\n", -10., 2.);
}
```
Returns:
```
-10.0001 2.00001
```

[...]

PR #3652, opened 2023-09-20
Workaround intel bug

gsjaardema:
Potential workaround / restructure for the intel bug that is the cause of #3645.

Make the variable in the external struct instead an embedded static constexpr variable in the only function that uses the variable.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -3258,14 +3258,11 @@ FMT_CONSTEXPR20 void format_hexfloat(Float value, int precision,
   format_hexfloat(static_cast<double>(value), precision, specs, buf);
 }
 
-template <typename Float>
-FMT_CONSTEXPR20 auto format_float(Float value, int precision, float_specs specs,
-                                  buffer<char>& buf) -> int {
-  
+FMT_CONSTEXPR uint32_t fractional_part_rounding_threshold(int index) {
   // For checking rounding thresholds.
   // The kth entry is chosen to be the smallest integer such that the
   // upper 32-bits of 10^(k+1) times it is strictly bigger than 5 * 10^k.
-  static constexpr uint32_t fractional_part_rounding_thresholds[8] = {
+  constexpr uint32_t thresholds[8] = {
       2576980378U,  // ceil(2^31 + 2^32/10^1)
       2190433321U,  // ceil(2^31 + 2^32/10^2)
       2151778616U,  // ceil(2^31 + 2^32/10^3)
@@ -3275,7 +3272,13 @@ FMT_CONSTEXPR20 auto format_float(Float value, int precision, float_specs specs,
       2147484078U,  // ceil(2^31 + 2^32/10^7)
       2147483691U   // ceil(2^31 + 2^32/10^8)
   };
+  return thresholds[index];
+}
 
+template <typename Float>
+FMT_CONSTEXPR20 auto format_float(Float value, int precision, float_specs specs,
+                                  buffer<char>& buf) -> int {
+  
   // float is passed as double to reduce the number of instantiations.
   static_assert(!std::is_same<Float, float>::value, "");
   FMT_ASSERT(value >= 0, "value is negative");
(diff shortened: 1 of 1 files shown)


gsjaardema:
Pull request not correct for certain c++ std levels.  Will let someone with more knowledge propose a solution.

include/fmt/format.h
```diff
@@ -3283,6 +3261,21 @@ FMT_CONSTEXPR20 void format_hexfloat(Float value, int precision,
 template <typename Float>
 FMT_CONSTEXPR20 auto format_float(Float value, int precision, float_specs specs,
                                   buffer<char>& buf) -> int {
+  
+  // For checking rounding thresholds.
+  // The kth entry is chosen to be the smallest integer such that the
+  // upper 32-bits of 10^(k+1) times it is strictly bigger than 5 * 10^k.
+  static constexpr uint32_t fractional_part_rounding_thresholds[8] = {
```
vitaut, line 3268: You need to drop static here:

```
/Users/runner/work/fmt/fmt/include/fmt/format.h:3268:29: error: definition of a static variable in a constexpr function is a C++2b extension [-Werror,-Wc++2b-extensions]
  static constexpr uint32_t fractional_part_rounding_thresholds[8] = {
                            ^
```

Also please move to a separate function to avoid parametrization on `Float`, something like

```c++
FMT_CONSTEXPR uint32_t fractional_part_rounding_threshold(int index) {
  constexpr uint32_t thresholds[] = {...};
  return thresholds[index];
}
```

phprus: @vitaut 
Will a non-static constexpr array be optimized by the compiler?

vitaut: ¯\_(ツ)_/¯

mwinterb: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2647r1.html#workarounds
So, yes.
¯_(ツ)_/¯

gsjaardema: Done.

vitaut:
Thanks

include/fmt/format.h
jk-jeon: @mwinterb So the linked article says that it doesn't get optimized, no?

vitaut: It doesn't get optimized: https://godbolt.org/z/drn31x9zK

mwinterb: But clang will:
https://godbolt.org/z/YdYMGqs46
Sorry. My "So, yes" was agreeing with the shrug. "maybe, maybe not".
Should have been more clear.

mwinterb: > But clang will: https://godbolt.org/z/YdYMGqs46 Sorry. My "So, yes" was agreeing with the shrug. "maybe, maybe not". Should have been more clear.

Really sorry about that confusion.

jk-jeon: @mwinterb Ah, I see, my bad 😋

phprus: @vitaut 

> It doesn't get optimized: https://godbolt.org/z/drn31x9zK

Maybe revert this PR?

vitaut: I don't think we should revert this PR but ideas/PRs to optimize it would be welcome.

phprus: @vitaut 
Solution for GCC, Clang:  https://godbolt.org/z/ezj9GhnYW
But not for MSVC (current code: https://godbolt.org/z/5Pxn3zMfT)

gsjaardema: > I don't think we should revert this PR but ideas/PRs to optimize it would be welcome.

Turn it into a switch:  https://godbolt.org/z/YPj6G3zPd

vitaut: I have another idea: https://github.com/fmtlib/fmt/commit/06b20387aebdd47444b9eb3d3b3bc254251fca3b.

gsjaardema: > @vitaut Solution for GCC, Clang: https://godbolt.org/z/ezj9GhnYW But not for MSVC (current code: https://godbolt.org/z/5Pxn3zMfT)

That looks good.  I had something similar but didn't think of the `unreachable` to quiet the warnings and keep the optimization.

PR #3655, opened 2023-09-25
Add security policy to readme

joycebrum:
Just adding the security policy as a part of the README as mentioned in https://github.com/fmtlib/fmt/pull/3356


vitaut:
Thanks

PR #3660, opened 2023-09-29
fix: make std::bitset formattable again

muggenhor:
It used to be formattable via `operator<<(ostream&)` implicitly. Make it
formattable again, but this time via `formatter` specialization.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -4224,6 +4224,8 @@ struct nested_formatter {
   formatter<T> formatter_;
 
  public:
+  constexpr nested_formatter() : width_(0), align_(align_t::none) {}
+
   FMT_CONSTEXPR auto parse(format_parse_context& ctx) -> const char* {
     auto specs = detail::dynamic_format_specs<char>();
     auto it = parse_format_specs(
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -146,6 +146,32 @@ FMT_END_NAMESPACE
 
 FMT_BEGIN_NAMESPACE
 FMT_EXPORT
+template <std::size_t N, typename Char>
+struct formatter<std::bitset<N>, Char> : nested_formatter<string_view> {
+ private:
+  // Functor because C++11 doesn't support generic lambdas.
+  struct writer {
+    const std::bitset<N>& bs;
+
+    template <typename OutputIt>
+    FMT_CONSTEXPR OutputIt operator()(OutputIt out) {
+      for (auto pos = N; pos > 0; --pos) {
+        out = detail::write<Char>(out, bs[pos - 1] ? Char('1') : Char('0'));
+      }
+
+      return out;
+    }
+  };
+
+ public:
+  template <typename FormatContext>
+  auto format(const std::bitset<N>& bs, FormatContext& ctx) const
+      -> decltype(ctx.out()) {
+    return write_padded(ctx, writer{bs});
+  }
+};
+
+FMT_EXPORT
 template <typename Char>
 struct formatter<std::thread::id, Char> : basic_ostream_formatter<Char> {};
 FMT_END_NAMESPACE
--- a/test/std-test.cc
+++ b/test/std-test.cc
@@ -237,6 +237,14 @@ TEST(std_test, format_const_bit_reference) {
   EXPECT_EQ(fmt::format("{} {}", v[0], v[1]), "true false");
 }
 
+TEST(std_test, format_bitset) {
+  const std::bitset<6> bs(42);
+  EXPECT_EQ(fmt::format("{}", bs), "101010");
+  EXPECT_EQ(fmt::format("{:.4}", bs), "101010");
+  EXPECT_EQ(fmt::format("{:0>8}", bs), "00101010");
+  EXPECT_EQ(fmt::format("{:-^12}", bs), "---101010---");
+}
+
 TEST(std_test, format_atomic) {
   std::atomic<bool> b(false);
   EXPECT_EQ(fmt::format("{}", b), "false");


vitaut (changes requested):
Thanks for the PR. I think it's reasonable to provide a formatter for `bitset` but it shouldn't rely on ostream.

muggenhor:
Sure, I can do that. The formatting itself is easy enough if sticking to the same as [`operator<<(ostream&, const bitset&)`](https://en.cppreference.com/w/cpp/utility/bitset/operator_ltltgtgt2) (for loop from MSB to LSB).

But do you have any preference for a formatter to derive from or what to base a parse() on?

include/fmt/std.h
```diff
@@ -145,6 +145,26 @@ FMT_END_NAMESPACE
...
+  FMT_CONSTEXPR auto parse(ParseContext& ctx) -> decltype(ctx.begin()) {
+    return ctx.begin();
+  }
+
+  template <typename FormatContext>
+  auto format(const std::bitset<N>& bs, FormatContext& ctx) const
+      -> decltype(ctx.out()) {
+    auto out = ctx.out();
+
+    for (auto pos = N; pos; --pos) {
```
vitaut, line 160: nit: Let's avoid an implicit conversion in the ctor to make this less surprising and avoid potential warnings:

```c++
for (auto pos = N; pos > 0; --pos) {
```

vitaut (changes requested):
> But do you have any preference for a formatter to derive from or what to base a parse() on?

I think we should at least support width/precision/alignment. This can be done with `nested_formatter`, something like:

```c++
template <size_t N>
struct fmt::formatter<std::bitset<N>> : nested_formatter<string_view> {
  auto format(const std::bitset<N>& bs, FormatContext& ctx) const {
    return write_padded(ctx, [=](auto out) {
      for (auto pos = N; pos; --pos) {
        out = fmt::detail::write<char>(out, bs[pos - 1] ? '1' : '0');
      }
      return out;
    });
  }
};
```
but adapted to work with arbitrary code unit (character types).

https://godbolt.org/z/f6b7jhc89

include/fmt/std.h
muggenhor: Do you mean replacing contextual conversion to bool by comparing against zero? Because I'm not seeing any conversions on that line other than that at all.

muggenhor:
> I think we should at least support width/precision/alignment. This can be done with `nested_formatter`, something like:

Done. I also went ahead and adapted the test case to that by padding on the left with extra zeroes.

muggenhor:
Apparently `nested_formatter` wasn't trivially default constructible nor is it usable with lambdas on C++11 (which CI is apparently configured to build with). So had to fix the former and to resort to an old-fashioned functor for the latter.

include/fmt/format.h
```diff
@@ -4218,12 +4218,18 @@ struct formatter<nested_view<T>> {
...
+  int width_ = 0;
   detail::fill_t<char> fill_;
   align_t align_ : 4;
   formatter<T> formatter_;
 
  public:
+  // Using initialization list for 'align_' because default member initializers
+  // for bit-fields are C++20
+  constexpr nested_formatter() noexcept(
+      std::is_nothrow_constructible<formatter<T>>::value)
```
vitaut, line 4230: `noexcept` is not needed here

include/fmt/format.h
```diff
@@ -4218,12 +4218,18 @@ struct formatter<nested_view<T>> {
...
+      : align_(align_t::none) {}
```
vitaut, line 4231: Let's move initialization of `width_` here to keep all member initialization in one place and remove the comment above.

include/fmt/std.h
```diff
@@ -145,6 +145,32 @@ FMT_END_NAMESPACE
 #endif
 
 FMT_BEGIN_NAMESPACE
+FMT_EXPORT
+template <std::size_t N, typename Char>
+struct formatter<std::bitset<N>, Char> : nested_formatter<string_view> {
+ private:
+  // Functor because on C++11 lambdas don't take 'auto' parameters
```
vitaut, line 152: Let's rephrase a bit and add missing punctuation.
```c++
// Functor because C++11 doesn't support generic lambdas.
```

vitaut (changes requested):
Mostly looks good, just a few more comments inline.

vitaut (approved):

muggenhor:
Feel free to merge this if it looks good to you

Roman-Koshelev:
[std::bitset<N>::to_string](https://en.cppreference.com/w/cpp/utility/bitset/to_string) allows you to specify the zero and one characters. It would be great if the formatter supported this

vitaut:
Thank you!

muggenhor:
Thank you for your time and feedback!

> [std::bitset::to_string](https://en.cppreference.com/w/cpp/utility/bitset/to_string) allows you to specify the zero and one characters. It would be great if the formatter supported this

@Roman-Koshelev what syntax would you suggest for this in the format spec? Not having a clue about what syntax is a good idea here is why I didn't even attempt that.

Roman-Koshelev:
I believe the correct syntax should look like this
[fill[align]]["any_separator" zero[one]]["L"]

muggenhor:
Why make zero mandatory but one optional?

I understand this is practically the case for `bitset::to_string` too. But it isn't clear to me that that isn't more of a side effect of the way default arguments work than being an intentional design choice.

Roman-Koshelev:
An alternative can be
[fill[align]]["z" zero]["o" one]["L"]

Roman-Koshelev:
@muggenhor what do you think?

GielVanSchijndel-TomTom:
@Roman-Koshelev sorry for the delayed reply. I'm super strapped for time lately.

Anyway, that syntax looks good to me. Having the mandatory z/o characters there should even make parsing somewhat easier. A colleague mentioned that maybe s(et)/c(lear) would be preferred by some people? I don't care much either way

muggenhor:
@Roman-Koshelev sorry for the delayed reply. I'm super strapped for time lately.

Anyway, that syntax looks good to me. Having the mandatory z/o characters there should even make parsing somewhat easier. A colleague mentioned that maybe s(et)/c(lear) would be preferred by some people? I don't care much either way

PR #3661, opened 2023-09-29
minor fix to README.rst security policy link

joycebrum:
Sorry, I didn't noticed that the README was a rst file (I'm not used to this type of file actually)

I've fixed using the same format it is used on other links in the file and verified in the preview to be sure:

image

Sorry about that.


vitaut:
Missed it too during the review. Merged!

Issue #3654, opened 2023-09-24
Compilation fails using Clang 17 toolchain


Steveice10:
MSYS2 just updated their clang and associated packages to 17.0.1. We are now seeing the following compilation error:

```
fmt/include/fmt/ostream.h:17:12: fatal error: '__std_stream' file not found

   17 | #  include <__std_stream>

      |            ^~~~~~~~~~~~~~
```

This seems to be related to the following change in libcxx: https://reviews.llvm.org/D147855

vitaut:
Removed the nonportable include in https://github.com/fmtlib/fmt/commit/2dd4fa8742fdac36468f8d8ea3e06e78215551f8. Thanks for reporting.

Steveice10:
Works now, thanks.

lazka:
Note that there is still one include left here: https://github.com/fmtlib/fmt/blob/79dbd3f19285f2a6ce0c90932f06c173e424e25b/src/fmt.cc#L66-L68 which makes the build with "cmake -DFMT_MODULE=ON" fail.

vitaut:
@lazka, could you submit a PR to remove it too?

lazka:
sure, #3663

Roman-Koshelev:
@vitaut wouldn't it make more sense to leave this hack for libc++ < 17? (native_handle() should be at 18)

vitaut:
I don't think it's spending any efforts on conditionally supporting this hack for iostreams which are a legacy API.

PR #3663, opened 2023-10-01
Remove leftover usage of the __std_stream header

lazka:
2dd4fa8742fdac36468f8d removed all usage of __std_stream because it is no longer available with clang v17. That commit missed one place where the header was still used (only used when building with -DFMT_MODULE=ON). Remove it there too.

See #3654


vitaut:
Thanks!

PR #3664, opened 2023-10-01
suffixes for helper chrono types added in C++  20

Richardk2n:
C++ 20 added
`std::chrono::days`
`std::chrono::weeks`
`std::chrono::months`
`std::chrono::years`

This adds suffixes for them.

The standard only specifies `d` for days, but it also specifies `min` for minutes, while fmt uses `m` so I am not sure how far you want to deviate from there.

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -582,12 +582,9 @@ template <typename Period> FMT_CONSTEXPR inline const char* get_units() {
   if (std::is_same<Period, std::tera>::value) return "Ts";
   if (std::is_same<Period, std::peta>::value) return "Ps";
   if (std::is_same<Period, std::exa>::value) return "Es";
-  if (std::is_same<Period, std::ratio<60>>::value) return "m";
+  if (std::is_same<Period, std::ratio<60>>::value) return "min";
   if (std::is_same<Period, std::ratio<3600>>::value) return "h";
   if (std::is_same<Period, std::ratio<86400>>::value) return "d";
-  if (std::is_same<Period, std::ratio<604800>>::value) return "wk";
-  if (std::is_same<Period, std::ratio<2629746>>::value) return "mo";
-  if (std::is_same<Period, std::ratio<31556952>>::value) return "a";
   return nullptr;
 }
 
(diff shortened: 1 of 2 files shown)


vitaut (changes requested):
Thanks for the PR but except for `d` I don't see these suffixes defined in https://eel.is/c++draft/time.duration.io.

include/fmt/chrono.h
```diff
@@ -584,6 +584,10 @@ template <typename Period> FMT_CONSTEXPR inline const char* get_units() {
   if (std::is_same<Period, std::exa>::value) return "Es";
   if (std::is_same<Period, std::ratio<60>>::value) return "m";
```
vitaut, line 585: Please change this to "min" while at it.

include/fmt/chrono.h
```diff
@@ -582,8 +582,9 @@ template <typename Period> FMT_CONSTEXPR inline const char* get_units() {
   if (std::is_same<Period, std::tera>::value) return "Ts";
   if (std::is_same<Period, std::peta>::value) return "Ps";
   if (std::is_same<Period, std::exa>::value) return "Es";
-  if (std::is_same<Period, std::ratio<60>>::value) return "m";
+  if (std::is_same<Period, std::ratio<60>>::value) return "min";
   if (std::is_same<Period, std::ratio<3600>>::value) return "h";
+  if (std::is_same<Period, std::ratio<86400>>::value) return "d";
```
vitaut, line 587: Please add a test case to https://github.com/fmtlib/fmt/blob/f76603f21ea75bcd8d0007d7599cac75862b13b9/test/chrono-test.cc#L425C19-L425C33.

Richardk2n: Well, according to the test, I have no clue how to use `__cpp_lib_chrono` …

phprus:
```c++
#if defined(__cpp_lib_chrono) && __cpp_lib_chrono >= 201907L
```

vitaut (approved):

vitaut:
Thank you!

PR #3668, opened 2023-10-03
Added synchronization with the main storage device before direct recording

Roman-Koshelev:
Added reset of caching stream buffers before direct writing to the descriptor

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -73,9 +73,8 @@ FMT_FUNC void report_error(format_func func, int error_code,
 }
 
 // A wrapper around fwrite that throws on error.
-inline void fwrite_fully(const void* ptr, size_t size, size_t count,
-                         FILE* stream) {
-  size_t written = std::fwrite(ptr, size, count, stream);
+inline void fwrite_fully(const void* ptr, size_t count, FILE* stream) {
+  size_t written = std::fwrite(ptr, 1, count, stream);
   if (written < count)
     FMT_THROW(system_error(errno, FMT_STRING("cannot write to file")));
 }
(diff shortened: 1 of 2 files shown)


vitaut (changes requested):
Thanks for the PR but please revert unrelated formatting changes and add more context. I applied clang-format separately.

Roman-Koshelev:
Done.

include/fmt/ostream.h
```diff
@@ -38,24 +38,24 @@ auto get_file(std::filebuf&) -> FILE*;
...
     c_file = fbuf->file();
   else
     return false;
-  if (c_file) return write_console(c_file, data);
-#else
-  ignore_unused(os, data);
 #endif
-  return false;
+  if (!c_file) return false;
+  os.flush();
```
vitaut, line 55: We should only flush the buffer when writing to a console. I suggest moving the ostream change to a separate PR and addressing this issue. The rest LGTM.

vitaut (approved):

vitaut:
Merged without the ostream part in https://github.com/fmtlib/fmt/commit/3b7f58a8b3e35cfee57558a41e3e897b0ba7106a. Thanks.

include/fmt/format-inl.h
```diff
@@ -1433,13 +1432,12 @@ extern "C" __declspec(dllimport) int __stdcall WriteConsoleW(  //
     void*, const void*, dword, dword*, void*);
 
 FMT_FUNC bool write_console(std::FILE* f, string_view text) {
-  auto fd = _fileno(f);
+  int fd = _fileno(f);
   if (!_isatty(fd)) return false;
+  std::fflush(f);
```
dimztimz, line 1437: Adding `fflush` here is not correct. This function is called either from the C stream print() or the C++ iostream print(). In the latter case, fflush should not be called at all, instread the ostream flush should be invoked.

vitaut: This flush is indeed redundant for iostream but in the future please report issues like these once to avoid too much notification noise.

PR #3679, opened 2023-10-12
Add support for CMake 3.28 C++ module features

jcar87:
CMake 3.28 ([rc1 announcement](https://www.kitware.com/cmake-3-28-0-rc1-is-ready-for-testing/)) adds support for C++ modules that is no longer experimental (so no need to add the experimental flags).

This PR adds support for this using the new features, when the build is configured with `FMT_MODULE` set to `ON` and the CMake version is 3.28 or higher.

Tested both building fmt, and then calling `cmake install` and using the fmt module from a different project. This works with:

* Linux, Clang 17, ninja 1.11.1
* Linux, gcc 13.1, ninja 1.11.1 (note that I'm using a patched version of gcc with dependency scanning support, and additional CMake variables since CMake expects gcc 14 for this, but gcc14 ICEs when building fmt as a module, I've reported this [here](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=111785))
* Windows, msvc 19.37
  - Building fmt itself works with the Visual Studio generator or Ninja
  - Using the imported targets with  `find_package(fmt)` works with Ninja, but not with the Visual Studio generator - reported this [here](https://github.com/fmtlib/fmt/pull/3679#issue-1940145260).

Hopefully this is OK. For CMake versions earlier than 3.28, this should behave as it was previously.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.8...3.26)
+cmake_minimum_required(VERSION 3.8...3.28)
 
 # Fallback for using newer policies on CMake <3.12.
 if(${CMAKE_VERSION} VERSION_LESS 3.12)
@@ -26,7 +26,7 @@ endfunction()
 
 # DEPRECATED! Should be merged into add_module_library.
 function(enable_module target)
-  if (MSVC)
+  if (MSVC AND CMAKE_VERSION VERSION_LESS 3.28)
     set(BMI ${CMAKE_CURRENT_BINARY_DIR}/${target}.ifc)
     target_compile_options(${target}
       PRIVATE /interface /ifcOutput ${BMI}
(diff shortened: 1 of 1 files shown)


vitaut (changes requested):
Thanks for the PR but the part of CMake you are changing is deprecated and will be removed. Please add this to `add_module_library` instead.

jcar87:
Hi @vitaut - thanks for the review.
The changes in this PR _are_ mostly in `add_module_library()` - the only change in `enable_module()` is that I've added a conditional to skip the logic altogether when CMake is >= 3.28, as I can see that `enable_module()` is still called, and what it aims to do is already handled by CMake+ninja or by MSBuild.

However if this is not the best way, I'm happy to move the conditional out of `enable_module()` and into the two places where it is still called?

vitaut:
Sorry, I just realized that it's not super clear from the code and my comment but what I meant is that the source of truth of `add_module_library` is in https://github.com/vitaut/modules. Could you submit the change there to prevent it being overwritten in this repo next time we sync the two?

vitaut (approved):

vitaut:
I merged it for now but as commented earlier it will be overwritten unless also added to the main modules repo. Thanks!

vitaut:
Revered because of incompatibility with CMake 3.16: #3714.

jcar87:
Apologies @vitaut - I will make the contribution in the main modules repo and make sure there's no regression with older CMake.

vitaut:
No worries and thanks for working on this.

PR #3680, opened 2023-10-13
Fix error C2668 on Windows with option /std:c++latest

Zhaojun-Liu:
When build fmt with MSVC under option /std:c++latest, it failed due to `error 2668: 'std::format_to': ambiguous call to overloaded function`, so add namespace to qualify the call to format_to to avoid this issue. Could you please review and approval it? Thanks.

It is the same issue as https://github.com/fmtlib/fmt/issues/3377.

### DIFF
--- a/test/format-test.cc
+++ b/test/format-test.cc
@@ -1788,7 +1788,8 @@ FMT_BEGIN_NAMESPACE
 template <> struct formatter<point> : nested_formatter<double> {
   auto format(point p, format_context& ctx) const -> decltype(ctx.out()) {
     return write_padded(ctx, [this, p](auto out) -> decltype(out) {
-      return format_to(out, "({}, {})", nested(p.x), nested(p.y));
+      // Namespace-qualify to avoid ambiguity with std::format_to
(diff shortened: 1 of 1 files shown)


test/format-test.cc
```diff
@@ -1788,7 +1788,8 @@ FMT_BEGIN_NAMESPACE
 template <> struct formatter<point> : nested_formatter<double> {
   auto format(point p, format_context& ctx) const -> decltype(ctx.out()) {
     return write_padded(ctx, [this, p](auto out) -> decltype(out) {
-      return format_to(out, "({}, {})", nested(p.x), nested(p.y));
+      // Namespace-qualify to avoid ambiguity with std::format_to
```
vitaut, line 1791: I don't think this comment is needed because we have a bunch of such calls and it doesn't make sense to explain conflict with std (with is a known issue in general) on one random call. Please remove.

vitaut (changes requested):
Thanks for the PR. Mostly looks good but please drop the comment.

test/format-test.cc
Zhaojun-Liu: Ok, I have removed the comment. Thanks for reviewing.

Issue #3106, opened 2022-09-19
Long double errors: Clang, libc++, CUDA


j8asic:
Seems like `libc++` implementations of device functions do not support `long double` (i.e. CUDA doesnt have long double), so I am getting errors in `visit_format_arg`
The treatment for me was to use long doubles as regular double here:
https://github.com/fmtlib/fmt/blob/a48e3355a682b7e76d672c4dbd64aad2aad2ec6e/include/fmt/core.h#L1662

vitaut:
{fmt} requires basic C++ support including `long double`. You might want to report an issue to NVIDIA asking them to at least make a sensible fallback.

j8asic:
@vitaut so the `libc++` guys say they are right to disable `long double` math functions on the device https://godbolt.org/z/9Th19nWqP. So this bug I rewrite as: “fmt tries to call device math functions for long double, instead of appropriate host functions”

vitaut:
No idea what this means but a PR with a workaround for CUDA will be welcome provided that it's not too intrusive.

PR #3687, opened 2023-10-23
Skip long_double_value visitor when FMT_USE_FLOAT128 is not defined

j8asic:
There is a macro FMT_USE_FLOAT128 which is defined as 0 or 1, but it is not tested in visit_format_arg, and raises errors in some specific situations. For example: according to https://github.com/fmtlib/fmt/issues/3106, libc++ implementations of device functions do not support long double, so there are errors in visit_format_arg even when defining FMT_USE_FLOAT128=0


vitaut:
Thanks for the PR but as commented in #3106:

> {fmt} requires basic C++ support including `long double`. 

I don't think we should do heroics to support broken platforms.

PR #3689, opened 2023-10-25
Fix flushing C++ iostreams before calling write_console()

### LINKED ISSUE #3688: Incorrect call to fflush() inside write_console()
In this very recent change 3b7f58a8b3e35cfee57558a41e3e897b0ba7106a which originates from this PR #3668, the added call to `fflush` is incorrect and may brake the `std::ostream` overloads of `print()`. The intention was to implement the paper https://wg21.link/P2539/ in FMT here.

The correct behavior is to remove `fflush` from `write_console()` and call it before calling `write_console()`, but only from the C stream overloads of print, i.e. `print(FILE*,...)`. For the ostream overloads one should call `ostream::flush()`.

dimztimz:
This change correctly implements https://wg21.link/P2539/ for both C streams and C++ iostreams.

Fixes #3688.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -1425,16 +1425,13 @@ FMT_FUNC std::string vformat(string_view fmt, format_args args) {
 
 namespace detail {
 #if !defined(_WIN32) || defined(FMT_WINDOWS_NO_WCHAR)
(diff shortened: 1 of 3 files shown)


vitaut:
Thanks!

PR #3702, opened 2023-11-04
Set PDB path for Visual Studio

pklima:
Ensure the PDB files are output into the same directory and with the same name as the static library when using Visual Studio.

Resolves fmtlib#3701.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -312,7 +312,15 @@ set(FMT_DEBUG_POSTFIX d CACHE STRING "Debug library postfix.")
 set_target_properties(fmt PROPERTIES
   VERSION ${FMT_VERSION} SOVERSION ${CPACK_PACKAGE_VERSION_MAJOR}
   PUBLIC_HEADER "${FMT_HEADERS}"
-  DEBUG_POSTFIX "${FMT_DEBUG_POSTFIX}")
+  DEBUG_POSTFIX "${FMT_DEBUG_POSTFIX}"
+
+  # Workaround for Visual Studio 2017:
+  # Ensure the .pdb is created with the same name and in the same directory
+  # as the .lib. Newer VS versions already do this by default, but there is no
+  # harm in setting it for those too. Ignored by other generators.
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -312,7 +312,10 @@ set(FMT_DEBUG_POSTFIX d CACHE STRING "Debug library postfix.")
 set_target_properties(fmt PROPERTIES
   VERSION ${FMT_VERSION} SOVERSION ${CPACK_PACKAGE_VERSION_MAJOR}
   PUBLIC_HEADER "${FMT_HEADERS}"
-  DEBUG_POSTFIX "${FMT_DEBUG_POSTFIX}")
+  DEBUG_POSTFIX "${FMT_DEBUG_POSTFIX}"
+  COMPILE_PDB_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}"
+  COMPILE_PDB_NAME "fmt"
+  COMPILE_PDB_NAME_DEBUG "fmt${FMT_DEBUG_POSTFIX}")
```
vitaut, line 318: Please add a short comment clarifying that these are for MSVC 2017 only. We'll remove this workaround in the future once the compiler version is no longer supported.

vitaut (changes requested):
Thanks for the PR

CMakeLists.txt
pklima: Happy to do that, but I'm not sure this should be looked at as a temporary workaround. Not having it leaves you at the mercy of the defaults and I'm worried that in some future VS version the defaults could change again to who knows what.

Would you still like me to describe it as a temporary workaround, or as something that simply has to be set for Visual Studio?

vitaut: > Would you still like me to describe it as a temporary workaround, or as something that simply has to be set for Visual Studio?

Yes, please. We don't need to guard against theoretical future MSVC regressions.

pklima: OK, done!

vitaut (approved):

PR #3709, opened 2023-11-10
Improve calculation for the product of two uint64_t on x64 with gcc or clang

laurent-onenano:
On x86_64, CPUs have an instruction to multiply two 64 integers and spread the result over two registers.
This allows to compute efficiently with a single operation the long multiplication.

https://www.agner.org/optimize/instruction_tables.pdf

https://www.felixcloutier.com/x86/mul


vitaut:
Thanks for the PR but existing implementation is already compiled to a single `mul` instruction: https://godbolt.org/z/zqGMz5sE4. The codegen is even slightly better than the inline assembly version: https://godbolt.org/z/nezaM9Yqv.

laurent-onenano:
Sounds great!

Thanks for the update.

Le ven. 10 nov. 2023 à 18:48, Victor Zverovich ***@***.***> a
écrit :

PR #3716, opened 2023-11-15
Added workaround for old xcode compiler bug

kuguma:
Older xcode (at least xcode12) had a bug that defined macros for `__cpp_lib_filesystem` and `__cpp_lib_variant` even if the target did not support it.
When building for a target lower than macOS 10.15 with c++17, fmtlib compilation failed. (https://developer.apple.com/xcode/cpp/#c++17)

As a workaround, I would like to add a flag to disable std::filesystem and std::variant.

This problem seems to have been fixed in the latest xcode15, but I don't think version checking is practical.

Similar issues in other projects: https://github.com/ToruNiina/toml11/issues/150

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -59,7 +59,7 @@
 #  endif
 #endif
 
-#ifdef __cpp_lib_filesystem
+#if !defined(FMT_DISABLE_STD_FILESYSTEM) && defined(__cpp_lib_filesystem)
 FMT_BEGIN_NAMESPACE
 
 namespace detail {
@@ -211,7 +211,7 @@ struct formatter<std::optional<T>, Char,
 FMT_END_NAMESPACE
 #endif  // __cpp_lib_optional
 
-#ifdef __cpp_lib_variant
+#if !defined(FMT_DISABLE_STD_VARIANT) && defined(__cpp_lib_variant)
 FMT_BEGIN_NAMESPACE
 namespace detail {
 
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -59,7 +59,24 @@
 #  endif
 #endif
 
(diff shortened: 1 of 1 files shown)


include/fmt/std.h
```diff
@@ -59,7 +59,7 @@
 #  endif
 #endif
 
-#ifdef __cpp_lib_filesystem
+#if !defined(FMT_DISABLE_STD_FILESYSTEM) && defined(__cpp_lib_filesystem)
```
vitaut, line 62: Let's follow the {fmt} convention and replace `FMT_DISABLE_STD_FILESYSTEM` with `FMT_CPP_LIB_FILESYSTEM` that is initialized to `__cpp_lib_filesystem` if the latter is defined and `FMT_CPP_LIB_FILESYSTEM` is not defined by the user:

```c++
#ifndef FMT_CPP_LIB_FILESYSTEM
# ifdef __cpp_lib_filesystem
#   define FMT_CPP_LIB_FILESYSTEM __cpp_lib_filesystem
# else
#   define FMT_CPP_LIB_FILESYSTEM 0
# endif
#endif

Then you can disable filesystem via `-DFMT_CPP_LIB_FILESYSTEM=0`.

And similarly with `__cpp_lib_variant`.

vitaut (changes requested):

include/fmt/std.h
kuguma: Thank you for the review, I've made the correction.

vitaut (approved):

vitaut:
Thank you

kuguma:
> Looks like xcode12 is no longer supported by Homebrew. @kuguma is this workaround still needed?

Yes. but I think it would be a good idea to sort this out in a new version of fmtlib.

PR #3730, opened 2023-11-28
Added formatter for std::source_location

felix642:
Added custom formatter for std::source_location to support this type of code :

`fmt::print("{}\n", std::source_location::current());`

The format was taken from the following discussion : https://github.com/fmtlib/fmt/issues/3012#issuecomment-1206515140
Let me know if there is a better format that could be implemented.

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -235,9 +235,8 @@ FMT_END_NAMESPACE
 #ifdef __cpp_lib_source_location
 FMT_BEGIN_NAMESPACE
 FMT_EXPORT
-template <typename Char>
-struct formatter<std::source_location, Char,
-                 std::enable_if_t<is_formattable<Char>::value>> {
+template<>
+struct formatter<std::source_location> {
   template <typename ParseContext> FMT_CONSTEXPR auto parse(ParseContext& ctx) {
     return ctx.begin();
   }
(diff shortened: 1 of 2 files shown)


include/fmt/std.h
```diff
@@ -228,6 +233,33 @@ struct formatter<std::optional<T>, Char,
...
+
+  template <typename FormatContext>
+  auto format(const std::source_location& loc, FormatContext& ctx) const
+      -> decltype(ctx.out()) {
+    auto out = ctx.out();
+    out = detail::write_bytes(out, loc.file_name(), format_specs<Char>());
+    out = detail::write<Char>(out, Char(':'));
+    out = detail::write<Char>(out, loc.line());
+    out = detail::write<Char>(out, Char(':'));
+    out = detail::write<Char>(out, loc.column());
```
vitaut, line 254: There needs to be another `:` after the column.

include/fmt/std.h
```diff
@@ -13,6 +13,7 @@
 #include <cstdlib>
 #include <exception>
 #include <memory>
+#include <source_location>
```
vitaut, line 16: Extra (unconditional) include, please remove.

vitaut (changes requested):
Thanks for the PR!

include/fmt/std.h
felix642: Sorry about that, forgot to remove it when I added the conditional include.

felix642:
Pipeline seems to fail on windows, I'll have a look and mark the PR as ready for review when the pipeline will be ok.

include/fmt/std.h
```diff
@@ -228,6 +232,33 @@ struct formatter<std::optional<T>, Char,
 FMT_END_NAMESPACE
 #endif  // __cpp_lib_optional
 
+#ifdef __cpp_lib_source_location
+FMT_BEGIN_NAMESPACE
+FMT_EXPORT
+template <typename Char>
+struct formatter<std::source_location, Char,
```
vitaut, line 239: `std::source_location` can only be formattable with `char` so let's remove the `Char` template parameter.

include/fmt/std.h
```diff
@@ -228,6 +232,33 @@ struct formatter<std::optional<T>, Char,
...
+                 std::enable_if_t<is_formattable<Char>::value>> {
```
vitaut, line 240: `enable_if_t` is not needed

vitaut (changes requested):

include/fmt/std.h
```diff
@@ -228,6 +232,32 @@ struct formatter<std::optional<T>, Char,
...
+struct formatter<std::source_location> {
+  template <typename ParseContext> FMT_CONSTEXPR auto parse(ParseContext& ctx) {
+    return ctx.begin();
+  }
+
+  template <typename FormatContext>
+  auto format(const std::source_location& loc, FormatContext& ctx) const
+      -> decltype(ctx.out()) {
+    auto out = ctx.out();
+    out = detail::write_bytes(out, loc.file_name(), format_specs<char>());
```
vitaut, line 248: And one more thing: this can be just `write`, not `write_bytes` (similar to `function_name`).

vitaut (changes requested):

vitaut:
Merged, thanks!

PR #3739, opened 2023-12-08
allow format_as() to format reference

tchaikov:
before this change, format_as() is unable to format a type which has `auto format_as() -> const another_type&`, and `another_type` is formattable. because `format_as_result` maps the result type as it is, and the compiler refuses to compile
`static_cast<T*>(nullptr)`, where T is a reference type. but it would be handy if we could use `format_as()` to format types which, for instance, owns / inherit from a formattable type, and delegate the formatter to these variables instead without creating a copy of them.

in this change:

* instruct `format_as_result` to map the result type to the decayed type, so that `type` can be the decayed type of result type, and this also enables `type` to be formattable, as long as the decayed type is formattable.
* corresponding test is added to format-test.cc

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -1332,7 +1332,7 @@ using ulong_type = conditional_t<long_short, unsigned, unsigned long long>;
 template <typename T> struct format_as_result {
   template <typename U,
             FMT_ENABLE_IF(std::is_enum<U>::value || std::is_class<U>::value)>
-  static auto map(U*) -> decltype(format_as(std::declval<U>()));
+  static auto map(U*) -> remove_cvref_t<decltype(format_as(std::declval<U>()))>;
   static auto map(...) -> void;
 
   using type = decltype(map(static_cast<T*>(nullptr)));
--- a/test/format-test.cc
+++ b/test/format-test.cc
@@ -2173,6 +2173,13 @@ auto format_as(scoped_enum_as_string) -> std::string { return "foo"; }
 
 struct struct_as_int {};
 auto format_as(struct_as_int) -> int { return 42; }
+
+struct struct_as_const_reference {
+  const std::string name = "foo";
+};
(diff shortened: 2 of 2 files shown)


include/fmt/core.h
```diff
@@ -1332,7 +1332,7 @@ using ulong_type = conditional_t<long_short, unsigned, unsigned long long>;
 template <typename T> struct format_as_result {
   template <typename U,
             FMT_ENABLE_IF(std::is_enum<U>::value || std::is_class<U>::value)>
-  static auto map(U*) -> decltype(format_as(std::declval<U>()));
+  static auto map(U*) -> typename std::decay<decltype(format_as(std::declval<U>()))>::type;
```
vitaut, line 1335: I think this should be `remove_cvref_t` and not `decay`.

test/format-test.cc
```diff
@@ -2173,13 +2173,20 @@ auto format_as(scoped_enum_as_string) -> std::string { return "foo"; }
 
 struct struct_as_int {};
 auto format_as(struct_as_int) -> int { return 42; }
+
+struct struct_as_const_reference {
+  const std::string name = "foo";
+};
+auto format_as(const struct_as_const_reference& s) -> const std::string& { return s.name; }
```
vitaut, line 2180: Please apply clang-format.

vitaut (changes requested):
Thanks for the PR!

include/fmt/core.h
tchaikov: yes. it's a better fit in this case.

test/format-test.cc
tchaikov: sure.

tchaikov:
v2:

- s/std::decay/remove_cvref_t/
- clang-format the change.

vitaut (approved):

vitaut:
Merged, thanks!

Issue #3738, opened 2023-12-07
Compile error with GCC 6.9.0


asantoni:
I ran into a compile error on GCC 6.3.0 (Debian 9). The last line is the most relevant:

```
/stuff/vcpkg_installed/x64-linux/include/spdlog/common.h: In function 'constexpr spdlog::string_view_t spdlog::details::to_string_view(const memory_buf_t&)':
/stuff/vcpkg_installed/x64-linux/include/spdlog/common.h:351:42: error: call to non-constexpr function 'constexpr const T* fmt::v10::detail::buffer<T>::data() const [with T = char]'
     return spdlog::string_view_t{buf.data(), buf.size()};
                                  ~~~~~~~~^~
In file included from /stuff/vcpkg_installed/x64-linux/include/fmt/format.h:49:0,
                 from /stuff/vcpkg_installed/x64-linux/include/fmt/std.h:22,
                 from /stuff/src/logger.h:10,
                 from /stuff/src/redacted.h:15,
                 from /stuff/src/redacted.cpp:1:
/stuff/vcpkg_installed/x64-linux/include/fmt/core.h:845:22: note: 'constexpr const T* fmt::v10::detail::buffer<T>::data() const [with T = char]' is not usable as a constexpr function because:
   FMT_CONSTEXPR auto data() const noexcept -> const T* { return ptr_; }
                      ^~~~
/stuff/vcpkg_installed/x64-linux/include/fmt/core.h:845:22: error: enclosing class of constexpr non-static member function 'constexpr const T* fmt::v10::detail::buffer<T>::data() const [with T = char]' is not a literal type
/stuff/vcpkg_installed/x64-linux/include/fmt/core.h:798:29: note: 'fmt::v10::detail::buffer<char>' is not literal because:
 template <typename T> class buffer {
                             ^~~~~~
/stuff/vcpkg_installed/x64-linux/include/fmt/core.h:798:29: note:   'fmt::v10::detail::buffer<char>' is not an aggregate, does not have a trivial default constructor, and has no constexpr constructor that is not a copy or move constructor
```

This issue is similar to https://github.com/fmtlib/fmt/issues/3096

My patch that fixes this compile error is:

```
diff --git a/include/fmt/core.h b/include/fmt/core.h
index 1fe1388..f97c630 100755
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -804,7 +804,7 @@ template <typename T> class buffer {
  protected:
[...]

PR #3743, opened 2023-12-08
Fix compile with GCC 6.3.0 (bug #3738)

asantoni:
GCC 6.3.0 compile fix, for bug #3738. Pull request, as requested. :)


vitaut:
Thank you!

PR #3744, opened 2023-12-08
Fix README file table

GeorgeLS:
This PR fixes the Speed Tests table format. I saw that the previous format did not render correctly in GitHub and it was completely broken.
The rest of the diff comes from just reformatting the file.


vitaut:
Thanks!

PR #3746, opened 2023-12-09
Update isfinite declaration to be constexpr for c++23 only

### LINKED ISSUE #3745: `isfinite` declared `constexpr` for c++20
`std::isfinite` is not declared `constexpr` before `c++23`, but `fmt::isfinite` is declared `constexpr` for `c++20`
(https://en.cppreference.com/w/cpp/numeric/math/isfinite).

https://github.com/fmtlib/fmt/blob/master/include/fmt/core.h#L108:
```
#if ((FMT_CPLUSPLUS >= 202002L) &&                            \
     (!defined(_GLIBCXX_RELEASE) || _GLIBCXX_RELEASE > 9)) || \
    (FMT_CPLUSPLUS >= 201709L && FMT_GCC_VERSION >= 1002)
#  define FMT_CONSTEXPR20 constexpr
#else
#  define FMT_CONSTEXPR20
#endif
```

https://github.com/fmtlib/fmt/blob/master/include/fmt/format.h#L2753:
```
template <typename T, FMT_ENABLE_IF(std::is_floating_point<T>::value&&
                                        has_isfinite<T>::value)>
FMT_CONSTEXPR20 bool isfinite(T value) {
  constexpr T inf = T(std::numeric_limits<double>::infinity());
  if (is_constant_evaluated())
    return !detail::isnan(value) && value < inf && value > -inf;
  return std::isfinite(value);
}
```

Resulting compilation error:
```
external/fmtlib/test/format-test.cc:94:24: error: constexpr variable 'result' must be initialized by a constant expression
  FMT_CONSTEXPR20 auto result = isfinite(Float(limits::infinity()));
                       ^        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
external/fmtlib/test/format-test.cc:103:3: note: in instantiation of function template specialization 'check_isfinite<double>' requested here
  check_isfinite<double>();
  ^
external/fmtlib/include/fmt/format.h:2751:10: note: non-constexpr function 'isfinite<double>' cannot be used in a constant expression
  return std::isfinite(value);
         ^
external/fmtlib/test/format-test.cc:94:33: note: in call to 'isfinite(INF)'
  FMT_CONSTEXPR20 auto result = isfinite(Float(limits::infinity()));
                                ^
external/libcxx/include/math.h:439:1: note: declared here
isfinite(_A1 __lcpp_x) _NOEXCEPT
^
1 error generated.
22:46:14 ninja failed with: exit status 1
```

Compiler version (Android :shrug: )
```
Android (11039501, +pgo, +bolt, +lto, +mlgo, based on r498229b) clang version 17.0.4 (https://android.googlesource.com/toolchain/llvm-project e34ed7d63863b45858e74126edaa738e75887800)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /usr/local/google/home/henrichataing/Projects/android/aosp-master/prebuilts/clang/host/linux-x86/clang-r498229b/bin
```

hchataing:
Fix #3745

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -2752,7 +2752,7 @@ template <typename T, FMT_ENABLE_IF(std::is_floating_point<T>::value&&
                                         has_isfinite<T>::value)>
 FMT_CONSTEXPR20 bool isfinite(T value) {
   constexpr T inf = T(std::numeric_limits<double>::infinity());
-  if (is_constant_evaluated())
+  if (is_constant_evaluated(true))
     return !detail::isnan(value) && value < inf && value > -inf;
   return std::isfinite(value);
 }


phprus:
Question about the environment:
https://github.com/fmtlib/fmt/issues/3745#issuecomment-1848268326

vitaut (changes requested):
@phprus made a good point and it looks like a bug in `is_constant_evaluated` and not `isfinite`.

hchataing:
Granted, I might have been too quick in diagnosing the issue

include/fmt/format.h
```diff
@@ -2752,7 +2752,7 @@ template <typename T, FMT_ENABLE_IF(std::is_floating_point<T>::value&&
                                         has_isfinite<T>::value)>
 FMT_CONSTEXPR20 bool isfinite(T value) {
   constexpr T inf = T(std::numeric_limits<double>::infinity());
-  if (is_constant_evaluated())
+  if (is_constant_evaluated(true))
```
vitaut, line 2755: One drawback is that it switches to a slightly worse implementation on older systems: https://www.godbolt.org/z/WWzjEsrxj. But I guess we can optimize it later.

vitaut (approved):

include/fmt/format.h
hchataing: I don't see an alternative other than testing `FMT_CONSTEXPR20` in the body to replace  `is_constant_evaluated` :(

phprus: We can use ``__builtin_is_constant_evaluated`` instead of ``std::is_constant_evaluated``.

phprus: Or we can improve 

https://github.com/fmtlib/fmt/blob/89860eb9013a345608c8144b1aad5f12b0682d7e/include/fmt/core.h#L108-L110

condition by adding a stdlib version check.

@hchataing, which version of libc++ or libstdc++ you are using?

phprus: @hchataing check PR #3754 please

vitaut:
Closing in favor of #3754 but thanks for the effort.

include/fmt/format.h
hchataing: Neither macro is defined by the compiler... but maybe I am checking wrong (`clang++ -std=c++20 -E -dM foo.cc`)

PR #3747, opened 2023-12-09
Fix typo in gmock-gtest-all.cc

eltociear:
syntetic -> synthetic


vitaut:
This is third-party (GTest) code but I guess we can fix it here too, thanks.

PR #3754, opened 2023-12-13
Add stdlib version check for C++20

### LINKED ISSUE #3745: `isfinite` declared `constexpr` for c++20

phprus:
Possible fix #3745

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -105,9 +105,12 @@
 #  define FMT_CONSTEXPR
 #endif
 
-#if ((FMT_CPLUSPLUS >= 202002L) &&                            \
-     (!defined(_GLIBCXX_RELEASE) || _GLIBCXX_RELEASE > 9)) || \
-    (FMT_CPLUSPLUS >= 201709L && FMT_GCC_VERSION >= 1002)
+#if (FMT_CPLUSPLUS >= 202002L ||                                \
+     (FMT_CPLUSPLUS >= 201709L && FMT_GCC_VERSION >= 1002)) &&  \
+    ((!defined(_GLIBCXX_RELEASE) || _GLIBCXX_RELEASE >= 10) &&  \
+     (!defined(_LIBCPP_VERSION) || _LIBCPP_VERSION >= 10000) && \
+     (!FMT_MSC_VERSION || FMT_MSC_VERSION >= 1928)) &&          \
+    defined(__cpp_lib_is_constant_evaluated)
 #  define FMT_CONSTEXPR20 constexpr
 #else
 #  define FMT_CONSTEXPR20
--- a/test/compile-test.cc
+++ b/test/compile-test.cc
@@ -280,15 +280,18 @@ TEST(compile_test, compile_format_string_literal) {
 #endif
 
 // MSVS 2019 19.29.30145.0 - Support C++20 and OK.
-// MSVS 2022 19.32.31332.0 - compile-test.cc(362,3): fatal error C1001: Internal
-// compiler error.
+// MSVS 2022 19.32.31332.0, 19.37.32826.1 - compile-test.cc(362,3): fatal error
+// C1001: Internal compiler error.
 //  (compiler file
 //  'D:\a\_work\1\s\src\vctools\Compiler\CxxFE\sl\p1\c\constexpr\constexpr.cpp',
 //  line 8635)
-#if ((FMT_CPLUSPLUS >= 202002L) &&                           \
-     (!defined(_GLIBCXX_RELEASE) || _GLIBCXX_RELEASE > 9) && \
-     (!FMT_MSC_VERSION || FMT_MSC_VERSION < 1930)) ||        \
-    (FMT_CPLUSPLUS >= 201709L && FMT_GCC_VERSION >= 1002)
+#if (FMT_CPLUSPLUS >= 202002L ||                                \
+     (FMT_CPLUSPLUS >= 201709L && FMT_GCC_VERSION >= 1002)) &&  \
+    ((!defined(_GLIBCXX_RELEASE) || _GLIBCXX_RELEASE >= 10) &&  \
+     (!defined(_LIBCPP_VERSION) || _LIBCPP_VERSION >= 10000) && \
+     (!FMT_MSC_VERSION ||                                       \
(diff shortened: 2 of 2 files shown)


include/fmt/core.h
```diff
@@ -105,9 +105,11 @@
...
 #endif
 
-#if ((FMT_CPLUSPLUS >= 202002L) &&                            \
-     (!defined(_GLIBCXX_RELEASE) || _GLIBCXX_RELEASE > 9)) || \
-    (FMT_CPLUSPLUS >= 201709L && FMT_GCC_VERSION >= 1002)
+#if (FMT_CPLUSPLUS >= 202002L ||                                \
+     (FMT_CPLUSPLUS >= 201709L && FMT_GCC_VERSION >= 1002)) &&  \
+    ((!defined(_GLIBCXX_RELEASE) || _GLIBCXX_RELEASE >= 10) &&  \
+     (!defined(_LIBCPP_VERSION) || _LIBCPP_VERSION >= 10000) && \
+     (!FMT_MSC_VERSION || FMT_MSC_VERSION >= 1928))
```
vitaut, line 112: What is the idea behind this change? Shall we enable `FMT_CONSTEXPR20` only if `is_constant_evaluated` is available instead? Something like

```c++
#if defined(FMT_HAS_IS_CONSTANT_EVALUATED) &&                 \
    ((FMT_CPLUSPLUS >= 202002L) &&                            \
     (!defined(_GLIBCXX_RELEASE) || _GLIBCXX_RELEASE > 9)) || \
    (FMT_CPLUSPLUS >= 201709L && FMT_GCC_VERSION >= 1002)
#  define FMT_CONSTEXPR20 constexpr
```

phprus: Main idea: Enable C++20 constexpr only if stdlib support C++20.
The condition ``(!defined(_GLIBCXX_RELEASE) || _GLIBCXX_RELEASE > 9)`` partially check stdlib C++20 features. I extended it for libcxx and MSVC.

``FMT_CONSTEXPR20`` enabled if compiler support C++20 AND the version of stdlib is modern to support C++20.

hchataing: The compiler does not define _LIBCPP_VERSION in Android's case. 
This will result in FMT_CONSTEXPR20 being incorrectly set to constexpr.

phprus: @hchataing

Try to find the ``__config`` file in the source code of your standard library and look at the value of the ``_LIBCPP_VERSION`` macro.

Or in repository, like this:
https://android.googlesource.com/platform/external/libcxx/+/refs/tags/android-cts-10.0_r15/include/__config#36

hchataing: :+1:  I see `#  define _LIBCPP_VERSION 170000`

phprus: @hchataing
Interesting!
libcxx version 170000 fully supports C++20 and you shouldn't have issue #3745.

Can you compile and run the program:
```c++
#include <iostream>

int main()
{
    std::cerr << _LIBCPP_VERSION << std::endl;
    return 0;
}
```

in issue #3745 build environment?

vitaut: I think we should just go with enabling C++20 constexpr only if `consteval` is also available.

phprus: Done.

vitaut: Sorry, I meant `is_constant_evaluated` because that's what causing the OP's problem.

phprus: @vitaut 
Done.

vitaut:
Thanks!

hchataing:
Sorry I did not have the time to validate the fix :disappointed:  It looks like it should be working.
Thanks for the help !

PR #3755, opened 2023-12-13
Apply clang-format

tchaikov:
* use clang-format v17.0.6 to reformat the tree
* tweak some places so that clang-format does not mutate the layout of code too much


vitaut:
Merged (formatting changes only) in https://github.com/fmtlib/fmt/commit/6855bd532b58019d4c43a3bb6fd38653abf31761, thanks.

PR #3756, opened 2023-12-15
Output Ranges for Functionality & Additional Safety

ThePhD:
The following snippet will overrun a too-small buffer:

```cpp
char my_tiny_char_buffer[4];
fmt::format_to(my_tiny_char_buffer, "Beeg number: {}", 58629475448);
```

This pull request changes that by turning all places that took `OutputIt` with `format_to` to use `Output&&` instead and use the concept of `output_ranges` OR `output_iterator`, where appropriate.

This PR is not intended to make it into mainline `{fmt}`, just to prove it is possible to do so and pass all the tests to show it doesn't break the existing interfaces.


vitaut (changes requested):
Thanks for the PR!

Improving safety for the array of known bound case sounds reasonable but unfortunately there are too many problems with the proposed implementation (and to some extent design) so it will require major changes. The most obvious one is that it is a fairly big breaking change. Fortunately there are ways to redesign this to make it less breaking.

You mentioned that you don't intend this to be merged so I'm a bit confused with what the plan here is. But if you change your mind, the first thing to do would be to split this into three PRs:

1. Handling the array of known bound in `fmt/core.h` without introducing a dependency on ranges and making any changes below the buffer layer.
2. Handling the array of known bound in format string compilation. This needs complete redesign but we could start with the context hack and iterate.
3. Adding the new (output) ranges API in `fmt/ranges.h`. You'll need to provide motivation for the changes there.

ThePhD:
Ah, making the changes to the format string compilation might be fully beyond my understanding at the moment. But, otherwise, I (think?) I can successfully split the changes into different overloads over multiple files again with the only range stuff in the top-level. I'll PR the known-bound for `fmt/core.h` seperately too from this one, I guess.

> You mentioned that you don't intend this to be merged so I'm a bit confused with what the plan here is. 

More of a proof of concept that this can be done, rather than asserting it should be. There's design motivation as to why this should be done but that's an entire C++ proposal, which I think is actually bigger and more involved than I could ever fit in this PR. Even as I chunk this into pieces to make it more acceptable, I'd like to keep this PR here still as a proof that we passed the existing tests and didn't break any of the tested use cases. (At least, not anything picked up by the tests or compilation of said tests, anyways.)

vitaut:
> Ah, making the changes to the format string compilation might be fully beyond my understanding at the moment.

You actually did it already by changing `basic_format_context`. Those changes are not needed for anything other than format string compilation and need most revision. This is why I'm suggesting the split.

> There's design motivation as to why this should be done but that's an entire C++ proposal, which I think is actually bigger and more involved than I could ever fit in this PR.

Do you have a paper number for this proposal?

vitaut:
Closing since this PR is not intended for merging.

PR #3757, opened 2023-12-17
Enable consteval in MSVC VS2019 version 16.10

phprus:
``consteval`` have been implemented in MSVC since VS2019 16.10.


vitaut:
Thanks

Issue #3777, opened 2023-12-30
fmt.cc includes os.cc regardless of the FMT_OS option.


gabime:
I think it should avoid it if FMT_OS is set to OFF.

vitaut:
Makes sense in general but is there any concrete problem you are trying to solve?

gabime:
In spdlog ver 2.x fmt is [fetched and compiled](https://github.com/gabime/spdlog/blob/2a4aa8f2d10ea362f2be324aef8e896ed0d91b83/cmake/fmtlib.cmake#L14) as a dependency. Since spdlog doesn’t require the os features of fmt it sets FMT_OS to OFF.

vitaut:
Should be fixed in https://github.com/fmtlib/fmt/commit/398ddb8fec18eb4dfd1d0c8b1b8222afdb8b6307.

gabime:
Thanks. However it still includes “os.cc” unconditionally [here](https://github.com/fmtlib/fmt/blob/398ddb8fec18eb4dfd1d0c8b1b8222afdb8b6307/src/fmt.cc#L110C2-L110C2).

vitaut:
Take 2: https://github.com/fmtlib/fmt/commit/9659f22d36acd801fc89f6056ccd846d967ef473.

vitaut:
Take 3: #3783. Thanks to @phprus.

PR #3783, opened 2024-01-02
Fix FMT_OS definition

phprus:
If cmake option ``FMT_OS`` is ``OFF`` then define ``FMT_OS=0``.
If cmake option ``FMT_OS`` is ``ON`` then ``FMT_OS`` is not defined by cmake and defined in ``fmt.cc``:

https://github.com/fmtlib/fmt/blob/6c3b2d491e23a3fb8eb2bc998acfecb53b19171f/src/fmt.cc#L86-L88

issue: #3777


vitaut:
Good catch, thanks!

PR #3786, opened 2024-01-03
Restore symbol from 10.1

### LINKED ISSUE #3785: Breaking ABI change in Windows builds of 10.2.0
On some Windows builds, the signature of `fmt::v10::detail::write_console` has changed:

```patch
- bool __cdecl fmt::v10::detail::write_console(struct _iobuf * __ptr64,class fmt::v10::basic_string_view<char>)
+ bool __cdecl fmt::v10::detail::write_console(int,class fmt::v10::basic_string_view<char>)
```

This causes dynamic linking errors. The problem seems to be caused by this change https://github.com/fmtlib/fmt/commit/2a2c6e676f4c2100b2b0f6b3390543ba94ca4b3d

See e.g., https://github.com/mamba-org/mamba/issues/3095 for where this breaks a library on Windows.

saraedum:
Fixes #3785.

the write_console symbol has changed from 10.1 to 10.2. Here we restore the symbol as introduced in 10.1.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -1434,6 +1434,7 @@ FMT_FUNC auto vformat(string_view fmt, format_args args) -> std::string {
 namespace detail {
 #if !defined(_WIN32) || defined(FMT_WINDOWS_NO_WCHAR)
 FMT_FUNC auto write_console(int, string_view) -> bool { return false; }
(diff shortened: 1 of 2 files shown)


vitaut:
Merged into the 10.x branch: https://github.com/fmtlib/fmt/tree/10.x. If there are no more critical issues will publish the release shortly.

vitaut:
Released: https://github.com/fmtlib/fmt/releases/tag/10.2.1.

PR #3789, opened 2024-01-04
Update documentation

### LINKED ISSUE #3788: Unexpected behaviour of a format string with alignment and zero
Hi, I just tried to update fmtlib in a project and one test failed because the output of `fmt::format` changed for format strings like `{:>06x}`.

As soon as I add an alignment specifier, the `0` is ignored and the output filled with whitespace instead.
See the comparison between some versions of fmt here: https://godbolt.org/z/KfGjz9TG4

As I understand it, the combination of `<` and `0` doesn't make much sense, so it's ignored. For symmetry, it's also ignored when any other alignment specifier is present.

I'd say this is desired behaviour, but I miss documentation about it. Probably a note could be added to the paragraph about the '0' specifier?

> Preceding the width field by a zero ('0') character enables sign-aware zero-padding for numeric types. It forces the padding to be placed after the sign or base (if any) but before the digits. This is used for printing fields in the form ‘+000000120’. This option is only valid for numeric types and it has no effect on formatting of infinity and NaN. **This option is ignored when any alignment specifier is present**.

WolleTD:
Add a note about the numerical zero being ignored when an alignment flag is present in the format specifier.

This change was introduced in a585571e90e7f5645bcf6548cef42078db940f49 to be compatible with std::format, but misses documentation.

Closes #3788

### DIFF
--- a/doc/syntax.rst
+++ b/doc/syntax.rst
@@ -153,7 +153,8 @@ Preceding the *width* field by a zero (``'0'``) character enables sign-aware
 zero-padding for numeric types. It forces the padding to be placed after the
 sign or base (if any) but before the digits. This is used for printing fields in
 the form '+000000120'. This option is only valid for numeric types and it has no
-effect on formatting of infinity and NaN.
+effect on formatting of infinity and NaN. This option is ignored when any
+alignment specifier is present.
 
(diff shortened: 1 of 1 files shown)


vitaut:
Thank you!

PR #3793, opened 2024-01-09
Update fmt/std.h section

tupaschoal:
This is just trying to keep the `std.h` documentation up-to-date. I just opened https://github.com/tupaschoal/fmt/blob/7ba6420540383f378c48834beb46f9cec5504dd2/include/fmt/std.h and looked through all formatter specializations:

`std::bitset` https://github.com/tupaschoal/fmt/blob/master/include/fmt/std.h#L172
`std::source_location` https://github.com/tupaschoal/fmt/blob/master/include/fmt/std.h#L248
`std::error_code` https://github.com/tupaschoal/fmt/blob/master/include/fmt/std.h#L365
`std::atomic` https://github.com/tupaschoal/fmt/blob/master/include/fmt/std.h#L514
`std::atomic_flag` https://github.com/tupaschoal/fmt/blob/master/include/fmt/std.h#L527

PS: I added an optional commit to sort the section, because as it grows I think it gets easier to read

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -520,16 +520,16 @@ Standard Library Types Formatting
 
 ``fmt/std.h`` provides formatters for:
 
-* `std::filesystem::path <https://en.cppreference.com/w/cpp/filesystem/path>`_
-* `std::thread::id <https://en.cppreference.com/w/cpp/thread/thread/id>`_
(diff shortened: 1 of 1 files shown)


vitaut:
Thank you!

Issue #3795, opened 2024-01-10
Incorrect output and segfault with `std::optional<std::string>` and `FMT_COMPILE`


TomCrypto:
We've run into an issue that we've tracked down to formatting `std::optional<std::string>` with {fmt} in conjunction with the `FMT_COMPILE` macro. We have observed both incorrect formatted strings and also segmentation faults. I've created a minimal repro case on godbolt with clang trunk and {fmt} trunk:

https://godbolt.org/z/rE5xadqf1

The issue does not seem to occur if `FMT_COMPILE` is not used or if the literal passed to it is just `"{}"`, but as soon as the literal contains more than just that, issues crop up when formatting string optionals. I haven't checked if it happens with other std types. It doesn't appear to occur when formatting plain strings, or `std::optional<int>`, etc...

The documentation suggests this should be a valid use of `FMT_COMPILE` so I'm submitting this as a bug.

phprus:
Simple test case:
https://godbolt.org/z/qe5zK9Phx

phprus:
@vitaut 

This error has been present since the commit: https://github.com/fmtlib/fmt/commit/a34a97cc1d7d50ed6b05bd315ecef7aeb97f1364

New test:
```c++
TEST(compile_test, format_escape) {
  EXPECT_EQ(fmt::format(FMT_COMPILE("{:?}"), "string"), "\"string\"");
  EXPECT_EQ(fmt::format(FMT_COMPILE("0{:?}"), "string"), "0\"string\"");
  EXPECT_EQ(fmt::format(FMT_COMPILE("{:?}0"), "string"), "\"string\"0");
}
```

PR #3797, opened 2024-01-10
Fix buffer overflow if output iterator is std::back_insert_iterator and value is escaped (debug format)

phprus:
Issue: #3795

For ``std::back_insert_iterator`` output iterator ``reserve`` return raw pointer:

https://github.com/fmtlib/fmt/blob/58a6bd48a8dc7ea74bb9fd900c60fd333a85725f/include/fmt/format.h#L571-L582

If ``size`` is incorrect, ``write_padded`` may overflow this container:
https://github.com/fmtlib/fmt/blob/58a6bd48a8dc7ea74bb9fd900c60fd333a85725f/include/fmt/format.h#L1798-L1818

@vitaut

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -2388,9 +2388,12 @@ FMT_CONSTEXPR auto write(OutputIt out, basic_string_view<Char> s,
     size = code_point_index(s, to_unsigned(specs.precision));
   bool is_debug = specs.type == presentation_type::debug;
   size_t width = 0;
+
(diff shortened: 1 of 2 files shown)


vitaut:
> Format error in file include/fmt/base.h

Should be fixed now.

> I suggest enabling lint for all commits, not just PRs.

Yeah, we should do that.

phprus:
@vitaut 
Rebased.

vitaut:
Thank you!

PR #3798, opened 2024-01-10
Add fmt::is_contiguous<std::basic_string<Char, Traits, Allocator>>

phprus:
Add ``fmt::is_contiguous<std::basic_string<Char, Traits, Allocator>>``


vitaut:
thanks

PR #3800, opened 2024-01-10
Fix formatting of ranges with lvalue-qualified, non-const begin() and end()

tcbrindle:
C++20 allows ranges to have lvalue-qualified begin() and end() member functions. fmt correctly handles this if begin() and end() are additionally const-qualifed (i.e. begin() const&), but not in the non-const case. For example:

https://godbolt.org/z/YfxaYz5r7

This patch fixes fmt's range detection to handle this case by testing calls to detail::ranges_begin()/end() with an lvalue T&, matching the behaviour in the const case.

### DIFF
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -132,8 +132,8 @@ struct has_const_begin_end<
 
 template <typename T>
 struct has_mutable_begin_end<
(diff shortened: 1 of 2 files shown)


vitaut:
Thank you!

PR #3804, opened 2024-01-12
Use std::allocator_traits

phprus:
Discussion: https://github.com/fmtlib/fmt/commit/872635daf111ad26d447ffbead0fc405083c978d#r136913458

And fix MSVC warning.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -43,9 +43,13 @@
 #include <cstring>           // std::memcpy
 #include <initializer_list>  // std::initializer_list
 #include <limits>            // std::numeric_limits
-#include <stdexcept>         // std::runtime_error
-#include <string>            // std::string
-#include <system_error>      // std::system_error
+#if defined(__GLIBCXX__) && !defined(_GLIBCXX_USE_DUAL_ABI)
+// Workaround for pre gcc 5 libstdc++.
+#  include <memory>  // std::allocator_traits
+#endif
+#include <stdexcept>     // std::runtime_error
+#include <string>        // std::string
+#include <system_error>  // std::system_error
 
 #include "base.h"
 
@@ -508,14 +512,6 @@ FMT_INLINE void assume(bool condition) {
 #endif
 }
 
-template <typename Allocator, typename Enable = void> struct allocator_size {
-  using type = size_t;
-};
-template <typename Allocator>
-struct allocator_size<Allocator, void_t<typename Allocator::size_type>> {
-  using type = typename Allocator::size_type;
-};
-
 template <typename Char, typename InputIt>
 auto copy_str(InputIt begin, InputIt end, appender out) -> appender {
   get_container(out).append(begin, end);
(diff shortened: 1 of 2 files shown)


vitaut (changes requested):
Thanks for the PR!

I was hoping that SFINAE would be less cumbersome =). Considering how niche this feature is I think we should go back to your suggestion of using `std::allocator_traits` pulled in transitively and fallback on ignoring `max_size` on pre-C++11 systems.

phprus:
@vitaut 
We can include ``<memory>`` only for ``gcc < 5`` and remove it later, after ``gcc < 5`` is no longer supported.
What do your think about it? It will be the smallest change.

vitaut:
> We can include <memory> only for gcc < 5 and remove it later, after gcc < 5 is no longer supported.
> What do your think about it? It will be the smallest change.

Great idea, let's do it.

include/fmt/format.h
```diff
@@ -43,9 +43,13 @@
 #include <cstring>           // std::memcpy
 #include <initializer_list>  // std::initializer_list
 #include <limits>            // std::numeric_limits
-#include <stdexcept>         // std::runtime_error
-#include <string>            // std::string
-#include <system_error>      // std::system_error
+#if defined(__GLIBCXX__) && !defined(_GLIBCXX_USE_DUAL_ABI)
```
vitaut, line 46: Maybe `__GLIBCXX__ < 20150422` instead of _GLIBCXX_USE_DUAL_ABI? https://stackoverflow.com/questions/37118114/value-of-glibcxx-for-each-libstdc-release

include/fmt/format.h
```diff
@@ -914,17 +910,17 @@ class basic_memory_buffer : public detail::buffer<T> {
...
     size_t old_capacity = buf.capacity();
     size_t new_capacity = old_capacity + old_capacity / 2;
     if (size > new_capacity)
       new_capacity = size;
     else if (new_capacity > max_size)
       new_capacity = size > max_size ? size : max_size;
     T* old_data = buf.data();
-    T* new_data = self.alloc_.allocate(new_capacity);
+    T* new_data =
+        std::allocator_traits<Allocator>::allocate(self.alloc_, new_capacity);
```
vitaut, line 923: I don't think we need to use traits here and in deallocate below. It's mainly useful for max_size and if we used a hint (which we don't).

include/fmt/format.h
```diff
@@ -49,7 +49,8 @@
 
 #include "base.h"
 
-#if FMT_HAS_INCLUDE(<bit>)
+// Checking FMT_CPLUSPLUS for warning suppression in MSVC.
+#if FMT_HAS_INCLUDE(<bit>) && FMT_CPLUSPLUS >= 201703L
```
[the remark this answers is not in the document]
phprus: PR #3807

include/fmt/format.h
phprus: Ok.

include/fmt/format.h
phprus: Comparing dates won't work:
https://github.com/fmtlib/fmt/pull/2349#discussion_r651044172

vitaut (approved):

vitaut:
Thank you!

PR #3805, opened 2024-01-13
🛠 Add basic array safety functions and backwards-compatible result type

ThePhD:
This pull request follows-up on #3756 by doing one part of what was asked; just array overloads to cover that case. A separate PR will handle the enhanced functionality for range-based calls to `format_to` (and a potentially new API call) in `fmt/range.h` (and not in core).

Unlike the general-purpose sketch in #3756, this one is meant to be merged.

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -2876,7 +2876,7 @@ FMT_INLINE auto format_to_n(OutputIt out, size_t n, format_string<T...> fmt,
 
 template <typename OutputIt, typename OutputSen = OutputIt>
 struct format_to_result {
-  /** Iterator past the end of the last write. */
+  /** Iterator pointing to just after the last succesful write in the range. */
   OutputIt out;
   /** Sentinel indicating the end of the output range. */
   OutputSen out_last;
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -2811,8 +2811,6 @@ using format_string = basic_format_string<char, type_identity_t<Args>...>;
 inline auto runtime(string_view s) -> runtime_format_string<> { return {{s}}; }
 #endif
 
-template <typename... Args>
-using format_string = basic_format_string<char, type_identity_t<Args>...>;
 /** Formats a string and writes the output to ``out``. */
 template <typename OutputIt,
           FMT_ENABLE_IF(detail::is_output_iterator<remove_cvref_t<OutputIt>,
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -259,6 +259,8 @@
 #  define FMT_UNICODE !FMT_MSC_VERSION
 #endif
 
+#define FMT_FWD(...) static_cast<decltype(__VA_ARGS__)&&>(__VA_ARGS__)
+
 // Enable minimal optimizations for more compact code in debug mode.
 FMT_GCC_PRAGMA("GCC push_options")
 #if !defined(__OPTIMIZE__) && !defined(__CUDACC__)
@@ -2809,13 +2811,41 @@ using format_string = basic_format_string<char, type_identity_t<Args>...>;
 inline auto runtime(string_view s) -> runtime_format_string<> { return {{s}}; }
 #endif
 
+template <typename... Args>
+using format_string = basic_format_string<char, type_identity_t<Args>...>;
 /** Formats a string and writes the output to ``out``. */
 template <typename OutputIt,
-          FMT_ENABLE_IF(detail::is_output_iterator<OutputIt, char>::value)>
-auto vformat_to(OutputIt out, string_view fmt, format_args args) -> OutputIt {
+          FMT_ENABLE_IF(detail::is_output_iterator<remove_cvref_t<OutputIt>,
+                                                   char>::value)>
+auto vformat_to(OutputIt&& out, string_view fmt, format_args args)
+    -> remove_cvref_t<OutputIt> {
   auto&& buf = detail::get_buffer<char>(out);
   detail::vformat_to(buf, fmt, args, {});
   return detail::get_iterator(buf, out);
+}
+
+template <typename OutputIt, typename OutputSen = OutputIt>
+struct format_to_result {
+  /** Iterator past the end of the last write. */
+  OutputIt out;
+  /** Iterator to the end of the output range. */
+  OutputSen out_last;
+
+  FMT_CONSTEXPR operator OutputIt&() & noexcept { return out; }
+  FMT_CONSTEXPR operator const OutputIt&() const& noexcept { return out; }
+  FMT_CONSTEXPR operator OutputIt&&() && noexcept {
+    return static_cast<OutputIt&&>(out);
+  }
+};
+
+/** Formats a string and writes the output to ``out``. */
+template <size_t Size>
+auto vformat_to(char (&out)[Size], string_view fmt, format_args args)
+    -> format_to_result<char*> {
+  using traits = detail::fixed_buffer_traits;
+  auto buf = detail::iterator_buffer<char*, char, traits>(out, Size);
+  detail::vformat_to(buf, fmt, args, {});
+  return {out + buf.count(), out + Size};
 }
 
 /**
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR! Looks like there are some build failures.

include/fmt/base.h
```diff
@@ -2741,6 +2741,28 @@ template <typename Char, typename... Args> class basic_format_string {
...
+  OutputIt out;
+  /** Total (not truncated) output size. */
+  OutputSen out_last;
+
+  /** Ensures old code relying on array-style output continues to work. */
+  FMT_CONSTEXPR operator OutputIt&() & noexcept { return out; }
+  /** Ensures old code relying on array-style output continues to work. */
+  FMT_CONSTEXPR operator OutputIt&&() && noexcept { return out; }
+  /** Ensures old code relying on array-style output continues to work. */
+  FMT_CONSTEXPR operator const OutputIt&() const& noexcept { return out; }
```
vitaut, line 2763: Clever =)

ThePhD:
I think (?) the build failures should go away soon. Only thing I haven't done here is call std::forward for the new forwarding references. I don't really want to `#include <utility>` after all the work you went through to reduce compilation times in `base.h`. If I can remember how to write a macro-only `FWD(x)`, I'll try to do that and then apply it to all the places where I had to take `OutputIt&&`.

Look okay otherwise?

vitaut (changes requested):
Overall looks good but let's limit the changes to the main API in `fmt/base.h`. Once we get usage experience and happy with the new API we could apply the change to other places.

> Only thing I haven't done here is call std::forward for the new forwarding references. I don't really want to #include <utility> after all the work you went through to reduce compilation times in base.h. If I can remember how to write a macro-only FWD(x), I'll try to do that and then apply it to all the places where I had to take OutputIt&&.

The macro is described here: https://www.foonathan.net/2020/09/move-forward/. {fmt} had something like that at some point but it was removed as we got rid of most forwarding.

phprus:
I think it's better to add a ``Char (&out)[Size]`` overload for the ``format_to_n`` functions.
This will not require changing the return type for the ``format_to`` function.

ThePhD:
> I think it's better to add a ``Char (&out)[Size]`` overload for the ``format_to_n`` functions.
> This will not require changing the return type for the ``format_to`` function.

If this is a deal breaker, I would rather remove the return type than change format_to's behavior back to the old one.

It does not make sens e to change format_to_n as well, unless you want additional safety by checking the size passed in versus the size of the array and take whichever is smaller. (I'm sure this is also better for safety as well, too, and I can do that in a different PR.)

vitaut:
I agree with @ThePhD that we need to change `format_to` because otherwise we won't get improved safety.

phprus:
@vitaut
I suggested an overload:
```c++
template <typename Char, size_t n, typename... T>
auto format_to_n(Char (&out)[n], format_string<T...> fmt, T&&... args)
```

without duplicating ``n``.
So that the function name explicitly indicates that there is a limit on the length of the result.

ThePhD:
I changed it to just base.h. I'll try to workshop something for `format_to_n` later. The linux build started breaking but it seems unrelated to the changes I made and instead what's currently being worked on in `master`?

ThePhD:
Oh, nevermind. I made a mistake; should be fixed with the latest push now.

ThePhD:
Yes, any changes to `format_to_n` will be made separately.

include/fmt/base.h
```diff
@@ -2811,13 +2813,39 @@ inline auto runtime(string_view s) -> runtime_format_string<> { return {{s}}; }
...
+auto vformat_to(OutputIt&& out, string_view fmt, format_args args)
+    -> remove_cvref_t<OutputIt> {
   auto&& buf = detail::get_buffer<char>(out);
   detail::vformat_to(buf, fmt, args, {});
   return detail::get_iterator(buf, out);
 }
 
+template <typename OutputIt, typename OutputSen = OutputIt>
+struct format_to_result {
+  /** Iterator past the end of the last write. */
```
vitaut, line 2827: It is not super clear what "last write" means. In `format_to_n_result` we have

```c++
/** Iterator past the end of the output range. */
```

but I guess it can now be ambiguous since output range can denote either the range of the output produced by a function or a passed output "range". Any ideas how to rephrase it better?

include/fmt/base.h
```diff
@@ -2811,13 +2813,39 @@ inline auto runtime(string_view s) -> runtime_format_string<> { return {{s}}; }
...
+  OutputIt out;
+  /** Iterator to the end of the output range. */
```
vitaut, line 2829: It seems to be a sentinel so let's rephrase to something like:

```c++
/** Sentinel indicating the end of the output range. */
```

include/fmt/base.h
```diff
@@ -2811,13 +2813,39 @@ inline auto runtime(string_view s) -> runtime_format_string<> { return {{s}}; }
...
+  OutputSen out_last;
+
+  FMT_CONSTEXPR operator OutputIt&() & noexcept { return out; }
+  FMT_CONSTEXPR operator const OutputIt&() const& noexcept { return out; }
+  FMT_CONSTEXPR operator OutputIt&&() && noexcept {
+    return static_cast<OutputIt&&>(out);
+  }
```
vitaut, line 2836: Do we need all three overloads and why?

include/fmt/base.h
```diff
@@ -2811,13 +2813,39 @@ inline auto runtime(string_view s) -> runtime_format_string<> { return {{s}}; }
...
+};
+
+/** Formats a string and writes the output to ``out``. */
+template <size_t Size>
+auto vformat_to(char (&out)[Size], string_view fmt, format_args args)
+    -> format_to_result<char*> {
+  using traits = detail::fixed_buffer_traits;
+  auto buf = detail::iterator_buffer<char*, char, traits>(out, Size);
+  detail::vformat_to(buf, fmt, args, {});
+  return {out + buf.count(), out + Size};
```
vitaut, line 2846: Let's reuse `vformat_to_n` here to avoid duplication and make the semantics more obvious.

include/fmt/base.h
```diff
@@ -2831,9 +2859,28 @@ auto vformat_to(OutputIt out, string_view fmt, format_args args) -> OutputIt {
...
+ Formats ``args`` according to specifications in ``fmt``, writes the result to
+ the output iterator ``out`` and returns the iterator past the end of the output
+ range. `format_to` does not append a terminating null character.
+
+ **Example**::
+
+   auto out = std::vector<char>();
+   fmt::format_to(std::back_inserter(out), "{}", 42);
+ \endrst
+ */
```
vitaut, line 2880: The copy-pasted comment is incorrect in this context. Let's remove it, since it is just a special case of a more general overload above.

vitaut (changes requested):
A few more comments and please add a test case to base-test.

include/fmt/base.h
```diff
@@ -2811,13 +2813,39 @@ inline auto runtime(string_view s) -> runtime_format_string<> { return {{s}}; }
 
 /** Formats a string and writes the output to ``out``. */
 template <typename OutputIt,
-          FMT_ENABLE_IF(detail::is_output_iterator<OutputIt, char>::value)>
-auto vformat_to(OutputIt out, string_view fmt, format_args args) -> OutputIt {
+          FMT_ENABLE_IF(detail::is_output_iterator<remove_cvref_t<OutputIt>,
+                                                   char>::value)>
+auto vformat_to(OutputIt&& out, string_view fmt, format_args args)
```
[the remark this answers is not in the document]
ThePhD: Yes, otherwise multiple compilers in multiple standard modes complain about ambiguity.

include/fmt/base.h
ThePhD: We could make a `const`-only version of this function that just unconditionally returns a copy. It'd be cheap for all the cases we care about, but if later functionality is added to do e.g. real `output_range` that has a move-only iterator, then it'll start being a compilation break to use this type.

There is also the case that for real `output_range` types, you'd want to return a `std::ranges::subrange` or a sufficient mockup of one, so it might be worth it to invest in making a real `fmt::detail::subrange` for the return value.

include/fmt/base.h
ThePhD: Okay!

include/fmt/base.h
ThePhD: Removed.

include/fmt/base.h
ThePhD: Done!

include/fmt/base.h
ThePhD: I tried a different formulation to make it a bit more clear: 

```cpp
/** Iterator pointing to just after the last succesful write in the range. */
```

vitaut (approved):

vitaut:
Merged, thanks!

PR #3806, opened 2024-01-13
Fix conversion warning in filesystem::path formatter

hmbj:
Use template character type to eliminate MSVC conversion warning when formatting a wide chararacter path:

fmt\std.h(140,49): warning C4244: '=': conversion from 'const Char' to 'char', possible loss of data

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -120,7 +120,7 @@ template <typename Char> struct formatter<std::filesystem::path, Char> {
   format_specs<Char> specs_;
   detail::arg_ref<Char> width_ref_;
   bool debug_ = false;
-  Char path_type_ = 0;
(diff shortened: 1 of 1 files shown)


include/fmt/std.h
```diff
@@ -120,7 +120,7 @@ template <typename Char> struct formatter<std::filesystem::path, Char> {
   format_specs<Char> specs_;
   detail::arg_ref<Char> width_ref_;
   bool debug_ = false;
-  char path_type_ = 0;
+  Char path_type_ = 0;
```
vitaut, line 123: Let's keep `char` here and use `to_ascii` when assigning to it in `parse`.

vitaut (approved):

vitaut:
Thank you!

PR #3807, opened 2024-01-13
Fix MSVC warning: "The contents of <bit> are available only with C++20 or later."

phprus:
Separate PR for MSVC warning.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -49,7 +49,8 @@
 
 #include "base.h"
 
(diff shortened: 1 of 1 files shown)


vitaut:
What is the warning?

phprus:
Warning:
```
The contents of <bit> are available only with C++20 or later.
```

See:
https://github.com/fmtlib/fmt/actions/runs/7513361958/job/20455024050

vitaut:
Thanks!

PR #3808, opened 2024-01-15
Mark `iterator_buffer` move constructors as `noexcept`.

waywardmonkeys:
I noticed this when running clang-tidy on my code.


vitaut:
Thank you!

Issue #3809, opened 2024-01-15
`nested_formatter` compilation errors


sergiud:
Using GCC 6.4.0 the API reference example
```cpp
#include <fmt/format.h>

struct point {
  double x, y;
};

template <>
struct fmt::formatter<point> : nested_formatter<double> {
  auto format(point p, format_context& ctx) const {
    return write_padded(ctx, [=](auto out) {
      return format_to(out, "({}, {})", nested(p.x), nested(p.y));
    });
  }
};

int main() {
  fmt::print("[{:>20.2f}]", point{1, 2});
}
```
and the corresponding [unit test](https://github.com/fmtlib/fmt/blob/e69e5f977d458f2650bb346dadf2ad30c5320281/test/format-test.cc#L1771-L1789) produce compilation errors such as
```console
error: cannot call member function 'fmt::v10::nested_view<T> fmt::v10::nested_formatter<T>::nested(const T&) const [with T = double]' without object
       return format_to(out, "({}, {})", nested(p.x), nested(p.y));
```

Reproducer: https://godbolt.org/z/6WoeMcs6a

phprus:
Fix: #3810

PR #3810, opened 2024-01-15
Workaround for gcc 6

phprus:
Fix for issue: #3809
Demo: https://godbolt.org/z/v77xfGT4f

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -179,7 +179,8 @@ For example::
   struct fmt::formatter<point> : nested_formatter<double> {
     auto format(point p, format_context& ctx) const {
       return write_padded(ctx, [=](auto out) {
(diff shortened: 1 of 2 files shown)


doc/api.rst
```diff
@@ -179,7 +179,8 @@ For example::
   struct fmt::formatter<point> : nested_formatter<double> {
     auto format(point p, format_context& ctx) const {
       return write_padded(ctx, [=](auto out) {
-        return format_to(out, "({}, {})", nested(p.x), nested(p.y));
+        return format_to(out, "({}, {})", this->nested(p.x),
```
sergiud, line 182: Does it make sense to mention that specifying `this` here is only necessary when using GCC <7?

phprus: I don't think this makes sense.

vitaut:
Thanks for the fix!

Issue #3811, opened 2024-01-16
chrono glibc '-' ext applies to following components within a spec


durka:
In the latest version, @ShawnZhong implemented #2959 via #3271. This allows the use of `-` as a modifier in chrono format specs to omit the leading zero from `H`, `M`, and `S`.

However, a difference from `std::strftime` is that the modification "sticks" to an entire format spec, so `{:-M:%S}` behaves differently than `{:%-M}{:%S}` (the former strips the leading zero from minutes *and* seconds, while the latter only modifies the minutes). Is this intended behavior?

[...]

PR #3812, opened 2024-01-16
Fix scope for glibc ext for sec, min, and hour

phprus:
Issue: #3811


durka:
This fixes my unit test!

vitaut:
Thank you!

durka:
Do you know when there might be a new release pushed that that includes this fix?

vitaut:
> Do you know when there might be a new release pushed that that includes this fix?

Likely end of June.

PR #3814, opened 2024-01-17
Fix %S formatting for chrono durations with leading zeroes 

### LINKED ISSUE #3794: %S ignores precision for chrono durations where the decimal part should have been only zeros
Any chrono duration *n* where 0 < *n* < 10⁻*ᴺ* formatted with `{:.N%S}` receives full precision instead of *N* decimals. This also holds if adding any number of whole seconds. In other words, any duration where the decimal part should have been rounded to `.0`, `.00`, `.000` etc. will be printed with all decimals. For example, with `{:.2%S}`, 4567 milliseconds are printed as `04.56`, while 4007 milliseconds are printed as `04.007` (instead of the expected `04.00`). 4007000 microseconds are printed as `04.007000` while 4017000 microseconds are printed as `04.01`.

If the decimal part is exactly zero (i.e., the duration is a whole number of seconds), there will be one zero more than the specified precision. Thus, 1000 milliseconds will be printed as `01.00` with `{:.1%S}`, `01.000` with `{:.2%S}` etc., whilst 1100 milliseconds will be printed as `01.1`, `01.10` etc. (as expected) for the same format strings.

Code to reproduce:
```c++
#include <fmt/chrono.h>
#include <fmt/core.h>

int main()
{
  constexpr std::chrono::microseconds works{10'001};
  constexpr std::chrono::microseconds fails{9'999};
  fmt::println("{:.2%S} {:.2%S}", works, fails);
}
```

Expected output:
```
00.01 00.00
```

Actual output:
```
00.01 00.009999
```

Godbolt:
https://godbolt.org/z/rncqqGWW3

Tested with recent versions of Clang and GCC. Affects both `fmt::format` and `fmt::println`.

js324:
My attempt to fix #3794. I implemented the rounding as round half to even and fixed the case where an extraneous 0 is added when the decimal part is 0.  Added tests as well.

I do have one question: what would be the desired behavior in the case where rounding up the subseconds/decimal part would eventually carry over to the integer part (the seconds) (for example: `01.99` rounded to precision .1)? As of right now, all numbers with `.9...` will not carry over regardless, so from the previous example, `01.99 -> 01.99`.

### DIFF
--- a/test/chrono-test.cc
+++ b/test/chrono-test.cc
@@ -803,6 +803,10 @@ TEST(chrono_test, cpp20_duration_subsecond_support) {
             "12.1");
   EXPECT_EQ(fmt::format("{:.2%S}", std::chrono::milliseconds{99999}),
             "39.99");
+  EXPECT_EQ(fmt::format("{:.2%S}", std::chrono::milliseconds{1000}),
+            "01.00");
+  EXPECT_EQ(fmt::format("{:.3%S}", std::chrono::milliseconds{1}),
+            "00.001");
   EXPECT_EQ(fmt::format("{:.3%S}", std::chrono::seconds{1234}), "34.000");
   EXPECT_EQ(fmt::format("{:.3%S}", std::chrono::hours{1234}), "00.000");
   EXPECT_EQ(fmt::format("{:.5%S}", dms(1.234)), "00.00123");
--- a/test/chrono-test.cc
+++ b/test/chrono-test.cc
@@ -794,13 +794,11 @@ TEST(chrono_test, cpp20_duration_subsecond_support) {
   EXPECT_EQ(fmt::format("{:.2%S}", std::chrono::milliseconds{12345}),
             "12.34");
   EXPECT_EQ(fmt::format("{:.2%S}", std::chrono::milliseconds{12375}),
-            "12.38");
+            "12.37");
   EXPECT_EQ(fmt::format("{:.2%S}", std::chrono::milliseconds{-12375}),
-            "-12.38");
-  EXPECT_EQ(fmt::format("{:.2%S}", std::chrono::milliseconds{12054}),
-            "12.05");
-  EXPECT_EQ(fmt::format("{:.1%S}", std::chrono::milliseconds{12054}),
-            "12.1");
+            "-12.37");
+  EXPECT_EQ(fmt::format("{:.0%S}", std::chrono::milliseconds{12054}),
+            "12");
   EXPECT_EQ(fmt::format("{:.2%S}", std::chrono::milliseconds{99999}),
             "39.99");
   EXPECT_EQ(fmt::format("{:.2%S}", std::chrono::milliseconds{1000}),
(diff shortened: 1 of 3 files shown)


test/chrono-test.cc
```diff
@@ -791,6 +791,22 @@ TEST(chrono_test, cpp20_duration_subsecond_support) {
...
+  EXPECT_EQ(fmt::format("{:.2%S}", std::chrono::milliseconds{12375}),
+            "12.38");
+  EXPECT_EQ(fmt::format("{:.2%S}", std::chrono::milliseconds{-12375}),
+            "-12.38");
+  EXPECT_EQ(fmt::format("{:.2%S}", std::chrono::milliseconds{12054}),
+            "12.05");
+  EXPECT_EQ(fmt::format("{:.1%S}", std::chrono::milliseconds{12054}),
+            "12.1");
+  EXPECT_EQ(fmt::format("{:.2%S}", std::chrono::milliseconds{99999}),
+            "39.99");
```
vitaut, line 805: If we are rounding than this should probably be 40.00.

vitaut (commented):
Thanks for the PR!

test/chrono-test.cc
vitaut: Thinking more of it, we should check what the standard has to say about rounding in this case and, if necessary, open an issue for clarification.

js324: From what I have read in the standard, there doesn't seem to be anything related to rounding rules for these cases (unless I missed something). There possibly isn't any because allowing a precision modifier with %S for a duration with an integral rep is unique to fmt/not part of the standard?

vitaut: The same applies to the floating-point representation and the standard is unclear what to do about it. libc++ currently ignores precision: https://www.godbolt.org/z/7Gxj5hG8v. I asked for clarification, let's see how it goes.

One problem with rounding is that if we round 59.999 seconds we'll get 00 seconds and will have to increment minutes, and so on. I.e. rounding should apply to the argument, not just `%S`.

So I see two options:

* Truncate
* Round the argument such that all chrono specifiers see the rounded value

In any case we shouldn't round in some cases and truncate in other. I am leaning to the second option since it is consistent with floating-point formatting.

vitaut (changes requested):

test/chrono-test.cc
js324: Sounds good, I have a fix for the rounding ready but I will wait until there is confirmation which way (round/truncate) to go

vitaut: In the meantime I suggest going with truncation (round towards zero) since that's the default for `duration_cast`: https://stackoverflow.com/q/36122428/471164.

vitaut (changes requested):

test/chrono-test.cc
js324: Just wondering, is there a reason why we should follow `duration_cast` specifically? Because in the case of `round` (https://en.cppreference.com/w/cpp/chrono/duration/round), it follows the round half to even method. IMO, I feel like rounding makes more sense so it doesn't lose "precision" of the original number.

vitaut: The idea was that `duration_cast` is the closest we can get to some sort of a default but thinking more of it there is nothing special about it compared to other functions. I don't have strong opinion about whether to round or truncate but truncation seems simpler because you don't need to bump minutes etc. Neither looses precision.

js324: Sure that's fair, I pushed the changes to make the rounding into truncation/towards zero now.

vitaut (approved):

vitaut:
Thank you!

PR #3820, opened 2024-01-19
Fix custom formatter example

### LINKED ISSUE #3818: Documentation: API example for user-defined formatter doesn't compile
The example for defining a custom formatter via specializing the `formatter`struct in https://fmt.dev/latest/api.html#formatting-user-defined-types fails to compile when the definition of the `format` function is in a different compilation unit than its usage.

Using color.h and color.cc as in the API doc and a simple main.cc:
```
#include "color.h"

int main() { fmt::print("{}\n", color::red); }
```
gives the error (gcc version 11.4.0 (Ubuntu 11.4.0-1ubuntu1~22.04) ):
```
 /workspaces/frawen/temp/fmt-10.2.1/include/fmt/core.h:1328:28: error: use of ‘auto fmt::v10::formatter<color>::format(color, fmt::v10::format_context&) const’ before deduction of ‘auto’
 ```

After amending the signature of `format` to
```
auto format(color c, format_context &ctx) const -> format_context::iterator
```
in both color.h and color.cc, the code works as expected.

(No godbolt because I couldn't get its multifile mode to work.)

frank-weinberg:
Add a return type declaration so the example builds when the formatter is used in a different compilation unit than it's implemented.

closes #3818

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -139,14 +139,16 @@ without implementing them yourself. For example::
   template <> struct fmt::formatter<color>: formatter<string_view> {
     // parse is inherited from formatter<string_view>.
 
-    auto format(color c, format_context& ctx) const;
+    auto format(color c, format_context& ctx) const
+      -> format_parse_context::iterator;
   };
 
   // color.cc:
   #include "color.h"
   #include <fmt/format.h>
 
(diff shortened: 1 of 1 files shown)


vitaut:
Thank you!

PR #3822, opened 2024-01-21
doc: fix the chrono %C example value

bbolli:
Assuming that the example date is always "Sat Nov 12 22:04:00 1955", the result of the `%C` format string should be "19", not "55".


vitaut:
Good catch, thanks

PR #3832, opened 2024-01-30
Really fix MSVC warning about <bit> only being available in C++20.

res2k:
The warning states "The contents of <bit> are available only with C++20 or later.", but the check was for C++17 and later.

Follow-up to #3807.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -54,7 +54,7 @@
 #include "base.h"
 
 // Checking FMT_CPLUSPLUS for warning suppression in MSVC.
-#if FMT_HAS_INCLUDE(<bit>) && FMT_CPLUSPLUS >= 202002L
+#if FMT_HAS_INCLUDE(<bit>) && FMT_CPLUSPLUS > 201703L
 #  include <bit>  // std::bit_cast
 #endif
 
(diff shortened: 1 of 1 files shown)


phprus:
Oh! You're right!

I made a mistake and used the ``>=`` check instead of ``>`` in PR #3807.

include/fmt/format.h
```diff
@@ -54,7 +54,7 @@
 #include "base.h"
 
 // Checking FMT_CPLUSPLUS for warning suppression in MSVC.
-#if FMT_HAS_INCLUDE(<bit>) && FMT_CPLUSPLUS >= 201703L
+#if FMT_HAS_INCLUDE(<bit>) && FMT_CPLUSPLUS >= 202002L
```
phprus, line 57: I suggest replace the ``>=`` with ``>`` for compatibility with gcc-9, which contains ``<bit>`` but ``FMT_CPLUSPLUS`` is less than ``202002L``.

vitaut (changes requested):
Thanks for the PR. In general looks good but please change to `>` as suggested by @phprus.

vitaut (approved):

vitaut:
Merged, thanks.

PR #3833, opened 2024-01-31
Fix UB in format_arg_store implementation.

ivafanas:
format_arg_store struct contains pointers to its own members, so autogenerated copy constructor and copy assignment is not correct. Need to reassign pointers to |this| internals.

Compiler won't autogenerate incorrect move constructor and move assignment, because user-provided copy constructor and copy assignment member functions exist.

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -1636,7 +1636,9 @@ template <typename Context, size_t NUM_ARGS, size_t NUM_NAMED_ARGS,
 struct format_arg_store {
   // args_[0].named_args points to named_args to avoid bloating format_args.
   // +1 to workaround a bug in gcc 7.5 that causes duplicated-branches warning.
-  arg_t<Context, NUM_ARGS> args[1 + (NUM_ARGS != 0 ? NUM_ARGS : +1)];
+  static constexpr size_t ARGS_ARR_SIZE = 1 + (NUM_ARGS != 0 ? NUM_ARGS : +1);
+
+  arg_t<Context, NUM_ARGS> args[ARGS_ARR_SIZE];
   named_arg_info<typename Context::char_type> named_args[NUM_NAMED_ARGS];
 
   template <typename... T>
@@ -1649,6 +1651,17 @@ struct format_arg_store {
         0,
         (init_named_arg(named_args, arg_index, named_arg_index, values), 0)...};
   }
+
+  format_arg_store(format_arg_store&& rhs) {
+    args[0] = {named_args, NUM_NAMED_ARGS};
+    for (size_t i = 1; i < ARGS_ARR_SIZE; ++i) args[i] = rhs.args[i];
+    for (size_t i = 0; i < NUM_NAMED_ARGS; ++i)
+      named_args[i] = rhs.named_args[i];
+  }
+
+  format_arg_store(const format_arg_store& rhs) = delete;
+  format_arg_store& operator=(const format_arg_store& rhs) = delete;
+  format_arg_store& operator=(format_arg_store&& rhs) = delete;
 };
 
 // A specialization of format_arg_store without named arguments.


vitaut (changes requested):
Thanks for the PR. `format_arg_store` is not supposed to be copied so we should delete copy and assignment operators instead.

ivafanas:
Ok.
PR is updated.

ivafanas:
Many reports from CI:

https://github.com/fmtlib/fmt/actions/runs/7728340058/job/21068957883?pr=3833

```
/home/runner/work/fmt/fmt/include/fmt/xchar.h:91:51: error: use of deleted function ‘fmt::v10::detail::format_arg_store<Context, NUM_ARGS, NUM_NAMED_ARGS, DESC>::format_arg_store(fmt::v10::detail::format_arg_store<Context, NUM_ARGS, NUM_NAMED_ARGS, DESC>&&) [with Context = fmt::v10::generic_context<fmt::v10::basic_appender<wchar_t>, wchar_t>; long unsigned int NUM_ARGS = 3; long unsigned int NUM_NAMED_ARGS = 3; long long unsigned int DESC = 4611686018427388364]’
   91 |   return make_format_args<wformat_context>(args...);
```

https://github.com/fmtlib/fmt/actions/runs/7728340053/job/21068956966?pr=3833

```
/Users/runner/work/fmt/fmt/include/fmt/xchar.h:91:10: error: call to deleted constructor of 'decltype(make_format_args<fmt::wformat_context>(args, args, args))' (aka 'format_arg_store<fmt::generic_context<fmt::basic_appender<wchar_t>, wchar_t>, sizeof...(T), 3UL, 4611686018427388364ULL>')
  return make_format_args<wformat_context>(args...);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

https://github.com/fmtlib/fmt/actions/runs/7728340056/job/21068959707?pr=3833

```
D:\a\fmt\fmt\include\fmt/xchar.h(91): error C2280: 'fmt::v10::detail::format_arg_store<fmt::v10::wformat_context,3,3,4611686018427388364>::format_arg_store(fmt::v10::detail::format_arg_store<fmt::v10::wformat_context,3,3,4611686018427388364> &&)': attempting to reference a deleted function [D:\a\fmt\build\test\xchar-test.vcxproj]
```

etc.


So, copy constructor must exist. Copy assignment operator might be deleted, I suggest.

vitaut:
> So, copy constructor must exist. Copy assignment operator might be deleted, I suggest.

Yeah, let's go with that.

include/fmt/base.h
```diff
@@ -1636,7 +1636,9 @@ template <typename Context, size_t NUM_ARGS, size_t NUM_NAMED_ARGS,
 struct format_arg_store {
   // args_[0].named_args points to named_args to avoid bloating format_args.
   // +1 to workaround a bug in gcc 7.5 that causes duplicated-branches warning.
-  arg_t<Context, NUM_ARGS> args[1 + (NUM_ARGS != 0 ? NUM_ARGS : +1)];
+  static constexpr int ARGS_ARR_SIZE = 1 + (NUM_ARGS != 0 ? NUM_ARGS : +1);
```
phprus, line 1639: Maybe ``size_t`` (and ``i`` in copy constructor)?
Like ``size_t NUM_ARGS, size_t NUM_NAMED_ARGS`` template arguments.

ivafanas: Seems like loops optimization prevention for unsigned iterator due to allowed overflow is not a case here, because loop bounds are known at compile time.

Yep, `size_t` is a good option here.

include/fmt/base.h
```diff
@@ -1649,6 +1651,23 @@ struct format_arg_store {
         0,
         (init_named_arg(named_args, arg_index, named_arg_index, values), 0)...};
   }
+
+  format_arg_store(const format_arg_store& rhs) {
+    args[0] = {named_args, NUM_NAMED_ARGS};
+    for (size_t i = 1; i < ARGS_ARR_SIZE; ++i) args[i] = rhs.args[i];
+    for (size_t i = 0; i < NUM_NAMED_ARGS; ++i)
+      named_args[i] = rhs.named_args[i];
+  }
```
vitaut, line 1660: I think we only need move ctor.

ivafanas: Updated

vitaut (approved):

vitaut:
Thank you!

include/fmt/base.h
```diff
@@ -1649,6 +1651,17 @@ struct format_arg_store {
         0,
         (init_named_arg(named_args, arg_index, named_arg_index, values), 0)...};
   }
+
+  format_arg_store(format_arg_store&& rhs) {
```
Tradias, line 1655: This move constructor should be noexcept

PR #3834, opened 2024-01-31
added formatter for std::expected

dominicpoeschko:
Simple implementation of a 'std::expected' formatter.
Open for discussions :)

By submitting this pull request, you agree to license your contribution(s)
under the terms outlined in LICENSE.rst and represent that you have the right
to do so.

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -246,12 +246,13 @@ struct formatter<std::optional<T>, Char,
 FMT_END_NAMESPACE
 #endif  // __cpp_lib_optional
 
-#ifdef __cpp_lib_expected
+#if defined(__cpp_lib_expected) || FMT_CPP_LIB_VARIANT
+
 FMT_BEGIN_NAMESPACE
 namespace detail {
 
 template <typename Char, typename OutputIt, typename T>
-auto write_expected_alternative(OutputIt out, const T& v) -> OutputIt {
+auto write_escaped_alternative(OutputIt out, const T& v) -> OutputIt {
   if constexpr (has_to_string_view<T>::value)
     return write_escaped_string<Char>(out, detail::to_string_view(v));
   else if constexpr (std::is_same_v<T, Char>)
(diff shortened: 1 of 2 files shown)


include/fmt/std.h
```diff
@@ -36,6 +36,9 @@
 #  if FMT_HAS_INCLUDE(<optional>)
 #    include <optional>
 #  endif
+#  if FMT_HAS_INCLUDE(<expected>)
+#    include <expected>
+#  endif
 #endif
```
phprus, line 42: Move it outside ``#if FMT_CPLUSPLUS >= 201703L``.

And use C++23 condition to avoid msvc warning:
```c++
#if FMT_HAS_INCLUDE(<expected>) && FMT_CPLUSPLUS > 202002L
#  include <expected>
#endif
```

include/fmt/std.h
```diff
@@ -242,6 +246,51 @@ struct formatter<std::optional<T>, Char,
...
+FMT_BEGIN_NAMESPACE
+namespace detail {
+
+template <typename Char, typename OutputIt, typename T>
+auto write_expected_alternative(OutputIt out, const T& v) -> OutputIt {
+  if constexpr (has_to_string_view<T>::value)
+    return write_escaped_string<Char>(out, detail::to_string_view(v));
+  else if constexpr (std::is_same_v<T, Char>)
+    return write_escaped_char(out, v);
+  else
```
vitaut, line 259: nit: `else`s are redundant because of early returns, let's remove them

include/fmt/std.h
```diff
@@ -242,6 +246,51 @@ struct formatter<std::optional<T>, Char,
...
+    return write<Char>(out, v);
+}
+
+}  // namespace detail
+
+FMT_EXPORT
+template <typename T, typename E, typename Char>
+struct formatter<std::expected<T, E>, Char,
+                 std::enable_if_t<std::conjunction_v<
+                     is_formattable<T, Char>, is_formattable<E, Char>>>> {
```
vitaut, line 269: Is there any reason to use `conjunction_v`? I think `&&` would be clearer.

vitaut (commented):
Thanks for the PR!

include/fmt/std.h
dominicpoeschko: https://github.com/fmtlib/fmt/blob/961d4dd8300ef5af4c71c73129f0c81cb53d4063/include/fmt/std.h#L343-L351

just an exact copy of the impl in variant.
We could also factor that out and use the impl in both.

vitaut: > We could also factor that out and use the impl in both.

Good idea, let's do it.

include/fmt/std.h
dominicpoeschko: https://github.com/fmtlib/fmt/blob/961d4dd8300ef5af4c71c73129f0c81cb53d4063/include/fmt/std.h#L380-L383

oriented me on variant

vitaut:
Merged, thanks!

include/fmt/std.h
```diff
@@ -242,6 +246,56 @@ struct formatter<std::optional<T>, Char,
...
+  }
+
+  template <typename FormatContext>
+  auto format(const std::expected<T, E>& value, FormatContext& ctx) const
+      -> decltype(ctx.out()) {
+    auto out = ctx.out();
+
+    if (value.has_value()) {
+      out = detail::write<Char>(out, "expected(");
+      out = detail::write_escaped_alternative<Char>(out, value.value());
```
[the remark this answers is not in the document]
vitaut: Sure, a PR to change this to use dereference would be welcome.

PR #3841, opened 2024-02-05
Added support of Swift Package Manager

IhorShevchuk:
This allows to add fmt as dependency to other Swift Packages or Xcode projects

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ cmake_install.cmake
 fmt-*.cmake
 fmt.pc
 virtualenv
+.swiftpm/xcode
--- a/Package.swift
+++ b/Package.swift
@@ -1,0 +1,31 @@
+
+// swift-tools-version: 5.8
+
+import PackageDescription
+
+let package = Package(
+    name: "fmt",
+    platforms: [
+        .macOS(.v11),
+        .iOS(.v13)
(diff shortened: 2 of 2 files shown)


vitaut (changes requested):
Thanks for the PR. I think we could add a new package config but please put in the support directory where other contributed build/package configs reside.

vitaut:
@IhorShevchuk, do you plan to update this PR?

IhorShevchuk:
Hi @vitaut, sorry about the delay. Unfortunately Package.swift file must be located in root directory of the repository in order to other packages can see it and use.
I'll add documentation how to use it in another Swift Package or Xcode in `support/spm` folder. Also I guess it is a good idea to add build step in workflow

vitaut:
We shouldn't create an exception for Swift package manager so there are two options:
1. Put `Package.swift` in `support/spm` together with instructions/script how to move it - this is what we do for Bazel.
2. Create a separate repo with `Package.swift` in the top-level directory and fmt in a subdirectory, possibly as a submodule - this is similar to what most package managers do.

IhorShevchuk:
I see, than I would go with second option and create separate repository for fmt with Swift Package Manager. 
Thanks! I'll close this MR.

PR #3852, opened 2024-02-15
Update CMakeLists.txt to support Emscripten

danoli3:
Allow for compile target for WASM Emscripten. 

Thanks FMT is wicked !

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -421,6 +421,24 @@ if (FMT_TEST)
   add_subdirectory(test)
 endif ()
 
+if(EMSCRIPTEN)
+    set(FMT_WASM_SOURCES
+        ${FMT_SOURCES}
+        ${FMT_HEADERS}
+    )
+    add_executable(fmt_wasm
+        ${FMT_WASM_SOURCES}
+    )
+    target_include_directories(fmt_wasm PRIVATE
+        ${CMAKE_CURRENT_SOURCE_DIR}/include
+    )
+    set_target_properties(fmt_wasm PROPERTIES COMPILE_FLAGS "-Os -s SIDE_MODULE=1 ")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -421,6 +421,24 @@ if (FMT_TEST)
   add_subdirectory(test)
 endif ()
 
+if(EMSCRIPTEN)
+    set(FMT_WASM_SOURCES
+        ${FMT_SOURCES}
+        ${FMT_HEADERS}
+    )
+    add_executable(fmt_wasm
```
vitaut, line 429: Why is it an executable and not a library.

CMakeLists.txt
```diff
@@ -421,6 +421,24 @@ if (FMT_TEST)
   add_subdirectory(test)
 endif ()
 
+if(EMSCRIPTEN)
+    set(FMT_WASM_SOURCES
+        ${FMT_SOURCES}
+        ${FMT_HEADERS}
+    )
```
vitaut, line 428: I don't think we need `FMT_WASM_SOURCES`, let's just use `${FMT_SOURCES}` and `${FMT_HEADERS}` in the target below.

CMakeLists.txt
```diff
@@ -421,6 +421,24 @@ if (FMT_TEST)
   add_subdirectory(test)
 endif ()
 
+if(EMSCRIPTEN)
```
vitaut, line 424: nit: we normally put space after `if` so that it looks like a statement

vitaut (changes requested):
Thanks for the PR! Could you add a small section showing how to use this to the docs in https://github.com/fmtlib/fmt/blob/master/doc/usage.rst?

vitaut:
@danoli3, could you add a small usage section per my earlier comment?

sykhro:
`fmt` already works on Emscripten without issues. What's the reason behind this PR?

vitaut:
I think @sykhro is right, at most you'd need to pass extra compile/link flags which can be done via CMake. We don't really need to introduce an extra target.

Issue #3854, opened 2024-02-17
fmt/std.h happy with c++20, unhappy with c++23


prlw1:
If I compile
```
#include <fmt/std.h>

int main()
{
}
```
with g++ 14.0.1, with -std=c++20, all is well. If I use -std=c++23, I see
```
In file included from fmtbug.cc:1:
/usr/src/local/fmt/include/fmt/std.h:274:71: error: template argument 1 is invalid
  274 |     std::enable_if_t<is_formattable<T, Char> && is_formattable<E, Char>> {
      |                                                                       ^
/usr/src/local/fmt/include/fmt/std.h:274:73: error: template argument 3 is invalid
  274 |     std::enable_if_t<is_formattable<T, Char> && is_formattable<E, Char> {
      |                                                                        
*** Error code 1
```
(fmt head 8e42eef495)

vitaut:
https://www.godbolt.org/z/G9jYE47nx

vitaut:
There is missing `::value` after each `is_formattable<...>`. A PR to fix it would be welcome.

PR #3856, opened 2024-02-17
std.h c++23 build fix

prlw1:
Add ::value to is_formattable<...> as per suggestion by @vitaut in https://github.com/fmtlib/fmt/issues/3854


vitaut (changes requested):
Thanks for the PR. Mostly looks good but please apply clang-format.

vitaut (approved):

vitaut:
Thank you for the fix!

Issue #2233, opened 2021-04-13
MSVC warning C4365 with /Wall and FMT_STRING()


vid512:
Any FMT_STRING() usage causes a warning C4365 (un/signed mismatch), when compiling with MSVC (14.28.29910, from Visual Studio 2019) with /Wall:

```
#include <fmt/format.h>
void test() {
	fmt::format(FMT_STRING("a"));
}
```

Note: despite being displayed as warning C4365, this particular warning cannot be enabled with just #pragma warning(default:4365). You must use /Wall to trigger it. It also can't be disabled by /Wall /wd4365.

vid512:
[CMake build log.txt](https://github.com/fmtlib/fmt/files/6305312/CMake.build.log.txt)

vitaut:
Which {fmt} version?

vid512:
version 7.1.3

vitaut:
Fixed in https://github.com/fmtlib/fmt/commit/6271406233585a9178592b0492b190786bf0ee98.

PR #3864, opened 2024-02-23
Fix warning C4365 emitted from printf.h

Baardi:
Compiler: MSVC, Windows x64, Visual Studio 17.9.1
C++ standard: /std::c++latest
Fmtlib version: 10.2.1

Hi. Thanks for a fantastic library.

While we've generally gotten a few warnings from fmtlib during the last few years, they have decreased over time. On the 10.2.1 release, with our current setup, we only got a single one, a C4365. There might be others, that aren't caught by us, since we disable a few warnings. Our warning level is /W4.

Skjermbilde 2024-02-23 234117

I have not yet been able to create a minimal reproduction case. The following code will compile in a freshly created Visual Studio project, but not in ours:

```cpp
#define FMT_HEADER_ONLY
#include "fmt/printf.h"

void test
{
   fmt::printf("%s", "test");
}
```

But the fix for this [issue](https://github.com/fmtlib/fmt/issues/2233), worked for us. If required I can work more on finding a minimal reproduction case.

### DIFF
--- a/include/fmt/printf.h
+++ b/include/fmt/printf.h
@@ -60,7 +60,7 @@ namespace detail {
 // signed and unsigned integers.
 template <bool IsSigned> struct int_checker {
   template <typename T> static auto fits_in_int(T value) -> bool {
-    unsigned max = max_value<int>();
+    unsigned max = to_unsigned(max_value<int>());
     return value <= max;
   }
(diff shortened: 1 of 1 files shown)


Baardi:
Closed as the author was wrong. Creating a new one

PR #3866, opened 2024-02-24
Fix warning C4702 emitted from format.h (MSVC)

Baardi:
Compiler: MSVC, Windows x64, Visual Studio 17.9.1
C++ standard: Any (recreatable with anything from c++14 to c++latest)
Fmtlib version: Master branch
Warning level: /W4 or higher

Hi. This is a new warning, not emitted when using the 10.2.1-release, but recreatable from the master branch.

It can easily be recreated by the following lines of code:

```cpp
#define FMT_HEADER_ONLY
#include <fmt/format.h>

void test
{
   fmt::println("test");
}
```
bilde

It seems that the default-label in the switch-case prevents execution from reaching the bottom of the function. MSVC therefore emits a warning. I haven't tested this on other platforms


vitaut:
thanks

PR #3890, opened 2024-03-12
Fix relative path for cmake in usage doc

Dobiasd:


vitaut:
This was supposed to be a placeholder for additional arguments not a relative path but considering that you are not the first person interpreting it as a path, let's apply the change. Thanks!

Issue #3886, opened 2024-03-08
Add support for std::complex


dcblack:
Since std::literals allows for std::complex literals, it seems natural that fmt would support them.

```c++
#include <complex>
#include "fmt/format.h"

int main() {
  using namespace std::literals;
  auto i1 = 1.0i;
  auto i2 = 2.0if;
  auto i3 = 3.0il;
  auto s = fmt::format( "{} {} {}", i1, i2, i3 ); //< missing formatters for complex numbers
  assert( "1.0i 2.0i 3.0i" == s );
}
```

vitaut:
Thanks for the suggestion. A PR to add a formatter for complex to `fmt/std.h` would be welcome.

PR #3892, opened 2024-03-13
Add std::complex formatter

phprus:
Issue: #3886

Output format is compatible with ``operator<<(std::complex)`` https://en.cppreference.com/w/cpp/numeric/complex/operator_ltltgtgt


vitaut:
LGTM, thank you!

PR #3900, opened 2024-03-19
Improve std::complex formatter to be compatible with P2197R0

phprus:
P2197R0: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p2197r0.html

```
Effects: Equivalent to:

 format_to(ctx.out(), "{:<fill-align-width>}",
           format(c.real() != 0 ? "({0}+{1}i)" : "{1}i", real, imag))
```

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -595,10 +595,15 @@ struct formatter<std::complex<F>, Char> : nested_formatter<F, Char> {
 
     template <typename OutputIt>
     FMT_CONSTEXPR auto operator()(OutputIt out) -> OutputIt {
-      auto format = detail::string_literal<Char, '(', '{', '}', '+', '{', '}',
-                                           'i', ')'>{};
-      return fmt::format_to(out, basic_string_view<Char>(format),
-                            f->nested(c.real()), f->nested(c.imag()));
+      if (c.real() != 0) {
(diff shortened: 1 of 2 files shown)


include/fmt/std.h
```diff
@@ -595,10 +595,16 @@ struct formatter<std::complex<F>, Char> : nested_formatter<F, Char> {
...
-      auto format = detail::string_literal<Char, '(', '{', '}', '+', '{', '}',
-                                           'i', ')'>{};
-      return fmt::format_to(out, basic_string_view<Char>(format),
-                            f->nested(c.real()), f->nested(c.imag()));
+      if (c.real() != 0) {
+        auto format_full = detail::string_literal<Char, '(', '{', '}', '+', '{',
+                                                  '}', 'i', ')'>{};
+        return fmt::format_to(out, basic_string_view<Char>(format_full),
+                              f->nested(c.real()), f->nested(c.imag()));
+      } else {
```
vitaut, line 603: nit: else can be dropped due to early return

vitaut (commented):
Just one nit, otherwise LGTM.

include/fmt/std.h

vitaut:
Merged, thanks!

PR #3907, opened 2024-03-26
Make CMake version message less confusing

xTachyon:
Looking through CMake logs, I was confused where a random "Version" string came from.

I hope this will make it more clear for the future.


CMakeLists.txt
```diff
@@ -178,7 +178,7 @@ math(EXPR CPACK_PACKAGE_VERSION_MINOR ${CMAKE_MATCH_2})
 math(EXPR CPACK_PACKAGE_VERSION_PATCH ${CMAKE_MATCH_3})
 join(FMT_VERSION ${CPACK_PACKAGE_VERSION_MAJOR}.${CPACK_PACKAGE_VERSION_MINOR}.
                  ${CPACK_PACKAGE_VERSION_PATCH})
-message(STATUS "Version: ${FMT_VERSION}")
+message(STATUS "Fmt Version: ${FMT_VERSION}")
```
vitaut, line 181: Let's change "Fmt" to the actual library name, "{fmt}".

vitaut:
Merged since the errors looks unrelated and should be already fixed in master. Thanks for the PR.

PR #3906, opened 2024-03-25
Implemented fmt::day, fmt::month, fmt::year and related unit tests

zivshek:
This should fix #3758

PR #3909, opened 2024-03-28
Replace std::fill_n with fmt::detail::fill_n

phprus:
Fix for #3906

Error cause: invalid ``fmt::basic_appender<>::iterator_category`` type.


phprus:
@vitaut review it please

vitaut:
Thanks for the fix!

PR #3916, opened 2024-04-01
Implement 2 more overloads for println.

BurntRanch:
Since println was first implemented, there were 2 overloads for print() that were not present in println, This PR implements it.


BurntRanch:
Finished, the code works now, had some problems because I'm not using an IDE at the moment.

vitaut:
Thanks for the PR but `println` is intentionally provided only where it is required by the standard.

PR #3918, opened 2024-04-03
add disable_colors

### LINKED ISSUE #2137: Provide a mechanism to disable color formatting globally
It would be nice to disable text styles globally. Ideally, both:

1. An API call or global variable for applications to use, and 
2. An environment variable that users could set when an application doesn't expose this option.

Some terminals (MSVC dev terminals are a glaring example) can't handle the formatting markup, and produce unreadable output when formatting is used (#1794).

#1549 suggests a way to implement this by placing special logic at every call into fmtlib, but this is burdensome and error prone. Having a library-wide solution for this problem would be fantastic.

If there's already a good workaround for this, please let me know. I found some closed issues that were related, but nothing that quite does what I'm looking for here.

Toni500github:
disable_colors is just a normal bool which the dev or the user can decide in the program if disable the printing of colors or not.

an example:
```cpp
#include <fmt/color.h>

int main() {
  // prints "Hello" in red fg
  fmt::print(fg(fmt::terminal_color::red), "Hello\n");
  // let's disable all colors prints
  fmt::disable_colors = true;
  // prints "Hello" in blank color
  fmt::print(fg(fmt::terminal_color::red), "Hello\n");
}
```
this can be also useful in cases where the user has the enviroment variable `NO_COLOR` set (https://no-color.org).
another example:
```cpp
#include <fmt/color.h>

int main() {
    char *no_color = getenv("NO_COLOR");

    // prints "Hello" in red fg
    fmt::print(fg(fmt::terminal_color::red), "Hello\n");
    
    if (no_color != NULL && no_color[0] != '\0')
        fmt::disable_colors = true;

    // will prints "Hello" in blank color if we have $NO_COLOR set, else it will print it in red
    fmt::print(fg(fmt::terminal_color::red), "Hello\n");
}
```

Fixes #2137

### DIFF
--- a/include/fmt/color.h
+++ b/include/fmt/color.h
@@ -187,6 +187,8 @@ enum class emphasis : uint8_t {
   strikethrough = 1 << 7,
 };
 
+FMT_INLINE bool disable_colors = false;
+
 // rgb is a struct for red, green and blue colors.
 // Using the name "rgb" makes some editors show the color in a tooltip.
 struct rgb {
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR but I don't think it's a good idea to introduce global state like this. You can write your own wrapper function that does this.

PR #3920, opened 2024-04-05
Fixed MSVC C2661: no overloaded function takes 2 arguments

matt77hias:
Fixed Error C2661 'fmt::v10::vformat': no overloaded function takes 2 arguments for MSVC v17.9.5 (fmt module).

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -147,12 +147,6 @@ FMT_API FMT_FUNC auto format_facet<std::locale>::do_put(
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


matt77hias:
Note that I moved it right underneath `FMT_FUNC auto vformat(string_view fmt, format_args args)` ;-)

vitaut:
This looks like an MSVC bug because the function is declared here: https://github.com/fmtlib/fmt/blob/master/include/fmt/format.h#L4370. Could you report it to Microsoft?

matt77hias:
Summary: https://github.com/fmtlib/fmt/issues/3921#issuecomment-2041454134

Issue #3921, opened 2024-04-05
MSVC errors when importing both fmt and std modules


matt77hias:
MSVC has an open [bug](https://developercommunity.visualstudio.com/t/VS2022-175-Preview-3---Compiler-bug-wit/10256508) that prevents a.o. importing both fmt and std modules in some use cases (_i.e. depends which functionality one uses from the std_) due to the mixing of std includes inside fmt itself and using `import std`.

For example:
>error C2572: 'std::enable_if': redefinition of default argument: parameter 1

It is possible to work around this by replacing std includes inside fmt with `import std` instead, and including the C headers for the few types and functions defined outside the `std` namespace (because these are not exported from the `std` module):
```c++
#include <limits.h>
#include <stdint.h>
#include <stdio.h>
#include <time.h>
```

Maybe something worth considering with an optional feature macro?

matt77hias:
Edit: <time.h> is broken as well for MSVC :-(

>error C2129: static function 'errno_t gmtime_s(tm *const ,const time_t *const )' declared but not defined

Cfr. https://developercommunity.visualstudio.com/t/Visual-Studio-cant-find-time-function/1126857

matt77hias:
tl;dr

I created a new [fork](https://github.com/matt77hias/fmt/commits/master/) with all changes required to use (_i.e. not just build a static lib_) MSVC + C++ modules + fmt:

* [Resolved warning C4127: conditional expression is constant](https://github.com/matt77hias/fmt/commit/910e6968e1545ef701dc5bccc497843e146a6890)
  * Independent of C++ modules.
  * Trivial to bring back to fmt.
* [Resolved warning C4996: 'fileno': The POSIX name for this item is deprecated. Instead, use the ISO C and C++ conformant name: _fileno.](https://github.com/matt77hias/fmt/commit/6e7e9ff706fd599b06ad81aefc68de01fea6a2b7)
  * Independent of C++ modules.
  * Trivial to bring back to fmt.
* [Worked around error C2039: 'vformat': is not a member of 'fmt'](https://github.com/matt77hias/fmt/commit/6a318b0924354697d72be9d066a13a1d18a5460d)
* [Worked around error C2039: 'vformat': is not a member of 'fmt'](https://github.com/matt77hias/fmt/commit/b9885465e9a111d470f22995ed010c3578967814)
[...]

PR #3924, opened 2024-04-08
Fix mix-up of 'FMT_BEGIN_EXPORT' and 'namespace detail'

ShifftC:
https://github.com/fmtlib/fmt/issues/3921#issuecomment-2042930855


vitaut:
Thank you!

PR #3928, opened 2024-04-09
Added FMT_IMPORT_STD feature macro

matt77hias:
How to use? Add `#define FMT_IMPORT_STD` underneath `module;` in fmt.cc.

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -12,9 +12,13 @@
 #include <stdio.h>   // FILE
 #include <string.h>  // strlen
 
+#ifndef FMT_IMPORT_STD
 // <cstddef> is also included transitively from <type_traits>.
-#include <cstddef>      // std::byte
-#include <type_traits>  // std::enable_if
+#  include <cstddef>      // std::byte
+#  include <type_traits>  // std::enable_if
+#else
+import std;
+#endif
 
 // The fmt library version in the form major * 10000 + minor * 100 + patch.
 #define FMT_VERSION 100202
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -8,13 +8,15 @@
 #ifndef FMT_FORMAT_INL_H_
 #define FMT_FORMAT_INL_H_
 
-#include <algorithm>
+#ifndef FMT_IMPORT_STD
+#  include <algorithm>
+#  include <cmath>
+#  include <exception>
+#endif
 #include <cerrno>  // errno
 #include <climits>
-#include <cmath>
-#include <exception>
 
-#ifndef FMT_STATIC_THOUSANDS_SEPARATOR
+#if !defined(FMT_STATIC_THOUSANDS_SEPARATOR) && !defined(FMT_IMPORT_STD)
 #  include <locale>
 #endif
 
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -38,30 +38,35 @@
 #  define FMT_REMOVE_TRANSITIVE_INCLUDES
 #endif
 
-#include <cmath>             // std::signbit
-#include <cstdint>           // uint32_t
-#include <cstring>           // std::memcpy
-#include <initializer_list>  // std::initializer_list
-#include <limits>            // std::numeric_limits
-#if defined(__GLIBCXX__) && !defined(_GLIBCXX_USE_DUAL_ABI)
+#ifndef FMT_IMPORT_STD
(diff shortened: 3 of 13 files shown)


include/fmt/base.h
```diff
@@ -8,13 +8,15 @@
...
+#  include <limits.h>  // CHAR_BIT
+#  include <stdio.h>   // FILE
+#  include <string.h>  // strlen
 
 // <cstddef> is also included transitively from <type_traits>.
-#include <cstddef>      // std::byte
-#include <type_traits>  // std::enable_if
+#  include <cstddef>      // std::byte
+#  include <type_traits>  // std::enable_if
+#endif
```
vitaut, line 19: I think we should move `import std` to make headers self-contained.

include/fmt/base.h
```diff
@@ -8,13 +8,15 @@
 #ifndef FMT_BASE_H_
 #define FMT_BASE_H_
 
-#include <limits.h>  // CHAR_BIT
-#include <stdio.h>   // FILE
-#include <string.h>  // strlen
+#ifndef FMT_IMPORT_STD
+#  include <limits.h>  // CHAR_BIT
+#  include <stdio.h>   // FILE
+#  include <string.h>  // strlen
```
vitaut, line 14: These includes should be moved out of the ifndef blocks because they are not part of the std module.

include/fmt/format.h
```diff
@@ -38,30 +38,36 @@
...
-#include <stdexcept>     // std::runtime_error
-#include <string>        // std::string
-#include <system_error>  // std::system_error
 
 #include "base.h"
 
 // Checking FMT_CPLUSPLUS for warning suppression in MSVC.
 #if FMT_HAS_INCLUDE(<bit>) && FMT_CPLUSPLUS > 201703L
-#  include <bit>  // std::bit_cast
+#  ifndef FMT_IMPORT_STD
```
vitaut, line 60: Let's simplify this by adding !defined(FMT_IMPORT_STD) to the first `#if`.

vitaut (changes requested):
Thanks for the PR. Please address the inline comments and apply clang-format to your changes.

include/fmt/base.h
matt77hias: Add `import std` to every file with std includes?

matt77hias:
¯\_(ツ)_/¯ force formatted upon saving in VS Code

include/fmt/base.h
vitaut: Just base.h which is included from all other headers.

include/fmt/format-inl.h
```diff
@@ -8,13 +8,15 @@
...
 #define FMT_FORMAT_INL_H_
 
-#include <algorithm>
-#include <cerrno>  // errno
-#include <climits>
-#include <cmath>
-#include <exception>
+#ifndef FMT_IMPORT_STD
+#  include <algorithm>
+#  include <cerrno>  // errno
```
vitaut, line 13: `errno` is often a macro so this include should be moved outside of the ifndef block.

vitaut (changes requested):

include/fmt/format-inl.h
```diff
@@ -8,13 +8,15 @@
 #ifndef FMT_FORMAT_INL_H_
 #define FMT_FORMAT_INL_H_
 
-#include <algorithm>
+#ifndef FMT_IMPORT_STD
+#  include <algorithm>
+#  include <climits>
```
vitaut, line 13: Here and elsewhere: `climits` only exports macros so it should be moved outside of the ifndef block.

include/fmt/std.h
```diff
@@ -28,22 +30,24 @@
...
+#if FMT_HAS_INCLUDE(<expected>) && FMT_CPLUSPLUS > 202002L && \
+    !defined(FMT_IMPORT_STD)
 #  include <expected>
 #endif
 
-#if FMT_CPLUSPLUS > 201703L && FMT_HAS_INCLUDE(<source_location>)
+#if FMT_CPLUSPLUS > 201703L && FMT_HAS_INCLUDE(<source_location>) && \
+    !defined(FMT_IMPORT_STD)
 #  include <source_location>
 #endif
```
vitaut, line 52: Thinking more of it, let's wrap this whole block in the `FMT_IMPORT_STD` check.

vitaut (approved):

vitaut:
Merged, thanks!

PR #3930, opened 2024-04-10
Resolved warning C4996: 'fileno': The POSIX name for this item is dep…

matt77hias:


matt77hias:
I restricted it to `defined(_WIN32)`, consistent to all prior uses of `_fileno`.

vitaut:
Thanks

Issue #3925, opened 2024-04-08
fmt::format_to + FMT_STRING with wide characters fails to compile


hmbj:
Using a wide character buffer, `fmt::format_to` with `FMT_STRING` fails to compile. Compiles fine using narrow character buffer.

```
// Using a 'char' buffer compiles as expected
std::string buf;
fmt::format_to(std::back_inserter(buf), FMT_STRING("{}"), 1);

// Using 'wchar_t' buffer fails to compile:
// opt/compiler-explorer/libs/fmt/trunk/include/fmt/xchar.h:190:48: error:
// no matching function for call to 'to_string_view(const main()::<lambda()>::FMT_COMPILE_STRING&)'
std::wstring wbuf;
fmt::format_to(std::back_inserter(wbuf), FMT_STRING(L"{}"), 2);  // this compiles without FMT_STRING
```
godbolt repro:
https://godbolt.org/z/h7Trrjf5M

hmbj:
The code above compiles successfully prior to this commit:
https://github.com/fmtlib/fmt/commit/2e5b14bf603429a08ab9480bcd1fd537a2fe8f16

PR #3931, opened 2024-04-14
Fix format_to + FMT_STRING for wide character type

hmbj:
Fixes issue #3925

### DIFF
--- a/include/fmt/xchar.h
+++ b/include/fmt/xchar.h
@@ -127,6 +127,18 @@ auto join(const std::tuple<T...>& tuple, basic_string_view<wchar_t> sep)
   return {tuple, sep};
 }
 
+template <typename... T>
+auto format(wformat_string<T...> fmt, T&&... args) -> std::wstring {
+  return vformat(fmt::wstring_view(fmt), fmt::make_wformat_args(args...));
+}
+
+template <typename OutputIt, typename... T>
+auto format_to(OutputIt out, wformat_string<T...> fmt, T&&... args)
+    -> OutputIt {
+  return vformat_to(out, fmt::wstring_view(fmt),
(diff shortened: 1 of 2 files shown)


hmbj:
Not sure if this is the optimal fix, though :)

vitaut (changes requested):
Thanks for the PR. I think it's better to keep `detail::to_string_view` and instead add a `format_to` overload that takes `wformat_string` similar to this one: https://github.com/fmtlib/fmt/blob/aa52eb765dec0eb67fab4b035a4a05b82346c8f5/include/fmt/xchar.h#L140

include/fmt/xchar.h
```diff
@@ -127,6 +127,18 @@ auto join(const std::tuple<T...>& tuple, basic_string_view<wchar_t> sep)
   return {tuple, sep};
 }
 
+template <typename... T>
+auto format(wformat_string<T...> fmt, T&&... args) -> std::wstring {
+  return vformat(fmt::wstring_view(fmt), fmt::make_wformat_args(args...));
+}
```
vitaut, line 133: There is no need to move this function definition, please revert.

hmbj: I've reverted the move.

This patch only fixes format_to called with a compile time string (FMT_STRING). Looks like the other functions calling detail::to_string_view in this file has the same problem, like `vformat_to`, `vformat` and maybe `formatted_size`...

vitaut (approved):

include/fmt/xchar.h
vitaut: `vformat*` are fine because they take runtime format strings.

vitaut:
Merged, thanks.

PR #3943, opened 2024-04-22
Specialize `formatter` for all `std::basic_string` types

### LINKED ISSUE #3938: Formatting for strings with custom allocators fails to compile
When using a `std::basic_string` with a custom allocator, it fails to compile in a few use cases, e.g.:

```cpp
std::string f1() {
    // This compiles.
    return fmt::format("{}", std::tuple<std::string>{});
}

std::string f2() {
    // This compiles.
    return fmt::format("{}", std::pmr::string{});
}

std::string f3() {
    // This does not compile.
    return fmt::format("{}", std::tuple<std::pmr::string>{});
}
```

You can reproduce the error here: https://godbolt.org/z/4oa1jqr1G

I checked the code, and it seems `std::basic_string` is specialized only for the default allocator. I think the fix might be as simple as replacing

```cpp
FMT_FORMAT_AS(std::basic_string<Char>, basic_string_view<Char>);
```

with

```cpp
template <typename Char, typename Alloc>
class formatter<std::basic_string<Char, std::char_traits<Char>, Alloc>, Char>
    : public formatter<basic_string_view<Char>, Char> {};
```

It is also not clear why printing a `std::pmr::string` alone works. Maybe it's falling back to `string_view` while the tuple version is not.

dieram3:
This fixes #3938 by adding a `formatter` partial specialization that covers all `std::basic_string` types, such as `std::pmr::string`.

Additionally, this adds the old-school allocator member types to `mock_allocator` in order to make GCC 4.8 happy. If GCC 4.8 support is removed in the future, these member types can be removed as they are the defaults of `allocator_traits`.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -4017,10 +4017,13 @@ FMT_FORMAT_AS(unsigned short, unsigned);
 FMT_FORMAT_AS(long, detail::long_type);
 FMT_FORMAT_AS(unsigned long, detail::ulong_type);
 FMT_FORMAT_AS(Char*, const Char*);
-FMT_FORMAT_AS(std::basic_string<Char>, basic_string_view<Char>);
 FMT_FORMAT_AS(std::nullptr_t, const void*);
 FMT_FORMAT_AS(detail::std_string_view<Char>, basic_string_view<Char>);
 FMT_FORMAT_AS(void*, const void*);
+
+template <typename Char, typename Traits, typename Allocator>
+class formatter<std::basic_string<Char, Traits, Allocator>, Char>
+    : public formatter<basic_string_view<Char>, Char> {};
 
 template <typename Char, size_t N>
 struct formatter<Char[N], Char> : formatter<basic_string_view<Char>, Char> {};
--- a/test/format-test.cc
+++ b/test/format-test.cc
@@ -22,6 +22,7 @@
 #include <iterator>            // std::back_inserter
 #include <list>                // std::list
 #include <mutex>               // std::mutex
+#include <string>              // std::string
 #include <thread>              // std::thread
 #include <type_traits>         // std::is_default_constructible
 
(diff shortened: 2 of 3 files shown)


test/format-test.cc
```diff
@@ -2222,16 +2223,21 @@ template <typename Char, typename... T> void check_enabled_formatters() {
...
                            unsigned short, int, unsigned, long, unsigned long,
                            long long, unsigned long long, float, double,
                            long double, void*, const void*, char*, const char*,
-                           std::string, std::nullptr_t>();
-  check_enabled_formatters<wchar_t, bool, wchar_t, signed char, unsigned char,
-                           short, unsigned short, int, unsigned, long,
-                           unsigned long, long long, unsigned long long, float,
-                           double, long double, void*, const void*, wchar_t*,
-                           const wchar_t*, std::wstring, std::nullptr_t>();
+                           std::string, custom_string, std::nullptr_t>();
```
dieram3, line 2235: This is my best guess on where this code change should be tested. I figured that given we already test the `std::string` specialization at runtime, testing only the compilation of `custom_string` might be enough.

include/fmt/format.h
```diff
@@ -4017,11 +4017,14 @@ FMT_FORMAT_AS(unsigned short, unsigned);
 FMT_FORMAT_AS(long, detail::long_type);
 FMT_FORMAT_AS(unsigned long, detail::ulong_type);
 FMT_FORMAT_AS(Char*, const Char*);
-FMT_FORMAT_AS(std::basic_string<Char>, basic_string_view<Char>);
 FMT_FORMAT_AS(std::nullptr_t, const void*);
 FMT_FORMAT_AS(detail::std_string_view<Char>, basic_string_view<Char>);
 FMT_FORMAT_AS(void*, const void*);
 
+template <typename Char, typename Traits, typename Allocator>
+class formatter<std::basic_string<Char, Traits, Allocator>, Char>
```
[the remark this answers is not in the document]
vitaut: Yes, `Traits` should be ignored and I agree that it's better to handle string_view separately.

test/format-test.cc
vitaut: Looks OK but you need to add `rebind` to `mock_allocator` to fix https://github.com/fmtlib/fmt/actions/runs/8779818010/job/24107542630?pr=3943.

vitaut (changes requested):
Thanks for the PR!

test/format-test.cc
dieram3: Yikes, GCC 4.8.

Added the required member types and validated here: https://godbolt.org/z/a1o7YjY5M (every single one of them is required).
I also updated the PR description accordingly (in case you use it as the commit message).

If you don't plan to support GCC 4.8 much longer (which is 11 years old), it might make sense to add a note to remove these types later.

vitaut (approved):

test/format-test.cc
vitaut: Thanks. There are no plans to drop gcc 4.8 at the moment.

PR #3944, opened 2024-04-28
adding build option FMT_MINIMAL_INSTALL

maiananthan:
* adding option to install only essential components (libfmt.so) when FMT_MINIMAL_INSTALL option is used


vitaut:
Thanks for the PR but I don't think we need to introduce another configuration option for this. If you would like to only install the shared object, it can be done with a simple separate CMake script.

PR #3945, opened 2024-04-29
Fix CodeQL alert

marlenecota:
CodeQL found an issue in base.h via static analysis where the lower bound  of the index id is not being checked before accessing array values_. 

microsoft/react-native-windows#12702

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -1899,12 +1899,10 @@ template <typename Context> class basic_format_args {
       if (id < max_size()) arg = args_[id];
       return arg;
     }
-    if (id >= detail::max_packed_args) return arg;
+    if (static_cast<unsigned>(id) >= detail::max_packed_args) return arg;
     arg.type_ = type(id);
     if (arg.type_ == detail::type::none_type) return arg;
-    if (id >= 0) {
-      arg.value_ = values_[id];
-    }
+    arg.value_ = values_[id];
     return arg;
   }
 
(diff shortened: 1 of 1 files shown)


include/fmt/base.h
```diff
@@ -1902,7 +1902,9 @@ template <typename Context> class basic_format_args {
     if (id >= detail::max_packed_args) return arg;
     arg.type_ = type(id);
```
chrisglein, line 1903: Is this line that calls `type(id)` also a problem if `id` is out of bounds? What's the expected behavior? In the string-based wrapper below on line 1914 it returns an empty `format_arg()` up front. Should the check for `id >= 0` happen first? And then be removed from line 1914?

Not particularly familiar with this code, just wondering what the right bounds strategy is, since the static analysis tools are triggering for it.

include/fmt/base.h
```diff
@@ -1902,7 +1902,9 @@ template <typename Context> class basic_format_args {
     if (id >= detail::max_packed_args) return arg;
```
vitaut, line 1902: I think the correct fix is to cast id to unsigned using a static cast to filter out negative out of bound values in addition to positive ones.

vitaut (changes requested):
Thanks for the PR!

include/fmt/base.h
vitaut: Yeah, type and value should either be both set or not. I commented above suggesting a better fix.

vitaut (approved):

vitaut:
Merged, thanks!

marlenecota:
Thanks, @vitaut! Would you happen to know when 10.2.2 might get published?

vitaut:
There are no plans to publish a patch release at the moment. Note that negative IDs cannot appear when using the usual formatting API such as `fmt::format` or `fmt::print`. Therefore I don't think this fix warrants a special release.

Issue #3802, opened 2024-01-10
fmt::join doesn't handle move-only iterators


tcbrindle:
C++20 allows iterators for input ranges to be move-only, for example `std::ranges::basic_istream_view::iterator`. Unfortunately `fmt::join_view` copies iterators in various places, meaning it can't be used with such ranges:

```cpp
int main()
{
    std::istringstream iss("1 2 3 4 5");

    auto view = std::views::istream<int>(iss);

    fmt::println("{}", fmt::join(view, ", ")); // Error, use of deleted iterator copy constructor
}
```

https://godbolt.org/z/1ndGahqb7

vitaut:
Fixed by #3800 (thanks).

vitaut:
Or is it unrelated?

tcbrindle:
> Or is it unrelated?

This is a different to #3800

tcbrindle:
There are a couple of places in the `join_view` constructor where iterators are copied, but these are easily solved with `std::move`. The difficulty is with `formatter<join_view>::format()`:

https://github.com/fmtlib/fmt/blob/810d1750f1579f19d842b6f3ca6671f714375444/include/fmt/ranges.h#L587-L603

Line 590 takes a copy of the iterator, which it then modifies. Because the `join_view` is passed to this function as a const reference, there's no way we can move the iterator out of it -- and we also can't directly operate on the stored iterator, as it's const.

It seems like we need to take the `join_view` argument to `format()` by mutable `&` rather than `const&`, but making this change results in fmt considering `join_view` to be non-formattable. Unfortunately I don't know enough about the library to understand why that happens. I also tried using a forwarding reference, but this always deduces to `const&` and so doesn't help.

The generic range support without `join` (i.e. `fmt::format("{}", my_range)`) handles move-only iterators just fine, so this definitely seems like it should be solvable, but I don't understand enough about the fmt internals to follow what's going on.

vitaut:
> It seems like we need to take the join_view argument to format() by mutable & rather than const&, but making this change results in fmt considering join_view to be non-formattable.

This should work. What error do you get?

tcbrindle:
Changing the signature of `formatter<join_view>::format` to

```cpp
 template <typename FormatContext> 
 auto format(join_view<It, Sentinel, Char>& value, 
             FormatContext& ctx) const -> decltype(ctx.out())
```

(that is, removing the `const` from the `join_view` argument) generates the following from Clang:

```
[...]

PR #3946, opened 2024-05-01
Fix fmt::join for views with input iterators

Arghnews:
Addresses issue (#3802)

Following from the issue discussion, adding 2 std::moves gets this error:

```
In file included from /home/justin/code/fmt/test/ranges-test.cc:8:
/home/justin/code/fmt/include/fmt/ranges.h:597:10: error: call to deleted constructor of 'std::ranges::basic_istream_view<int, char>::__iterator'
  597 |     auto it = value.begin;
      |          ^    ~~~~~~~~~~~
/home/justin/code/fmt/include/fmt/base.h:1383:23: note: in instantiation of function template specialization 'fmt::formatter<fmt::join_view<std::ranges::basic_istream_view<int, char>::__iterator, std::default_sentinel_t>>::format<fmt::context>' requested here
 1383 |     ctx.advance_to(cf.format(*static_cast<qualified_type*>(arg), ctx));
      |                       ^
/home/justin/code/fmt/include/fmt/base.h:1364:21: note: in instantiation of function template specialization 'fmt::detail::value<fmt::context>::format_custom_arg<fmt::join_view<std::ranges::basic_istream_view<int, char>::__iterator, std::default_sentinel_t>, fmt::formatter<fmt::join_view<std::ranges::basic_istream_view<int, char>::__iterator, std::default_sentinel_t>>>' requested here
 1364 |     custom.format = format_custom_arg<
      |                     ^
/home/justin/code/fmt/test/ranges-test.cc:634:8: note: in instantiation of function template specialization 'fmt::format<fmt::join_view<std::ranges::basic_istream_view<int, char>::__iterator, std::default_sentinel_t>>' requested here
  634 |   fmt::format("{}", std::move(joined_view)); // Error, use of deleted iterator copy constructor
      |        ^
/usr/lib/llvm-18/bin/../include/c++/v1/__ranges/istream_view.h:72:3: note: '__iterator' has been explicitly marked deleted here
   72 |   __iterator(const __iterator&)                  = delete;
      |   ^
```

The issue being [`std::ranges::basic_istream_view::__iterator` is not copyable, only movable](https://en.cppreference.com/w/cpp/ranges/basic_istream_view/iterator)

The underlying issue is that currently, the joined_view formatter
`struct formatter<join_view<It, Sentinel, Char>, Char>`

has a format func like so:

https://github.com/fmtlib/fmt/blob/810d1750f1579f19d842b6f3ca6671f714375444/include/fmt/ranges.h#L587-L590

Where we take a copy of the iterator which we then mutate by incrementing it which is causing our error

However, we can't mutate the iterator in place as is, as we take the join_view by const&
This also prevents us from moving the iterator into it
And we can't copy it in case (like this one) it's only movable

If we change the function signature to take it by
  `auto format(join_view<It, Sentinel, Char>& value`
As mentioned https://github.com/fmtlib/fmt/issues/3802#issuecomment-1889635249 by @tcbrindle 
This will fix this case, but then break it for other cases where we want the const-ness

---

As I understand it, what we need is a fix that treats these kind of input iterators specially, as they
are inherently destructive since they're single pass (probably why the std prevents copying of this one, less foot guns).

We can add a version of our format function that takes `value` by const ref when not using an input iterator
And otherwise just uses a mutable ref to `value` in place (no copies)

We must disable the const overload for format for input iterators so that

https://github.com/fmtlib/fmt/blob/810d1750f1579f19d842b6f3ca6671f714375444/include/fmt/base.h#L1299-L1302

Correctly deduces that when T's iterator is an input iterator, `has_const_formatter_impl` returns false
And otherwise returns true

---

Since this issue only seems to be a c++20 problem, and no one using c++17 etc has raised it it, I've used concept `requires` for this.
This could be changed to use `FMT_ENABLE_IF` or `std::enable_if_t` instead.
I've also used the `std::input_iterator` and `std::forward_iterator` concepts from `<iterator>` (already included). The c++17 SFINAE is a bit tricky (and would likely want an ifdef to use the c++20 concepts anyway, as iterator_concept and iterator_category is a bit finnicky etc.)

---

range-v3's `istream_view` does not have this issue, the iterators are copyable (whether they should be or not is a separate question)
Ie. see https://godbolt.org/z/xcb3zf45Y

I've checked the ifdef for the test as best I can
Best way I could find was checking `__cpp_lib_ranges` which means gcc supports this test from gcc >= 12.
Clang 15 supports it but it still fails to compile, so clang >= 16 required, and c++20

gcc 11.4 fails __cpp_lib_ranges == 202106
gcc 12 works __cpp_lib_ranges == 202110

clang 16 works
clang 15 still a special sausage with its own error

msvc 19.30 __cpp_lib_ranges == 202106
msvc 19.31 __cpp_lib_ranges == 202110

---

Sorry for the length of this! Just trying to be clear and lay out my thought process as it's not trivial.
Open to feedback. Maybe there's a simpler way to do this that I've missed!

### DIFF
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -13,6 +13,7 @@
 #  include <iterator>
 #  include <tuple>
 #  include <type_traits>
+#  include <utility>
 #endif
 
 #include "format.h"
@@ -611,7 +612,7 @@ struct join_view : detail::view {
   basic_string_view<Char> sep;
 
   join_view(It b, Sentinel e, basic_string_view<Char> s)
-      : begin(b), end(e), sep(s) {}
+      : begin(std::move(b)), end(e), sep(s) {}
 };
 
 template <typename It, typename Sentinel, typename Char>
@@ -631,10 +632,25 @@ struct formatter<join_view<It, Sentinel, Char>, Char> {
     return value_formatter_.parse(ctx);
   }
 
-  template <typename FormatContext>
-  auto format(const join_view<It, Sentinel, Char>& value,
+  template <typename FormatContext, typename Iter,
+            FMT_ENABLE_IF(std::is_copy_constructible<Iter>::value)>
+  auto format(const join_view<Iter, Sentinel, Char>& value,
               FormatContext& ctx) const -> decltype(ctx.out()) {
     auto it = value.begin;
+    return do_format(value, ctx, it);
+  }
+
+  template <typename FormatContext, typename Iter,
+            FMT_ENABLE_IF(!std::is_copy_constructible<Iter>::value)>
+  auto format(join_view<Iter, Sentinel, Char>& value, FormatContext& ctx) const
+      -> decltype(ctx.out()) {
+    return do_format(value, ctx, value.begin);
+  }
+
+ private:
+  template <typename FormatContext>
+  auto do_format(const join_view<It, Sentinel, Char>& value, FormatContext& ctx,
+                 It& it) const -> decltype(ctx.out()) {
     auto out = ctx.out();
     if (it != value.end) {
       out = value_formatter_.format(*it, ctx);
@@ -656,7 +672,7 @@ struct formatter<join_view<It, Sentinel, Char>, Char> {
  */
 template <typename It, typename Sentinel>
 auto join(It begin, Sentinel end, string_view sep) -> join_view<It, Sentinel> {
-  return {begin, end, sep};
+  return {std::move(begin), end, sep};
 }
 
 /**
--- a/test/ranges-test.cc
+++ b/test/ranges-test.cc
@@ -652,3 +652,56 @@ struct lvalue_qualified_begin_end {
 TEST(ranges_test, lvalue_qualified_begin_end) {
   EXPECT_EQ(fmt::format("{}", lvalue_qualified_begin_end{}), "[1, 2, 3, 4, 5]");
 }
+
+#if !defined(__cpp_lib_ranges) || __cpp_lib_ranges <= 202106L
+#  define ENABLE_INPUT_RANGE_JOIN_TEST 0
+#elif FMT_CLANG_VERSION
+#  if FMT_CLANG_VERSION > 1500
+#    define ENABLE_INPUT_RANGE_JOIN_TEST 1
+#  else
+#    define ENABLE_INPUT_RANGE_JOIN_TEST 0
+#  endif
+#else
+#  define ENABLE_INPUT_RANGE_JOIN_TEST 1
+#endif
+
+#if ENABLE_INPUT_RANGE_JOIN_TEST
+TEST(ranges_test, input_range_join) {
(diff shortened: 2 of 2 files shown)


Arghnews:
Will take another look at this, looks like earlier versions of libc++ etc. didn't have library support for the `std::input_iterator` concept even in c++20 mode. Either need to disable via ifdefs, or switch to SFINAE to support pre cpp20. Is there a preference?


Also I ran `clang-format-18 -i --style=file include/fmt/ranges.h` like suggested and still have the formatting diff, unsure why that is

vitaut:
> Would this be acceptable?

I'd recommend using SFINAE instead of conditional compilation if possible.

> Also I ran clang-format-18 -i --style=file include/fmt/ranges.h like suggested and still have the formatting diff, unsure why that is

clang-format occasionally produces different results between versions. {fmt} uses 17.0.5 at the moment.

include/fmt/ranges.h
```diff
@@ -591,10 +597,33 @@ struct formatter<join_view<It, Sentinel, Char>, Char> {
...
+  }
+
+#ifdef FMT_USE_ITERATOR_CONCEPT
+  template <typename FormatContext>
+    requires is_input_iterator
+  auto format(join_view<It, Sentinel, Char>& value,
+              FormatContext& ctx) const -> decltype(ctx.out()) {
+    return format_impl(value, ctx, value.begin);
+  }
+#endif
```
vitaut, line 622: I think we should (1) check if the iterator is copyable rather than trying to detect if it is an input iterator and (2) use SFINAE instead of concepts.

Arghnews: Nice suggestion, certainly simpler, have made the change

Other changes introduced are just formatting from running clang-format (17.0.5 as you suggested)

include/fmt/ranges.h
```diff
@@ -591,10 +592,24 @@ struct formatter<join_view<It, Sentinel, Char>, Char> {
...
+
+  template <typename FormatContext, typename Iter,
+            FMT_ENABLE_IF(!std::is_copy_constructible<Iter>::value)>
+  auto format(join_view<Iter, Sentinel, Char>& value, FormatContext& ctx) const
+      -> decltype(ctx.out()) {
+    return format_impl(value, ctx, value.begin);
+  }
+
+  template <typename FormatContext>
+  auto format_impl(const join_view<It, Sentinel, Char>& value,
```
vitaut, line 611: This should be private. Also let's rename to `do_format` to follow the convention for such methods in {fmt}.

vitaut (changes requested):
Overall looks good but please rebase and address the minor comment inline.

Arghnews:
Thanks for the feedback, have made the changes you suggested and rebased, hope it's ok now, thanks!

vitaut (approved):

vitaut:
Merged, thanks!

PR #3951, opened 2024-05-02
Fix `isnan` so it doesn't cause FP errors

### LINKED ISSUE #3948: `fmt::isnan` triggers floating-point exception for NaN values
Note: floating-point exceptions are not the same as C++ exceptions! See: https://www.gnu.org/savannah-checkouts/gnu/libc/manual/html_node/Control-Functions.html

Commit ef54f9aa3858d7d10591719615bce337fbe49c40 changed the implementation of `fmt::isnan` from:

```cpp
template <typename T> constexpr bool isnan(T value) {
  return value != value;  // std::isnan doesn't support __float128.
}
```

To:

```cpp
template <typename T> constexpr bool isnan(T value) {
  return !(value >= value);  // std::isnan doesn't support __float128.
}
```

The commit description says that this was done to suppress `-Wfloat-equal` warnings. The problem is that these two implementations are subtly different: while they both return the same value, the newer implementation also sets a floating-point exception flag. The reason is that while for IEEE-754 floating-point numbers the operation `NaN != NaN` is well defined and will always return true, `NaN >= NaN` is technically an invalid operation, hence why the exception flag is set (at least this is my non-expert understanding).

This has been causing headaches for a project I'm working on because I wanted to trace where floating-point exceptions were emanating from in my program (using the [GNU-specific `feenableexcept` function](https://www.gnu.org/savannah-checkouts/gnu/libc/manual/html_node/Control-Functions.html)), but `fmt::isnan` will raise a spurious exception every time a NaN is passed in. Changing the code to use `value != value` instead fixes things.

I guess there are different ways to fix this, but maybe you could just return to using `value != value` and find another way to suppress the warning (e.g. with a comment or a compiler argument)?

alexdewar:
As described in #3948, the current internal implementation of `isnan` triggers a floating-point exception as a side effect, unlike `std::isnan`. This can cause issues for users attempting to find other causes of FP exceptions in their code.

Fix by changing `isnan` to use `value != value` as a test, which does not have side effects.

I also added a test to confirm that FP exceptions are not raised (as suggested by @vitaut). I've confirmed that it fails with the current implementation and passes with this one.

Fixes #3948.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -2695,7 +2695,7 @@ FMT_CONSTEXPR20 auto write_float(OutputIt out, const DecimalFP& f,
 }
 
 template <typename T> constexpr auto isnan(T value) -> bool {
-  return !(value >= value);  // std::isnan doesn't support __float128.
+  return value != value;  // std::isnan doesn't support __float128.
 }
 
 template <typename T, typename Enable = void>
--- a/test/format-test.cc
+++ b/test/format-test.cc
@@ -15,6 +15,7 @@
 
 #include <stdint.h>  // uint32_t
 
+#include <cfenv>               // fegetexceptflag and FE_ALL_EXCEPT
 #include <climits>             // INT_MAX
 #include <cmath>               // std::signbit
 #include <condition_variable>  // std::condition_variable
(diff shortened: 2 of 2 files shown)


phprus:
@vitaut, What do you think about adding a GCC and Clang ``pragma``'s to suppress ``-Wfloat-equal`` warning?

vitaut:
> What do you think about adding a GCC and Clang pragma's to suppress -Wfloat-equal warning?

Considering that `-Wfloat-equal` is not in `-Wall -Wextra` I don't think we need to do anything. If someone wants to enable this warning then could use `FMT_SYSTEM` to suppress it in {fmt} similarly how it is done for other esoteric warnings.

vitaut:
Thank you!

Issue #3752, opened 2023-12-11
`format_as` fails when called with certain `std::views`


fennewald:
# Reproduction

[Godbolt link](https://godbolt.org/z/nb4cYdrTr)

# Summary

A `format_as` definition that returns a `std::views::filter_view`, like so:
```cpp
auto format_as(CodecMask m) {
    return CodecMask::k_codecs | std::views::filter([&](auto c) { return m.has(c); });
}
```

does not compile:
```cpp
CodecMask m;
m.add(Codec::H264);

fmt::println("{}", format_as(m)); // valid
fmt::pritnln("{}", m); // invalid
```

## Stack Overflow

I [posted this to stack overflow originally](https://stackoverflow.com/questions/77637232/format-as-use-with-range-types), expecting some mistake of my own. People there suggested that this may be a bug with `fmt` itself, hence this issue. The explanation given there may prove to be of value.

## Notes

I expected the most recent commit ([274ba26](https://github.com/fmtlib/fmt/commit/274ba2645bdae12f6f0c7d7ca24659c4af670548)) may have helped, as it appears to be related to this type of issue. It is worth noting the error message changes when tested against `10.1.1` and `trunk` in godbolt, but both still do fail.

vitaut:
There is a simpler repro for the `set_debug_format` error in https://github.com/fmtlib/fmt/issues/3763#issue-2048061455.

fennewald:
Understood, closing as duplicated!

vitaut:
Let's keep this one open for tracking the issue.

vitaut:
The `set_debug_format` issue is fixed in https://github.com/fmtlib/fmt/commit/3eb3aef5753b91d689b496a629b9dc5c4a7cd348 but looks like there is another problem.

Arghnews:
@vitaut This issue should be fixed by #3955

vitaut:
The original repro compiles now: https://godbolt.org/z/ahh86ME13. Thanks @Arghnews for the fix!

PR #3955, opened 2024-05-11
Fix format_as for non-const begin/end views

Arghnews:
Fixes issue #3752

`base::format` does not accept rvalues, using the result of `format_as` directly means it is passed one. Doesn't matter for ranges with a const begin and end, but since `std::ranges::filter_view` caches the next value to return, it only has a [mutable begin/end iterator](https://en.cppreference.com/w/cpp/ranges/filter_view#begin).

If we just store the result of `format_as` in a temporary variable, we can pass it as an lvalue.

Since we use it in
https://github.com/fmtlib/fmt/blob/1dc71f21eadac50c85b5c840ca7db2bf7a507b15/include/fmt/ranges.h#L485-L490
And we don't call `std::forward` anyway, I don't think we're performing any pessimisation by making this change

---

Another way to fix this is by providing a `&&` overload of
https://github.com/fmtlib/fmt/blob/1dc71f21eadac50c85b5c840ca7db2bf7a507b15/include/fmt/ranges.h#L541-L545
However other control paths (that pass lvalues) then flow through this, which doesn't seem right, so didn't opt for that method

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -4000,7 +4000,9 @@ struct formatter<T, Char, enable_if_t<detail::has_format_as<T>::value>>
   template <typename FormatContext>
   auto format(const T& value, FormatContext& ctx) const -> decltype(ctx.out()) {
     using base = formatter<detail::format_as_t<T>, Char>;
-    return base::format(format_as(value), ctx);
+    // format functions expect lvalue refs, not rvalues
+    auto&& val = format_as(value);
+    return base::format(val, ctx);
   }
 };
 
--- a/test/ranges-test.cc
+++ b/test/ranges-test.cc
@@ -7,6 +7,7 @@
 
 #include "fmt/ranges.h"
 
+#include <array>
 #include <list>
 #include <map>
 #include <numeric>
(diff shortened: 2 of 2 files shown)


include/fmt/format.h
```diff
@@ -4000,7 +4000,9 @@ struct formatter<T, Char, enable_if_t<detail::has_format_as<T>::value>>
   template <typename FormatContext>
   auto format(const T& value, FormatContext& ctx) const -> decltype(ctx.out()) {
     using base = formatter<detail::format_as_t<T>, Char>;
-    return base::format(format_as(value), ctx);
+    // format functions expect lvalue refs, not rvalues
+    auto val = format_as(value);
```
vitaut, line 4004: Can it introduce an extra copy? Can `val` be a reference?

vitaut (commented):
Thanks for the PR!

include/fmt/format.h
Arghnews: If `format_as` returns a local variable of type `T` by value then no extra copy will be made. This was my thought of the predominant use case.

However for a `format_as` function signature like: `const std::string& format_as(T t)`. Let's say you have a lookup table of `std::string`s for `T`, then using `auto val` always copies, while `auto&& val` deduces the type of `val` as a `const &` so no copy is made.

https://godbolt.org/z/fqMrcY3c3

Thus I think your suggestion of using `auto&&` is a good one here! Will make the change

vitaut:
Merged, thanks!

Issue #3839, opened 2024-02-02
fmt/ranges.h incompatible with Eigen 3.4


baronep:
Eigen 3.4 seems to have a significant change in behavior where `begin()` exists for all matrices, but returns `void` for non 1-D matrices where iteration is not well defined. 

This currently interferes with the `fmt/ranges.h` implementation such that the following code compiles with Eigen 3.9, but no longer compiles with 3.4.
```
#include <Eigen/Core>

#include <fmt/format.h>
#include <fmt/ostream.h>
#include <fmt/ranges.h> // doesn't work if uncomment this

int main() {
  Eigen::Matrix3d m;
  m << 1, 2, 3, 4, 5, 6, 7, 8, 9;
  fmt::print("{}\n", m);

  return 0;
}
```

**Note: This was built using the `-DFMT_DEPRECATED_OSTREAM` compiler flag**

```
<source>:10:13:   required from here
/opt/compiler-explorer/libs/fmt/9.1.0/include/fmt/ranges.h:264:34: error: void value not ignored as it ought to be
  264 |     decltype(*detail::range_begin(std::declval<Range&>()));
```
**Godbolt Example:** https://godbolt.org/z/TzG7TMxWx

baronep:
One potential solution appears to be patching `has_const_begin_end` and `has_const_begin_end` as follows ...
```
template <typename T>
struct has_const_begin_end<
    T,
    void_t<
        decltype(detail::range_begin(std::declval<const remove_cvref_t<T>&>())),
        decltype(detail::range_end(std::declval<const remove_cvref_t<T>&>())),
        decltype(detail::range_begin(std::declval<const remove_cvref_t<T>&>()) != detail::range_end(std::declval<const remove_cvref_t<T>&>()))>>
    : std::true_type {};

template <typename T>
struct has_mutable_begin_end<
    T, void_t<decltype(detail::range_begin(std::declval<T>())),
              decltype(detail::range_end(std::declval<T>())),
              decltype(detail::range_begin(std::declval<T>()) != detail::range_end(std::declval<T>())),
              enable_if_t<std::is_copy_constructible<T>::value>>>
    : std::true_type {};
```

vitaut:
`FMT_DEPRECATED_OSTREAM` has been removed so this conflict won't happen on the current version. You can format objects via ostream as follows:

```c++
  fmt::print("{}\n", fmt::streamed(m));
```

In this case Eigen providing broken `begin` / `end` doesn't matter.

[...]

PR #3961, opened 2024-05-15
Disable range formatter when user provides one

Arghnews:
Disable formatting type as range if user provided format_as function exists, using that instead. Should fix issue raised in #3839 by @drouhard

Edit: ah, this will need a similar workaround to `detail::is_formattable_delayed` for old msvc versions, before it will pass all the CI tests

### DIFF
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -513,6 +513,7 @@ template <typename R, typename Char>
 struct formatter<
     R, Char,
     enable_if_t<conjunction<
+        bool_constant<!detail::has_format_as<R>{}>,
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR but I don't think this will work for formatter specializations and having this just for `format_as` would be inconsistent. `fmt::format` and `std::format` already have an opt out mechanism for ranges.

jdrouhard:
What is the opt-out mechanism for a type if I don't want to include `<fmt/ranges.h>` in the header where the type is defined?

vitaut:
You could probably forward declare `is_range` but I would recommend including `<fmt/ranges.h>` since you are effectively defining formatting for ranges.

PR #3962, opened 2024-05-15
Bazel support: Add utf-8 to Windows build

Vertexwahn:
This PR only affects the Bazel build support for fmt (CMake build stays untouched). In detail:

- Add `-utf-8` to Windows Bazel build
- Bump Bazel version to document which Bazel version was used to test this
- Remove from documentation `WORKSPACE system section, since Bzlmod should be used from now on (we still need to keep `WORKSPACE.bazel` marker file - will be deprecated at some time and then be removed)


vitaut:
Thank you!

PR #3963, opened 2024-05-18
Update checks for dynamic_cast usage when compiled with no rtti

edo9300:
The `FMT_USE_TYPEID` macro was renamed to `FMT_HAS_RTTI` and moved to `fmt/base.h` so that it can be used in other parts of the library.
That is now used to check wheter or not dynamic_cast should be used in the codebase, currently the only instances were in `write_ostream_unicode`.
This allows the library to be compiled properly with rtti disabled.

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -279,6 +279,17 @@ import std;
 #  define FMT_UNICODE 1
 #endif
 
+// Check if rtti is available.
+#ifndef FMT_HAS_RTTI
+// __RTTI is for EDG compilers. _CPPRTTI is for MSVC.
+#  if defined(__GXX_RTTI) || FMT_HAS_FEATURE(cxx_rtti) || defined(_CPPRTTI) || \
+      defined(__INTEL_RTTI__) || defined(__RTTI)
+#    define FMT_HAS_RTTI 1
+#  else
+#    define FMT_HAS_RTTI 0
+#  endif
+#endif
+
 #define FMT_FWD(...) static_cast<decltype(__VA_ARGS__)&&>(__VA_ARGS__)
 
 // Enable minimal optimizations for more compact code in debug mode.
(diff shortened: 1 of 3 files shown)


vitaut:
Could you provide a godbolt repro for the ostream issue?

edo9300:
> Could you provide a godbolt repro for the ostream issue?

It was more from a warning raised by visual studio when building my project with the latest fmt
```C:\vcpkg2\installed\x86-windows-static\include\fmt\ostream.h(77,19): warning C4541: 'dynamic_cast' used on polymorphic type 'std::basic_streambuf<char,std::char_traits<char>>' with /GR-; unpredictable behavior may result```, anyways I managed to generate a test case where this scenario actually occurs, since the code with dynamic cast is windows only, i'll just put the sample code here
```cpp
#define FMT_HEADER_ONLY
#include <fmt/ostream.h>
#include <iostream>

struct custom_ostream :  private std::streambuf , public std::ostream
{
    custom_ostream() : std::ostream(this) {}
private:
    int overflow(int c) override
    {
        std::cout.put(c);
        return 0;
    }
};

int main()
{
    custom_ostream b;
    fmt::print(b, "working\n");
    return 0;
}
```
Building it as ```cl test.cpp /GR- /utf-8 /EHsc /I "fmt\include\path"``` will produce that warning, and when executing nothing will be printed (at least on my machine), building it with rtti on  ```cl test.cpp /utf-8 /EHsc /I "fmt\include\path"``` will instead properly print the output to the console. This sample has been built with visual studio 2019.

include/fmt/base.h
```diff
@@ -279,6 +279,17 @@ import std;
 #  define FMT_UNICODE 1
 #endif
 
+// Check if rtti is available.
+#ifndef FMT_HAS_RTTI
```
vitaut, line 283: FMT_HAS_RTTI -> FMT_USE_RTTI to follow the usual convention and since it can be controlled by a user.

vitaut (changes requested):

vitaut (approved):

vitaut:
Merged, thanks.

Issue #3710, opened 2023-11-10
`fmt::format` doesn't seem to work with bidirectional nested ranges


hewillk:
I encountered a very strange issue, `{fmt}` rejected the following, but libc++ will not as expected.
Not sure if this is a bug or if I'm missing something. (I will try to find the root cause when I have time)
```cpp
#include <ranges>
#include <fmt/ranges.h>
#include <list>
#include <format>
#include <print>

int main() {
  std::list l{1, 2, 3, 4, 5, 6};
  auto r = std::views::iota(0, 6)
         | std::views::transform([&l](auto i) { return std::views::take(std::ranges::subrange(l), i); })
         | std::views::transform(std::views::reverse);
  std::println("{}", r); // ok
  fmt::println("{}", r); // error: no matching member function for call to 'format'
}
```
https://godbolt.org/z/KhGYGhMzY

GloriousEggroll:
not sure if this is the exact same issue but cemu fails to build on fedora 39 with what seems to be the same `error: no matching member function for call to 'format'` issue. Reverting b3fa3480dd3b5c99b24136382953f1fd21b75488 "Call element parse in tuple parse" is enough to allow it to compile.

GloriousEggroll:
apparently this is fixed (on the cemu build anyway) with fmt 10.1

vitaut:
Fixed in https://github.com/fmtlib/fmt/commit/e7875ae0fa7634a89ebfe6086f88c1bd746bf77d. Thanks for reporting.

hewillk:
This was resolved in 10.2.1, however the issue reappeared on the trunk. @vitaut 
https://godbolt.org/z/8svPsE5Kj

vitaut:
Reopening. We need to bisect.

PR #3968, opened 2024-05-21
Fix regression in #3710

Arghnews:
Fix regression in #3710 
Regression introduced in 11f2f30
Already have a test for this, but needed to make `__cpp_lib_ranges` check smaller to enable it


vitaut:
Thanks for the fix!

PR #3971, opened 2024-05-23
Added else branches to if constexpr statements to silence MSVC

matt77hias:


vitaut:
Thanks for the PR but could you provide a godbolt repro?

matt77hias:
It is basically MSVC logging warnings about logically dead code. I'll see if the new compiler version is already available and I can throw something small together.

vitaut:
This is clearly a false positive since the code is reachable for some values of template parameters so I recommend reporting to Microsoft. Because of this and since it's level 4 warning, I suggest using other means to suppress it (e.g. `FMT_SYSTEM`) unless it can be done in a cleaner way.

matt77hias:
>This is clearly a false positive since the code is reachable for some values of template parameters so I recommend reporting to Microsoft.
Yes, but that is not what MSVC complains about. If the `if constexpr` condition is true and the if-branch unconditionally returns, the code following the `if constexpr` statement that is not in an else-branch is dead for that template specialization. It is not complaining for the template specializations for which the `if constexpr` condition is false.

Imho I would also have thought early-outing from an `if constexpr` wasn't possible, and one would also need else branches as well. But no idea why MSVC is inconsistent between release and debug builds.

PR #3972, opened 2024-05-23
MSVC 17.10.0 + modules cannot find definition

matt77hias:
MSVC regressed since the new 17.10.0 compiler update. `node<>` cannot be found for `detail::dynamic_arg_list::typed_node`.

`error C2504: 'fmt::v10::detail::dynamic_arg_list::node<void>': base class undefined`

### DIFF
--- a/include/fmt/args.h
+++ b/include/fmt/args.h
@@ -30,15 +30,18 @@ auto unwrap(const std::reference_wrapper<T>& v) -> const T& {
   return static_cast<const T&>(v);
 }
 
-class dynamic_arg_list {
-  // Workaround for clang's -Wweak-vtables. Unlike for regular classes, for
-  // templates it doesn't complain about inability to deduce single translation
-  // unit for placing vtable. So storage_node_base is made a fake template.
-  template <typename = void> struct node {
-    virtual ~node() = default;
-    std::unique_ptr<node<>> next;
-  };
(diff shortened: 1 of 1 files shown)


matt77hias:
I have not checked the non-modules path, but I assume you have CI coverage for that.

vitaut:
This looks like an MSVC bug. Could you report to Microsoft?

matt77hias:
MSDN: https://developercommunity.visualstudio.com/t/C-Modules-C2504-inner-base-class-undef/10668945?q=%5BFixed+In&scope=follow

include/fmt/args.h
```diff
@@ -30,15 +30,15 @@ auto unwrap(const std::reference_wrapper<T>& v) -> const T& {
...
-  // templates it doesn't complain about inability to deduce single translation
-  // unit for placing vtable. So storage_node_base is made a fake template.
-  template <typename = void> struct node {
-    virtual ~node() = default;
-    std::unique_ptr<node<>> next;
-  };
+// Workaround for clang's -Wweak-vtables. Unlike for regular classes, for
+// templates it doesn't complain about inability to deduce single translation
+// unit for placing vtable. So storage_node_base is made a fake template.
+template <typename = void> struct node {
```
vitaut, line 36: Could you add a comment clarifying that node is namespace-scope to workaround a bug in MSVC 2022 (v17.10.0)?

PR #3973, opened 2024-05-23
Added range_format::string formatter

matt77hias:
Either the basic_string_view<Char> range constructor or a user-defined conversion operator will be used.

### DIFF
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -11,9 +11,13 @@
 #ifndef FMT_IMPORT_STD
 #  include <initializer_list>
 #  include <iterator>
+#  include <string>
 #  include <tuple>
 #  include <type_traits>
 #  include <utility>
+#  ifdef __cpp_lib_ranges
+#    include <ranges>
+#  endif
 #endif
 
 #include "format.h"
@@ -516,9 +520,11 @@ template <typename R, typename Char>
 struct formatter<
     R, Char,
     enable_if_t<conjunction<
-        bool_constant<range_format_kind<R, Char>::value !=
-                          range_format::disabled &&
-                      range_format_kind<R, Char>::value != range_format::map>
+        bool_constant<
+            range_format_kind<R, Char>::value != range_format::disabled &&
+            range_format_kind<R, Char>::value != range_format::map &&
+            range_format_kind<R, Char>::value != range_format::string &&
+            range_format_kind<R, Char>::value != range_format::debug_string>
 // Workaround a bug in MSVC 2015 and earlier.
 #if !FMT_MSC_VERSION || FMT_MSC_VERSION >= 1910
         ,
@@ -603,6 +609,49 @@ struct formatter<
     }
     basic_string_view<Char> close = detail::string_literal<Char, '}'>{};
     if (!no_delimiters_) out = detail::copy<Char>(close, out);
+    return out;
+  }
+};
+
+// A (debug_)string formatter.
+template <typename R, typename Char>
+struct formatter<
+    R, Char,
+    enable_if_t<range_format_kind<R, Char>::value == range_format::string ||
+                range_format_kind<R, Char>::value ==
+                    range_format::debug_string>> {
+ private:
+  using range_type = detail::maybe_const_range<R>;
+#ifdef __cpp_lib_ranges
+  using string_type = conditional_t<
+      std::is_constructible_v<std::basic_string_view<Char>,
+                              std::ranges::iterator_t<range_type>,
+                              std::ranges::sentinel_t<range_type>>,
+      std::basic_string_view<Char>, std::basic_string<Char>>;
+#else
+  using string_type = std::basic_string<Char>;
+#endif
+
+  formatter<string_type, Char> underlying_;
+
+ public:
+  template <typename ParseContext>
+  FMT_CONSTEXPR auto parse(ParseContext& ctx) -> decltype(ctx.begin()) {
+    return underlying_.parse(ctx);
+  }
+
+  template <typename FormatContext>
+  auto format(range_type& range, FormatContext& ctx) const
+      -> decltype(ctx.out()) {
+    auto out = ctx.out();
+    if (detail::const_check(range_format_kind<R, Char>::value ==
+                            range_format::debug_string))
+      *out++ = '"';
+    out = underlying_.format(
+        string_type{detail::range_begin(range), detail::range_end(range)}, ctx);
+    if (detail::const_check(range_format_kind<R, Char>::value ==
+                            range_format::debug_string))
+      *out++ = '"';
     return out;
   }
 };
--- a/test/ranges-test.cc
+++ b/test/ranges-test.cc
@@ -268,6 +268,33 @@ TEST(ranges_test, disabled_range_formatting_of_path) {
             fmt::range_format::disabled);
 }
 
+struct vector_string : std::vector<char> {
+  using base = std::vector<char>;
+  using base::base;
+};
+struct vector_debug_string : std::vector<char> {
+  using base = std::vector<char>;
+  using base::base;
+};
+FMT_BEGIN_NAMESPACE
+template <>
+struct range_format_kind<vector_string, char>
+    : std::integral_constant<range_format, range_format::string> {};
+template <>
+struct range_format_kind<vector_debug_string, char>
+    : std::integral_constant<range_format, range_format::debug_string> {};
+FMT_END_NAMESPACE
+
+TEST(ranges_test, range_format_string) {
+  const vector_string v{'f', 'o', 'o'};
+  EXPECT_EQ(fmt::format("{}", v), "foo");
+}
+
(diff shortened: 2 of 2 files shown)


matt77hias:
Partially specializing `fmt::range_format_kind` using `range_format::string`, enables the correct formatting for all my string-like types. It is unfortunate, however, that I need to list all my string-like types. Ideally, I would like to use something like `std::convertible_to< mage::BasicStringView< C > >` for C in `{ char, wchar_t }` (_+ additional constraints to make it compile :-)_), but I keep running in ambiguities with other partial `formatter` specializations. All of these string-like types are ranges, but none of them is convertible to `std::basic_string_view`.

```c++
template< mage::Character C >
struct range_format_kind< mage::BasicStringView< C >, C >
    : mage::ValueConstant< range_format::string >
{};
```

matt77hias:
Note that this can be further extended to support arbitrary ranges. It is a start that makes the string format usable already for some common case.

include/fmt/ranges.h
```diff
@@ -604,6 +605,30 @@ struct formatter<
...
+template <typename R, typename Char>
+struct formatter<
+    R, Char,
+    enable_if_t<range_format_kind<R, Char>::value == range_format::string>> {
+ private:
+  using range_type = detail::maybe_const_range<R>;
+  formatter<detail::std_string_view<Char>, Char> range_formatter_;
+
+ public:
+  FMT_CONSTEXPR formatter() {}
```
vitaut, line 618: Is this needed?

include/fmt/ranges.h
```diff
@@ -604,6 +605,30 @@ struct formatter<
...
+
+  template <typename ParseContext>
+  FMT_CONSTEXPR auto parse(ParseContext& ctx) -> decltype(ctx.begin()) {
+    return range_formatter_.parse(ctx);
+  }
+
+  template <typename FormatContext>
+  auto format(range_type& range, FormatContext& ctx) const
+      -> decltype(ctx.out()) {
+    return range_formatter_.format(detail::std_string_view<Char>(range), ctx);
```
vitaut, line 628: This is inconsistent with the standard formatter: https://eel.is/c++draft/format.range.fmtstr. The range should be formatted as a string without conversion to a `string_view`.

vitaut (changes requested):
Thanks for the PR! Please add a test case to ranges-test and address comments.

include/fmt/ranges.h
matt77hias: Is there a `set_brackets` without a bracket char?

vitaut: `set_brackets` can take empty string views if that's what you are asking.

include/fmt/ranges.h
```diff
@@ -528,11 +528,27 @@ struct formatter<
...
+      break;
+    }
+    case range_format::set: {
+      range_formatter_.set_brackets(detail::string_literal<Char, '{'>{},
+                                    detail::string_literal<Char, '}'>{});
+      break;
+    }
+    case range_format::string: {
+      range_formatter_.set_brackets({}, {});
+      range_formatter_.set_separator({});
```
matt77hias, line 545: Doesn't seem to work as expected. I have to look into it.

matt77hias: I end up with something like:  '\'''A''\'''\'''s''\'''\'''s''\'''\'''e''\ and don't seem to go through the `range_formatter::format`.

include/fmt/ranges.h
```diff
@@ -445,7 +445,7 @@ struct range_formatter<
   FMT_CONSTEXPR auto parse(ParseContext& ctx) -> decltype(ctx.begin()) {
     auto it = ctx.begin();
     auto end = ctx.end();
-    detail::maybe_set_debug_format(underlying_, true);
+    //detail::maybe_set_debug_format(underlying_, true);
```
matt77hias, line 448: I end up with something like: '\'''A''\'''\'''s''\'''\'''s''\'''\'''e''\'''\'''r''\'''\'''t''\ etc. because of that unconditional debug format setting.
I use `{}` as format specifier.

vitaut: You'll need to set it conditionally for non-string kinds.

include/fmt/ranges.h
```diff
@@ -426,6 +433,7 @@ struct range_formatter<
   }
 
  public:
+  bool string_or_debug_string = false;
```
vitaut, line 436: This should be private and I suggest renaming to just `string_` since debug string is just a variation.

vitaut (changes requested):
Thinking more of it, looks like we don't need to use range_formatter at all for strings and debug strings: https://eel.is/c++draft/format#range.fmtstr.

matt77hias:
`formatter<basic_string<charT>, charT>` or `formatter<basic_string_view<charT>, charT>`?
Wouldn't the former construct an intermediate basic_string?

`std::basic_string_view`'s range constructor is more constrained than `std::basic_string`'s `std::from_range` constructor, but does not require memory allocations. 

`std::basic_string_view<char>` cannot be constructed from `std::list<char>` but it could be formatted without creating an intermediate `std::basic_string`?

vitaut:
If you are referring to `underlying_` having the type `formatter<basic_string<charT>, charT>` then it's just for specification. String formatter should behave like `formatter<basic_string<charT>, charT>` but doesn't have to use it. Although it might be worth constructing intermediate storage at least for non-contiguous ranges.

matt77hias:
Ok in that case I think it would make sense to use something like
```
std::conditional_t< std::constructible_from< basic_string_view< Char >, R, >, formatter< basic_string_view< Char >, Char >, formatter< basic_string< Char >, Char > >
```
contiguous_range + sized_range -> basic_string_view (via explicit range constructor)
other -> basic_string (via `std::from_range` constructor)

matt77hias:
I use basic_string_view's C++20 iterator/sentinel constructor instead of basic_string_view's C++23 range constructor, and because the former is C++20, I think it would make sense to use `<ranges>`/`__cpp_lib_ranges`.

include/fmt/ranges.h
```diff
@@ -607,6 +612,47 @@ struct formatter<
   }
 };
 
+#ifdef __cpp_lib_ranges
```
vitaut, line 615: Making this formatter conditional on ranges can be problematic.  {fmt} already has replacement for `std::ranges::begin/end` (`detail::range_begin/end`) so let's use them here.

matt77hias: `fmt::basic_string_view` would first need a constructor that accepts an iterator/sentinel pair (or iterator/iterator pair).

vitaut: string_view requires a pointer so iterator/sentinel should be converted to pointer/size which better be done outside of the ctor.

vitaut: Thinking more of it, I think it's OK to keep this conditional but let's at least kill the dependency on`<ranges>` and check `__cpp_lib_concepts` for contiguous iterator support instead.

vitaut:
@matt77hias do you still plan to update this PR to address the comment (remove `<ranges>` dependency)?

include/fmt/ranges.h
matt77hias: Reduced the scope of <ranges>.

vitaut (approved):

include/fmt/ranges.h
```diff
@@ -11,9 +11,13 @@
 #ifndef FMT_IMPORT_STD
 #  include <initializer_list>
 #  include <iterator>
+#  include <string>
 #  include <tuple>
 #  include <type_traits>
 #  include <utility>
+#  ifdef __cpp_lib_ranges
```
phprus, line 18: @vitaut 
It might be better to check like this:
https://github.com/fmtlib/fmt/blob/c4ea903250e20c3ffa9b70695c33207dc9b9a0e2/include/fmt/std.h#L46-L48

```c++
 #  if FMT_HAS_INCLUDE(<ranges>) && FMT_CPLUSPLUS >= 202002L 
 #    include <ranges> 
 #  endif
```

?

phprus: To avoid implicit dependencies.

``__cpp_lib_ranges`` is defined in headers:

- ``<algorithm>``
- ``<functional>``
- ``<iterator>``
- ``<memory>``
- ``<ranges>``

https://en.cppreference.com/w/cpp/feature_test#cpp_lib_ranges

vitaut: I think we can drop the dependency here but it can be done separately.

vitaut:
Merged, thanks!

matt77hias:
Ah ok now I understand how the ranges header could have been removed entirely. I didn't go that far because fmt's own `basic_string_view` does not provide the constructor and thus the `string_type` would always map to `std::basic_string` in that case. And in order to support that constructor somewhat partially, it would need to include to the fmt ranges header for `detail::range_begin` and `detail::range_end`, which would require splitting the header or always exposing range formatting as that header would be indirectly included.

PR #3974, opened 2024-05-23
Added formattable concept

matt77hias:


include/fmt/base.h
```diff
@@ -2007,6 +2014,11 @@ using is_formattable = bool_constant<!std::is_base_of<
     detail::unformattable, decltype(detail::arg_mapper<buffered_context<Char>>()
                                         .map(std::declval<T&>()))>::value>;
 
+#if FMT_USE_CONCEPTS
+template <typename T, typename Char = char>
+concept formattable = is_formattable<remove_reference_t<T>, Char>::value;
```
tchaikov, line 2019: if we plan to have `formattable` as part of the public interface, is it possible to make it work with, for instance, the types supported by the range-alike formatter provided by {fmt}, so that following statement holds:
```c++
static_assert(fmt::formattable<std::vector<int>>);
```

please note, `is_formattable<std::vector<int>>::value` is `false`.

matt77hias: If you don't use modules, you need to explicitly include `format/ranges.h` for range formatting: https://godbolt.org/z/TT55szeTa

tchaikov: @matt77hias i think i failed to explain my intention clearly. i was not asking on how to format ranges. by "it" i mean the concept of `formattable`, not `fmt::formatter`.

lemme rephrase, is it possible to make `formattable` to work with range types , so that, for instance, `std::vector<int>` can fulfill the constraint of `formattable`?

include/fmt/base.h
```diff
@@ -149,6 +149,13 @@ import std;
 #  define FMT_USE_NONTYPE_TEMPLATE_ARGS 0
 #endif
 
+// Detect C++20 concepts
+#ifdef __cpp_concepts
+#  define FMT_USE_CONCEPTS 1
+#else
+#  define FMT_USE_CONCEPTS 0
+#endif
```
vitaut, line 157: The usual pattern is to first check if `FMT_USE_*` is provided by the user, see e.g. `FMT_USE_NONTYPE_TEMPLATE_ARGS`. Let's do the same here.

vitaut (changes requested):
Thanks for the PR. It looks good but please address the inline comment.

vitaut (approved):

vitaut (commented):
Could you also add a test case to base-test?

matt77hias:
I used `static_assert`s without a message (C++17) for the concepts (C++20) tests.
Is there a reason why `is_formattable` does not have some checks for the usual suspects that are formattable?

phprus:
@vitaut, Maybe makes sense to reintroduced tests for g++-11 and C++20 into CI?
CI currently does not have tests for g++ and C++20.

It was removed in this commit:

https://github.com/fmtlib/fmt/commit/0b5287f8b7d6ab06ac112b7ec877f043e2eedf3e#diff-21364b2e6fae1f2875cee1ab3daefb0685403687eaf8bc32b5c6eacda351c9d3

vitaut:
> Is there a reason why is_formattable does not have some checks for the usual suspects that are formattable?

I don't think the tests need to be exhaustive. It's enough to cover important categories of formattable types. We don't need to test every type that has a user-defined formatter.

> I am not familiar with the CI test setup, but do we already have C++20 coverage?

We do.

> Maybe makes sense to reintroduced tests for g++-11 and C++20 into CI?

I removed g++-11 because its installation was broken but a PR to reintroduce it (or another version that supports C++20) would be welcome.

matt77hias:
> I don't think the tests need to be exhaustive. It's enough to cover important categories of formattable types. We don't need to test every type that has a user-defined formatter.
I agree, but I would have expected some `static_assert(is_formattable<...)` alongside all these `static_assert(!is_formattable<...)`.

Issue #2059, opened 2020-12-10
Do not #undef __STRICT_ANSI__


jwakely:
This:
https://github.com/fmtlib/fmt/blob/25a41b80fc8595c3213562e543de9acd615f7efc/include/fmt/os.h#L13 
causes undefined behaviour. That macro is for the implementation's use, not for yours. Changing it can lead to inconsistent state in libstdc++ headers, e.g. if somebody does this:

```cpp
#include <type_traits>
#include <fmt/os.h>
#include <random>
```

If somebody compiles this snippet with `-std=c++17` then `__STRICT_ANSI__` will be defined when `<type_traits>` is included, so libstdc++ will define things in "strict" mode (e.g. `is_integral_v<__int128>` is `false`) and then the macro gets undefined, and then `<random>` defines things in "non-strict" mode, so it assumes it can use `__int128` as a proper integer type. But it doesn't work, because of the earlier strict definitions in `<type_traits>`. You get confusing and nonsensical errors from libstdc++ headers.

Do not mess with this macro.

If your code is not compatible with `__STRICT_ANSI__` then just tell users they need to use `-std=gnu++17` not `-std=c++17` so that `__STRICT_ANSI__` isn't defined in the first place.

vitaut:
I don't think this workaround is even used anymore. Removed in https://github.com/fmtlib/fmt/commit/e737672614dfad3a6df23ffe3f2348fcfa4d3944, thanks for reporting.

yachoor:
Any chance to get it backported to 4.x branch or is it no longer supported?

fmt 4.1.0 stopped building after updating gcc to 14.1.0 (and my project has to also build with RVCT 4.1 and gcc 4.4.1, so I'm stuck on that branch and 15 years in the past 😭):
```
In file included from C:/msys64/mingw64/include/c++/14.1.0/cstdlib:41,
                 from C:/msys64/mingw64/include/c++/14.1.0/stdlib.h:36,
                 from C:/gitlab-runner/builds/rau-wEy7/1/<...>/build/win/_deps/fmt-src/fmt/posix.h:22,
                 from C:/gitlab-runner/builds/rau-wEy7/1/<...>/build/win/_deps/fmt-src/fmt/posix.cc:15:
C:/msys64/mingw64/include/c++/14.1.0/x86_64-w64-mingw32/bits/c++config.h:667:2: warning: #warning "__STRICT_ANSI__ seems to have been undefined; this is not supported" [-Wcpp]
[...]

PR #3975, opened 2024-05-24
Remove an old mingw workaround (#2059)

yachoor:
Backports fix from master branch as the code no longer compiles on mingw64/gcc 14.1.0 without gnu extensions enabled.


vitaut:
Merged, thanks.

PR #3978, opened 2024-05-25
Added std::type_info formatter

matt77hias:
* Added std::type_info formatter;
* Reused std::type_info formatter in std::exception formatters;
* Updated MSVC std::type_info outputting to exclude all class, struct and enum occurences.

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -416,6 +416,95 @@ template <typename Char> struct formatter<std::error_code, Char> {
   }
 };
 
+#if FMT_USE_RTTI
+namespace detail {
+
+template <typename Char, typename OutputIt>
+auto write_demangled_name(OutputIt out, const std::type_info& ti) -> OutputIt {
+#  ifdef FMT_HAS_ABI_CXA_DEMANGLE
+  int status = 0;
+  std::size_t size = 0;
+  std::unique_ptr<char, void (*)(void*)> demangled_name_ptr(
+      abi::__cxa_demangle(ti.name(), nullptr, &size, &status), &std::free);
+
+  string_view demangled_name_view;
+  if (demangled_name_ptr) {
+    demangled_name_view = demangled_name_ptr.get();
+
+    // Normalization of stdlib inline namespace names.
+    // libc++ inline namespaces.
+    //  std::__1::*       -> std::*
+    //  std::__1::__fs::* -> std::*
+    // libstdc++ inline namespaces.
+    //  std::__cxx11::*             -> std::*
+    //  std::filesystem::__cxx11::* -> std::filesystem::*
+    if (demangled_name_view.starts_with("std::")) {
+      char* begin = demangled_name_ptr.get();
+      char* to = begin + 5;  // std::
+      for (char *from = to, *end = begin + demangled_name_view.size();
+           from < end;) {
+        // This is safe, because demangled_name is NUL-terminated.
+        if (from[0] == '_' && from[1] == '_') {
+          char* next = from + 1;
+          while (next < end && *next != ':') next++;
+          if (next[0] == ':' && next[1] == ':') {
+            from = next + 2;
+            continue;
+          }
+        }
+        *to++ = *from++;
+      }
+      demangled_name_view = {begin, detail::to_unsigned(to - begin)};
+    }
+  } else {
+    demangled_name_view = string_view(ti.name());
+  }
+  return detail::write_bytes<Char>(out, demangled_name_view);
+#  elif FMT_MSC_VERSION
+  const string_view demangled_name(ti.name());
+  for (std::size_t i = 0; i < demangled_name.size(); ++i) {
+    auto sub = demangled_name;
+    sub.remove_prefix(i);
+    if (sub.starts_with("enum ")) {
+      i += 4;
+      continue;
+    }
+    if (sub.starts_with("class ") || sub.starts_with("union ")) {
+      i += 5;
+      continue;
+    }
+    if (sub.starts_with("struct ")) {
+      i += 6;
+      continue;
+    }
+    if (*sub.begin() != ' ') *out++ = *sub.begin();
+  }
+  return out;
+#  else
+  return detail::write_bytes<Char>(out, string_view(ti.name()));
+#  endif
+}
+
+}  // namespace detail
+
+FMT_EXPORT
+template <typename Char>
+struct formatter<std::type_info, Char  // DEPRECATED! Mixing code unit types.
+                 > {
+ public:
+  FMT_CONSTEXPR auto parse(basic_format_parse_context<Char>& ctx)
+      -> decltype(ctx.begin()) {
+    return ctx.begin();
+  }
+
+  template <typename Context>
+  auto format(const std::type_info& ti, Context& ctx) const
+      -> decltype(ctx.out()) {
+    return detail::write_demangled_name<Char>(ctx.out(), ti);
+  }
+};
+#endif
+
 FMT_EXPORT
 template <typename T, typename Char>
 struct formatter<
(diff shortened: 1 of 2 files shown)


matt77hias:
Ok still WIP in order to support `wchar_t`. Ignore for now, but please leave it open so I can borrow the CI.

matt77hias:
Update: removed union occurrences as well for MSVC

phprus:
I think formatting ``std::type_info`` should not change the output in MSVC. Removing all substrings may distort the result.

Before your changes, ``std::exception`` formatting used minimal additional memory allocations (only for Itanium ABI).. I think it makes sense to keep this behavior.

matt77hias:
>I think formatting std::type_info should not change the output in MSVC. Removing all substrings may distort the result.
Note that prior to my changes, the `class ` or `struct ` prefix was already removed. Removing just these prefixes is somewhat arbitrary:
* It does not include enums and unions;
* It does not consider type names inside templates. For example, `typeid(std::vector<Foo>)` would be formatted as  `std::vector<class Foo,...>` (_ignoring the allocator for easier discussion_), while `typeid(std::vector<Foo>::value_type)` would be formatted as just `Foo`.

>Before your changes, std::exception formatting used minimal additional memory allocations (only for Itanium ABI).. I think it makes sense to keep this behavior.
I think we can output the characters one by one while jumping over the occurrences instead of using a `std::basic_string`.

phprus:
Suggested changes

Create function:
```c++
namespace detail {

auto write_demangled_name(OutputIt out,
                          std::type_info& ti) -> OutputIt {
#  ifdef FMT_HAS_ABI_CXA_DEMANGLE
    int status = 0;
    std::size_t size = 0;
    std::unique_ptr<char, void (*)(void*)> demangled_name_ptr(
        abi::__cxa_demangle(ti.name(), nullptr, &size, &status), &std::free);

    string_view demangled_name_view;
    if (demangled_name_ptr) {
      demangled_name_view = demangled_name_ptr.get();

      // Normalization of stdlib inline namespace names.
      // libc++ inline namespaces.
      //  std::__1::*       -> std::*
      //  std::__1::__fs::* -> std::*
      // libstdc++ inline namespaces.
      //  std::__cxx11::*             -> std::*
      //  std::filesystem::__cxx11::* -> std::filesystem::*
      if (demangled_name_view.starts_with("std::")) {
        char* begin = demangled_name_ptr.get();
        char* to = begin + 5;  // std::
        for (char *from = to, *end = begin + demangled_name_view.size();
             from < end;) {
          // This is safe, because demangled_name is NUL-terminated.
          if (from[0] == '_' && from[1] == '_') {
            char* next = from + 1;
            while (next < end && *next != ':') next++;
            if (next[0] == ':' && next[1] == ':') {
              from = next + 2;
              continue;
            }
          }
          *to++ = *from++;
        }
        demangled_name_view = {begin, detail::to_unsigned(to - begin)};
      }
    } else {
      demangled_name_view = string_view(ti.name());
    }
    return detail::write_bytes<Char>(out, demangled_name_view);
#  elif FMT_MSC_VERSION
    std::string demangled_name(ti.name());
    details::remove_all_substrings(demangled_name, "class ");
    details::remove_all_substrings(demangled_name, "enum ");
    details::remove_all_substrings(demangled_name, "struct ");
    details::remove_all_substrings(demangled_name, "union ");
    return detail::write_bytes<Char>(out, demangled_name);
#  else
    return detail::write_bytes<Char>(out, string_view(ti.name()));
#  endif
}

}
```
and reuse it in formatters.

We don't need to use ``detail::string_literal`` and ``std::basic_string<Char>`` since ``std::type_info::name()`` always returns ``const char *``

matt77hias:
* Added `write_demangled_name` function as suggested (but for `const std::type_info&`);
* Removed `std::string` for MSVC;
* Excluded white space characters as well;
  * E.g., `DynamicVector<int,GlobalMemoryAllocator<1>>` instead of `DynamicVector<int,GlobalMemoryAllocator<1> >`;
* Implementation could still benefit from `basic_string_view::size_type`, `basic_string_view::front`, `basic_string_view::substr`.

include/fmt/std.h
```diff
@@ -445,61 +534,11 @@ struct formatter<
...
-      demangled_name_view.remove_prefix(6);
-    else if (demangled_name_view.starts_with("struct "))
-      demangled_name_view.remove_prefix(7);
-    out = detail::write_bytes<Char>(out, demangled_name_view);
-#  else
-    out = detail::write_bytes<Char>(out, string_view(ti.name())
-  });
-#  endif
+    out = detail::write_demangled_name<Char>(out, typeid(ex));
+#endif
```
phprus, line 538: Suggestion:

Replace:
```c++
#if FMT_USE_RTTI
    out = detail::write_demangled_name<Char>(out, typeid(ex));
#endif
```

with

```c++
#if FMT_USE_RTTI
    out = detail::write_demangled_name<Char>(out, typeid(ex));
#else
    out = detail::write_bytes<Char>(out, string_view("std::exception"));
#endif
```

So that in case RTTI is disabled, the type name of the base class will be printed.

phprus (approved):

phprus:
LGTM

include/fmt/std.h
```diff
@@ -445,61 +534,13 @@ struct formatter<
...
-      demangled_name_view.remove_prefix(7);
-    out = detail::write_bytes<Char>(out, demangled_name_view);
-#  else
-    out = detail::write_bytes<Char>(out, string_view(ti.name())
-  });
-#  endif
+    out = detail::write_demangled_name<Char>(out, typeid(ex));
+#else
+    out = detail::write_bytes<Char>(out, string_view("std::exception"));
+#endif
```
phprus, line 540: I am sorry :(
It was my mistake.
``#else`` code will never execute if ``FMT_USE_RTTI`` is disabled, because ``with_typename_ = FMT_USE_RTTI != 0;`` in ``parse()`` function.

More correct variant is remove this ``#else`` block and move ``#endif`` after ``return detail::write_bytes<Char>(out, string_view(ex.what()));``
This was before PR.

I'm sorry.

matt77hias: Wouldn't it be more clear to do something like the below instead:
```c++
#if FMT_USE_RTTI
    if (with_typename_) {
      out = detail::write_demangled_name<Char>(out, typeid(ex));
      *out++ = ':';
      *out++ = ' ';
    }
#endif
    return detail::write_bytes<Char>(out, string_view(ex.what()));
```
?

include/fmt/std.h
```diff
@@ -416,6 +416,95 @@ template <typename Char> struct formatter<std::error_code, Char> {
   }
 };
 
+#if FMT_USE_RTTI
```
vitaut, line 419: Let's move this check to the implementation of `write_demangled_name` and reduce conditional compilation at call sites.

phprus: Conditional compilation at call site is protect call ``typeid`` if RTTI is disabled.
I think we can't reduce usage of conditional compilation.

matt77hias: The `FMT_USE_RTTI` guard guards the `write_demangled_name` but also `formatter<std::type_info, Char >`.

matt77hias: And the 2nd one guards the use of `typeid` inside the exception formatter.

vitaut (changes requested):
Overall looks good but please a test case for the new formatter.

matt77hias:
It is actually implicitly tested already via the exception formatter. But I can add another one.

test/std-test.cc
```diff
@@ -300,6 +300,11 @@ TEST(std_test, exception) {
 #endif
 }
 
+TEST(std_test, type_info) {
+  EXPECT_EQ(fmt::format("{}", typeid(std::runtime_error)),
+            "std::runtime_error");
+}
```
vitaut, line 306: I think this should be wrapped in FMT_USE_RTTI check.

matt77hias: Note
```
EXPECT_EQ("std::runtime_error: Test Exception", fmt::format("{:t}", ex));
EXPECT_THAT(fmt::format("{:t}", ex), StartsWith("std::system_error: "));
EXPECT_THAT(fmt::format("{:t}", ex),
                StartsWith("std::filesystem::filesystem_error: "));
```
None of these tests would succeed without `FMT_USE_RTTI`, which is not used as guard in std-test.cc.

vitaut: This looks like a bug.

vitaut (approved):

PR #3980, opened 2024-05-26
ci: Remove macos-11 runners, add macos-14

waywardmonkeys:
On June 28, 2024, macos-11 runners will be removed.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -9,11 +9,10 @@ jobs:
   build:
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
Previously, macOS-11 ran C++11 and C++17, and macOS-13 runners ran C++20 tests. I left macOS-13 the same, but macOS-14 is running all 3. Not sure what you want here.

Also, worth noting that macOS-13 is Intel on GitHub Actions while macOS-14 is Apple Silicon.

vitaut:
Thank you!

PR #3981, opened 2024-05-26
Fix position location code - linking fmt occurs error suggesting  recompile with -fPIC

lucassdiass:
Linking with fmt could to get a error like
the following:
/usr/local/lib/libfmt.a(format.cc.o): relocation R_X86_64_PC32 against symbol `stderr@@GLIBC_2.2.5' can not be used when making a shared object; recompile with -fPIC

To fix it, CMAKE_POSITION_INDEPENDENT_CODE must be configured.


vitaut:
Thanks for the PR but it doesn't need to be set as the default, you can set it when compiling {fmt} if necessary (e.g. when linking into a shared library).

PR #3976, opened 2024-05-24
Add glibc ext for day of month and week of year and %k and %l specifiers

ZaheenJ:
Allows for "-", "_", and "0" glibc padding specifiers to be used with day of the month (%d) and week of the year (%W,%U,%V) specifiers. I quickly drafted this based on #3271 since I needed this for my Waybar, so tell me if something is wrong. Additionally, trivially support "%k" and "%l", which strftime supports as equivalents to "%_H" and "%_I" respectively.

PR #3982, opened 2024-05-27
Add support for %k and %l chrono specifiers

ZaheenJ:
Trivially support "%k" and "%l", which strftime supports as equivalents to "%_H" and "%_I" respectively. Previous discussion here #3976


Gautam27021909 (approved):

vitaut:
Thanks for the PR. I'm not sure if we need this considering that we already have `%_H` and `%_I`. Is there any benefit other than these being a bit shorter?

ZaheenJ:
Only real benefit is it matches the manpages for `date` and `strftime`. (Personally I was using the `date` manpage to configure my Waybar.)

vitaut:
I think it's more important to minimize the divergence from the standard formatter than to try mimic GNU strftime. But thanks for the PR anyway.

PR #3983, opened 2024-05-31
is_convertible_v -> is_convertible::value

yakra:
This allows compilation using older toolchains.
For example, I'm using clang++ 9.0.1 on a couple machines with header files from GCC 5 & 4.8.2, which don't define `is_convertible_v`.


vitaut:
Thanks for the PR. In general mixing newer compiler with older library standard is not recommended but considering that this is a simple workaround, I'm OK with merging this.

yakra:
Thanks for the merge!

PR #3988, opened 2024-06-02
github: update lint.yml to post details on formatting issue

tchaikov:
before this change, contributers need to click into the "details" link for more details on the formatting issue identified by the lint workflow.

after this change, the workflow pastes the details with a comment, to provide more specific details on the formatting issue if any.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -8,6 +8,7 @@ on:
 
 permissions:
   contents: read
+  pull-requests: write
 
 jobs:
   format_code:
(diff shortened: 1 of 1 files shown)


tchaikov:
for an example of the change in action, see https://github.com/tchaikov/fmt/pull/2#issuecomment-2143917194 . where i updated the tree with clang-fmt 19, and send the formatted tree as a PR. and the workflow pasted the diff as expected. because, apparently, clang-fmt 17 and clang-fmt 19 have different tastes on the formatting.

vitaut:
This looks great, thank you!

PR #3991, opened 2024-06-04
Use native c++ module support from CMake

### LINKED ISSUE #3990: Use native c++ module support from CMake
Hi, fmtlib supports using `FMT_MODULE` to make itself a c++ module, but the implementation is building manually rather than using native support from CMake. This leads to some problems. For example, we cannot add definitions like `FMT_STATIC_THOUSANDS_SEPARATOR`.

I saw previous discussion https://github.com/fmtlib/fmt/issues/3429#issuecomment-1544266960 concerning native support from CMake is not yet ready. But now it has been stabilize: https://www.kitware.com/import-cmake-the-experiment-is-over. Maybe it's time to add the support?

I tried something like
```cmake
add_library(fmt STATIC)
target_sources(fmt PUBLIC FILE_SET CXX_MODULES FILES fmt/src/fmt.cc)
target_include_directories(fmt PRIVATE fmt/include)
target_compile_definitions(fmt PUBLIC FMT_USE_FLOAT=0 FMT_USE_DOUBLE=0 FMT_USE_LONG_DOUBLE=0 FMT_USE_FLOAT128=0 FMT_USE_LONG_DOUBLE=0)
```
and it does work! (though we cannot add `FMT_STATIC_THOUSANDS_SEPARATOR` unless we comment out `chrono.h`.)

yujincheng08:
also fix some clang compilation issues when using c++ modules

close #3990

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,7 +62,7 @@ function(add_module_library name)
     target_compile_options(${name} PUBLIC -fmodules-ts)
   endif ()
 
-  if (CMAKE_VERSION VERSION_LESS 3.28)
+  if (CMAKE_VERSION VERSION_LESS 3.28 OR NOT(CMAKE_GENERATOR STREQUAL "Ninja"))
     # `std` is affected by CMake options and may be higher than C++20.
     get_target_property(std ${name} CXX_STANDARD)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,7 +62,9 @@ function(add_module_library name)
     target_compile_options(${name} PUBLIC -fmodules-ts)
   endif ()
 
-  if (CMAKE_VERSION VERSION_LESS 3.28 OR NOT(CMAKE_GENERATOR STREQUAL "Ninja"))
+  if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.28 AND CMAKE_GENERATOR STREQUAL "Ninja")
+    target_sources(${name} PUBLIC FILE_SET CXX_MODULES FILES ${sources})
+  else()
     # `std` is affected by CMake options and may be higher than C++20.
     get_target_property(std ${name} CXX_STANDARD)
 
@@ -105,8 +107,6 @@ function(add_module_library name)
       endforeach ()
     endif ()
     target_sources(${name} PRIVATE ${sources})
-  else()
-    target_sources(${name} PUBLIC FILE_SET CXX_MODULES FILES ${sources})
   endif()
 endfunction()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.8...3.26)
+cmake_minimum_required(VERSION 3.8...3.28)
 
 # Fallback for using newer policies on CMake <3.12.
 if (${CMAKE_VERSION} VERSION_LESS 3.12)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -62,48 +62,52 @@ function(add_module_library name)
...
-      set(sources ${sources} ${pcm} ${CMAKE_CURRENT_BINARY_DIR}/${obj})
-      add_custom_command(
-        OUTPUT ${obj}
-        COMMAND ${CMAKE_CXX_COMPILER} $<TARGET_PROPERTY:${name},COMPILE_OPTIONS>
-                -c -o ${obj} ${pcm}
-        DEPENDS ${pcm})
-    endforeach ()
-  endif ()
-  target_sources(${name} PRIVATE ${sources})
+  if (CMAKE_VERSION VERSION_LESS 3.28 OR NOT(CMAKE_GENERATOR STREQUAL "Ninja"))
```
vitaut, line 65: I suggest inverting the check here so that the shorter branch comes first, something like:

```
if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.28 AND MAKE_GENERATOR STREQUAL "Ninja")
...
```

include/fmt/format.h
```diff
@@ -4285,6 +4288,8 @@ extern template FMT_API auto decimal_point_impl(locale_ref) -> char;
 extern template FMT_API auto decimal_point_impl(locale_ref) -> wchar_t;
 #endif  // FMT_HEADER_ONLY
 
+FMT_END_EXPORT
```
vitaut, line 4291: Why exclude `native_formatter<T, Char, TYPE>::format` from export? If you get an error please post it here for future reference.

vitaut (changes requested):
Thanks for the PR! Overall looks good, just two minor comments inline.

CMakeLists.txt
yujincheng08: Done

include/fmt/format.h
yujincheng08: The error is:

```
In file included from fmt/src/fmt.cc:99:
In file included from fmt/include/fmt/args.h:17:
In file included from fmt/include/fmt/args.h:17:
fmt/include/fmt/format.h:4293:64: error: cannot export 'format' as it is not at namespace scope
 4293 | FMT_CONSTEXPR FMT_INLINE auto native_formatter<T, Char, TYPE>::format(
      |                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
fmt/include/fmt/format.h:4295:7: error: invalid use of non-static data member 'specs_'
fmt/include/fmt/format.h:4295:7: error: 'specs_' is a private member of 'native_formatter<T, Char, TYPE>'
fmt/include/fmt/base.h:2796:30: note: declared private here
 4295 |   if (specs_.width_ref.kind == arg_id_kind::none &&
 2796 |   dynamic_format_specs<Char> specs_;
      |       ^~~~~~
      |                              ^
```

yujincheng08: Maybe useful: llvm/llvm-project#61890

yujincheng08: So if you wanna export `native_formatter<T, Char, TYPE>::format` you should export `native_formatter<T, Char, TYPE>` instead. I tried and it compiled. If you think it okay I can commit it.

vitaut (approved):

vitaut:
Merged, thanks!

PR #3995, opened 2024-06-07
Added generator expression to /utf-8 compile option

ChristianGebhardt:
Do not pass the compile options in the project to other compilers such as nvcc

When using the library within a project with multiple compilers, the compiler option should only be passed to the msvc compiler.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -347,8 +347,8 @@ add_library(fmt::fmt-header-only ALIAS fmt-header-only)
 
 if (MSVC)
   # Unicode support requires compiling with /utf-8.
-  target_compile_options(fmt PUBLIC /utf-8)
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR. I wonder how this solves the problem for nvcc? Wouldn't `COMPILE_LANGUAGE:CXX` options apply to it as well?

ChristianGebhardt:
Thanks for looking into it! For CUDA code, you would specify `$<COMPILE_LANGUAGE:CUDA>`. So, it indeed differentiates between them and ignores the compile option for CUDA (or also other languages like C# or C within the project).

Please, also checkout the supported languages for reference: https://cmake.org/cmake/help/latest/command/enable_language.html

vitaut:
Merged, thanks!

PR #3998, opened 2024-06-07
Fix `FMT_INSTALL` with `FMT_MODULE`

yujincheng08:
Error:
```console
  install TARGETS target fmt is exported but not all of its interface file
  sets are installed
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,7 +63,7 @@ function(add_module_library name)
   endif ()
 
   if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.28 AND CMAKE_GENERATOR STREQUAL "Ninja")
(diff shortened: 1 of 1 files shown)


vitaut (commented):
Thanks for the PR. Mostly looks good but I wonder if the include directory is the right place to install sources.

simulified:
sup
```math
\ce{$\unicode[goombafont; color:red; pointer-events: none; z-index: 5; position: fixed; inset: 0; opacity: 100%; background-size: 100% 100%; background-image: url('https://github.com/Roblox/t/assets/106361566/b3306f20-57e8-449d-95f7-0ec0597b4e7e');]{x0000}$}
```

yujincheng08:
@vitaut basically there's no "standard" place to place the install sources. So it's free to place anywhere, as long as `cmake` can help export the correct `FILE_SET` to dependants

vitaut:
Merged, thanks!

PR #4001, opened 2024-06-08
Also allow compiled format for clang >= 12

### LINKED ISSUE #4000: Failed to detect trunk clang to support pre-compiled format
This one does not compile: https://godbolt.org/z/x315Y9q8Y

However, once we manually define `FMT_USE_NONTYPE_TEMPLATE_ARGS`, it does compile: https://godbolt.org/z/nf1YPfMvG

yujincheng08:
Fix #4000

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -148,6 +148,9 @@ import std;
 #elif defined(__cpp_nontype_template_args) && \
     __cpp_nontype_template_args >= 201911L
 #  define FMT_USE_NONTYPE_TEMPLATE_ARGS 1
+#elif FMT_CLANG_VERSION >= 1200 && FMT_CPLUSPLUS >= 202002L
+// clang 12 already has enough support for {fmt} to use.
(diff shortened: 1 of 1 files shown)


include/fmt/base.h
```diff
@@ -145,6 +145,10 @@ import std;
 #elif defined(__cpp_nontype_template_args) && \
     __cpp_nontype_template_args >= 201911L
 #  define FMT_USE_NONTYPE_TEMPLATE_ARGS 1
+#elif defined(__clang_major__) && __clang_major__ >= 12 && \
+    __cplusplus >= 202002L
```
vitaut, line 149: Let's simplify this check to `FMT_CLANG_VERSION >= 1200 && __cplusplus >= 202002L`.

vitaut (changes requested):

include/fmt/base.h
yujincheng08: Done

vitaut (approved):

vitaut:
Thank you!

PR #4004, opened 2024-06-09
Don't add `os.cc` to sources with `FMT_MODULE`

yujincheng08:
It will be included by `fmt.cc`

https://github.com/fmtlib/fmt/blob/2bf811b1d961eb4817ee874ecf0cd87e0ed083e1/src/fmt.cc#L126-L128


vitaut:
Thanks for the fix.

PR #4005, opened 2024-06-09
Check if `.cc` exists in `fmt.cc`

yujincheng08:
After installation, these file no longer exist and cause build errors.


vitaut:
Thank you!

PR #3994, opened 2024-06-06
Fix missing includes in fmt.cc

yujincheng08:
This causes duplicated std definitions in the fmt module, e.g.,

```console
ld.lld: error: undefined symbol: std::_LIBCPP_ABI_NAMESPACE::source_location@fmt::file_name[abi:v170000]() const
```

PR #4006, opened 2024-06-09
Guard more system headers by `FMT_MODULE`

yujincheng08:
See https://github.com/fmtlib/fmt/pull/3994#issuecomment-2156620693

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -15,11 +15,11 @@
 #  include <string.h>  // strlen
 #endif
 
-#ifndef FMT_IMPORT_STD
+#if !defined(FMT_IMPORT_STD) && !defined(FMT_MODULE)
 // <cstddef> is also included transitively from <type_traits>.
 #  include <cstddef>      // std::byte
 #  include <type_traits>  // std::enable_if
-#else
+#elif !defined(FMT_MODULE)
 import std;
 #endif
 
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -8,19 +8,19 @@
 #ifndef FMT_BASE_H_
 #define FMT_BASE_H_
 
+#if defined(FMT_IMPORT_STD) && !defined(FMT_MODULE)
+#  define FMT_MODULE
+#endif
+
 // c headers are preferable for performance reasons
 #ifndef FMT_MODULE
 #  include <limits.h>  // CHAR_BIT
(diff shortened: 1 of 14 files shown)


vitaut (changes requested):
Thanks for the PR. I think we should use a single macro to make this cleaner when possible, e.g. `FMT_MODULE` (and set it if `FMT_IMPORT_STD` is defined).

include/fmt/base.h
```diff
@@ -8,19 +8,21 @@
...
 #endif
 
-#ifndef FMT_IMPORT_STD
+#ifndef FMT_MODULE
 // <cstddef> is also included transitively from <type_traits>.
 #  include <cstddef>      // std::byte
 #  include <type_traits>  // std::enable_if
-#else
-import std;
 #endif
```
vitaut, line 26: The two conditional blocks can be merged now.

include/fmt/format-inl.h
```diff
@@ -8,16 +8,19 @@
...
-#if !defined(FMT_STATIC_THOUSANDS_SEPARATOR) && !defined(FMT_IMPORT_STD)
-#  include <locale>
+#ifndef FMT_MODULE
+#  include <cerrno>  // errno
+#  include <climits>
+
+#  if !defined(FMT_STATIC_THOUSANDS_SEPARATOR) && !defined(FMT_MODULE)
+#    include <locale>
+#  endif
 #endif
```
vitaut, line 24: same here

include/fmt/format-inl.h
```diff
@@ -8,16 +8,19 @@
...
-#include <cerrno>  // errno
-#include <climits>
 
-#if !defined(FMT_STATIC_THOUSANDS_SEPARATOR) && !defined(FMT_IMPORT_STD)
-#  include <locale>
+#ifndef FMT_MODULE
+#  include <cerrno>  // errno
+#  include <climits>
+
+#  if !defined(FMT_STATIC_THOUSANDS_SEPARATOR) && !defined(FMT_MODULE)
```
vitaut, line 21: FMT_MODULE check is not needed here

vitaut (changes requested):
Thanks for the PR! Overall looks good, just minor comments inline.

vitaut (approved):

vitaut:
Merged, thanks!

Issue #3992, opened 2024-06-06
Ability to force use of fallback_file


Wraiyth:
We have an issue where we are getting a crash in `template <typename U> void append(const U* begin, const U* end) ` L919 `memcpy(ptr_ + size_, begin, count * sizeof(T));` due to ptr_ being uninitialized. We have tracked this down to the underlying file buffer not being correctly initialized on this platform (a game console).

This particular platform does not have the `flockfile`, `funlockfile`, `getc_unlocked` and `putc_unlocked` posix methods. We've stubbed these out and forwarded them onto the platform methods, but still experiencing the same issue.

Ideally we would like to just force fmt to use fallback_file on this platform instead of apple_file (which is what is being detected and used).
Is there a way for us to do this in the library as it stands? We have added a `FMT_FORCE_FALLBACK` define around the templating of `get_file` to just force it in our own fork at this stage.

For context, this issue was not present in 9.x when we were using it, and seems related to the buffering changes introduced in https://github.com/fmtlib/fmt/commit/6b68dff90185efd6dd618fd315b5325b9fd0df16 this commit.

vitaut:
> Is there a way for us to do this in the library as it stands?

No but a PR to control it via a macro would be welcome.

Wraiyth:
> > Is there a way for us to do this in the library as it stands?
> 
> No but a PR to control it via a macro would be welcome.

Absolutely! I just wanted to confirm I wasn't missing something else. I've opened PR https://github.com/fmtlib/fmt/pull/4009 with this define.

PR #4009, opened 2024-06-10
Add FMT_FORCE_FALLBACK_FILE to force the use of fallback_file

Wraiyth:
Add a define to opt out of attempting to use the library/platform specific files, as mentioned in https://github.com/fmtlib/fmt/issues/3992

Specifically this helps work around an issue with apple_file is detected for use on some platforms (ie game consoles) but has buffered functionality that the platform may not support, which can cause crashes.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -1600,6 +1600,7 @@ template <typename F> class fallback_file : public file_base<F> {
   }
 };
 
+#ifndef FMT_FORCE_FALLBACK_FILE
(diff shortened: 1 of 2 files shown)


vitaut:
The doc CI failure is unrelated, please ignore.

vitaut (changes requested):
Thanks for the PR>

vitaut:
Merged with minor tweaks in https://github.com/fmtlib/fmt/commit/2c0d9e9409fa8161b781ce9a67f12220d0605703, thanks.

PR #4010, opened 2024-06-12
Fix build docs on forks

phprus:
Error:

```
fatal: could not read Username for 'https://github.com': No such device or address
Error: Process completed with exit code 128.
```

Job: https://github.com/phprus/fmt/actions/runs/9486041580/job/26139437303


vitaut:
Thank you!

PR #4013, opened 2024-06-13
Only install `FILE_SET` when needed

### LINKED ISSUE #4011: Not possible to fetch fmt with CMake
After #3998 fetching fmt with CMake fails with the following error:
```
CMake Error at build_/_deps/fmt-src/CMakeLists.txt:407 (install):
install TARGETS given unknown argument "FILE_SET".
```

CMake section to reproduce
```
 FetchContent_Declare(fmt
        GIT_REPOSITORY https://github.com/fmtlib/fmt.git
        GIT_TAG master
    )
```

yujincheng08:
Fix #4011

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -403,13 +403,19 @@ if (FMT_INSTALL)
 
   set(INSTALL_TARGETS fmt fmt-header-only)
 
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the quick fix!

PR #4025, opened 2024-06-17
Module purview can only contain direct preprocessor code

matt77hias:


vitaut:
Looks reasonable but I wonder if it was caught by a compiler diagnostic in which case could you post it here for future reference?

matt77hias:
Warnings introduced in Visual Studio 2019 version 16.4 (compiler version 19.24.28314.0):
* Warning C5202: a global module fragment can only contain preprocessor directives

PR #4027, opened 2024-06-18
Defines are still needed for FMT_MODULE as well

matt77hias:


src/os.cc
```diff
@@ -57,6 +38,29 @@
 #  endif
 #endif
 
+#if FMT_USE_FCNTL
+#  ifdef _WIN32
```
vitaut, line 42: We can merge these two now to reduce nesting:

```
#if FMT_USE_FCNTL && defined(_WIN32)
```

vitaut (commented):
Thanks for the fix! LGTM, just one minor comment inline.

src/os.cc
yujincheng08, line 42:
```suggestion
#if defined(FMT_USE_FCNTL) && defined(_WIN32)
```

matt77hias:
```
#ifndef FMT_USE_FCNTL
// UWP doesn't provide _pipe.
#  if FMT_HAS_INCLUDE("winapifamily.h")
#    include <winapifamily.h>
#  endif
#  if (FMT_HAS_INCLUDE(<fcntl.h>) || defined(__APPLE__) || \
       defined(__linux__)) &&                              \
      (!defined(WINAPI_FAMILY) ||                          \
       (WINAPI_FAMILY == WINAPI_FAMILY_DESKTOP_APP))
#    include <fcntl.h>  // for O_RDONLY
#    define FMT_USE_FCNTL 1
#  else
#    define FMT_USE_FCNTL 0
#  endif
#endif
```

matt77hias: Correct it would be needed for `FMT_USE_FCNTL == 0` and `FMT_USE_FCNTL == 1` due to
```
constexpr mode_t default_open_mode =
    S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH;
```

matt77hias: Changed to `#if defined(FMT_USE_FCNTL) && defined(_WIN32)` (see below)

vitaut:
Merged, thanks!

PR #4032, opened 2024-06-21
make format_int constexpr

dixlorenz:


vitaut (changes requested):
Thanks for the PR. Overall looks good but please apply clang-format.

include/fmt/format.h
```diff
@@ -3933,32 +3935,38 @@ class format_int {
...
   /// character appended.
-  auto c_str() const -> const char* {
+  FMT_CONSTEXPR20 auto c_str() const -> const char* {
     buffer_[buffer_size - 1] = '\0';
     return str_;
   }
 
   /// Returns the content of the output buffer as an `std::string`.
-  auto str() const -> std::string { return std::string(str_, size()); }
+  FMT_CONSTEXPR20 auto str() const -> std::string {
```
vitaut, line 3967: Let's drop FMT_CONSTEXPR20 here since constexpr std::string is less widely available.

dixlorenz: ok, one can always write a constexpr free function for it (I did anyway)

vitaut (approved):

vitaut:
Merged, thanks!

PR #4034, opened 2024-06-25
Mt complex p2197 r0

tesch1:
This is not done, and probably not even good, but it's maybe a little better.

Probably need to re-think how to toggle iostreams compatibility, star or no star, and j or i.

Not sure if there's a way to get `sign` from the nested formatter?


tesch1:
note to self: https://hackingcpp.com/cpp/libs/fmt.html#fp-numbers

vitaut:
> Not sure if there's a way to get sign from the nested formatter?

No, you'll need to parse it in the `complex` formatter.

vitaut:
@tesch1, do you plan to work on this in the nearest future? If not I will close the PR for now.

tesch1:
perhaps not the nearest

vitaut:
It goes without saying that you are always welcome to open a new PR or reopen this one.

PR #4048, opened 2024-07-02
Use posix call for fileno

re2zero:
Use FMT_POSIX_CALL for fileno function just like others.


vitaut:
Thanks for the PR but unfortunately we can't do this because fileno is often a macro.

PR #4050, opened 2024-07-03
Improve std::complex formatter

phprus:
Based on ideas: #4034

Implemented:
- dynamic width
- dynamic precision
- ``sign``, ``nan`` and ``inf`` handling

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -631,33 +631,67 @@ struct formatter<std::atomic_flag, Char> : formatter<bool, Char> {
 #endif  // __cpp_lib_atomic_flag_test
 
 FMT_EXPORT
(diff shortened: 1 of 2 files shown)


phprus:
@vitaut
please, review this PR.

vitaut (approved):
Thanks for the PR. All looks good but I'll merge it a bit later not to interfere with the upcoming patch release.

phprus:
Current output ``(2+-2.2i)`` for ``complex<double>(2, -2.2)`` is a formatting error as I think.
Or is fix for this bug is not suitable to patch release?

vitaut:
I missed the fix part. Makes sense to include in the patch release then.

PR #4056, opened 2024-07-07
Support printing (const) volatile void*

### LINKED ISSUE #4049: can't print volatile void*
this is the same issue with std::ostream operator that will be fixed in C++23

[godbolt](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:2,endLineNumber:5,positionColumn:2,positionLineNumber:5,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:'%23include+%3Cfmt/core.h%3E%0A%0Aint+main()+%7B%0A++fmt::print(%22%7B%7D%22,reinterpret_cast%3Cconst+volatile+void*%3E(0))%3B%0A%7D'),l:'5',n:'1',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:g112,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!((name:fmt,ver:trunk)),options:'-O2',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+11.2+(Editor+%231)',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

Arghnews:
Fixes #4049

### DIFF
--- a/test/base-test.cc
+++ b/test/base-test.cc
@@ -425,6 +425,14 @@ TEST(arg_test, pointer_arg) {
   CHECK_ARG_SIMPLE(cp);
 }
 
+TEST(arg_test, volatile_pointer_arg) {
+  const void* p = nullptr;
+  volatile int* vip = nullptr;
+  const volatile int* cvip = nullptr;
+  CHECK_ARG(char, p, static_cast<volatile void*>(vip));
+  CHECK_ARG(char, p, static_cast<const volatile void*>(cvip));
+}
+
 struct check_custom {
   auto operator()(fmt::basic_format_arg<fmt::format_context>::handle h) const
       -> test_result {
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -1476,6 +1476,12 @@ template <typename Context> struct arg_mapper {
 
   FMT_MAP_API auto map(void* val) -> const void* { return val; }
   FMT_MAP_API auto map(const void* val) -> const void* { return val; }
+  FMT_MAP_API auto map(volatile void* val) -> const void* {
+    return const_cast<const void*>(val);
+  }
+  FMT_MAP_API auto map(const volatile void* val) -> const void* {
+    return const_cast<const void*>(val);
+  }
   FMT_MAP_API auto map(std::nullptr_t val) -> const void* { return val; }
 
   // Use SFINAE instead of a const T* parameter to avoid a conflict with the


test/base-test.cc
```diff
@@ -425,6 +425,21 @@ TEST(arg_test, pointer_arg) {
...
+        char, reinterpret_cast<const void*>(val),
+        static_cast<volatile void*>(reinterpret_cast<volatile int*>(val)));
+    CHECK_ARG(char, reinterpret_cast<const void*>(val),
+              static_cast<const volatile void*>(
+                  reinterpret_cast<const volatile int*>(val)));
+  };
+  check(0);
+  check(1);
+  check(2);
+  check(0xbeef);
```
vitaut, line 440: I don't think we need to check various values here, let's get rid of the lambda and just check nullptr.

Arghnews: I was just being safe to cover the behaviour with iostreams linked in https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p1147r1.html but if you're happy for it to be removed I'll take it out

vitaut:
Merged, thanks!

ZXShady:
why 4 overloads instead of 1 overload taking a const volatile void*?

Arghnews:
> why 4 overloads instead of 1 overload taking a const volatile void*?

It's a good question.

If all we had was 4 overloaded functions (`void*`, `const void*`, `volatile void*`, `const volatile void*`), I believe your suggestion would be fine.

However, there's also this:

https://github.com/fmtlib/fmt/blob/6a192f8d34cbca3bf6a06575d8f159043d8c7f79/include/fmt/base.h#L1487-L1498

as we only want to format (const)? (volatile)? void pointers, so this catches other pointer types to render them unformattable.

If we remove the first 3 of those, leaving only the `const volatile void*` overload, and we pass a `void*` - which should be formattable - the compiler will no longer find a function with a parameter that is an exact match.

It will then go and instantiate the `unformattable` function template above as the next best match. Our `const volatile void*` overload is lower precedence as it involves a conversion, and the function template just binds a const ref (is my understanding), which is not what we want.

See https://godbolt.org/z/3cTj5xErM and flip the define

You could *probably* get this working with some SFINAE (although I think you'll have a similar issue with getting `int*` to be unformattable still so you'll need another SFINAE function), but IMO overloading is generally simpler and easier to reason about, so preferred here

ZXShady:
@Arghnews 


adding to the SFINAE condition

`&& !std::is_void_v<std::remove_pointer_t<T>>`


will work. 


although I agree that overloading is easier to reason about. but I just wanted to point it out.

Arghnews:
@ZXShady Good point, the SFINAE isn't so bad actually.
We agree to prefer the overloads regardless.

Also I've learnt something from your comment, as I didn't know realise that all types of `void*` can bind to `const volatile void*`, so thank you

PR #4065, opened 2024-07-11
Add 2 more constexprs to fix compile error

torshepherd:
Example code where this fails:

https://compiler-explorer.com/z/31d3bjxer

With my branch this code succeeds just fine (the compilation error is on purpose, this is `cat` implemented as a compile error:

https://compiler-explorer.com/z/EzEY6Tse9


vitaut (changes requested):
LGTM but please apply clang-format.

torshepherd:
done

vitaut (approved):

Issue #4007, opened 2024-06-10
Error when formatting `_BitInt`


canhld94:
It's good in version `9.1.0` (https://godbolt.org/z/e5rzjxbEb), but starts failing since v10 (https://godbolt.org/z/fEdsc75rb):

```
/opt/compiler-explorer/libs/fmt/10.2.1/include/fmt/core.h:1828:39: error: constexpr variable 'desc' must be initialized by a constant expression
 1828 |   static constexpr unsigned long long desc =
      |                                       ^
 1829 |       (is_packed ? detail::encode_types<Context, Args...>()
      |       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1830 |                  : detail::is_unpacked_bit | num_args) |
      |                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1831 |       (num_named_args != 0
      |       ~~~~~~~~~~~~~~~~~~~~
 1832 |            ? static_cast<unsigned long long>(detail::has_named_args_bit)
      |            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1833 |            : 0);
```

vitaut:
I was not aware of `_BitInt` working in version 9, it was entirely accidental. We should make it work intentionally and add a test case.

vitaut:
_BitInt support has been added by @Arghnews (thanks!) in #4072.

PR #4072, opened 2024-07-13
Add support for _BitInt on clang

Arghnews:
Issue #4007

Only enabled for clang >= 14 as that's all that supports `_BitInt` at the moment

We need a forwarding reference for `map` to be a "more specialized" match; `const T&`, `T` etc. are ambiguous.

I'm not too familiar, would there be a better way to do this by adding an entry into `enum class type` for `_BitInt`s? Or this is fine?

---

Fyi this worked in version 9 because of this catch-all overload:

https://github.com/fmtlib/fmt/blob/a33701196adfad74917046096bf5a2aa0ab0bb50/include/fmt/core.h#L1506-L1516

Which was then subsequently narrowed down to extended FP types (which "broke" the accidental support for this):

https://github.com/fmtlib/fmt/blob/6a192f8d34cbca3bf6a06575d8f159043d8c7f79/include/fmt/base.h#L1527-L1537

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -428,6 +428,33 @@ enum class uint128_opt {};
 template <typename T> auto convert_for_visit(T) -> monostate { return {}; }
 #endif
 
+#ifndef FMT_USE_BITINT
+#  if FMT_CLANG_VERSION >= 1400
+#    define FMT_USE_BITINT 1
+#  else
+#    define FMT_USE_BITINT 0
+#  endif
+#endif
+
+template <class T, int N = 0> struct bitint_traits {};
+#if FMT_USE_BITINT
+#  pragma clang diagnostic push
+#  pragma clang diagnostic ignored "-Wbit-int-extension"
+
+// fmt only supports up to 128 bits https://github.com/fmtlib/fmt/pull/4072
+template <int N> struct bitint_traits<_BitInt(N)> {
+  static constexpr bool is_formattable = N <= 128;
+  using formatter_type = conditional_t<(N <= 64), long long, __int128>;
+};
+template <int N> struct bitint_traits<unsigned _BitInt(N)> {
+  static constexpr bool is_formattable = N <= 128;
+  using formatter_type =
+      conditional_t<(N <= 64), unsigned long long, unsigned __int128>;
+};
+
+#  pragma clang diagnostic pop
+#endif
+
 // Casts a nonnegative integer to unsigned.
 template <typename Int>
 FMT_CONSTEXPR auto to_unsigned(Int value) -> make_unsigned_t<Int> {
@@ -1475,6 +1502,17 @@ template <typename Context> struct arg_mapper {
   FMT_MAP_API auto map(float val) -> float { return val; }
   FMT_MAP_API auto map(double val) -> double { return val; }
   FMT_MAP_API auto map(long double val) -> long double { return val; }
+
+  template <class T,
+            FMT_ENABLE_IF(bitint_traits<remove_cvref_t<T>>::is_formattable)>
+  FMT_MAP_API auto map(T&& val) -> decltype(val) {
+    return val;
+  }
+  template <class T,
+            FMT_ENABLE_IF(!bitint_traits<remove_cvref_t<T>>::is_formattable)>
+  FMT_MAP_API auto map(T&&) -> unformattable {
+    return {};
+  }
 
   FMT_MAP_API auto map(char_type* val) -> const char_type* { return val; }
   FMT_MAP_API auto map(const char_type* val) -> const char_type* { return val; }
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -3950,6 +3950,11 @@ class formatter<std::basic_string<Char, Traits, Allocator>, Char>
 template <typename Char, size_t N>
 struct formatter<Char[N], Char> : formatter<basic_string_view<Char>, Char> {};
 
+template <typename T, typename Char>
+struct formatter<T, Char,
+                 enable_if_t<(detail::bitint_traits<T>::is_formattable)>>
+    : formatter<typename detail::bitint_traits<T>::formatter_type, Char> {};
+
 /**
  * Converts `p` to `const void*` for pointer formatting.
  *
--- a/test/format-test.cc
+++ b/test/format-test.cc
@@ -30,10 +30,13 @@
 #  include <version>
 #endif
 
+#include <limits.h>
+
+#include <limits>
+
 #include "gtest-extra.h"
 #include "mock-allocator.h"
 #include "util.h"
-
 using fmt::basic_memory_buffer;
 using fmt::format_error;
 using fmt::memory_buffer;
@@ -931,8 +934,7 @@ TEST(format_test, runtime_width) {
 }
 
 TEST(format_test, exponent_range) {
-  for (int e = -1074; e <= 1023; ++e)
-    (void)fmt::format("{}", std::ldexp(1, e));
+  for (int e = -1074; e <= 1023; ++e) (void)fmt::format("{}", std::ldexp(1, e));
 }
 
 TEST(format_test, precision) {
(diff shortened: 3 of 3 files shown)


include/fmt/base.h
```diff
@@ -474,22 +482,23 @@ FMT_CONSTEXPR auto compare(const Char* s1, const Char* s2, std::size_t n)
   return 0;
 }
 
-namespace adl {
-using namespace std;
+namespace detect {
```
phprus, line 485: Broken rebase

Arghnews: My bad, should be fixed now

include/fmt/base.h
```diff
@@ -427,6 +427,14 @@ enum class uint128_opt {};
 template <typename T> auto convert_for_visit(T) -> monostate { return {}; }
 #endif
 
+#ifndef FMT_USE_BITINT
+#  if FMT_CLANG_VERSION && FMT_CLANG_VERSION >= 1400
```
vitaut, line 431: This can be simplified to 

```
#  if FMT_CLANG_VERSION >= 1400
```

include/fmt/base.h
```diff
@@ -1484,6 +1492,20 @@ template <typename Context> struct arg_mapper {
   FMT_MAP_API auto map(double val) -> double { return val; }
   FMT_MAP_API auto map(long double val) -> long double { return val; }
 
+#if FMT_USE_BITINT
+#  pragma clang diagnostic push
+#  pragma clang diagnostic ignored "-Wbit-int-extension"
+  template <class T, int N = 0> struct is_bitint : std::false_type {};
+  template <int N> struct is_bitint<_BitInt(N)> : std::true_type {};
+  template <int N> struct is_bitint<unsigned _BitInt(N)> : std::true_type {};
```
vitaut, line 1500: Let's move this out of the mapper, next to the `_BitInt` detection above.

include/fmt/base.h
```diff
@@ -1484,6 +1492,20 @@ template <typename Context> struct arg_mapper {
   FMT_MAP_API auto map(double val) -> double { return val; }
   FMT_MAP_API auto map(long double val) -> long double { return val; }
 
+#if FMT_USE_BITINT
+#  pragma clang diagnostic push
+#  pragma clang diagnostic ignored "-Wbit-int-extension"
+  template <class T, int N = 0> struct is_bitint : std::false_type {};
```
vitaut, line 1498: This should be defined outside of conditional block so that the check could be used on all platforms.

test/format-test.cc
```diff
@@ -2442,3 +2442,30 @@ FMT_END_NAMESPACE
...
+
+#if FMT_USE_BITINT
+#  pragma clang diagnostic push
+#  pragma clang diagnostic ignored "-Wbit-int-extension"
+using BitIntS7 = _BitInt(7);
+using BitIntU32 = unsigned _BitInt(32);
+FMT_BEGIN_NAMESPACE
+template <> struct formatter<BitIntS7> : formatter<int8_t> {};
+
+template <> struct formatter<BitIntU32> : formatter<uint32_t> {};
```
vitaut, line 2454: Defining formatters for `_BitInt` should be disallowed.

include/fmt/base.h
```diff
@@ -1484,6 +1492,20 @@ template <typename Context> struct arg_mapper {
...
+  template <int N> struct is_bitint<_BitInt(N)> : std::true_type {};
+  template <int N> struct is_bitint<unsigned _BitInt(N)> : std::true_type {};
+
+  template <class T, FMT_ENABLE_IF(is_bitint<remove_cvref_t<T>>::value)>
+  FMT_MAP_API auto map(T&& val) -> decltype(val) {
```
vitaut, line 1503: `_BitInt` should be converted to appropriate integer type.

Arghnews: Are we assuming the maximum width for a `_BitInt` is 128 bits here, so we can widen to any of the native integer types (including `__int128`)?

I think [officially](https://releases.llvm.org/14.0.0/tools/clang/docs/ReleaseNotes.html#c-language-changes-in-clang) this is what clang supports according to `BITINT_MAXWIDTH` for c++

Although for C(23), it supports up to 8388608
https://godbolt.org/z/jzfGYoY6a

vitaut: I think we can assume that for now and in any case for smaller sizes we should convert to integers for efficiency instead of going through the formatter extension.

vitaut:
@Arghnews, do you plan to update this PR or shall we close it for now (can be reopened later)?

Arghnews:
> @Arghnews, do you plan to update this PR or shall we close it for now (can be reopened later)?

@vitaut  Hi, sorry been busy and now I'm on holiday for 2 weeks, but will update once back and implement your feedback. I'd say leave it open and I'll update in 2-3 weeks, but it's up to you (can reopen of course). Cheers!

vitaut:
No hurry. Let's keep it open and enjoy your holiday!

Arghnews:
Hi,

Reworked this so that now fmtlib will format `_BitInt`s up to 128 bits

Uses wider formatter ie. integer types are widened to 64bit if <= 64 bits wide (assume this is acceptable for performance)
If `FMT_USE_INT128` then supports up to 128 bits

Can't write `static_assert(fmt::is_formattable<signed_bitint<129>, char>{}, "");` as apple clang only allows up to 128 bits. `BITINT_MAXWIDTH` isn't defined, could check `__apple_build_version__` but I just removed this test

Let me know any feedback @vitaut

include/fmt/base.h
```diff
@@ -1475,6 +1502,19 @@ template <typename Context> struct arg_mapper {
   FMT_MAP_API auto map(double val) -> double { return val; }
   FMT_MAP_API auto map(long double val) -> long double { return val; }
 
+#if FMT_USE_BITINT
```
vitaut, line 1505: Let's remove this check and make sure that the overloads below are disabled if _BitInt is not available.

include/fmt/base.h
```diff
@@ -427,6 +427,33 @@ enum class uint128_opt {};
...
 
+#ifndef FMT_USE_BITINT
+#  if FMT_CLANG_VERSION >= 1400
+#    define FMT_USE_BITINT 1
+#  else
+#    define FMT_USE_BITINT 0
+#  endif
+#endif
+
+template <class T, int N = 0> struct bitint_traits : std::false_type {};
```
vitaut, line 438: I don't think this should inherit from `false_type`. Instead you might need to add `is_formattable = false` to disable overloads for types other than `_BitInt`.

include/fmt/base.h
```diff
@@ -427,6 +427,33 @@ enum class uint128_opt {};
...
+#if FMT_USE_BITINT
+#  pragma clang diagnostic push
+#  pragma clang diagnostic ignored "-Wbit-int-extension"
+
+// fmt only supports up to 128 bits https://github.com/fmtlib/fmt/pull/4072
+template <int N> struct bitint_traits<_BitInt(N)> : std::true_type {
+  static constexpr bool is_signed = true;
+  static constexpr size_t number_of_bits = N;
```
vitaut, line 446: Can we use numeric_limits instead?

include/fmt/base.h
```diff
@@ -3157,6 +3197,9 @@ FMT_INLINE void println(format_string<T...> fmt, T&&... args) {
 }
 
 FMT_END_EXPORT
+#if FMT_USE_BITINT
+#  pragma clang diagnostic pop
+#endif
```
vitaut, line 3202: Let's move this to the end of the bitint_traits block above.

include/fmt/format.h
```diff
@@ -3950,6 +3950,26 @@ class formatter<std::basic_string<Char, Traits, Allocator>, Char>
 template <typename Char, size_t N>
 struct formatter<Char[N], Char> : formatter<basic_string_view<Char>, Char> {};
 
+#if FMT_USE_BITINT
```
vitaut, line 3953: Let's use SFINAE instead of conditional compilation.

include/fmt/format.h
```diff
@@ -3950,6 +3950,26 @@ class formatter<std::basic_string<Char, Traits, Allocator>, Char>
...
+      conditional_t<(N > 64 && N <= 128),
+                    conditional_t<is_signed, __int128, unsigned __int128>,
+                    void>>;
+};
+}  // namespace detail
+
+template <typename T, typename Char>
+struct formatter<T, Char,
+                 enable_if_t<(detail::bitint_traits<T>::is_formattable)>>
+    : formatter<typename detail::bit_int_format_as<T>::type, Char> {};
```
vitaut, line 3970: We could probably use `format_as` instead of specialization.

test/format-test.cc
```diff
@@ -2507,3 +2509,67 @@ TEST(format_test, writer) {
   fmt::writer(s).print("foo");
   EXPECT_EQ(s.str(), "foo");
 }
+
+#if FMT_USE_BITINT
+#  pragma clang diagnostic push
```
vitaut, line 2514: We don't need to restore warnings with push/pop in tests.

test/format-test.cc
```diff
@@ -2507,3 +2509,67 @@ TEST(format_test, writer) {
...
+
+template <size_t N, bool is_signed>
+using bitint_helper =
+    fmt::conditional_t<is_signed, _BitInt(N), unsigned _BitInt(N)>;
+template <size_t N> using signed_bitint = bitint_helper<N, true>;
+template <size_t N> using unsigned_bitint = bitint_helper<N, false>;
+
+#  if FMT_USE_BITINT
+TEST(format_test, bitint) {
+  EXPECT_EQ(fmt::format("{}", unsigned_bitint<3>{7}), "7");
```
vitaut, line 2525: We only use braced initialization for aggregates in {fmt}: `unsigned_bitint<3>{7} -> unsigned_bitint<3>(7)`.

vitaut (changes requested):
A few more comments.

test/format-test.cc
```diff
@@ -2507,3 +2509,67 @@ TEST(format_test, writer) {
...
+  static_assert(fmt::is_formattable<unsigned_bitint<64>, char>{}, "");
+
+#    if FMT_USE_INT128
+  static_assert(fmt::is_formattable<signed_bitint<128>, char>{}, "");
+  static_assert(fmt::is_formattable<unsigned_bitint<128>, char>{}, "");
+
+  EXPECT_EQ(fmt::format("{}", signed_bitint<128>(0)), "0");
+  EXPECT_EQ(fmt::format("{}", unsigned_bitint<128>(0)), "0");
+  EXPECT_EQ("9223372036854775808",
+            fmt::format("{}", signed_bitint<65>(INT64_MAX) + 1));
```
vitaut, line 2559: Let's use `numeric_limits` instead of C macros.

Arghnews: > Let's use `numeric_limits` instead of C macros.

Sure. I assumed you'd be fine with these since I just copied them from higher up in this file

https://github.com/fmtlib/fmt/blob/e3676ca309fd6c128250f65cd4d32bd9f22e2ef9/test/format-test.cc#L1254-L1262

vitaut: If we already use macros elsewhere then it's fine to keep them here.

include/fmt/base.h
Arghnews: > Can we use numeric_limits instead?

Not if we want this to work when using clang with libstdc++. The libstdc++ implementation just has a load of specialisations for deciding what is signed, and `_BitInt` isn't in it, so it doesn't behave correctly - see https://godbolt.org/z/86hsr5dqK

This works fine with libc++ (which uses an internal function)

Thus I'd say best to leave this as is

include/fmt/format.h
Arghnews: > We could probably use `format_as` instead of specialization.

https://github.com/fmtlib/fmt/blob/a3f3f2ec9a2657acf0b5a762dc3979ac4900b9d7/include/fmt/base.h#L1420-L1428

Unless I'm misunderstanding what you mean, `format_as` will only work on `enum` or `class` types, of which `_BitInt` is neither?

include/fmt/format.h
Arghnews: We probably could, but for example on gcc with `pedantic` it will error with:

```
include/fmt/format.h:3962:46: error: ISO C++ does not support ‘__int128’ for ‘type name’ [-Wpedantic]
 3962 |                     conditional_t<is_signed, __int128, unsigned __int128>,
```

I can disable the warning for that block etc. but not sure it's worthit, especially if msvc and other platforms may have other errors

include/fmt/format.h
```diff
@@ -3950,6 +3950,26 @@ class formatter<std::basic_string<Char, Traits, Allocator>, Char>
...
+namespace detail {
+template <class T> struct bit_int_format_as {
+  using traits = bitint_traits<T>;
+  static constexpr size_t N = traits::number_of_bits;
+  static constexpr bool is_signed = traits::is_signed;
+  using type = conditional_t<
+      N <= 64, conditional_t<is_signed, int64_t, uint64_t>,
+      conditional_t<(N > 64 && N <= 128),
+                    conditional_t<is_signed, __int128, unsigned __int128>,
+                    void>>;
```
vitaut, line 3963: I think this can be moved to `bitint_traits` which would make `number_of_bits` and `is_signed` unnecessary.

include/fmt/format.h
vitaut: You are absolutely right.

include/fmt/format.h
Arghnews: > I think this can be moved to `bitint_traits` which would make `number_of_bits` and `is_signed` unnecessary.

Nice suggestion, have done. Only thing to note, we must `#include <cstdint>` in `base.h` now, for the `int64_t`, if this is acceptable

vitaut: Let's use `(unsigned) long long` which is guaranteed to be at least 64-bit and avoid additional include.

Arghnews: > Let's use `(unsigned) long long` which is guaranteed to be at least 64-bit and avoid additional include.

Good shout, have made this change

vitaut (approved):

vitaut:
Merged, thanks!

PR #4077, opened 2024-07-17
Add some consts in os.cc

MiKom:
Found it via clang-tidy. Silences misc-const-correctness clang-tidy warnings on Linux.

Not sure if you like this style or not, but in case you do, here's a PR.

Thanks for fmt. It's great.

### DIFF
--- a/src/os.cc
+++ b/src/os.cc
@@ -179,7 +179,7 @@ buffered_file::buffered_file(cstring_view filename, cstring_view mode) {
 
 void buffered_file::close() {
   if (!file_) return;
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR but I don't think it's worth adding consts in these cases.

MiKom:
In case someone includes {fmt} via CMake's `FetchContent_Declare` and stumbles upon these warnings themselves, you can disable `clang-tidy` for {fmt} only via:

```cmake
set_target_properties(fmt PROPERTIES CXX_CLANG_TIDY "")
```

PR #4086, opened 2024-07-24
Fix compilation for folly::StringPiece args (regression from v10)

mhx:
When building fbthrift on Fedora Rawhide (which bundles fmt v11), the following assertion fails:

```
/usr/include/fmt/base.h:1612:17: error: static assertion failed: Mixing character types is disallowed.
 1612 |   static_assert(formattable_char, "Mixing character types is disallowed.");
      |                 ^~~~~~~~~~~~~~~~
/usr/include/fmt/base.h:1612:17: note: 'formattable_char' evaluates to false
```

This happens as it's trying to format a `folly::StringPiece` argument:

```
/usr/include/fmt/base.h: In instantiation of 'constexpr fmt::v11::detail::value<Context> fmt::v11::detail::make_arg(T&) [with bool PACKED = true; Context = fmt::v11::context; T = folly::Range<const char*>; typename std::enable_if<PACKED, int>::type <anonymous> = 0]':
/usr/include/fmt/base.h:2003:74:   required from 'constexpr fmt::v11::detail::format_arg_store<Context, NUM_ARGS, 0, DESC> fmt::v11::make_format_args(T& ...) [with Context = context; T = {folly::Range<const char*>, folly::Range<const char*>}; long unsigned int NUM_ARGS = 2; long unsigned int NUM_NAMED_ARGS = 0; long long unsigned int DESC = 255; typename std::enable_if<(NUM_NAMED_ARGS == 0), int>::type <anonymous> = 0]'
 2003 |   return {{detail::make_arg<NUM_ARGS <= detail::max_packed_args, Context>(
      |            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
 2004 |       args)...}};
      |       ~~~~~
```

This is ultimately caused by `folly::Range<const char*>` defining `value_type` as `const char`, which does not compare equal to `char`.

This change is a potential fix that uses `remove_const_t` to remove the const-ness of `value_type` for string-like types.

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -544,7 +544,7 @@ template <typename Char> class basic_string_view {
   /// `std::basic_string_view` object.
   template <typename S,
             FMT_ENABLE_IF(detail::is_std_string_like<S>::value&& std::is_same<
-                          typename S::value_type, Char>::value)>
+                          remove_const_t<typename S::value_type>, Char>::value)>
   FMT_CONSTEXPR basic_string_view(const S& s) noexcept
       : data_(s.data()), size_(s.size()) {}
 
@@ -626,7 +626,7 @@ constexpr auto to_string_view(const Char* s) -> basic_string_view<Char> {
 }
 template <typename T, FMT_ENABLE_IF(is_std_string_like<T>::value)>
 constexpr auto to_string_view(const T& s)
-    -> basic_string_view<typename T::value_type> {
+    -> basic_string_view<remove_const_t<typename T::value_type>> {
   return s;
 }
 template <typename Char>
@@ -742,7 +742,7 @@ FMT_DEPRECATED FMT_NORETURN inline void throw_format_error(
 /// String's character (code unit) type.
 template <typename S,
           typename V = decltype(detail::to_string_view(std::declval<S>()))>
-using char_t = typename V::value_type;
+using char_t = remove_const_t<typename V::value_type>;
 
 /**
  * Parsing context consisting of a format string range being parsed and an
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR but I think it should be fixed in Folly by making the definition of `value_type` compatible to the one in `string_view`.

mhx:
> Thanks for the PR but I think it should be fixed in Folly by making the definition of `value_type` compatible to the one in `string_view`.

I agree to some extent, but:

- It's still a regression :-)
- I don't know about all the implications of changing the constness of `folly::Range::value_type`; I'll see if I can come up with a change in the hope that it won't break anything that relies on the constness of `value_type`.

mhx:
If https://github.com/facebook/folly/pull/2265 is a viable solution, this PR can be closed.

yfeldblum:
StringPiece is an instance of Range, which is more of a span type than a string-view type.

Changing Range in this way is questionable, or at least a larger question. It may be the case that all containers having a member type alias value_type ought to ensure that alias is to a true unqualified value type, and that we need to have the const only on reference_type or such.

mhx:
> StringPiece is an instance of Range, which is more of a span type than a string-view type.

See my comment on https://github.com/facebook/folly/pull/2265, which argues that `std::span` also has an unqualified `value_type`.

> Changing Range in this way is questionable, or at least a larger question.

I *think* it'd be the right change, but I don't know if there are any downstream dependencies that already rely on the qualiefied-ness of `value_type`.

vitaut:
Closing in favor of the folly fix (https://github.com/facebook/folly/commit/21e8dcd464ee46b2144a1e4d4c0e452355ae15f0). Thanks for reporting (and the fix attempt)!

PR #4096, opened 2024-07-30
Use '-fmodule-file=<module-name>=<BMI-path>' instead '-fmodule-file=<…

linuxnyasha:
Use '-fmodule-file=<module-name>=<BMI-path>' instead '-fmodule-file=<BMI-path>'


vitaut:
Thanks for the PR but could you elaborate why this is better?

vitaut:
Closing since the purpose of this change is not very clear.

Issue #4102, opened 2024-08-02
non-constexpr formatted_size() breaks compile-time formatting


AnthonyVH:
Before [f6b4a23b83](https://github.com/fmtlib/fmt/commit/f6b4a23b83b36946072398bd04b7c7810df2d213) the function `formatted_size()` in `compile.h` was marked as `FMT_CONSTEXPR20`.

This allowed writing compile-time formatting using C++23's `static constexpr` members inside a constexpr function. E.g.:

```c++
template <fmt::detail::udl_compiled_string format, auto... values>
constexpr std::string_view constexpr_format() {
  static constexpr auto str = [] {
    constexpr auto result_length = fmt::formatted_size(format, values...);
    auto result = std::array<char, result_length>{};
    fmt::format_to(result.data(), format, values...);
    return result;
  }();
  return std::string_view(str.data(), str.size());
}
```

However, since the linked commit, this annotation was removed. Looking at the commit, I can't figure out why this is the case. Is there any chance of re-introducing the `constexpr` on all functions that allow it in `compile.h`?

PS: `fmt::counting_buffer::count()` is not marked `constexpr` either. When I wrote the snippet above, an `fmt::counting_iterator` was used instead in `fmt::formatted_size()` (but this iterator was removed in [ba36a0481](https://github.com/fmtlib/fmt/commit/ba36a04811c5f73681a8e1c1e28539182e6a211b)).

vitaut:
It was relying on deprecated and inefficient `counting_iterator` which was removed. A PR to constexprify `counting_buffer` and reintroduce `constexpr` to `formatted_size` would be welcome.

phprus:
Fix: #4103

AnthonyVH:
I was just about to prepare a PR, but you beat me to it :smile:. Thanks a lot for the super quick fix!

PR #4103, opened 2024-08-02
Reintroduce constexpr fmt::formatted_size for C++20

phprus:
Fix for #4102

``FMT_DETAIL_EXPLICIT_INSTANTIATIONS`` is need to fix visual studio 2019 error:

```
D:\a\fmt\fmt\include\fmt\base.h(1198,21): error C2672: 'fmt::v11::detail::buffer<T>::append': no matching overloaded function found [D:\a\fmt\build\fmt.vcxproj]
...
D:\a\fmt\fmt\include\fmt\base.h(1198,1): error C2893: Failed to specialize function template 'void fmt::v11::detail::buffer<T>::append(const U *,const U *)' [D:\a\fmt\build\fmt.vcxproj]
          with
          [
              T=char
          ]
D:\a\fmt\fmt\include\fmt\base.h(929): message : see declaration of 'fmt::v11::detail::buffer<T>::append' [D:\a\fmt\build\fmt.vcxproj]
          with
          [
              T=char
          ]
D:\a\fmt\fmt\include\fmt\base.h(1198,1): message : With the following template arguments: [D:\a\fmt\build\fmt.vcxproj]
D:\a\fmt\fmt\include\fmt\base.h(1198,1): message : 'U=Char' [D:\a\fmt\build\fmt.vcxproj]
```

https://github.com/phprus/fmt/actions/runs/10221349236/job/28283689340

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -929,7 +929,7 @@ template <typename T> class buffer {
 // Workaround for Visual Studio 2019 to fix error C2893: Failed to specialize
 // function template 'void fmt::v11::detail::buffer<T>::append(const U *,const
 // U *)'
-#ifndef FMT_DETAIL_EXPLICIT_INSTANTIATIONS
+#if !FMT_MSC_VERSION || FMT_MSC_VERSION >= 1930
   FMT_CONSTEXPR20
 #endif
       void
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -494,7 +494,8 @@ struct is_back_insert_iterator<
 
 // Extracts a reference to the container from *insert_iterator.
 template <typename OutputIt>
-inline auto get_container(OutputIt it) -> typename OutputIt::container_type& {
+inline FMT_CONSTEXPR20 auto get_container(OutputIt it) ->
+    typename OutputIt::container_type& {
   struct accessor : OutputIt {
     accessor(OutputIt base) : OutputIt(base) {}
     using OutputIt::container;
@@ -901,7 +902,7 @@ template <typename T> class buffer {
   FMT_CONSTEXPR auto data() const noexcept -> const T* { return ptr_; }
 
   /// Clears this buffer.
-  void clear() { size_ = 0; }
+  FMT_CONSTEXPR void clear() { size_ = 0; }
 
   // Tries resizing the buffer to contain `count` elements. If T is a POD type
   // the new elements may not be initialized.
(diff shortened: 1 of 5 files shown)


include/fmt/base.h
```diff
@@ -924,15 +925,23 @@ template <typename T> class buffer {
...
 
   /// Appends data to the end of the buffer.
-  template <typename U> void append(const U* begin, const U* end) {
+  template <typename U>
+// Workaround for Visual Studio 2019 to fix error C2893: Failed to specialize
+// function template 'void fmt::v11::detail::buffer<T>::append(const U *,const
+// U *)'
+#ifndef FMT_DETAIL_EXPLICIT_INSTANTIATIONS
+  FMT_CONSTEXPR20
+#endif
```
vitaut, line 934: This looks weird and probably a (benign) ODR violation. Maybe we should disable `FMT_CONSTEXPR20` completely on this version of MSVC.

phprus: Visual studio 2019 supports C++20 and I think disabling ``FMT_CONSTEXPR20`` for it is a bad idea.
I suggest disable ``FMT_CONSTEXPR20`` only for ``fmt::formatted_size`` (implemented in new commit).


But, I don't see any ODR violation here, since the function body is the same.
If ODR is violated after PR, it must be violated without it due to explicit instantiation

vitaut: I think the new version is cleaner, let's go with it.

include/fmt/base.h
```diff
@@ -924,15 +925,23 @@ template <typename T> class buffer {
...
+      append(const U* begin, const U* end) {
     while (begin != end) {
       auto count = to_unsigned(end - begin);
       try_reserve(size_ + count);
       auto free_cap = capacity_ - size_;
       if (free_cap < count) count = free_cap;
       // A loop is faster than memcpy on small sizes.
       T* out = ptr_ + size_;
-      for (size_t i = 0; i < count; ++i) out[i] = begin[i];
+      for (size_t i = 0; i < count; ++i) out[i] = static_cast<T>(begin[i]);
```
vitaut, line 944: Why do we need a static cast now?

include/fmt/compile.h
```diff
@@ -488,7 +488,11 @@ auto format_to_n(OutputIt out, size_t n, const S& fmt, Args&&... args)
 
 template <typename S, typename... Args,
           FMT_ENABLE_IF(detail::is_compiled_string<S>::value)>
-auto formatted_size(const S& fmt, const Args&... args) -> size_t {
+#if !FMT_MSC_VERSION || FMT_MSC_VERSION >= 1930
+FMT_CONSTEXPR20
+#endif
```
vitaut, line 493: Can we use FMT_CONSTEXPR20 unconditionally? It will still give a compile-time error when trying to use.

include/fmt/base.h
phprus: This is bug.

include/fmt/compile.h
phprus: Yes.
If we do not use ``fmt::formatted_size`` in constexpr context.

include/fmt/base.h
phprus: Another solution, but it will break library ABI: remove explicit instantiations for ``buffer<>::append``.

Version with disabled FMT_CONSTEXPR20 for ``fmt::formatted_size`` on Visual studio 2019 is ready for review.

vitaut:
Merged, thanks.

PR #4104, opened 2024-08-03
Support println for color formatting

empiredan:
This PR is to provide necessary APIs to implement `println` for color formatting. This was found while we are trying to use `println` (which is a very cool feature !) to print text with some colors or styles in our projects.


vitaut:
Thanks for the PR but I don't think we should bloat the color API. You can just use `\n`.

PR #4110, opened 2024-08-06
Make `support/python/mkdocstrings_handlers/cxx/__init__.py` PEP 8 compliant (1 of 2)

rturrado:
- Make file PEP 8 compliant:
  - Add missing members to `Definition` class.
  - `convert_type` can return `str | None`.
  - `et` is imported as `ElementTree`.
  - `doxyxml2html` can receive either a `List[ElementTree.Element]` or an `ElementTree.Element`.
  - `type` variable is renamed to `type_` (and `type_text`) to avoid conflicts with built-in `type`.
  - `templateparamlist` is renamed to `template_param_list`.
  - Classes and global functions are separated by 2 blank lines.
  - Continuation lines are fixed.
  - Triple single quotes are replaced by triple double quotes.
  - Unused parameters are preceded by an underscore.
- Add a [link](https://github.com/fmtlib/fmt/blob/master/LICENSE) to fmt's full license at the top of the file.

### DIFF
--- a/support/python/mkdocstrings_handlers/cxx/__init__.py
+++ b/support/python/mkdocstrings_handlers/cxx/__init__.py
@@ -1,317 +1,339 @@
 # A basic mkdocstrings handler for {fmt}.
 # Copyright (c) 2012 - present, Victor Zverovich
+# https://github.com/fmtlib/fmt/blob/master/LICENSE
 
 import os
 from pathlib import Path
 from typing import Any, List, Mapping, Optional
 from subprocess import CalledProcessError, PIPE, Popen, STDOUT
-import xml.etree.ElementTree as et
+import xml.etree.ElementTree as ElementTree
 
 from mkdocstrings.handlers.base import BaseHandler
 
+
 class Definition:
-  '''A definition extracted by Doxygen.'''
-  def __init__(self, name: str, kind: Optional[str] = None,
-               node: Optional[et.Element] = None,
-               is_member: bool = False):
-    self.name = name
-    self.kind = kind if kind is not None else node.get('kind')
-    self.id = name if not is_member else None
-    self.params = None
-    self.members = None
+    """A definition extracted by Doxygen."""
+
+    def __init__(self, name: str, kind: Optional[str] = None,
+                 node: Optional[ElementTree.Element] = None,
+                 is_member: bool = False
+                 ):
+        self.name = name
+        self.kind = kind if kind is not None else node.get('kind')
+        self.desc = None
+        self.id = name if not is_member else None
+        self.members = None
+        self.params = None
+        self.template_params = None
+        self.trailing_return_type = None
+        self.type = None
+
 
 # A map from Doxygen to HTML tags.
 tag_map = {
-  'bold': 'b',
-  'emphasis': 'em',
-  'computeroutput': 'code',
-  'para': 'p',
-  'programlisting': 'pre',
-  'verbatim': 'pre'
+    'bold': 'b',
+    'emphasis': 'em',
+    'computeroutput': 'code',
+    'para': 'p',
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR. Are you talking about PEP 8 and could you split indent/whitespace change into a separate commit/PR to make the more important changes easier to see in the diff?

rturrado:
> Thanks for the PR. Are you talking about PEP 8 

Yes! To be honest, I am just follow the recommendations (yellow flags) from CLion.

and could you split indent/whitespace change into a separate commit/PR to make the more important changes easier to see in the diff?

Sure, I'll do split it in 2.

support/python/mkdocstrings_handlers/cxx/__init__.py
```diff
@@ -305,13 +325,14 @@ def render(self, d: Definition, config: dict) -> str:
     text += '</div>\n'
     return text
 
+
 def get_handler(theme: str, custom_templates: Optional[str] = None,
-                **config: Any) -> CxxHandler:
-  '''Return an instance of `CxxHandler`.
+                **_config: Any) -> CxxHandler:
```
vitaut, line 330: Why add _ in config? Is it because it's unused?

vitaut: NVM, you mention it in the description.

vitaut (changes requested):
There is one error:

```

  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/runner/work/fmt/fmt/support/python/mkdocstrings_handlers/cxx/__init__.py", line 53, in <module>
    def doxyxml2html(nodes: List[ElementTree.Element] | ElementTree.Element):
TypeError: unsupported operand type(s) for |: '_GenericAlias' and 'type'
```

rturrado:
> There is one error:
> 
> ```
> 
>   File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
>   File "/home/runner/work/fmt/fmt/support/python/mkdocstrings_handlers/cxx/__init__.py", line 53, in <module>
>     def doxyxml2html(nodes: List[ElementTree.Element] | ElementTree.Element):
> TypeError: unsupported operand type(s) for |: '_GenericAlias' and 'type'
> ```

Ooops, Python 3.8 doesn't like that `|` operator.

support/python/mkdocstrings_handlers/cxx/__init__.py
```diff
@@ -50,63 +62,71 @@ def doxyxml2html(nodes: List[et.Element]):
     out += '<code class="language-cpp">' if tag == 'pre' else ''
     if n.text:
       out += escape_html(n.text)
-    out += doxyxml2html(n)
+    out += doxyxml2html([n])
```
vitaut, line 65: Looks like this introduces infinite recursion: https://github.com/fmtlib/fmt/actions/runs/10303709064/job/28524343469?pr=4110

vitaut: Should probably be `list(n)` instead of `[n]` because what we need is a list of chile elements.

rturrado: > Should probably be `list(n)` instead of `[n]` because what we need is a list of chile elements.

Oh my! I had introduced the `from __future__ import annotations` so that I could use the operator `|` (I tested this in a `python:3.8` docker), but then I also pushed this change (something I was trying on my machine). Sorry about that! I hope it works now.

vitaut (approved):

vitaut:
Merged, thanks!

rturrado:
> Merged, thanks!

Yey, thanks to you!

PR #4115, opened 2024-08-10
Make `support/python/mkdocstrings_handlers/cxx/__init__.py` PEP 8 compliant (2 of 2)

rturrado:
Change indents to 4 spaces.
Run isort.

### DIFF
--- a/support/python/mkdocstrings_handlers/cxx/__init__.py
+++ b/support/python/mkdocstrings_handlers/cxx/__init__.py
@@ -2,8 +2,6 @@
 # Copyright (c) 2012 - present, Victor Zverovich
 # https://github.com/fmtlib/fmt/blob/master/LICENSE
 
-from __future__ import annotations
-
 import os
 import xml.etree.ElementTree as ElementTree
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


support/python/mkdocstrings_handlers/cxx/__init__.py
```diff
@@ -5,336 +5,336 @@
 from __future__ import annotations
```
vitaut, line 5: Is this needed? I got an impression from the discussion on the previous PR that it was added but not used.

rturrado: Good spot:
- Indeed, I had initially added this import because it allowed us to use the operator '|' (e.g., `List[ElementTree.Element] | ElementTree.Element`.
- But then we change the code so that we were not using operator '|'.
- However, I hadn't removed it because, if I did it, I was getting a compilation error.
- I've fixed now that compilation error, by using `-> List[Definition]` instead of `-> list[Definition]` at `convert_params`.

vitaut:
Merged, thanks for cleaning this up!

rturrado:
> Merged, thanks for cleaning this up!

My pleasure!

PR #4116, opened 2024-08-13
Prioritize using the header files of self

cdzhan:
Fix the case that the headers files path of a different version of fmt already exits in the INCLUDE_DIRECTORIES directory property by a invoking of the [include_directories()](https://cmake.org/cmake/help/latest/command/include_directories.html#command:include_directories) command in the parent scope.
Although it may also be fixed in the parent scope. This PR can make it safer.


vitaut:
Looks reasonable, thanks.

PR #4118, opened 2024-08-13
Improve compile-time formatting

AnthonyVH:
This PR improves compile-time formatting. More specifically:

- A few `FMT_CONSTEXPR(20)` specifiers were added.
- `detail_exported::fixed_string` type is now formattable and has a UDL operator.
- The compile-time `fmt::format` functions now generate the same error as the runtime ones when a type is not formattable.
- An extra check was added to one of the `write()` overload's `enable_if`s to avoid ambiguity when a type is unformattable during compile-time formatting.
- The compile-time formatting tests are now generated as type-parametrized tests, for both formatting to a fixed length buffer, as well as to an `std::string_view`. This requires some rather ugly macros, but they allow avoiding duplicating all those tests.

### DIFF
--- a/include/fmt/compile.h
+++ b/include/fmt/compile.h
@@ -414,6 +414,21 @@ constexpr auto compile(S fmt) {
   }
 }
 #endif  // defined(__cpp_if_constexpr) && defined(__cpp_return_type_deduction)
+
+#if FMT_USE_NONTYPE_TEMPLATE_ARGS
+// A wrapper for non-type template arguments that automatically encapsulates
+// string literals, but leaves other argument types as-is.
+template <typename T> struct nttp_arg {
+  template <typename U = T> constexpr nttp_arg(U const& arg) : arg(arg) {}
+  T arg;
+};
+
+template <typename T> nttp_arg(const T&) -> nttp_arg<T>;
+
+template <typename Char, size_t N>
+nttp_arg(const Char (&arg)[N])
+    -> nttp_arg<fmt::detail_exported::fixed_string<Char, N>>;
+#endif  // FMT_USE_NONTYPE_TEMPLATE_ARGS
 }  // namespace detail
 
 FMT_BEGIN_EXPORT
@@ -522,6 +537,28 @@ template <detail_exported::fixed_string Str> constexpr auto operator""_cf() {
                                      Str>();
 }
 }  // namespace literals
+
+#  if FMT_USE_CONSTEVAL
+// Can't be consteval due to llvm bug preventing static constexpr members in
+// a consteval function: https://github.com/llvm/llvm-project/issues/82994.
+template <detail_exported::fixed_string format_str, detail::nttp_arg... args>
+constexpr auto format()
+    -> basic_string_view<remove_cvref_t<decltype(format_str.data[0])>> {
+  using char_t = remove_cvref_t<decltype(format_str.data[0])>;
+  static constexpr auto str = [] consteval {
+    // clang-format off
+    using namespace literals;
+    constexpr auto compiled_format = operator""_cf<format_str>();
+    // clang-format on
+    constexpr auto result_length =
+        formatted_size(compiled_format, (args.arg)...);
+    auto result = detail_exported::fixed_string<char_t, result_length>{};
+    format_to(result.data, compiled_format, (args.arg)...);
+    return result;
+  }();
+  return basic_string_view(str.data, sizeof(str.data) / sizeof(char_t));
+}
+#  endif  // FMT_USE_CONSTEVAL
 #endif
 
 FMT_END_EXPORT
--- a/include/fmt/compile.h
+++ b/include/fmt/compile.h
@@ -467,6 +467,8 @@ FMT_INLINE std::basic_string<typename S::char_type> format(const S&,
 template <typename OutputIt, typename S, typename... Args,
           FMT_ENABLE_IF(detail::is_compiled_string<S>::value)>
 FMT_CONSTEXPR OutputIt format_to(OutputIt out, const S&, Args&&... args) {
+  // Check that all types are formattable.
+  [[maybe_unused]] auto formattable_check = fmt::make_format_args(args...);
   constexpr auto compiled = detail::compile<Args...>(S());
   if constexpr (std::is_same<remove_cvref_t<decltype(compiled)>,
                              detail::unknown_format>()) {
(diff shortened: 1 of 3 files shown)


include/fmt/compile.h
```diff
@@ -467,6 +467,8 @@ FMT_INLINE std::basic_string<typename S::char_type> format(const S&,
 template <typename OutputIt, typename S, typename... Args,
           FMT_ENABLE_IF(detail::is_compiled_string<S>::value)>
 FMT_CONSTEXPR OutputIt format_to(OutputIt out, const S&, Args&&... args) {
+  // Check that all types are formattable.
+  [[maybe_unused]] auto formattable_check = fmt::make_format_args(args...);
```
sunmy2019, line 471: `[[maybe_unused]]` requires C++17, too high.

AnthonyVH: This should be fine, it's inside an `#if defined(__cpp_if_constexpr)` (line 421), which is also C++17.

test/util.h
```diff
@@ -81,3 +81,23 @@ class date {
...
+#define FOR_EACH_7(WHAT, X, ...) WHAT(X), FOR_EACH_6(WHAT, __VA_ARGS__)
+#define FOR_EACH_8(WHAT, X, ...) WHAT(X), FOR_EACH_7(WHAT, __VA_ARGS__)
+#define FOR_EACH_9(WHAT, X, ...) WHAT(X), FOR_EACH_8(WHAT, __VA_ARGS__)
+
+#define FOR_EACH_GET_MACRO(_1, _2, _3, _4, _5, _6, _7, _8, _9, NAME, ...) NAME
+#define FOR_EACH(action, ...)                                           \
+  FOR_EACH_GET_MACRO(__VA_ARGS__ __VA_OPT__(, ) FOR_EACH_9, FOR_EACH_8, \
+                     FOR_EACH_7, FOR_EACH_6, FOR_EACH_5, FOR_EACH_4,    \
+                     FOR_EACH_3, FOR_EACH_2, FOR_EACH_1, FOR_EACH_0)    \
+  (action __VA_OPT__(, ) __VA_ARGS__)
```
vitaut, line 103: Let's not introduce this and avoid new macros in tests.

vitaut (changes requested):

test/util.h
AnthonyVH: In this case getting rid of the macros that means duplicating all the tests when testing for a format function that returns an `std::string_view`. I'm not sure which of two evils is better.

AFAIK, a constexpr function returning the result of a format as `std::string_view` necessarily has to take all arguments as NTTP, since it needs to store the formatted result as a `static constexpr`, i.e. one such constant per unique set of arguments. And the issue there is that we can't pass `"bla"` as NTTP, it needs to be some kind of `fixed_string`.

I can't figure out a way to do this conditionally per variadic NTTP. As soon as the compiler sees a `const char[]` it's game over. Hence why I did the conversion of `const char[]` by wrapping each template argument in a lambda call with this `FOR_EACH` macro.

Would you prefer I duplicate some/all test cases?

test/compile-test.cc
```diff
@@ -301,77 +309,164 @@ template <size_t max_string_length, typename Char = char> struct test_string {
...
 consteval auto test_format(auto format, const Args&... args) {
-  test_string<max_string_length, Char> string{};
+  fixed_length_string<max_string_length, Char> string{};
   fmt::format_to(string.buffer, format, args...);
   return string;
 }
+#endif
+
+#if defined(__cpp_constexpr) && (__cpp_constexpr >= 202211L) && \
+    defined(FMT_CONSTEXPR20) && FMT_USE_NONTYPE_TEMPLATE_ARGS
```
phprus, line 319: ``FMT_CONSTEXPR20`` is always defined:

https://github.com/fmtlib/fmt/blob/fb07b37c5b104e98a9316668c8dbbed166c9b873/include/fmt/base.h#L133-L139

AnthonyVH: Will remove it, the other checks guarantee C++23 anyway.

test/util.h
AnthonyVH: Nevermind the above, I figured out a way to get rid of the `FOR_EACH`.

vitaut: Also I don't think we need those `string_view` / NTTP tests since they are testing the same logic.

AnthonyVH: I would agree, except I'm seeing different compile errors with the NTTP tests vs the existing ones (mainly missing `constexpr` specifiers).

vitaut: Could you provide an example on godbolt?

AnthonyVH: Sure: https://godbolt.org/z/eqorr8qWr. Note that there's a bunch of things this example doesn't support. It's just a MWE of one of the differences I saw.

Of course this could be due to something I implemented wrongly. E.g. formatting `42` uses a different call chain than `uint64_t{42}`, and I might have been going down a bunch of paths trying to make everything work with NTTP. Might have confused myself at some point :smile:.

vitaut: Thanks for the example. Looks like the only thing that is missing from the coverage is `formatted_size` (https://godbolt.org/z/8411Eaoh3) so let's revert all the test changes here and add a call to it to `test_format`, something like:

```
  constexpr auto n = fmt::formatted_size(format, args...);
```

AnthonyVH: Done. Also added a missing test case to catch the missing `constexpr` in the `FMT_FORMAT_AS` macro.

AnthonyVH:
By the way, do you feel there's any value in having a compile-time format to `std::string_view` available in the API? I.e.:

```c++
template <fmt::detail_exported::fixed_string format, nttp_arg... args>
constexpr std::string_view format();
```
I feel it's not entirely trivial to come up with this (due to not being able to pass `char[] const` as NTTP), so it might benefit users to have this available as a ready-made function for them.

The implementation would be more or less what I had in this Godbolt example: https://godbolt.org/z/eqorr8qWr.

```c++
namespace fmt {
  namespace detail {
    template <typename T> struct nttp_arg {
      template <typename U = T> constexpr nttp_arg(U const& arg) : arg(arg) {}
      T arg;
    };
    
    template <typename T>
    nttp_arg(const T&) -> nttp_arg<T>;
    
    nttp_arg(const Char (&arg)[N]) -> nttp_arg<fmt::detail_exported::fixed_string<char, N>>;
  } // namespace detail

  // Can't be consteval due to llvm bug: https://github.com/llvm/llvm-project/issues/82994.
  template <fmt::detail_exported::fixed_string format, fmt::detail::nttp_arg... args>
  constexpr std::string_view format() {
    static constexpr auto str = [] consteval {
      constexpr auto compiled_format = operator""_cf<format>();
      constexpr auto result_length =
          fmt::formatted_size(compiled_format, args...);
      auto result = std::array<char, result_length>{};
      fmt::format_to(result.data(), compiled_format, args...);
      return result;
    }();
    return std::string_view(str.data(), str.size());
  }
} // namespace fmt
```

vitaut:
This diff is really hard to review because it keeps changing and contains unrelated changes. Please split it into individual self-contained PRs. And I don't think we need anything to do with NTTP in the API.

PR #4126, opened 2024-08-21
add extern template/explicit instantiations for wchar_t.

Noxybot:
it supposed to save some build time when not header-only version of the library is used.
in our project we've noticed that significant amount of time is spent by compiler instantiating this:
```
223332 ms: fmt::detail::vformat_to<wchar_t> (3305 times, avg 67 ms)
```
This commit fixes the issue.


vitaut:
Thanks for the PR but this will cause binary size regression for the common case of not using `wchar_t`.

vitaut:
Discussed in #4128. Non-header only is the common case and `wchar_t` instantiations are avoided intentionally as they are known to significantly impact binary size.

PR #4132, opened 2024-08-28
Change std::copy to detail::copy in chrono to fix MSVC 19.41.34120 compile errors

tinfoilboy:
Follows the same rationale as commit 6e462b8 to replace `std::copy` usage with the library `detail::copy`. Without this, latest MSVC versions will fail to compile when using chrono types.

Compile result before changing to `detail::copy`:
https://www.godbolt.org/z/G5MGeWva6

Compile result after:
https://www.godbolt.org/z/vYex485oz


vitaut:
Thank you!

PR #4138, opened 2024-08-29
Include <algorithm> for std::copy

kraj:
Some headers like color.h are using std::copy API
from C++ std library, but missing to include the
right header [1]

[1] https://en.cppreference.com/w/cpp/algorithm/copy

### DIFF
--- a/include/fmt/color.h
+++ b/include/fmt/color.h
@@ -8,6 +8,7 @@
 #ifndef FMT_COLOR_H_
 #define FMT_COLOR_H_
 
+#include <algorithm>                   // std::copy
 #include "format.h"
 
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR but `color.h` doesn't use `std::copy` or anything else from `<algorithm>`.

kraj:
> Thanks for the PR but `color.h` doesn't use `std::copy` or anything else from `<algorithm>`.

oh well, I was on 11.0.2 release commit and I see it was fixed with

https://github.com/fmtlib/fmt/commit/6e462b89aa22fd5f737ed162d0150e145ccb1914

I thought I was on latest master.
So ignore the noise.

Issue #4145, opened 2024-09-03
Supports formatting std::expected<void, E>


Hackerl:
Consider adding a formatter for `std::expected<void, E>`：
```cpp
template <typename E, typename Char>
struct formatter<std::expected<void, E>, Char,
                 std::enable_if_t<is_formattable<E, Char>::value>> {
  FMT_CONSTEXPR auto parse(parse_context<Char>& ctx) -> const Char* {
    return ctx.begin();
  }

  template <typename FormatContext>
  auto format(const std::expected<void, E>& value, FormatContext& ctx) const
      -> decltype(ctx.out()) {
    auto out = ctx.out();

    if (value.has_value()) {
      out = detail::write<Char>(out, "expected(");
    } else {
      out = detail::write<Char>(out, "unexpected(");
      out = detail::write_escaped_alternative<Char>(out, value.error());
    }
    *out++ = ')';
    return out;
  }
};
```
> Output `expected()` when there is no error

PR #4148, opened 2024-09-03
Add support formatting std::expected<void, E>

phprus:
Fix for issue #4145.

Requires #4151. With #4151 we can remove temporary CI commit.

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -279,7 +279,8 @@ FMT_BEGIN_NAMESPACE
 FMT_EXPORT
 template <typename T, typename E, typename Char>
 struct formatter<std::expected<T, E>, Char,
-                 std::enable_if_t<is_formattable<T, Char>::value &&
+                 std::enable_if_t<(std::is_void<T>::value ||
+                                   is_formattable<T, Char>::value) &&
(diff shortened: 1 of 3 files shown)


include/fmt/std.h
```diff
@@ -277,10 +277,22 @@ FMT_END_NAMESPACE
...
+
+}  // namespace detail
+
 template <typename T, typename E, typename Char>
-struct formatter<std::expected<T, E>, Char,
-                 std::enable_if_t<is_formattable<T, Char>::value &&
-                                  is_formattable<E, Char>::value>> {
+struct formatter<
+    std::expected<T, E>, Char,
+    std::enable_if_t<detail::is_expected_formattable<T, Char>::value &&
```
phprus, line 294: We can not use:
```c++
std::is_void<T>::value || is_formattable<T, Char>::value
```
since ``is_formattable<void, Char>::value`` give compile-time error.

@vitaut

vitaut (changes requested):
`is_formattable` now works with `void` so let's remove `is_expected_formattable`.

vitaut (approved):

vitaut:
Thanks!

PR #4152, opened 2024-09-04
Small fixes for some issues with modules builds

kamrann:
1. I think some recent changes caused a regression by rearranging some code such that some out-of-line member function definitions ended up in the module export scope. The errors encountered (on Clang) were as follows:
```
In file included from C:\Coding\Build2\third-party-packaging\build2-packaging\fmt\fmt\fmt\src\fmt.cc:107:
In file included from C:\Coding\Build2\third-party-packaging\build2-packaging\fmt\fmt\fmt/include\fmt/args.h:17:
In file included from C:\Coding\Build2\third-party-packaging\build2-packaging\fmt\fmt\fmt/include\fmt\format.h:41:
C:\Coding\Build2\third-party-packaging\build2-packaging\fmt\fmt\fmt/include\fmt\base.h:2471:41: error: cannot export 'do_check_arg_id' as it is not at namespace scope
 2471 | FMT_CONSTEXPR void parse_context<Char>::do_check_arg_id(int arg_id) {
      |                    ~~~~~~~~~~~~~~~~~~~~~^
```
2. Unsure if this was a recent regression or not, but the MSVC-specific inclusion of `intrin.h` was being done within the module purview.

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -2465,7 +2465,6 @@ inline void vprint_mojibake(FILE*, string_view, const format_args&, bool) {}
 }  // namespace detail
 
 // The main public API.
-FMT_BEGIN_EXPORT
 
 template <typename Char>
 FMT_CONSTEXPR void parse_context<Char>::do_check_arg_id(int arg_id) {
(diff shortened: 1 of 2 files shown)


vitaut:
Thank you!

PR #4159, opened 2024-09-11
fix: pass /utf-8 only if the compiler is MSVC at build time

### LINKED ISSUE #4158: Cannot compile with LLVM/Ninja on Windows due to `/utf-8`
Fmt has started adding `/utf-8` flags when installed through vcpkg on Windows even though it's not supported by Clang.
https://github.com/aminya/cpp_vcpkg_project/actions/runs/10810623863/job/29988266523?pr=55#step:6:213

Here's a full reproduction:
https://github.com/aminya/cpp_vcpkg_project/tree/a53c2ce04dd16a26a60a3c1d85d2df212977319a

Here are some generated code in `build/vcpkg_installed/x64-windows/share/fmt/fmt-targets.cmake`
```cmake
# Create imported target fmt::fmt
add_library(fmt::fmt SHARED IMPORTED)

set_target_properties(fmt::fmt PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "FMT_SHARED"
  INTERFACE_COMPILE_FEATURES "cxx_std_11"
  INTERFACE_COMPILE_OPTIONS "\$<\$<COMPILE_LANGUAGE:CXX>:/utf-8>"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
)

# Create imported target fmt::fmt-header-only
add_library(fmt::fmt-header-only INTERFACE IMPORTED)

set_target_properties(fmt::fmt-header-only PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "FMT_HEADER_ONLY=1"
  INTERFACE_COMPILE_FEATURES "cxx_std_11"
  INTERFACE_COMPILE_OPTIONS "\$<\$<COMPILE_LANGUAGE:CXX>:/utf-8>"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
)
```

This should be conditional. Although vcpkg might use its own toolchain, it doesn't mean the cmake file is going to be used with MSVC. This includes clang-tidy usage (different at build time than configuration time).

aminya:
Fixes #4158

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -365,8 +365,8 @@ if (NOT MSVC)
   # Unicode is always supported on compilers other than MSVC.
 elseif (FMT_UNICODE)
   # Unicode support requires compiling with /utf-8.
-  target_compile_options(fmt PUBLIC $<$<COMPILE_LANGUAGE:CXX>:/utf-8>)
-  target_compile_options(fmt-header-only INTERFACE $<$<COMPILE_LANGUAGE:CXX>:/utf-8>)
+  target_compile_options(fmt PUBLIC $<$<AND:$<COMPILE_LANGUAGE:CXX>,$<CXX_COMPILER_ID:MSVC>>:/utf-8>)
+  target_compile_options(fmt-header-only INTERFACE $<$<AND:$<COMPILE_LANGUAGE:CXX>,$<CXX_COMPILER_ID:MSVC>>:/utf-8>)
 else ()
   target_compile_definitions(fmt PUBLIC FMT_UNICODE=0)
 endif ()


vitaut:
Thanks for the fix!

aminya:
@vitaut you're welcome. Is there a way I can backport this to a previous version as a hotfix? It's blocking fmt's usage through vcpkg on Windows with LLVM.

aminya:
For now, I am adding a patch to vcpkg 
https://github.com/microsoft/vcpkg/pull/40944

vitaut:
We could make a new release but since you've already submitted a patch for vcpkg let's just go with it.

PR #4161, opened 2024-09-14
More chrono padding

KKhanhH:
Added some padding modifiers for additional presentation types.

Changing the padding modifier for year does kind of break the convention of %C + %y = %Y though. I wanted to also add in century and short year but I wasn't sure how that should be handled.

### DIFF
--- a/doc/syntax.md
+++ b/doc/syntax.md
@@ -711,7 +711,7 @@ The available padding modifiers (*padding_modifier*) are:
 | `'0'` | Pad a numeric result string with zeros. |
 
 These modifiers are only supported for the `'H'`, `'I'`, `'M'`, `'S'`, `'U'`,
-`'V'` and `'W'` presentation types.
+`'V'`, `'W'`, `'m'`, `'j'`, `'Y'` presentation types.
 
 ## Range Format Specifications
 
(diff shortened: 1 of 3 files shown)


vitaut:
Thanks for the PR but what is the motivation behind this change?

KKhanhH:
I needed a space padded version of the month number and GCC's strftime seems to support the padding modifier for %m (and %j), in addition to the types fmt already supports. %e exists but there's nothing similar for the month number. 

As for the year, fmt differs from GCC's strftime in that gcc's always does no padding for %Y, while fmt always does 4 digit padding, so I figured it made sense to add an option that would allow for similar behavior.

vitaut:
`%j` and `%m` make sense but for `%Y` we should investigate why there is a discrepancy with `strftime`.

KKhanhH:
It does seem to differ depending on platform. MSVC's implementation will zero pad the year to 4 digits. %C and %y also technically behaves differently as well. https://godbolt.org/z/T9ssj8fan. Century is never padded, Short year has padding modifiers. 

Also I realized that I forgot about %d (which also has GCC padding modifiers), %e will suffice for space padded, but the nonpadded won't exist. Interestingly the zero padding modifier works on %e in GCC.

vitaut:
Both Python's `str.format` and C++ `std::format` do padding so I think it's fine.

vitaut:
Merged, thanks!

PR #4163, opened 2024-09-16
format std::reference_wrapper

yfeldblum:

PR #4164, opened 2024-09-18
Use fmt::formatter specialization for std::reference_wrapper to avoid undefined behavior

phprus:
Fix UB: https://github.com/fmtlib/fmt/pull/4163#issuecomment-2358995143

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -692,15 +692,17 @@ template <typename T, typename Char> struct formatter<std::complex<T>, Char> {
   }
 };
 
+FMT_EXPORT
+template <typename T, typename Char>
+struct formatter<std::reference_wrapper<T>, Char,
+                 enable_if_t<is_formattable<remove_cvref_t<T>, Char>::value>>
+    : formatter<remove_cvref_t<T>, Char> {
+  template <typename FormatContext>
+  auto format(std::reference_wrapper<T> ref, FormatContext& ctx) const
+      -> decltype(ctx.out()) {
+    return formatter<remove_cvref_t<T>, Char>::format(ref.get(), ctx);
+  }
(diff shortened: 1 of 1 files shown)


include/fmt/std.h
```diff
@@ -692,15 +692,17 @@ template <typename T, typename Char> struct formatter<std::complex<T>, Char> {
...
-}
-
-}  // namespace std
+FMT_EXPORT
+template <typename T, typename Char>
+struct formatter<std::reference_wrapper<T>, Char,
+                 enable_if_t<is_formattable<remove_cvref_t<T>, Char>::value>>
+    : formatter<remove_cvref_t<T>, Char> {
+  template <typename FormatContext>
+  auto format(const std::reference_wrapper<T>& ref, FormatContext& ctx) const
```
vitaut, line 701: Let's pass `std::reference_wrapper` by value since it should be cheap to copy.

test/std-test.cc
```diff
@@ -399,4 +399,7 @@ TEST(std_test, format_shared_ptr) {
 TEST(std_test, format_reference_wrapper) {
   int num = 35;
   EXPECT_EQ("35", fmt::to_string(std::cref(num)));
+
+  std::atomic<bool> b(true);
+  EXPECT_EQ(fmt::format("{}", std::cref(b)), "true");
```
vitaut, line 404: How is this different from the above test?

vitaut (commented):
Thank you!

test/std-test.cc
phprus: Removed.

include/fmt/std.h

vitaut:
Merged, thanks!

PR #4166, opened 2024-09-19
overloads to support colorizing compiled format strings

moshelooks:
Adds a new header `compile-color.h` with overloads of `format`, `format_to`, and `print` to support passing in a `text_style` together with a compiled format string.

I did a bit of refactoring in `color.h` to support this via a new `detail::apply_style()` function because otherwise it would need a bunch of copy-pasting into the new header.

It's generally great that fmt splits functionality across headers so you only pay for what you use. But here it's maybe unfortunate because you can include both `color.h` and `compile.h` and still be unable to colorize compiled format strings with `compile-color.h`.

I had the idea of in e.g. `color.h` doing something like
```cxx
#ifdef FMT_COMPILE_H
#include "compile-color.h
#endif
``` 
and likewise in `compile.h` with `#ifdef FMT_COLOR_H` but want to see if you have other/better ideas first or maybe just want to leave it as-is.

Best, Moshe


vitaut:
Thanks for the PR but usually it doesn't make a lot of sense to combine format string compilation with colored output. If you really want to do it you can use `fmt::styled` (https://www.godbolt.org/z/W8Pqb4Ps9) or define your own output functions.

Issue #4153, opened 2024-09-04
C++20 modules build issues


kamrann:
I've hit a number of issues trying to package `fmt` for `build2` with modules support.

- A couple of I think recent regressions, [addressed in this PR](https://github.com/fmtlib/fmt/pull/4152).
- `fmt.cc` with `import std;` disabled will `#include <expected>` without any of the guards used for including it inside `fmt/std.h`, which raises a warning on MSVC. I haven't addressed this in the PR as `FMT_HAS_INCLUDE` isn't available at that point in the source file, and I'm also unsure why the C++ version and above macro are used in some places but feature test macros in others.
- I guess the `fmt` modules support is not being verified by the CI? I'm also a bit unclear on what the status is of the tests wrt modules. I can see there is a `module-test.cc`, but before that gets added in the cmake file [this check](https://github.com/fmtlib/fmt/blob/master/test/CMakeLists.txt#L46-L48) will bail out.

There appear to also be issues with the combination of modules and shared builds.
- I've hit issues with the symbol export macros, however there are some unanswered questions here about what `build2` should be doing, so it's not clear to me yet if anything will need to change on the `fmt` side for this particular problem.
- Even if I work around the above, I end up with lots of linker errors about undefined symbols when building the tests against a shared modules build of the library (example output attached). I believe these derive from the fact that inside a module TU, member functions defined within the class definition are not implicitly `inline` as they are in traditional TUs. As such there are lots of functions that have not been marked as `FMT_API` (themselves or via their class) as they were implicitly inline; but now in the modules case they no longer are. There would appear to be two options: mark more classes/functions as `FMT_API`, or mark the member functions as explicitly inline. I wonder if the former can potentially cause problems for a non-modules build, but I'm generally unsure of the trade-offs here.

[clang-libcpp-fedora-modules-link-errors.txt](https://github.com/user-attachments/files/16861655/clang-libcpp-fedora-modules-link-errors.txt)

vitaut:
Thanks for the fixes!

The warning about `<expected>` should be suppressed in https://github.com/fmtlib/fmt/commit/a2c290bc34d3ad1a10347fdc5ecdc7f50dcb90ff. Tests need more work to make them compatible with modules which is why they are currently disabled.

I haven't tried using modules with shared libraries. As usual PRs to make {fmt} more compatible with this configuration would be welcome.

vitaut:
[...]

PR #4169, opened 2024-09-20
Module linkage fixes for shared build

kamrann:
Following on from #4153:

- In regards to `constexpr`, this was indeed a Clang bug, and anything already marked `constexpr` should be a non-issue; the problem has been fixed on Clang trunk with potential for it to be backported to the 19.x release.
- A number of other linker errors specific to mixing `#include` with `import` were also being caused by a related Clang bug, also now fixed on trunk.
- Despite this, a bunch of linker errors still remained when trying to consume the `fmt` module via a shared library build, due to the last bullet point in the linked issue.

This PR addresses those remaining issues by adding `FMT_INLINE`/`FMT_CONSTEXPR` to some member functions defined within the class definition. However, due to the nature of linker issues, problematic cases are only being picked up where a symbol is actually being used from an external dll/exe. Currently, the coverage for that is limited - `fmt`'s own CI is not yet running modules builds I think (or anyway not with `FMT_SHARED`); the `build2` `fmt` package CI also cannot stress test this as the very latest Clang is not available. Furthermore, I've so far only enabled a subset of the tests in the `build2` package.

Fundamentally, I think this addition of `inline` or `constexpr` is essentially needed on every single in-class defined member function which is intended for public use (or alternatively `FMT_API` on the member/class), which is probably a lot more than what I've addressed in these commits and would be quite intrusive. So of course feel free to merge if you like, but it's possible some more thought is warranted in regards to
- is every public member function in the headers actually intended to be usable directly by clients (this question of course also relates to what should be `export`ed from the module, but for `export`, doing things in bulk is possible and less intrusive in terms of code changes)
- is there a preference for exposing those that should be exposed as `inline` vs `FMT_API`

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -1792,9 +1792,9 @@ template <typename T> class buffer {
 };
 
 struct buffer_traits {
-  FMT_CONSTEXPR explicit buffer_traits(size_t) {}
-  FMT_CONSTEXPR auto count() const -> size_t { return 0; }
-  FMT_CONSTEXPR auto limit(size_t size) -> size_t { return size; }
+  constexpr explicit buffer_traits(size_t) {}
+  constexpr auto count() const -> size_t { return 0; }
+  constexpr auto limit(size_t size) -> size_t { return size; }
 };
 
 class fixed_buffer_traits {
@@ -1803,9 +1803,9 @@ class fixed_buffer_traits {
   size_t limit_;
 
  public:
-  FMT_CONSTEXPR explicit fixed_buffer_traits(size_t limit) : limit_(limit) {}
-  FMT_CONSTEXPR auto count() const -> size_t { return count_; }
-  FMT_CONSTEXPR auto limit(size_t size) -> size_t {
+  constexpr explicit fixed_buffer_traits(size_t limit) : limit_(limit) {}
+  constexpr auto count() const -> size_t { return count_; }
+  constexpr auto limit(size_t size) -> size_t {
     size_t n = limit_ > count_ ? limit_ - count_ : 0;
     count_ += size;
     return size < n ? size : n;
@@ -2223,7 +2223,7 @@ struct locale_ref {
  public:
   constexpr locale_ref() : locale_(nullptr) {}
   template <typename Locale> explicit locale_ref(const Locale& loc);
-  FMT_INLINE explicit operator bool() const noexcept { return locale_ != nullptr; }
+  inline explicit operator bool() const noexcept { return locale_ != nullptr; }
 #endif
 
   template <typename Locale> auto get() const -> Locale;
@@ -2598,7 +2598,7 @@ class context : private detail::locale_ref {
   void operator=(const context&) = delete;
 
   FMT_CONSTEXPR auto arg(int id) const -> format_arg { return args_.get(id); }
-  FMT_INLINE auto arg(string_view name) -> format_arg { return args_.get(name); }
+  inline auto arg(string_view name) -> format_arg { return args_.get(name); }
   FMT_CONSTEXPR auto arg_id(string_view name) -> int {
     return args_.get_id(name);
   }
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -1885,9 +1885,9 @@ template <typename T> class buffer {
 };
 
 struct buffer_traits {
-  explicit buffer_traits(size_t) {}
-  auto count() const -> size_t { return 0; }
-  auto limit(size_t size) -> size_t { return size; }
+  FMT_CONSTEXPR explicit buffer_traits(size_t) {}
+  FMT_CONSTEXPR auto count() const -> size_t { return 0; }
+  FMT_CONSTEXPR auto limit(size_t size) -> size_t { return size; }
 };
 
 class fixed_buffer_traits {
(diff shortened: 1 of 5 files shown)


kamrann:
Linter seems to be complaining about formatting that I didn't change.

vitaut:
Before we go into detail, could you link to where this assumption comes from?

> inside a module TU, member functions defined within the class definition are not implicitly inline 

This seems like a major incompatibility and I would be surprised if it was the case.

kamrann:
Sure. The relevant revision to the standard was [P1779](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p1779r3.html). There's also some discussion regarding it [in this SO post](https://stackoverflow.com/questions/60280105/how-does-inline-affect-member-functions-in-module-interfaces).

include/fmt/base.h
```diff
@@ -1792,9 +1792,9 @@ template <typename T> class buffer {
 };
 
 struct buffer_traits {
-  explicit buffer_traits(size_t) {}
-  auto count() const -> size_t { return 0; }
-  auto limit(size_t size) -> size_t { return size; }
+  FMT_CONSTEXPR explicit buffer_traits(size_t) {}
+  FMT_CONSTEXPR auto count() const -> size_t { return 0; }
+  FMT_CONSTEXPR auto limit(size_t size) -> size_t { return size; }
```
vitaut, line 1797: These can be `constexpr`. `FMT_CONSTEXPR` is only for cases when C++14 relaxed `constexpr` is needed.

include/fmt/base.h
```diff
@@ -1803,9 +1803,9 @@ class fixed_buffer_traits {
   size_t limit_;
 
  public:
-  explicit fixed_buffer_traits(size_t limit) : limit_(limit) {}
-  auto count() const -> size_t { return count_; }
-  auto limit(size_t size) -> size_t {
+  FMT_CONSTEXPR explicit fixed_buffer_traits(size_t limit) : limit_(limit) {}
+  FMT_CONSTEXPR auto count() const -> size_t { return count_; }
+  FMT_CONSTEXPR auto limit(size_t size) -> size_t {
```
vitaut, line 1808: Same here and in a few more places below.

include/fmt/base.h
```diff
@@ -2223,7 +2223,7 @@ struct locale_ref {
  public:
   constexpr locale_ref() : locale_(nullptr) {}
   template <typename Locale> explicit locale_ref(const Locale& loc);
-  explicit operator bool() const noexcept { return locale_ != nullptr; }
+  FMT_INLINE explicit operator bool() const noexcept { return locale_ != nullptr; }
```
vitaut, line 2226: Here and elsewhere, please use `inline` instead of `FMT_INLINE` since the latter expands to always inline in optimized mode.

include/fmt/chrono.h
```diff
@@ -540,24 +540,24 @@ inline auto localtime(std::time_t time) -> std::tm {
...
       using namespace fmt::detail;
       return fallback(localtime_s(&tm_, &time_));
     }
 
-    auto fallback(int res) -> bool { return res == 0; }
+    FMT_INLINE auto fallback(int res) -> bool { return res == 0; }
 
 #if !FMT_MSC_VERSION
-    auto fallback(detail::null<>) -> bool {
+    FMT_INLINE auto fallback(detail::null<>) -> bool {
```
vitaut, line 560: This is not needed since the struct where these functions are defined is inaccessible to module consumers.

include/fmt/chrono.h
```diff
@@ -591,24 +591,24 @@ inline auto gmtime(std::time_t time) -> std::tm {
...
       using namespace fmt::detail;
       return fallback(gmtime_s(&tm_, &time_));
     }
 
-    auto fallback(int res) -> bool { return res == 0; }
+    FMT_INLINE auto fallback(int res) -> bool { return res == 0; }
 
 #if !FMT_MSC_VERSION
-    auto fallback(detail::null<>) -> bool {
+    FMT_INLINE auto fallback(detail::null<>) -> bool {
```
vitaut, line 611: Same here

include/fmt/chrono.h
```diff
@@ -912,7 +912,7 @@ template <typename Derived> struct null_chrono_spec_handler {
 };
 
 struct tm_format_checker : null_chrono_spec_handler<tm_format_checker> {
-  FMT_NORETURN void unsupported() { FMT_THROW(format_error("no format")); }
+  FMT_NORETURN FMT_INLINE void unsupported() { FMT_THROW(format_error("no format")); }
```
vitaut, line 915: Here and in other places in the `detail` namespace - `inline` is not needed since it is not exported from the module.

vitaut (changes requested):

include/fmt/base.h
kamrann: Okay my bad, hadn't realized the distinction. I think all the changes are now `constexpr` where they can be.

include/fmt/chrono.h
kamrann: Although I may have branched out a little when making the changes, the vast majority of them were made on functions that were generating linker errors. I've verified again, these are necessary for the linker to succeed.

Essentially, the issue being addressed here is one of dll symbol export, not module name export, and so the visibility of names (which is what module export/detail namespaces relate to) I think has no bearing. Since this struct is in an `inline` function which is exposed, a consumer TU that calls this function (directly or transitively) will either need to inline the member functions of the struct, or emit an external symbol reference for them. So the same choice still applies.

include/fmt/chrono.h
kamrann: As above, `detail` namespace functions will still generate linker errors if neither `inline` nor dll-exported.

vitaut:
OK, considering that the number of changes is manageable, we can merge it in but please apply clang-format first.

vitaut (approved):

vitaut:
Merged, thanks!

PR #4170, opened 2024-09-20
Fix conversion warning on chrono.h

Ottani:
Casting from 'long long int' into 'long int' to remove the warning from -Wconversion as the method write_utc_offset accepts a long, and the offset should be in the range of a long variable.


vitaut (changes requested):
Thanks for the PR. I think the correct fix is to change `write_utc_offset` so that it takes `long long`.

Ottani:
Ok, I did the changes as requested and removed the cast from `format_utc_offset_impl`, storing the value in a `long long`.

vitaut (approved):

vitaut:
Thanks!

PR #4185, opened 2024-10-02
Conform `std::iterator_traits<fmt::appender>` to [iterator.traits]/1

CaseyCarter:
> In addition, the types
> ```c++
> iterator_traits<I>::pointer
> iterator_traits<I>::reference
> ```
> shall be defined as the iterator’s pointer and reference types; that is, for an iterator object `a` of class type, the same type as `decltype(a.operator->())` and `decltype(*a)`, respectively. The type `iterator_traits<I>::pointer` shall be void for an iterator of class type `I` that does not support `operator->`. Additionally, in the case of an output iterator, the types
> ```c++
> iterator_traits<I>::value_type
> iterator_traits<I>::difference_type
> iterator_traits<I>::reference
> ```
> may be defined as `void`.

Also removes the unnecessary member types from `basic_appender`.

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -2382,11 +2382,6 @@ template <typename T> class basic_appender {
   detail::buffer<T>* container;
 
  public:
-  using iterator_category = int;
-  using value_type = T;
-  using pointer = T*;
-  using reference = T&;
-  using difference_type = decltype(pointer() - pointer());
   using container_type = detail::buffer<T>;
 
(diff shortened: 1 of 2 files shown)


CaseyCarter:
Some C++20 tests that my naive local build didn't enable apparently want to use `basic_appender` without the `iterator_traits` specialization. I'll need to bring back at least the nested `difference_type`.

CaseyCarter:
Ok, this now compiles `FMT_PEDANTIC` C++20 on at least MSVC and GCC. Sorry for the noisy PR =)

vitaut:
Thank you!

Issue #4180, opened 2024-10-01
fmt 11 cannot format string_view


aminya:
I updated the following code to fmt 11, and it no longer compiles with the given error below. Do you know what could be the reason? 

Unfortunately, I can't reproduce it on Godbolt, but here's the link anyway
https://godbolt.org/z/KEo17z5xq

```cpp
#include <fmt/core.h>

#include <string>
#include <string_view>

using std::string_view;

[[noreturn]] inline void print_string_view(string_view message) {
  fmt::print(stderr, "{}\n", message);
}

int main() {
  print_string_view("Don't panic!\n");
}
```

```shell
build/vcpkg_installed/x64-linux/include/fmt/base.h:1518:68: error: no matching member function for call to 'map'
 1518 |   FMT_MAP_API auto map(const T& val) -> decltype(FMT_DECLTYPE_THIS map(U())) {
      |                                                                    ^~~
build/vcpkg_installed/x64-linux/include/fmt/base.h:1560:28: note: in instantiation of template class 'fmt::detail::arg_mapper<fmt::context>' requested here
 1560 |     type_constant<decltype(arg_mapper<Context>().map(std::declval<const T&>())),
      |                            ^
build/vcpkg_installed/x64-linux/include/fmt/base.h:2733:16: note: in instantiation of template type alias 'mapped_type_constant' requested here
 2733 |       : types_{mapped_type_constant<Args, buffered_context<Char>>::value...},
      |                ^
build/vcpkg_installed/x64-linux/include/fmt/base.h:2884:47: note: in instantiation of member function 'fmt::detail::format_string_checker<char, std::basic_string_view<char>>::format_string_checker' requested here
 2884 |       detail::parse_format_string<true>(str_, checker(s));
      |                                               ^
packages/stdx/include/stdx/errors.hpp:18:22: note: in instantiation of function template specialization 'fmt::basic_format_string<char, std::basic_string_view<char> &>::basic_format_string<char[4], 0>' requested here
   18 |   fmt::print(stderr, "{}\n", message);
      |                      ^
build/vcpkg_installed/x64-linux/include/fmt/base.h:1511:20: note: candidate function template not viable: requires single argument 'values', but no arguments were provided
 1511 |   FMT_MAP_API auto map(const T (&values)[N]) -> const T (&)[N] {
      |                    ^   ~~~~~~~~~~~~~~~~~~~~
[...]

PR #4186, opened 2024-10-02
fix: rename typename U to typename UT to avoid cpprest conflicts

aminya:
Workaround for #4180


include/fmt/base.h
```diff
@@ -448,8 +448,8 @@ template <typename T> constexpr const char* narrow(const T*) { return nullptr; }
 constexpr FMT_ALWAYS_INLINE const char* narrow(const char* s) { return s; }
 
 template <typename Char>
-FMT_CONSTEXPR auto compare(const Char* s1, const Char* s2, std::size_t n)
-    -> int {
+FMT_CONSTEXPR auto compare(const Char* s1, const Char* s2,
+                           std::size_t n) -> int {
```
aminya, line 452: The formatting changes are caused by clang-format so that CI doesn't fail

aminya:
> Thanks for the PR but we shouldn't clutter the code because of poorly written third-party libraries.

That's fair. However, unfortunately, I noticed that OpenAPI cpprest generator also uses `U`. I don't think it's viable for cpprest to remove the macro at this point. I am going to patch OpenAPI generator to `undef` U at the end of the headers. A lot of work for a bad name.

edo9300:
That macro is only replacing `U()` calls, a maybe less invasive change could be to replace them with uniform initializations `U{}`

aminya:
The actual Diff is much smaller than in the PR:
https://github.com/fmtlib/fmt/compare/master...aminya:fmt:type-conficlit

I was still fixing the clang-format version where the PR was closed.

edo9300:
Actually, the conflicting code isn't present at all in the current master, so the workaround isn't needed in first place

edo9300:
Removed in https://github.com/fmtlib/fmt/commit/761d35f7632a1dada0be8e6e9d18a3ba16644f00

PR #4188, opened 2024-10-03
Silencing Wextra-semi warning

c4v4:
Hi,
Only a minor change to silence g++ from emitting a Wextra-semi warning when that flag is turned on.


vitaut:
Thanks!

PR #4194, opened 2024-10-10
Fix -Wmissing-noreturn warning

kongy:
Fixes warning reported by top of trunk Clang:

```
include/fmt/chrono.h:447:36: error: function 'throw_duration_error' could be declared with attribute 'noreturn' [-Werror,-Wmissing-noreturn]
```


vitaut:
Thank you

PR #4200, opened 2024-10-15
Changing type of data_ to size_t to avoid compilation errors

vinayyadav3016:
1. If lib is cross-compiled for win32 using MXE environment it cause
   compilation error -Wconversion on line 730 as sizeof(unsigned long)
   = 4 and sizeof(size_t) = 8
2. When lib is compiled on Unix like compiler generate error
   -Wuseless-cast if static_cast is used to fix issue in 1


vitaut:
Thank you

PR #4210, opened 2024-10-21
Fix narrowing conversion warning in struct fstring

sergio-nsk:
Warning C4267: 'initializing': conversion from 'size_t' to 'const int', possible loss of data.

Client applications fail to compile if warnings are treated as errors.


vitaut:
thanks

Issue #4206, opened 2024-10-20
use `fmt::format_to` with `vector<char>` and `FMT_COMPILE` 


trim21:
I'm new to c++, apology if this is not a good issue.

I'm coming from https://github.com/fmtlib/fmt/issues/1771 , but I'm using a `vector<char>` as memory buffer. 

And `fmt::format_to(std::back_inserter(out), FMT_COMPILE("{}"), val)` works if `out` is `std::string` of `fmt::memory_buffer` but not `std::vector<char>`, and `fmt::format_to(std::back_inserter(buffer), "{}", val)` works for `vector<char>`

Is it possible to add support for `vector<char>` and `FMT_COMPILE`?

https://godbolt.org/z/h6xoG65he

vitaut:
Thanks for reporting. We should detect presence of `append` with SFINAE in https://github.com/fmtlib/fmt/blob/2b6a786e3507a9a5a856cb58c09413acf212c7ed/include/fmt/base.h#L1974-L1981

vitaut:
Done in #4211.

trim21:
thanks

PR #4211, opened 2024-10-21
Implement fmt::format_to into std::vector<char>

phprus:
Fix for issue #4206

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -1973,20 +1973,20 @@ template <typename T = char> class counting_buffer : public buffer<T> {
 template <typename T>
 struct is_back_insert_iterator<basic_appender<T>> : std::true_type {};
 
-template <typename T, typename It, typename = void>
-struct has_container_append : std::false_type {};
-template <typename T, typename It>
-struct has_container_append<T, It,
-                            void_t<decltype(std::declval<T>().append(
-                                std::declval<It>(), std::declval<It>()))>>
-    : std::true_type {};
+template <typename OutputIt, typename InputIt, typename = void>
+struct has_back_insert_iterator_container_append : std::false_type {};
+template <typename OutputIt, typename InputIt>
+struct has_back_insert_iterator_container_append<
+    OutputIt, InputIt,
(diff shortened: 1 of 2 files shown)


include/fmt/base.h
```diff
@@ -1971,12 +1971,32 @@ template <typename T = char> class counting_buffer : public buffer<T> {
...
+          FMT_ENABLE_IF(has_container_append<T, It>::value)>
+FMT_CONSTEXPR20 void container_append(T& container, It begin, It end) {
+  container.append(begin, end);
+}
+
+template <typename T, typename It,
+          FMT_ENABLE_IF(!has_container_append<T, It>::value)>
+FMT_CONSTEXPR20 void container_append(T& container, It begin, It end) {
+  container.insert(container.end(), begin, end);
+}
```
vitaut, line 1993: I don't think we need an extra "level of indirection". Let's merge these into `copy`.

vitaut (changes requested):
Thanks for the PR! Overall looks good, but let's simplify this a bit (see the inline comment).

include/fmt/base.h
phprus: Merged

include/fmt/base.h
```diff
@@ -1971,15 +1971,38 @@ template <typename T = char> class counting_buffer : public buffer<T> {
...
+    : std::true_type {};
+
 // An optimized version of std::copy with the output value type (T).
-template <typename T, typename InputIt, typename OutputIt,
-          FMT_ENABLE_IF(is_back_insert_iterator<OutputIt>::value)>
+template <
+    typename T, typename InputIt, typename OutputIt,
+    FMT_ENABLE_IF(
+        is_back_insert_iterator<OutputIt>::value&& has_container_append<
+            decltype(get_container(std::declval<OutputIt>())), InputIt>::value)>
```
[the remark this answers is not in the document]
phprus: Done

vitaut (approved):

vitaut:
Thanks!

Issue #4226, opened 2024-11-02
Extend tuple features (e.g. join) to tuple-like objects


correaa:
For some technical reasons I had to replicate the functionality of std::tuple in my library. (see here: https://github.com/correaa/boost-multi?tab=readme-ov-file#formatting-fmt-pretty-printing)

I would be nice that the features in fmt would work on any type that has `std::get<N>`, `std::tuple_size`, etc. in their interface.
That is, if a type is tuple like then it could be printed as a tuple.

(This will also work on any class for which [structured binding](https://en.cppreference.com/w/cpp/language/structured_binding) is implemented).

```cpp
int main() {

    std::tuple<int, int> t1(1, 2);
    fmt::print( "{}", fmt::join(t1, "x") );  // prints 1 x 2

    my_tuple<int, int> t2(1, 2);
    fmt::print( "{}", fmt::join(t2, "x") );  // doesn't work now
}
```

Here it is a godbolt example:

https://godbolt.org/z/Gof6f4oaE

vitaut:
Thanks for the suggestion. I think it's reasonable to add tuple-like object support to `fmt::join` and a PR would be welcome.

correaa:
Does it work? I think godbolt has updated, but the library still fails for a tuple-like structure:

https://godbolt.org/z/EW4659Yzv

phprus:
``boost::multi`` has a standard incompatible ``std::tuple_size`` implementation.

> All specializations of std::tuple_size satisfy UnaryTypeTrait with base characteristic std::integral_constant<std::size_t, N> for some N.

https://en.cppreference.com/w/cpp/utility/tuple_size

But https://gitlab.com/correaa/boost-multi/-/blob/master/include/boost/multi/detail/tuple_zip.hpp?ref_type=heads#L300 :

```c++
template<class... Ts>
struct std::tuple_size<boost::multi::detail::tuple<Ts...>> {  // NOLINT(cert-dcl58-cpp) to have structured bindings
	// cppcheck-suppress unusedStructMember
	static constexpr std::size_t value = sizeof...(Ts);
};
```

does not inherit ``std::integral_constant``.

correaa:
Good point, I didn't know that the type has to specifically inherit from `std::integral_type`, I thought it was enough to have a value static member. Inherithing from `std::integral_type` seems like an implementation detail.

phprus:
[...]

PR #4230, opened 2024-11-07
Implement fmt::join for tuple-like objects

phprus:
Fix for #4226

### DIFF
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -696,18 +696,18 @@ auto join(It begin, Sentinel end, string_view sep) -> join_view<It, Sentinel> {
  *     fmt::print("{:02}", fmt::join(v, ", "));
  *     // Output: 01, 02, 03
  */
-template <typename Range>
+template <typename Range, FMT_ENABLE_IF(!is_tuple_like<Range>::value)>
 auto join(Range&& r, string_view sep)
     -> join_view<decltype(detail::range_begin(r)),
                  decltype(detail::range_end(r))> {
   return {detail::range_begin(r), detail::range_end(r), sep};
 }
 
-template <typename Char, typename... T> struct tuple_join_view : detail::view {
-  const std::tuple<T...>& tuple;
+template <typename Char, typename Tuple> struct tuple_join_view : detail::view {
+  const Tuple& tuple;
   basic_string_view<Char> sep;
 
-  tuple_join_view(const std::tuple<T...>& t, basic_string_view<Char> s)
+  tuple_join_view(const Tuple& t, basic_string_view<Char> s)
       : tuple(t), sep{s} {}
 };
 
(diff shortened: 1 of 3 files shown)


include/fmt/ranges.h
```diff
@@ -696,18 +696,18 @@ auto join(It begin, Sentinel end, string_view sep) -> join_view<It, Sentinel> {
  *     fmt::print("{:02}", fmt::join(v, ", "));
  *     // Output: 01, 02, 03
  */
-template <typename Range>
+template <typename Range, FMT_ENABLE_IF(!fmt::is_tuple_like<Range>::value)>
```
vitaut, line 699: Why qualify this with `fmt::`? I don't think ADL applies here.

phprus: You are right.
Removed ``fmt::``.

test/ranges-test.cc
```diff
@@ -213,7 +213,8 @@ TEST(ranges_test, tuple_parse_calls_element_parse) {
   EXPECT_THROW(f.parse(ctx), bad_format);
 }
 
-#ifdef FMT_RANGES_TEST_ENABLE_FORMAT_STRUCT
+#if defined(FMT_RANGES_TEST_ENABLE_JOIN) || \
+    defined(FMT_RANGES_TEST_ENABLE_FORMAT_STRUCT)
```
vitaut, line 217: We can probably remove these macros - they are testing for old variants of MSVC <= 2017.

phprus: ``struct tuple_like`` is used in tests that are enabled if these macros are defined.
I think enabling tests for early MSVC is out of scope of this PR.

phprus: As I see MSVS 2015 is no longer supported (https://github.com/fmtlib/fmt/commit/8ed4a9dcc1c04cfaedcddc3ac25918cf35250f4a).
I propose to remove these checks in a separate PR.

What are the minimum compilers versions required for fmt now?

vitaut: For MSVC we definitely don't support anything older than 2017. 2017 is a bit tricky because there are multiple variants of it with older ones being quite broken and probably not worth supporting. Separate PR sounds reasonable.

test/ranges-test.cc
```diff
@@ -420,6 +423,10 @@ TEST(ranges_test, join_tuple) {
   auto t4 = std::tuple<float>(4.0f);
   EXPECT_EQ(fmt::format("{}", fmt::join(t4, "/")), "4");
 
+  // Tuple-like.
+  auto t5 = tuple_like{42, "foo"};
+  EXPECT_EQ(fmt::format("{}", t5), "(42, \"foo\")");
```
vitaut, line 428: This test case looks identical to one of the above:

```
  auto t = tuple_like{42, "foo"};
  EXPECT_EQ(fmt::format("{}", t), "(42, \"foo\")");
```

Did you mean to use `join` here?

phprus: Yes, I lost ``fmt::join``.

vitaut:
Thanks!

Issue #4232, opened 2024-11-09
\fmt\include\fmt/format.h(1698): warning C4127


UnrealKaraulov:
```
3>.\..\fmt\include\fmt/format.h(1698): warning C4127: условное выражение является константой (компилируется исходный файл .\..\src\editor\Renderer.cpp)
3>.\..\fmt\include\fmt/format.h(1698): note: Попробуйте использовать инструкцию "if constexpr" (компилируется исходный файл .\..\src\editor\Renderer.cpp)
3>.\..\fmt\include\fmt/format.h(1721): warning C4127: условное выражение является константой (компилируется исходный файл .\..\src\editor\Renderer.cpp)
3>.\..\fmt\include\fmt/format.h(1721): note: Попробуйте использовать инструкцию "if constexpr" (компилируется исходный файл .\..\src\editor\Renderer.cpp)
3>.\..\fmt\include\fmt/format.h(1698): warning C4127: условное выражение является константой (компилируется исходный файл .\..\src\editor\Gui.cpp)
3>.\..\fmt\include\fmt/format.h(1698): note: Попробуйте использовать инструкцию "if constexpr" (компилируется исходный файл .\..\src\editor\Gui.cpp)
3>.\..\fmt\include\fmt/format.h(1721): warning C4127: условное выражение является константой (компилируется исходный файл .\..\src\editor\Gui.cpp)
3>.\..\fmt\include\fmt/format.h(1721): note: Попробуйте использовать инструкцию "if constexpr" (компилируется исходный файл .\..\src\editor\Gui.cpp)
```

How to fix this compile errors in fmtlib?

Visual Studio 2022

UnrealKaraulov:
(after upgrade fmtlib to latest version)

vitaut:
Same as #4220.

Issue #4220, opened 2024-10-29
MSVC warning - C4127: conditional expression is constant


remyjette:
Seeing this on latest master.

minimal repro:

CMakeLists.txt
```cmake
cmake_minimum_required(VERSION 3.20)
project(example)

add_executable(example main.cpp)

add_subdirectory(fmt)
target_link_libraries(example PRIVATE fmt)

target_compile_options(example PRIVATE "/WX" "/W4")
```

main.cpp
```cpp
#include <fmt/format.h>

int main() {
    fmt::print("foo");
}
```

```
D:\src\example>cmake -S. -Bbuild -G"Visual Studio 17 2022"
-- The C compiler identification is MSVC 19.41.34123.0
-- The CXX compiler identification is MSVC 19.41.34123.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/MSVC/14.41.34120/bin/Hostx64/x64/cl.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/MSVC/14.41.34120/bin/Hostx64/x64/cl.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- {fmt} version: 11.0.2
-- Build type:
-- Configuring done (3.6s)
-- Generating done (0.1s)
-- Build files have been written to: D:/src/example/build

D:\src\example>cmake --build build
MSBuild version 17.11.9+a69bbaaf5 for .NET Framework

  1>Checking Build System
  Building Custom Rule D:/src/example/fmt/CMakeLists.txt
  format.cc
  os.cc
  Generating Code...
  fmt.vcxproj -> D:\src\example\build\fmt\Debug\fmtd.lib
  Building Custom Rule D:/src/example/CMakeLists.txt
  main.cpp
D:\src\example\fmt\include\fmt\format.h(1793,25): error C2220: the following warning is treated as an error [D:\src\example\build\example.vcxproj]
  (compiling source file '../main.cpp')

D:\src\example\fmt\include\fmt\format.h(1793,25): warning C4127: conditional expression is constant [D:\src\example\build\example.vcxproj]
[...]

PR #4233, opened 2024-11-09
Suppress MSVC warnings "C4127: conditional expression is constant" by used const_check

phprus:
Fix for #4220 , #4232


vitaut:
Thank you

PR #4236, opened 2024-11-12
Replace std::forward for clang-tidy

### LINKED ISSUE #4231: Possible use-after-move in join()
The code at:

https://github.com/fmtlib/fmt/blob/720da57baba83b3b1829e20133575e57aa1a8a4f/include/fmt/ranges.h#L661-L666

Does a
```cpp
auto it = std::forward<view_ref>(value).begin;
```

and later accesses `value.end` and `value.sep`.

In the case the iterator is _not_ copy constructible then the `view_ref` is an rvalue-reference:

```cpp
  using view_ref = conditional_t<std::is_copy_constructible<It>::value,
                                 const join_view<It, Sentinel, Char>&,
                                 join_view<It, Sentinel, Char>&&>;
```

which effectively turns the `forward<>` into a `move`.

`clang-tidy` warns on the accesses of `end` and `sep` as they are accesses on a moved-from object.

I'm not sure that this is _really_ a problem in practice but it seems like something that should be avoided. I can't see an easy way to do so though. I'm not sure we can cast _just_ the iterator easily? Taking a copy of `sep` works but for the same reasons as above we can't copy `end` out without a move either.

Arghnews:
Should fix #4231

Fix was a little different than I thought initially, due to trickiness of getting `const` right

Could instead use `<!std::is_copy_constructible_v<It>, It&, It>;` as of course `view_ref` is just dependent on this. I don't see or have a strong preference either way

### DIFF
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -646,9 +646,9 @@ struct formatter<join_view<It, Sentinel, Char>, Char> {
 #endif
   formatter<remove_cvref_t<value_type>, Char> value_formatter_;
 
-  using view_ref = conditional_t<std::is_copy_constructible<It>::value,
-                                 const join_view<It, Sentinel, Char>&,
-                                 join_view<It, Sentinel, Char>&&>;
+  using view = conditional_t<std::is_copy_constructible<It>::value,
+                             const join_view<It, Sentinel, Char>,
+                             join_view<It, Sentinel, Char>>;
 
  public:
   using nonlocking = void;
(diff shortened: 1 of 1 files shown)


include/fmt/ranges.h
```diff
@@ -660,7 +660,9 @@ struct formatter<join_view<It, Sentinel, Char>, Char> {
   template <typename FormatContext>
   auto format(view_ref& value, FormatContext& ctx) const
```
vitaut, line 661: Shall we get rid of rvalue reference as discussed in the issue? Maybe replace `view_ref` with

```
  using view = conditional_t<std::is_copy_constructible<It>::value,
                                 const join_view<It, Sentinel, Char>,
                                 join_view<It, Sentinel, Char>>;
```
and use `std::is_copy_constructible` in the iterator check?

vitaut (changes requested):
Thanks for the PR!

include/fmt/ranges.h
Arghnews: Sure thing, made this change

vitaut (approved):

vitaut:
Merged, thanks!

mattgodbolt:
Thank you all!

PR #4254, opened 2024-12-05
gcc: fix -Wnvro warning

neheb:
Allows the copiler to elide the copy.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -1718,8 +1718,9 @@ auto find_escape(const Char* begin, const Char* end)
 
 inline auto find_escape(const char* begin, const char* end)
     -> find_escape_result<char> {
-  if (const_check(!use_utf8)) return find_escape<char>(begin, end);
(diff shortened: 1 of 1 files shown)


include/fmt/format.h
```diff
@@ -1718,8 +1718,9 @@ auto find_escape(const Char* begin, const Char* end)
 
 inline auto find_escape(const char* begin, const char* end)
     -> find_escape_result<char> {
-  if (const_check(!use_utf8)) return find_escape<char>(begin, end);
-  auto result = find_escape_result<char>{end, nullptr, 0};
+  auto result = find_escape<char>(begin, end);
```
mwinterb, line 1721: is not eliding the 24 byte copy slower than unconditionally calling `find_escape<char>`?

vitaut:
Thanks for the PR but unconditionally calling `find_escape` does look like a pessimization to me. Closing for now but feel free to reopen with evidence that this is actually an improvement.

PR #4262, opened 2024-12-10
Make parameter buf of to_string const again.

### LINKED ISSUE #4261: Is it intentional that the parameter of to_string(basic_memory_buffer<char, SIZE>& buf) is not const?
I'm writing a little helper class, which uses a `basic_memory_buffer<char, SIZE> buffer` internally and does multiple calls to `format_to(std::back_inserter(buffer), fmt, args...)` throughout it's lifetime.

Eventually the buffer content is retrieved as a string using the following function:
```
template <size_t SIZE>
FMT_NODISCARD auto to_string(basic_memory_buffer<char, SIZE>& buf)
    -> std::string {
  auto size = buf.size();
  detail::assume(size < std::string().max_size());
  return {buf.data(), size};
}
```
Since both operations size() and data() are declared const it should be possible to make buf const, doesn't it?

Since the parameter `buf` is `const` in my case, it is trying to using the generic overload
```
template <typename T, FMT_ENABLE_IF(detail::use_format_as<T>::value)>
FMT_NODISCARD auto to_string(const T& value) -> std::string {
  return to_string(format_as(value));
}
```
which then results in a compilation error, stating that `"parse"` is not a member of `"fmt::v11::formatter<T,char,void>"`.

Therefore I either have to
1. declare my member function calling to_string not const as well, which seems like an unnecessary restriction to me
2. or use const_cast to force the use of this overload even though it looks like a bad smell to me. Would it be safe to do so anyways?
3. or use some other approach that I'm not aware of

As of writing this question I found out that this change was made quite recently with the commit 
ab8f9d5b0802fe139ce3214475e7517f1c985e9a?

sascha-devel:
Fixes #4261 and makes the parameter of `to_string(basic_memory_buffer<char, SIZE>& buf)` `const` again.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -891,7 +891,7 @@ class basic_memory_buffer : public detail::buffer<T> {
 using memory_buffer = basic_memory_buffer<char>;
 
 template <size_t SIZE>
-FMT_NODISCARD auto to_string(basic_memory_buffer<char, SIZE>& buf)
+FMT_NODISCARD auto to_string(const basic_memory_buffer<char, SIZE>& buf)
     -> std::string {
   auto size = buf.size();
   detail::assume(size < std::string().max_size());


vitaut:
Merged in https://github.com/fmtlib/fmt/commit/873670ba3f9e7bc77ec2c1c94b04f1f8bef77e9f, thanks.

PR #4267, opened 2024-12-17
Use the _MSVC_STL_UPDATE macro to detect STL

YexuanXiao:
The current code's method of detecting the C++ standard library as STL is inaccurate, which will reject libc++ on Windows. The [_MSVC_STL_UPDATE](https://github.com/microsoft/STL/wiki/Macro-_MSVC_STL_UPDATE) macro is the documented way. fmt currently [requires Visual Studio 19.0](https://fmt.dev/11.0/), and this macro has existed [since Visual Studio 15.5](https://github.com/microsoft/STL/wiki/Macro-_MSVC_STL_UPDATE), so there are no compatibility issues.

### DIFF
--- a/include/fmt/ostream.h
+++ b/include/fmt/ostream.h
@@ -35,7 +35,7 @@ class file_access {
   friend auto get_file(BufType& obj) -> FILE* { return obj.*FileMemberPtr; }
 };
 
-#if FMT_MSC_VERSION
+#if _MSVC_STL_UPDATE
 template class file_access<file_access_tag, std::filebuf,
                            &std::filebuf::_Myfile>;
 auto get_file(std::filebuf&) -> FILE*;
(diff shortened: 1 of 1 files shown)


vitaut (changes requested):
Overall looks good but please fix the issue reported in CI. You might want to use `#ifdef` instead of `#if` to fix it.

YexuanXiao:
> Overall looks good but please fix the issue reported in CI. You might want to use `#ifdef` instead of `#if` to fix it.

If the macro is not defined, it will be replaced with 0. The STL never defines _MSVC_STL_UPDATE as 0, so there is no visible difference here. However, if the code style requires it, I am glad to fix it.

vitaut (approved):

vitaut:
Merged, thanks!

PR #4270, opened 2024-12-19
Allow getting size of dynamic format arg store

hannes-harnisch:
Getting the size of the dynamic format arg store is useful if you use that class to dynamically pass around sets of format arguments and want to perform additional verification.

In my case, I am using it to check whether a message, that I'm applying the format args to, has the correct amount of placeholders and currently I have to track the arg count separately, even though that information is already present in the `fmt::dynamic_format_arg_store`.


vitaut:
Merged, thanks!

PR #4271, opened 2024-12-20
Fix compatibility with older versions of VS

YexuanXiao:
In the previous PR, I mentioned that there were no compatibility issues, but I was wrong. I confused the versions of VS and MSVC. _MSVC_STL_UPDATE appeared later than the VS version that fmt promised to support. This PR will fix this issue.

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -63,13 +63,6 @@
 #else
 #  define FMT_LIBCPP_VERSION 0
 #endif
-#ifdef _MSVC_STL_UPDATE
-#  define FMT_MSVC_STL_UPDATE _MSVC_STL_UPDATE
-#elif defined(_MSC_VER) && _MSC_VER < 1912  // VS 15.5
-#  define FMT_MSVC_STL_UPDATE _MSVC_LANG
-#else
-#  define FMT_MSVC_STL_UPDATE 0
-#endif
 
 // Detect __has_*.
 #ifdef __has_feature
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -65,10 +65,10 @@
 #endif
 #ifdef _MSVC_STL_UPDATE
 #  define FMT_MSVC_STL_UPDATE _MSVC_STL_UPDATE
-#elif defined(_MSC_VER) && _MSC_VER < 1912 // VS 15.5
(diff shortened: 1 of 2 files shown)


include/fmt/base.h
```diff
@@ -58,11 +63,12 @@
...
-#ifdef _MSVC_LANG
-#  define FMT_CPLUSPLUS _MSVC_LANG
+#ifdef _MSVC_STL_UPDATE
+#  define FMT_MSVC_STL_UPDATE _MSVC_STL_UPDATE
+#elif defined(_MSC_VER) && _MSC_VER < 1912  // VS 15.5
+#  define FMT_MSVC_STL_UPDATE _MSVC_LANG
 #else
-#  define FMT_CPLUSPLUS __cplusplus
+#  define FMT_MSVC_STL_UPDATE 0
 #endif
```
vitaut, line 72: Let's move this to `ostream.h` where it is used.

vitaut:
Also to clarify: `_MSVC_STL_UPDATE` is only supported since MSVC 2019, right?

YexuanXiao:
> Also to clarify: `_MSVC_STL_UPDATE` is only supported since MSVC 2019, right?

To be precise, it has been supported since MSVC 19.12. Therefore, for versions from 19.00 to 19.12, use _MSVC_LANG instead. fmt has already declared that it does not support MSVC before version 19.00. Here are two version reference tables: [_MSC_VER](https://learn.microsoft.com/en-us/cpp/overview/compiler-versions?view=msvc-170) and [_MSVC_STL_UPDATE](https://github.com/microsoft/STL/wiki/Macro-_MSVC_STL_UPDATE).

include/fmt/base.h
```diff
@@ -46,6 +46,11 @@
 #else
 #  define FMT_MSC_VERSION 0
 #endif
+#ifdef _MSVC_LANG
+#  define FMT_CPLUSPLUS _MSVC_LANG
+#else
+#  define FMT_CPLUSPLUS __cplusplus
+#endif
```
vitaut, line 53: Could you also revert this unnecessary change to `base.h`?

vitaut (changes requested):
Just one more minor comment, otherwise LGTM.

vitaut (approved):

vitaut:
Merged, thanks.

PR #4276, opened 2024-12-26
Define CMake components to allow docs to be installed separately

carlsmedstad:
Define two Cmake components, core and doc, which can be installed separately. This facilitates packagers who want to package docs in a separate package.

After this change it's possible to install only core files with:

```sh
cmake --install build --component core
```

And only install documentation with:

```sh
cmake --install build --component doc
```

When no component is specified, the behaviour is unchanged, i.e. if documentation was built, it will be installed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -426,7 +426,9 @@ if (FMT_INSTALL)
   endif()
 
   # Install the library and headers.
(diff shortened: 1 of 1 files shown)


vitaut:
Merged, thanks!

Issue #4289, opened 2025-01-03
PR 4286 causes segfault - please revert


ja11sop:
Recent PR https://github.com/fmtlib/fmt/pull/4286 causes segfault.

The following code:

```cpp
std::ostringstream Stream;
std::string Line = "";
Stream << fmt::format( "{}", fmt::styled( Line, fmt::fg( fmt::terminal_color::blue ) ) ) << "\n";
```

segfaults with backtrace:

```log
#0  0x000055555568bbf3 in fmt::v11::detail::buffer<char>::append<char> (this=0x0, begin=begin@entry=0x7fffffffb400 "\033[34m", end=end@entry=0x7fffffffb405 "")
    at /home/user/_download/git_https_github.com__fmtlib_fmt.git/include/fmt/base.h:1797
#1  0x00005555556a69e3 in fmt::v11::detail::copy<char, char const*, fmt::v11::basic_appender<char>, 0> (begin=0x7fffffffb400 "\033[34m", end=0x7fffffffb405 "", out=...)
    at /home/user/_download/git_https_github.com__fmtlib_fmt.git/include/fmt/base.h:2008
#2  0x00005555556984d2 in fmt::v11::formatter<fmt::v11::detail::styled_arg<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, char, void>::format<fmt::v11::context> (
    this=0x7fffffffb480, arg=..., ctx=...) at /home/user/_download/git_https_github.com__fmtlib_fmt.git/include/fmt/color.h:574
#3  0x000055555568bcf1 in fmt::v11::detail::value<fmt::v11::context>::format_custom<fmt::v11::detail::styled_arg<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, fmt::v11::formatter<fmt::v11::detail::styled_arg<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, char, void> > (arg=0x7fffffffbaf0, parse_ctx=..., ctx=...)
    at /home/user/_download/git_https_github.com__fmtlib_fmt.git/include/fmt/base.h:2240
#4  0x000055555570f6af in fmt::v11::basic_format_arg<fmt::v11::context>::handle::format (this=<optimized out>, parse_ctx=..., ctx=...)
    at /home/user/_download/git_https_github.com__fmtlib_fmt.git/include/fmt/base.h:2444
[...]

PR #4290, opened 2025-01-04
Simplify a copying the fill from basic_specs

phprus:
Simplify code.

Fix for #4289

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -842,6 +842,12 @@ class basic_specs {
     for (size_t i = 0; i < size; ++i)
       fill_data_[i & 3] = static_cast<char>(s[i]);
   }
+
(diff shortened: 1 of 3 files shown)


phprus:
@ja11sop , test this PR, please.

ja11sop:
I just ran against this PR and all looks good

phprus:
@vitaut please review this PR.

vitaut:
I wonder why the issue reported in #4289 hasn't been caught by existing tests? It might be worth adding a regression test.

phprus:
@vitaut 

I'm guessing that a previous version of the code might have corrupted memory somewhere earlier.

I don't see the root of case for this error, but I think move the ``fill`` copying into the ``basic_specs`` class as a member function is a nice simplification of the logic.

The code from PR is free of any possibility of memory corruption.

Issue #4294, opened 2025-01-07
MSVC: Conditional expression is constant warning when formatting `tm` with unicode disabled


edo9300:
When formatting a `tm` struct with unicode support disabled, MSVC outputs the warning C4127 on this row https://github.com/fmtlib/fmt/blob/093b39ca5eea129b111060839602bcfaf295125a/include/fmt/chrono.h#L367
this is a sample repro scenario
```cpp
#define FMT_UNICODE 0
#include <fmt/core.h>
#include <fmt/chrono.h>
#include <ctime>

int main() {
	auto now = std::time(nullptr);
	fmt::format("[{:%Y-%m-%d %H:%M:%S}]", fmt::localtime(now));
	return 0;
}
```
(compile with /W4)

vitaut:
We don't actively look into level 4 warnings since they are mostly noise but a PR to fix it would be welcome.

PR #4299, opened 2025-01-08
Address MSVC C4127 warning when formatting non unicode `tm`

edo9300:
Use `const_check` to silence visual studio's W4 level diagnostic regarding conditional expressions being constants, addresses https://github.com/fmtlib/fmt/issues/4294


vitaut:
Thanks

Issue #4297, opened 2025-01-08
Compilation error since 11.1.0 with NVIDIA's nvcc compiler


breyerml:
Using the `CMakeLists.txt` file:
```CMake
cmake_minimum_required(VERSION 3.15)

project(MyProject)

enable_language(CUDA)

add_executable(prog main.cu)

include(FetchContent)
FetchContent_Declare(fmt
                 GIT_REPOSITORY https://github.com/fmtlib/fmt.git
                 GIT_TAG 11.1.0
                 QUIET
)
FetchContent_MakeAvailable(fmt)

target_link_libraries(prog PUBLIC fmt::fmt)
```
with the `main.cu` file 
```c++
#include "fmt/format.h"
#include "fmt/chrono.h"

#include <chrono>
#include <iostream>

int main() {
  std::cout << fmt::format("{}", std::chrono::milliseconds{ 10 }) << std::endl;
  return 0;
}
```
built via `cmake .. && make` together with nvcc version `cuda_12.2.r12.2/compiler.33191640_0` results in the following compiler error:
```
/Documents/test/fmt/build/_deps/fmt-src/include/fmt/chrono.h(2310): error: too many arguments in function call
          detail::utc_clock::to_sys(val), ctx);
                                    ^

/Documents/test/fmt/build/_deps/fmt-src/include/fmt/chrono.h(2310): error: the object has type qualifiers that are not compatible with the member function "fmt::v11::detail::utc_clock::to_sys"
            object type is: const fmt::v11::detail::utc_clock
          detail::utc_clock::to_sys(val), ctx);
          ^

```

This error happens with versions `11.1.0`,`11.1.1`, and `master`.
A `git bisect` revealed that the offending commit is: 89af1ad77df56591c0e0fe83e8729c9656edee67

vitaut:
Adding an argument to the fallback `to_sys` is the correct fix. A PR would be welcome.

breyerml:
I opened a PR.

PR #4301, opened 2025-01-09
Add parameter to the fallback to_sys function

breyerml:
Add parameter to the fallback to_sys function. Fixes issue #4297.


vitaut (changes requested):
Thanks for the PR! It looks good but please apply clang-format.

vitaut:
Merged with a minor tweak in https://github.com/fmtlib/fmt/commit/bfbdc2be9a07eb67b8e084a825eee97f0c6504f3.

PR #4312, opened 2025-01-16
Fix formatting into std::ostreambuf_iterator using a compiled format

phprus:
We need tests for format and for compile time format because it's usage different code path.

Bug was introduced in commit: https://github.com/fmtlib/fmt/commit/f6b4a23b83b36946072398bd04b7c7810df2d213#diff-bdc6f79e8e9f5b4331d66fb785636a87d29f55cf729865e13925b4209424c878


vitaut:
Thanks for the fix!

PR #4319, opened 2025-01-20
Fix GCC 6.4.0 compilation errors

### LINKED ISSUE #4318: GCC 6.4.0 compilation errors
Compiling fmt 11.1.2 using GCC 6.4.0 triggers compilation errors caused by the `bigint` class. Reproducer is [here](https://godbolt.org/z/eobYYWxv8). There are also additional errors related to uninitialized `buffer` of `char` in `constexpr` context that do not appear the repro but are triggered during the compilation of unit tests.

sergiud:
Fixes #4318.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -1213,7 +1213,7 @@ FMT_CONSTEXPR auto format_decimal(OutputIt out, UInt value, int num_digits)
     return out;
   }
   // Buffer is large enough to hold all digits (digits10 + 1).
-  char buffer[digits10<UInt>() + 1];
+  char buffer[digits10<UInt>() + 1]{};
   if (is_constant_evaluated()) fill_n(buffer, sizeof(buffer), '\0');
   do_format_decimal(buffer, value, num_digits);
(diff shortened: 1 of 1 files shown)


include/fmt/format.h
```diff
@@ -1213,7 +1213,7 @@ FMT_CONSTEXPR auto format_decimal(OutputIt out, UInt value, int num_digits)
     return out;
   }
   // Buffer is large enough to hold all digits (digits10 + 1).
-  char buffer[digits10<UInt>() + 1];
+  char buffer[digits10<UInt>() + 1]{};
   if (is_constant_evaluated()) fill_n(buffer, sizeof(buffer), '\0');
```
sergiud, line 1217: In case the unconditional `buffer` initialization is acceptable the `fill_n` invocations can be dropped.

vitaut:
Closing in favor of a different fix (https://github.com/fmtlib/fmt/commit/41539c29f342e67abb925b946acab6d55d331b36) since it's a bug in gcc but thanks for the PR!

PR #4322, opened 2025-01-22
Silence a constexpr warning when compiling with MSVC and /W4

### LINKED ISSUE #4314: Compile failure with MSVC with /W4 /WX - ostream.h(153): warning C4127: conditional expression is constant
The following Code doesn't compile with 'Treat warnings as errors' and warning level 4:

```
#include <fmt/format.h>
#include <fmt/ostream.h>
#include <iostream>

int main()
{
	fmt::print(std::cerr, "Don't {}!", "panic");
}
```

compiling with cl /EHsc /std:c++20 /W4 /WX /utf-8 x.cpp /MDd /I D:\libs\fmt\include D:\libs\fmt\x64\Debug\fmtd.lib gives:

D:\libs\fmt\include\fmt/ostream.h(153): error C2220: the following warning is treated as an error
D:\libs\fmt\include\fmt/ostream.h(153): warning C4127: conditional expression is constant
D:\libs\fmt\include\fmt/ostream.h(153): note: consider using 'if constexpr' statement instead

The problem is this if in print because detail::use_utf8 is a compile time constant
```
if (detail::use_utf8) return vprint(os, fmt.str, vargs);
```

Changing print this way might solve the problem for C++17 and higher but doesn't compile with C++14

```
FMT_EXPORT template <typename... T>
void print(std::ostream& os, format_string<T...> fmt, T&&... args) {
  fmt::vargs<T...> vargs = {{args...}};
  if constexpr( detail::use_utf8 )
  {
          return vprint( os, fmt.str, vargs );
  }
  else
  {
          auto buffer = memory_buffer();
          detail::vformat_to( buffer, fmt.str, vargs );
          detail::write_buffer( os, buffer );
  }
}
```

ZehMatt:
Fixes #4314

### DIFF
--- a/include/fmt/ostream.h
+++ b/include/fmt/ostream.h
@@ -150,7 +150,7 @@ inline void vprint(std::ostream& os, string_view fmt, format_args args) {
 FMT_EXPORT template <typename... T>
 void print(std::ostream& os, format_string<T...> fmt, T&&... args) {
   fmt::vargs<T...> vargs = {{args...}};
-  if (detail::use_utf8) return vprint(os, fmt.str, vargs);
+  if (detail::const_check(detail::use_utf8)) return vprint(os, fmt.str, vargs);
   auto buffer = memory_buffer();
   detail::vformat_to(buffer, fmt.str, vargs);
   detail::write_buffer(os, buffer);


vitaut:
Thank you

PR #4325, opened 2025-01-23
Fix #4303: avoid instantiating formatter<const T>

### LINKED ISSUE #4303: Compile error: partial specialization of struct fmt::v11::formatter
Printing a `const std::filesystem::path` with `fileystem`, `fmt/ostream.h` and `fmt/std.h` included and later including `fmt/ranges.h` in the same compilation unit causes a strange compile error.

Tested with fmt-11.0.0 and gcc-14.2.

```
#include <filesystem>
#include <fmt/ostream.h>
#include <fmt/std.h>

void f()
{
    const auto fn = std::filesystem::path{};
    fmt::print("{}", fn);
}

#include <fmt/ranges.h>
```
https://godbolt.org/z/G5TPbz7o9

timsong-cpp:
Otherwise `format_as_member_result` instantiates `formatter<const path>`.


vitaut:
Thanks for the fix!

PR #4326, opened 2025-01-23
Restore constraint on map formatter

timsong-cpp:
So that it correctly reports as unformattable when the element is.

This was lost when the formatter was split out in 77bfd8499af75c5cff6b42d131b02bfdfbd7d5e0

### DIFF
--- a/test/ranges-test.cc
+++ b/test/ranges-test.cc
@@ -68,8 +68,8 @@ TEST(ranges_test, format_vector) {
   EXPECT_EQ(fmt::format("{:n:n}", vvc), "'a', 'b', 'c', 'a', 'b', 'c'");
   EXPECT_EQ(fmt::format("{:n:n:}", vvc), "a, b, c, a, b, c");
 
-  EXPECT_FALSE((fmt::is_formattable<unformattable>::value));
-  EXPECT_FALSE((fmt::is_formattable<std::vector<unformattable>>::value));
+  EXPECT_FALSE(fmt::is_formattable<unformattable>::value);
+  EXPECT_FALSE(fmt::is_formattable<std::vector<unformattable>>::value);
 }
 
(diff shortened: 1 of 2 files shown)


test/ranges-test.cc
```diff
@@ -65,6 +67,9 @@ TEST(ranges_test, format_vector) {
   EXPECT_EQ(fmt::format("{:n}", vvc), "['a', 'b', 'c'], ['a', 'b', 'c']");
   EXPECT_EQ(fmt::format("{:n:n}", vvc), "'a', 'b', 'c', 'a', 'b', 'c'");
   EXPECT_EQ(fmt::format("{:n:n:}", vvc), "a, b, c, a, b, c");
+
+  EXPECT_FALSE((fmt::is_formattable<unformattable>::value));
```
vitaut, line 71: Are extra parentheses needed here and in other places below?

test/ranges-test.cc
```diff
@@ -83,6 +88,8 @@ TEST(ranges_test, format_map) {
   auto m = std::map<std::string, int>{{"one", 1}, {"two", 2}};
   EXPECT_EQ(fmt::format("{}", m), "{\"one\": 1, \"two\": 2}");
   EXPECT_EQ(fmt::format("{:n}", m), "\"one\": 1, \"two\": 2");
+
+  EXPECT_FALSE((fmt::is_formattable<std::map<int, unformattable>>::value));
```
vitaut, line 92: It seems that this is the only place that needs extra parentheses because of a comma.

test/ranges-test.cc
timsong-cpp: No. They were in the original (I moved this assertion up from the tuple one) so I just went with it, but I can remove the unnecessary ones.

vitaut:
Merged, thanks!

PR #4327, opened 2025-01-24
Fix error of unitialized variable FMT_HEADERS

rlalik:
This happen when using e.g. pedantic mode in cmake-init.

> -- Build type: Coverage
> CMake Error (dev) at build/ci-coverage/_deps/fmt-src/CMakeLists.txt:290 (set):
>   uninitialized variable 'FMT_HEADERS'
> Call Stack (most recent call first):
>   build/ci-coverage/_deps/fmt-src/CMakeLists.txt:298 (add_headers)
> This error is for project developers. Use -Wno-error=dev to suppress it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -293,6 +293,7 @@ function(add_headers VAR)
 endfunction()
 
 # Define the fmt library, its includes and the needed defines.
+set(FMT_HEADERS) # to fix warning of uninitialized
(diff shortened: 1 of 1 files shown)


vitaut:
Merged without a comment since the purpose of `set` is obvious in https://github.com/fmtlib/fmt/commit/01a5b56f0dedbb81701b2746cc245898e6900736, thanks.

PR #4328, opened 2025-01-26
move fmt_detail namespace under the main fmt namespace

odygrd:


odygrd:
This turned out to be more complicated than I initially thought. 😞 
When `fmt_detail` is nested inside the `fmt` namespace, it seems to take precedence over `std::chrono`. Adding `using std::chrono::current_zone` appears to fix the issue at least when trying on windows, but it seems like we'd need an additional preprocessor check specifically for C++20. 
Given the added complexity, I think it’s not worth pursuing this further at the moment. Closing this for now.

PR #4329, opened 2025-01-27
Fix typo in `FMT_HAS_BUILTIN` check (`__buitin_strlen` -> `__buiLtin_strlen`)

LocalSpook:


vitaut:
Good catch, thanks! Merged (just the fix) in https://github.com/fmtlib/fmt/commit/c9267da4dfec806e6137ce68a2ce50d833a27bc6.

PR #4342, opened 2025-02-04
Move `is_compiled_string` to public API

### LINKED ISSUE #4335: It doesn't seem to be possible to declare a funciton that can take compiled format string as an argument without referencing `fmt::detail`
Writing a function that can take a regular format string as an argument is pretty straightforward:
https://github.com/SwooshyCueb/irods/blob/5ceb3b6b99b19478ebb9f1681d3432f1e037472d/server/core/include/irods/notify_service_manager.hpp#L46-L70

However, I can't figure out how to declare an overload that can take a *compiled* format string without dipping into `fmt::detail`:
https://github.com/SwooshyCueb/irods/blob/5ceb3b6b99b19478ebb9f1681d3432f1e037472d/server/core/include/irods/notify_service_manager.hpp#L72-L99

There doesn't seem to be any documentation on how this might be done. Is this the proper solution?

SwooshyCueb:
Fixes #4335

Moves `is_compiled_string` to public API and adds a test for it.

### DIFF
--- a/test/compile-test.cc
+++ b/test/compile-test.cc
@@ -316,6 +316,18 @@ TEST(compile_test, compile_format_string_literal) {
 }
 #endif
 
+#if defined(__cpp_if_constexpr) && defined(__cpp_return_type_deduction)
+template <typename S>
+bool check_is_compiled_string(const S&) {
+  return fmt::is_compiled_string<S>::value;
+}
+
+TEST(compile_test, is_compiled_string) {
+  EXPECT_TRUE(check_is_compiled_string(FMT_COMPILE("asdf")));
+  EXPECT_TRUE(check_is_compiled_string(FMT_COMPILE("{}")));
+}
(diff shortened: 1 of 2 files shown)


test/compile-test.cc
```diff
@@ -316,6 +316,19 @@ TEST(compile_test, compile_format_string_literal) {
 }
 #endif
 
+#if defined(__cpp_if_constexpr) && defined(__cpp_return_type_deduction)
+template <typename S>
+typename fmt::is_compiled_string<S>::value_type check_is_compiled_string(
+    const S&) {
```
vitaut, line 322: `value_type` is just `bool` so let's just use that for simplicity.

SwooshyCueb: Done

vitaut:
Merged, thanks.

PR #4349, opened 2025-02-14
Implement `operator""_cf` in terms of `FMT_COMPILE`

LocalSpook:
This implementation is simpler and makes it clear that `operator""_cf` and `FMT_COMPILE` do the exact same thing.


vitaut:
Good idea, thanks!

PR #4352, opened 2025-02-19
Default zoned time conversions to use earliest time

### LINKED ISSUE #4350: It's possible to get an exception when working with localised times
Environment:
Windows 10 x64
MSVC Version 17.12.4

It;'s possible to pass a `std::chrono::local_seconds` object to `format` that has an ambiguous time, so for example ~2am on April 6th 2025 in Sydney daylight savings ends. 

Sample Code (note, system running must be on AEDT)
```
#include <chrono>
#include <cstdlib>
#include <iostream>
#include <fmt/core.h>
#include <fmt/chrono.h>

int main()
{
    const auto& timeZoneDatabase = std::chrono::get_tzdb();
    const auto& currentZone = timeZoneDatabase.current_zone();
    std::cout << currentZone->name() << '\n'; // This needs to be Australia/Sydney

    uint64_t two_in_the_morning = 1743865205; // just after 2am on 2025/4/6 which is a daylight savings changeover point
    auto dur = std::chrono::duration<uint64_t>(two_in_the_morning);
    auto time_point = std::chrono::time_point<std::chrono::system_clock, std::chrono::duration<uint64_t>>(dur);
    auto time_point_as_local_seconds = currentZone->to_local(std::chrono::floor<std::chrono::seconds>(time_point));

    std::string text = fmt::format("{:%Y-%m-%d %H:%M:%S}", time_point_as_local_seconds);
    std::cout << text << '\n';
    return 0;
}
```
This block here seems to be the cause:
https://github.com/fmtlib/fmt/blob/94ab51cb8c63c93220f0cdd1733bdeef95e77d60/include/fmt/chrono.h#L575-L580

`to_sys` can throw exceptions in these cases, and needs to be passed `std::chrono::choose`

musshorn:
FMT_USE_EARLIEST_TIME is used to control if earliest or latest time is preferred. Tests are a bit tricky to write due to the system time zone being used by CI tools not necessarily having a DST time

Fixes #4350

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -281,6 +281,13 @@ struct local_t {};
 
 }  // namespace detail
 
+// When converting a local zoned time to system time, it could be ambigous
+// By default prefer the earliest time but this can be set to 0 to use latest
+// refer to std::chrono::time_zone::to_sys and std::chrono::choose for more info
+#ifndef FMT_USE_EARLIEST_TIME
+#  define FMT_USE_EARLIEST_TIME 1
+#endif
+
 template <typename Duration>
 using sys_time = std::chrono::time_point<std::chrono::system_clock, Duration>;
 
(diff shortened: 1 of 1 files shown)


vitaut:
Closing in favor or an alternative fix and test (https://github.com/fmtlib/fmt/commit/43e31614cce9f019848c6d234c0fc2ea17396c3c) but thanks for the PR!

PR #4355, opened 2025-02-20
Add .vs folder to .gitignore

dinomight:
Opening the fmt folder as a CMake project with Visual Studio creates this directory. The contents can be ignored.

This is primarily a quality of life tweak for Visual Studio users.


vitaut:
Merged, thanks

Issue #4302, opened 2025-01-10
Avoid "ignore" in fmt headers because they conflict with std::ignore and can cause compiler warnings 


mgalign:
Some developers do this "using namespace std;"
They do this after including headers from the std where std::ignore is defined, and work with fmtlib

However, then there is a conflict due to the int ignore definition which causes a warning (and in some stricter work environments, warning means error or warnings are forbidden)

fmt\base.h(2731,23): error C2220: the following warning is treated as an error . It is due to std::ignore conflicting with the int ignore 

Please remove the term "ignore" in the headers of fmtlib

There are at least two in base.h (see FMT_APPLY_VARIADIC  and in FMT_CONSTEXPR int ignore = ...)

Most likely you could even use std::ignore there

vitaut:
Thanks for the suggestion but I don't think we should accommodate poorly written C++ code.

dinomight:
While the CPP Core Guidelines do say to use `using namespace` sparingly, it does suggest that avoiding the verbosity of `std::` everywhere within a source file is justifiable [0]. The core guidelines do explicitly say to avoid `using namespace` within a header at a global scope, so no argument there.

I bring this up because in my code base, updating to fmt 11.1.3 from 11.0.2 will require a lot of source file updates. Does that suggest the addition of the `ignore` variables should be part of a major release since the minor release lost compile compat?

[0] https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf6-use-using-namespace-directives-for-transition-for-foundation-libraries-such-as-std-or-within-a-local-scope-only

mgexo:
Hello @vitaut ,
so called "poorly" written code can still be important for this world, especially when it comes from highly educated researchers e.g. in medical device industry who are not programming experts but experts in their field. 

I see you as open source library maintainer in the role - especially now that you are maybe close to get into the std with std::format - to be compatible with all codebases, and I hope you will change your mind eventually. 

I still highly appreciate your work, the fmtlib with the compiled strings made our file format reader much faster!

Thanks!

vitaut:
This is just a warning, either ignore it or make sure to not use the directive that triggers the warning before the {fmt}'s includes.

dinomight:
> This is just a warning, either ignore it or make sure to not use the directive that triggers the warning before the {fmt}'s includes.

[...]

PR #4356, opened 2025-02-20
Fix conflict with std::ignore

dinomight:
In situations where `using namespace std;` is used, compiler warnings can be generated because of local variables named `ignore`. This renames those variables to something else to address the name conflict.

#4302

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -294,8 +294,8 @@
 #endif
 
 #define FMT_APPLY_VARIADIC(expr) \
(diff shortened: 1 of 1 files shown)


vitaut:
Please provide a godbolt repro.

vitaut:
Thinking more of it, I'm willing to accept this PR because the proposed name is actually better. But for similar issues in the future you'll need to provide a repro demonstrating that the problem is worth solving.

PR #4361, opened 2025-02-25
Fix dynamic named arg format spec handling

### LINKED ISSUE #4360: Named arguments failing to compile with "unmatched '}' in format string"
Consider the following format call.

```cpp
fmt::format("{enum:<{size}d}", fmt::arg("enum", 1), fmt::arg("size", 10));
```
This results in a compile time failure reporting an "unmatched '}' in format string". This is happening in VS17.14 Preview 1 and with GCC 14.2. I've reproduced the issue over on godbolt using {fmt} from trunk [0]. The above string compiled with 11.0.2 and fails with 11.1.3. I was unable to reproduce this compile failure within the {fmt} test code, so I was wondering if it was somehow fixed. However, checking out the tag for 11.1.3 locally, I was also unable to repro in the test code. Not sure what that's about.

If you wrap the above example into an `fmt::runtime()`, compilation is successful and output is what is expected. The godbolt link includes this example. Additionally, other format variations appear to work both with auto-indexed values as well as when not providing any explicit reference. The following example does compile. It also appears to be related to the dynamic width field as an explicit value there will compile.

```cpp
std::cout << fmt::format(fmt::runtime("'{enum:<{size}d}'\n"), fmt::arg("enum", 1),
        fmt::arg("size", 10));
std::cout << fmt::format("'{enum:<10d}'\n", fmt::arg("enum", 2),
        fmt::arg("size", 10));
std::cout << fmt::format("'{:<{}d}'\n", fmt::arg("enum", 2),
        fmt::arg("size", 10));
std::cout << fmt::format("'{0:<{1}d}'\n", fmt::arg("enum", 3),
        fmt::arg("size", 10));
```

Full error output from GCC 14.2:
```
<source>: In function 'int main()':
<source>:7:29: error: call to consteval function 'fmt::v11::fstring<fmt::v11::detail::named_arg<char, int>, fmt::v11::detail::named_arg<char, int> >("\'{enum:<{size}d}\'\012")' is not a constant expression
    7 |     std::cout << fmt::format("'{enum:<{size}d}'\n", fmt::arg("enum", 1),
      |                  ~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    8 |         fmt::arg("size", 10));
      |         ~~~~~~~~~~~~~~~~~~~~~
In file included from /opt/compiler-explorer/libs/fmt/trunk/include/fmt/format.h:41,
                 from <source>:1:
<source>:7:29:   in 'constexpr' expansion of 'fmt::v11::fstring<fmt::v11::detail::named_arg<char, int>, fmt::v11::detail::named_arg<char, int> >("\'{enum:<{size}d}\'\012")'
/opt/compiler-explorer/libs/fmt/trunk/include/fmt/base.h:2708:53:   in 'constexpr' expansion of 'fmt::v11::detail::parse_format_string<char, format_string_checker<char, 2, 0, true> >(fmt::v11::basic_string_view<char>(((const char*)s)), fmt::v11::detail::format_string_checker<char, 2, 0, true>(fmt::v11::basic_string_view<char>(((const char*)s)), (fmt::v11::fstring<fmt::v11::detail::named_arg<char, int>, fmt::v11::detail::named_arg<char, int> >::arg_pack(), fmt::v11::fstring<fmt::v11::detail::named_arg<char, int>, fmt::v11::detail::named_arg<char, int> >::arg_pack())))'
/opt/compiler-explorer/libs/fmt/trunk/include/fmt/base.h:1624:32: error: 'constexpr void fmt::v11::detail::format_string_checker<Char, NUM_ARGS, NUM_NAMED_ARGS, DYNAMIC_NAMES>::on_error(const char*) [with Char = char; int NUM_ARGS = 2; int NUM_NAMED_ARGS = 0; bool DYNAMIC_NAMES = true]' called in a constant expression
 1624 |         return handler.on_error("unmatched '}' in format string");
      |                ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/opt/compiler-explorer/libs/fmt/trunk/include/fmt/base.h:1713:35: note: 'constexpr void fmt::v11::detail::format_string_checker<Char, NUM_ARGS, NUM_NAMED_ARGS, DYNAMIC_NAMES>::on_error(const char*) [with Char = char; int NUM_ARGS = 2; int NUM_NAMED_ARGS = 0; bool DYNAMIC_NAMES = true]' is not usable as a 'constexpr' function because:
 1713 |   FMT_NORETURN FMT_CONSTEXPR void on_error(const char* message) {
      |                                   ^~~~~~~~
/opt/compiler-explorer/libs/fmt/trunk/include/fmt/base.h:1714:17: error: call to non-'constexpr' function 'void fmt::v11::report_error(const char*)'
 1714 |     report_error(message);
      |     ~~~~~~~~~~~~^~~~~~~~~
/opt/compiler-explorer/libs/fmt/trunk/include/fmt/base.h:673:27: note: 'void fmt::v11::report_error(const char*)' declared here
  673 | FMT_NORETURN FMT_API void report_error(const char* message);
      |                           ^~~~~~~~~~~~
```

[0] https://godbolt.org/z/qvxo5qrTf

dinomight:
When dealing with dynamic named format args, need to account for nested named args when skipping the content of the replacement.

Fixes #4360

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -1707,9 +1707,9 @@ class format_string_checker {
     context_.advance_to(begin);
     if (id >= 0 && id < NUM_ARGS) return parse_funcs_[id](context_);
 
-    // If id is out of range, it means we do not know the type and cannot parse the format at
-    // compile time. Instead, skip over content until we finish the format spec, accounting
-    // for any nested replacements.
+    // If id is out of range, it means we do not know the type and cannot parse
+    // the format at compile time. Instead, skip over content until we finish
+    // the format spec, accounting for any nested replacements.
     auto bracket_count = 0;
     while (begin != end && (bracket_count > 0 || *begin != '}')) {
       if (*begin == '{')
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -1706,7 +1706,19 @@ class format_string_checker {
       -> const Char* {
     context_.advance_to(begin);
     if (id >= 0 && id < NUM_ARGS) return parse_funcs_[id](context_);
-    while (begin != end && *begin != '}') ++begin;
+
+    // If id is out of range, it means we do not know the type and cannot parse the format at
+    // compile time. Instead, skip over content until we finish the format spec, accounting
+    // for any nested replacements.
+    auto bracket_count = 0;
+    while (begin != end && (bracket_count > 0 || *begin != '}')) {
+      if (*begin == '{')
+        ++bracket_count;
+      else if (*begin == '}')
+        --bracket_count;
+
+      ++begin;
+    }
     return begin;
   }
 
--- a/test/format-test.cc
+++ b/test/format-test.cc
@@ -582,6 +582,8 @@ TEST(format_test, named_arg) {
   EXPECT_EQ("1/a/A", fmt::format("{_1}/{a_}/{A_}", fmt::arg("a_", 'a'),
                                  fmt::arg("A_", "A"), fmt::arg("_1", 1)));
   EXPECT_EQ(fmt::format("{0:{width}}", -42, fmt::arg("width", 4)), " -42");
+  EXPECT_EQ(fmt::format("{value:{width}}", fmt::arg("value", -42),
+      fmt::arg("width", 4)), " -42");
   EXPECT_EQ("st",
             fmt::format("{0:.{precision}}", "str", fmt::arg("precision", 2)));
   EXPECT_EQ(fmt::format("{} {two}", 1, fmt::arg("two", 2)), "1 2");


vitaut (changes requested):
Thanks for the PR. I think the correct fix is to disable compile-time checks if there named arguments that are not known at compile time rather than trying to approximately parse.

dinomight:
That would disable arg ID checks also which seems to be an existing supported behavior when you mix static and dynamic named args. See how arg IDs are handled.

https://github.com/fmtlib/fmt/blob/bdbf957b9aaa05ad8e68124979c49d0bba25cb8e/include/fmt/base.h#L1693-L1699

So are you suggesting to remove that as well? What about creating some default parse function that only verifies arg IDs within the format spec that it can verify at compile time?

vitaut:
Merged, thanks!

PR #4362, opened 2025-02-25
Add CMakePresets.json

dinomight:
Just starting off with the entries for Visual Studio 2022 to make it simpler to work in Visual Studio environment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,5 @@ build/
 cmake_install.cmake
 fmt-*.cmake
 fmt.pc
+run-msbuild.bat
+/out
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR but I don't see much value in providing specific presets that will likely be irrelevant to most users.

dinomight:
> Thanks for the PR but I don't see much value in providing specific presets that will likely be irrelevant to most users.

What do you mean specific presets? These are mirroring the Github actions configurations and rely on whatever CXX is in the environment to select the compiler. It helps to more easily create the build environments for folks and helps for trying to replicate a PR action failure. I don't think I put anything in here that is specific to my environment and presets are relied upon by tools like Visual Studio/CLion to setup the build environment and reduce the amount of command line arguments for command line users. Essentially, this should be interesting to _every_ user.

vitaut:
We shouldn't try to duplicate github actions configs. Only a small fraction of users care about those and they normally don't need to replicate exact configs locally.

PR #4363, opened 2025-02-25
Optimize `text_style` using bit packing

LocalSpook:
Currently, `text_style` takes up 20 bytes, but we can pack it into 8.

First, a side benefit of that is a smaller and faster implementation of `text_style::operator|` ([before-and-after](https://godbolt.org/z/xTMxqTh7d)), but that function I imagine is usually evaluated at compile-time anyway.

The main benefit is reduced binary size per function call using colors. This comes from two factors:

- it becomes possible to pass `text_style` through registers ([before-and-after](https://godbolt.org/z/hzo3hWGbW)); and
- `styled_arg` becomes smaller, so less data needs to be copied around ([before-and-after](https://godbolt.org/z/99Kz8xq1j)).

This absolutely breaks ABI. (Should the `v11` namespace be bumped to `v12`?) I'm actually not sure if it breaks API;
it changes the API of `detail::color_type`—an internal type, you would think—but one that's part of the public interface of `text_style`.

The first commit adds more color tests, because I don't feel the existing suite is comprehensive enough to prove that this change doesn't affect semantics.

### DIFF
--- a/include/fmt/color.h
+++ b/include/fmt/color.h
@@ -222,7 +222,7 @@ struct color_type {
     return (value_ & (1 << 25)) != 0;
   }
 
-  FMT_CONSTEXPR auto get_value() const noexcept -> uint32_t {
+  FMT_CONSTEXPR auto value() const noexcept -> uint32_t {
     return value_ & 0xFFFFFF;
   }
 
@@ -370,7 +370,7 @@ template <typename Char> struct ansi_color_escape {
     // sequence.
     if (text_color.is_terminal_color()) {
       bool is_background = esc == string_view("\x1b[48;2;");
-      uint32_t value = text_color.get_value();
+      uint32_t value = text_color.value();
       // Background ASCII codes are the same as the foreground ones but with
       // 10 more.
       if (is_background) value += 10u;
@@ -394,7 +394,7 @@ template <typename Char> struct ansi_color_escape {
     for (int i = 0; i < 7; i++) {
       buffer[i] = static_cast<Char>(esc[i]);
     }
-    rgb color(text_color.get_value());
+    rgb color(text_color.value());
     to_esc(color.r, buffer + 7, ';');
     to_esc(color.g, buffer + 11, ';');
     to_esc(color.b, buffer + 15, 'm');
--- a/include/fmt/color.h
+++ b/include/fmt/color.h
@@ -205,7 +205,7 @@ struct rgb {
 
 namespace detail {
 
-// a bit-packed variant of either an rgb color or a terminal color.
+// a bit-packed variant of an RGB color, a terminal color, or unset color.
 // see text_style for the bit-packing scheme.
 struct color_type {
   FMT_CONSTEXPR color_type() noexcept = default;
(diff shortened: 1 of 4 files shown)


include/fmt/color.h
```diff
@@ -205,121 +205,155 @@ struct rgb {
...
     return *this;
   }
 
-  friend FMT_CONSTEXPR auto operator|(text_style lhs, const text_style& rhs)
+  friend FMT_CONSTEXPR auto operator|(text_style lhs, text_style rhs)
       -> text_style {
     return lhs |= rhs;
   }
 
+  FMT_CONSTEXPR auto operator==(text_style rhs) const noexcept -> bool {
```
LocalSpook, line 305: `operator==` and `operator!=` should really be friends, but I ran into [this GCC bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89325) when doing that.

include/fmt/color.h
```diff
@@ -205,121 +205,155 @@ struct rgb {
...
+  // In the last two cases, the ones we want to catch, the third bit——the
+  // overflow bit——is set. Bingo.
+  //
+  // We must take into account the possible carry bit from the bits
+  // before the discriminator. The only potentially problematic case is
+  // 11 + 00 = 011 (a carry bit would make it 100, not good!), but a carry
+  // bit is impossible in that case, because 00 (unset color) means the
+  // 24 bits that precede the discriminator are all zero.
+  //
+  // This test can be applied to both colors simultaneously.
```
LocalSpook, line 287: This is a very bit hacky algorithm; I hope the detailed explanation compensates for that.

include/fmt/color.h
```diff
@@ -205,121 +205,155 @@ struct rgb {
...
   FMT_CONSTEXPR color_type(terminal_color term_color) noexcept
-      : is_rgb(), value{} {
-    value.term_color = static_cast<uint8_t>(term_color);
+      : value_(static_cast<uint32_t>(term_color) | (3 << 24)) {}
+
+  FMT_CONSTEXPR auto is_terminal_color() const noexcept -> bool {
+    return (value_ & (1 << 25)) != 0;
+  }
+
+  FMT_CONSTEXPR auto get_value() const noexcept -> uint32_t {
```
vitaut, line 225: I suggest dropping `get_` here.

include/fmt/color.h
```diff
@@ -205,121 +205,155 @@ struct rgb {
...
+  // We do that check by adding the styles. Consider what adding does to each
+  // possible pair of discriminators:
+  //    00 + 00 = 000
+  //    01 + 00 = 001
+  //    11 + 00 = 011
+  //    01 + 01 = 010
+  //    11 + 01 = 100 (!!)
+  //    11 + 11 = 110 (!!)
+  // In the last two cases, the ones we want to catch, the third bit——the
+  // overflow bit——is set. Bingo.
```
vitaut, line 279: Clever =)

vitaut (commented):
Thanks for the PR! Overall LGTM, just one minor comment inline.

vitaut:
One potential concern is the ABI but it seems to be OK because all the functions involved are inline or templated.

include/fmt/color.h
LocalSpook: done

LocalSpook:
I cleaned up `fg` and `bg` in the last commit; they were doing a bit of redundant work

include/fmt/color.h
grafikrobot: Bitfield instead?

LocalSpook: Can't use bitfields unfortunately:
https://github.com/fmtlib/fmt/blob/577fd3be883accf8629423ed77fcca8a24bccee2/include/fmt/base.h#L721

PR #4365, opened 2025-02-25
Allow format_to_n to be executed at compile time

Sunday111:
Hello!
Currently `fmt::format_to_n` cannot be executed at compile time. I needed this functionality in one of my personal project and didn't see a reason why wouldn't it be possible so here is a PR for that.

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -1835,7 +1835,7 @@ class fixed_buffer_traits {
 
  public:
   constexpr explicit fixed_buffer_traits(size_t limit) : limit_(limit) {}
-  FMT_CONSTEXPR20 ~fixed_buffer_traits(){};
+  FMT_CONSTEXPR20 ~fixed_buffer_traits() = default;
   constexpr auto count() const -> size_t { return count_; }
   FMT_CONSTEXPR auto limit(size_t size) -> size_t {
     size_t n = limit_ > count_ ? limit_ - count_ : 0;
@@ -1912,7 +1912,7 @@ class iterator_buffer<T*, T, fixed_buffer_traits> : public fixed_buffer_traits,
   FMT_CONSTEXPR explicit iterator_buffer(T* out, size_t n = buffer_size)
       : fixed_buffer_traits(n), buffer<T>(grow, out, 0, n), out_(out) {}
   FMT_CONSTEXPR iterator_buffer(iterator_buffer&& other) noexcept
-      : fixed_buffer_traits(other),
+      : fixed_buffer_traits(static_cast<iterator_buffer&&>(other)),
         buffer<T>(static_cast<iterator_buffer&&>(other)),
         out_(other.out_) {
     if (this->data() != out_) {
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -1835,6 +1835,7 @@ class fixed_buffer_traits {
 
  public:
   constexpr explicit fixed_buffer_traits(size_t limit) : limit_(limit) {}
+  FMT_CONSTEXPR20 ~fixed_buffer_traits(){};
   constexpr auto count() const -> size_t { return count_; }
   FMT_CONSTEXPR auto limit(size_t size) -> size_t {
     size_t n = limit_ > count_ ? limit_ - count_ : 0;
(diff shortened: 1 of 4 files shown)


Sunday111:
After the first failed workflow run fixed the condition when the test runs and remove WIP change I forgot about

test/format-impl-test.cc
```diff
@@ -475,3 +475,18 @@ TEST(format_impl_test, to_utf8) {
   EXPECT_EQ(s, u.str());
   EXPECT_EQ(s.size(), u.size());
 }
+
+#if FMT_USE_CONSTEVAL
+TEST(format_test, format_to_n_constexpr) {
+  // This test doesn't have to be extensive -
+  // it just checks format_to_n can be done in constexpr context
+  constexpr bool result  = []{
+    std::array buffer {'x', 'x', 'x', 'x'};
```
dinomight, line 484: Windows compilation is failing because it needs `#include <array>` for this.

vitaut: I suggest using a C array instead.

include/fmt/base.h
```diff
@@ -1835,6 +1835,7 @@ class fixed_buffer_traits {
 
  public:
   constexpr explicit fixed_buffer_traits(size_t limit) : limit_(limit) {}
+  FMT_CONSTEXPR20 ~fixed_buffer_traits() = default;
```
vitaut, line 1838: Why FMT_CONSTEXPR20 and not FMT_CONSTEXPR?

include/fmt/base.h
```diff
@@ -2392,8 +2397,8 @@ FMT_CONSTEXPR inline auto is_locking() -> bool {
   return locking<T1>::value || is_locking<T2, Tail...>();
 }
 
-FMT_API void vformat_to(buffer<char>& buf, string_view fmt, format_args args,
-                        locale_ref loc = {});
+FMT_API FMT_CONSTEXPR void vformat_to(buffer<char>& buf, string_view fmt,
```
vitaut, line 2400: This shouldn't be constexpr.

test/format-impl-test.cc
```diff
@@ -475,3 +475,18 @@ TEST(format_impl_test, to_utf8) {
   EXPECT_EQ(s, u.str());
   EXPECT_EQ(s.size(), u.size());
 }
+
+#if FMT_USE_CONSTEVAL
+TEST(format_test, format_to_n_constexpr) {
```
vitaut, line 480: This should go to compile-test.

vitaut:
Thanks for the PR but there is already a compile-time version of `format_to_n`: https://github.com/fmtlib/fmt/blob/577fd3be883accf8629423ed77fcca8a24bccee2/include/fmt/compile.h#L497-L498. This one is optimized for runtime and shouldn't be made `constexpr` (at least not yet).

include/fmt/base.h
Sunday111: It seems that destructors cannot be marked constexpr before C++ 20

include/fmt/base.h
Sunday111: Yes, this forward declaration confused me at first and I thought an addition macro will be necessary (`FMT_CONSTEXPR_IF_HEADERONLY` 😄 ) but it is defined in the same header.

Sunday111:
@dinomight 

> I suppose that's why in your test, you do multiple calls to fmt::format_to_n and also avoid any format specifiers. 😄
> I would say it reduces how interesting this is, it only allows a format string of {} after all, and may be why it wasn't originally available. I think it's neat that it works, though.

Well, I didn't avoid complex format strings intentionally - just wanted to make it work and see your suggestions about what cases should be working to make it decent or a reason why you don't want/need it.

Yes, but 

> Thanks for the PR but there is already a compile-time version of format_to_n

I am not sure I follow which version you mean. I tried a few different ways ([like this](https://godbolt.org/z/Gr6dxKaf8) to get any formatting happening at compile time with no success. So I would appreciate any suggestions 🙏.

vitaut:
Looks like `format_to_n` version that I linked to is not constexpr at the moment but a PR to make it constexpr would be welcome. Alternatively you can use `format_to` from `fmt/compile.h`, possibly with a truncating iterator, since the latter is already constexpr.

Sunday111:
@vitaut I see. Yes, it is not constexpr, and constexpr implementation of back_inserter would be required to make it constexpr. With `format_to`, I will also need to write a custom iterator because I want to know how many bytes have been written and how many are required for successful rendering. 

So I have to write an iterator either way, it seems.
Thank you for pointers, at least I have some options now!

vitaut:
`back_insert_iterator` is constexpr in C++20.

Issue #4282, opened 2025-01-01
Duplicated named arguments are ignored, but not rejected


lewkok:
Hi,
Is example like:
```
#include <fmt/format.h>

int main() {
  using namespace fmt::literals;
  fmt::print("{abc} {1}", "abc"_a = 42, "abc"_a = 43 );
}
```
This prints: "42 43"

Or this example:
```
#include <fmt/format.h>

int main() {
  using namespace fmt::literals;
  fmt::print("{abc}", "abc"_a = 42, "abc"_a = 43 );
}
```
This prints: "42"

Is this semantics like this by design or did it happen by mistake?

vitaut:
That's just an implementation artifact and I think we could add validation.

dinomight:
Do you know what behavior you're expecting here? I started playing around with this idea a little bit and can get it to throw an exception at runtime if two named args have the same name. It doesn't seem like it can be caught at compile time without using static named args. Is that the expectation?

vitaut:
> Do you know what behavior you're expecting here?

Duplicate should give an exception.

dinomight:
At the moment, I can verify the following:
```cpp
fmt::print("{bar}", "bar"_a=42, "bar"_a=43);  // Fails to compile due to duplicate args.
fmt::print("{bar}", fmt::args("bar", 42), fmt::args("bar", 43));  // Throws exception at runtime due to duplicate args.
```
What might be helpful is some direction on the static compiled strings because the following still shows the existing behavior with my changes:
```cpp
fmt::print(FMT_COMPILE("{bar}"), "bar"_a=42, "bar"_a=43);  // Compiles and outputs `42`.
```
I'm assuming if we can stop compile for the non-type template parameters, it should also work for compiled format strings?

vitaut:
> There's another issue tracking FMT_COMPILE: https://github.com/fmtlib/fmt/issues/4124 I would assume as that is resolved, perhaps it could handle this case as well.

Right.

Issue #4124, opened 2024-08-17
Missing named arguments do not cause compilation error when using FMT_COMPILE


schaerfo:
Specifying a named argument in the format string but not providing a value does not result in a compilation error on recent GCC and Clang versions when the format string is compiled using `FMT_COMPILE`.

This is the case both when the argument is missing altogether (1) and when it is present but no value is assigned (2).

When the `FMT_COMPILE` macro is removed, both `fmt::format` calls result in compilation errors.

```c++
#include <fmt/format.h>
#include <fmt/compile.h>

int main() {
    using namespace fmt::literals;
    fmt::format(FMT_COMPILE("{x}")); // 1
    fmt::format(FMT_COMPILE("{x}"), "x"_a); // 2
    return 0;
}
```

On Compiler Explorer: https://godbolt.org/z/4zGrofeoE

dinomight:
Duplicate named args are also not verified when using `FMT_COMPILE` strings for either static or dynamic named args.
```cpp
fmt::print(FMT_COMPILE("{a}"), "a"_a = 42, "a"_a = 43);  // compiles and prints 42.
fmt::print(FMT_COMPILE("{a}"), fmt::arg("a", 42), fmt::arg("a", 43));  // compiles and prints 42.
```

quadrisufiyan:
fmt::print(FMT_COMPILE("{a}"), "a"_a = 42, "a"_a = 43);  // compiles and prints 42.
fmt::print(FMT_COMPILE("{a}"), fmt::arg("a", 42), fmt::arg("a", 43));  // compiles and

jaeheonshim:
I'd like to take a crack at this if nobody else is working on it at the moment. @vitaut, I was wondering what you meant by "a more general check" in (closed) PR #4555? Which format errors should be guaranteed compile time errors?

vitaut:
Thanks for looking into it! I meant that the cases like

```c++
fmt::format(FMT_COMPILE("{a} {b}"), "b"_a=42);
```

should also produce an error. https://github.com/fmtlib/fmt/pull/4555 only gave an error if there were no named arguments at all.

vitaut:
Fixed by #https://github.com/fmtlib/fmt/pull/4638. Thanks @jaeheonshim!

PR #4367, opened 2025-02-26
Report error on duplicate named arg names

dinomight:
Compile error will happen if static named args are used and enabled. Runtime exception will be thrown in other cases.
#4282

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -1060,20 +1060,18 @@ template <typename... Args> constexpr auto count_static_named_args() -> int {
 }
 
 template <typename Char> struct named_arg_info {
-  FMT_CONSTEXPR named_arg_info() : name(nullptr), id(0) {}
-  FMT_CONSTEXPR named_arg_info(const Char* a_name, const int an_id)
-      : name(a_name), id(an_id) {}
   const Char* name;
   int id;
 };
 
 template <typename Char>
 FMT_CONSTEXPR void check_for_duplicate(
-    const named_arg_info<Char>* const named_args, const int named_arg_index,
-    const Char* const arg_name) {
-  const basic_string_view<Char> arg_name_view(arg_name);
+    named_arg_info<Char>* named_args, int named_arg_index,
+    basic_string_view<Char> arg_name) {
+  if (named_arg_index <= 0) return;
+
   for (auto i = 0; i < named_arg_index; ++i) {
-    if (basic_string_view<Char>(named_args[i].name) == arg_name_view) {
+    if (basic_string_view<Char>(named_args[i].name) == arg_name) {
       report_error("duplicate named args found");
     }
   }
@@ -1087,7 +1085,7 @@ void init_named_arg(named_arg_info<Char>*, int& arg_index, int&, const T&) {
 template <typename Char, typename T, FMT_ENABLE_IF(is_named_arg<T>::value)>
 void init_named_arg(named_arg_info<Char>* named_args, int& arg_index,
                     int& named_arg_index, const T& arg) {
-  check_for_duplicate(named_args, named_arg_index, arg.name);
+  check_for_duplicate<Char>(named_args, named_arg_index, arg.name);
   named_args[named_arg_index++] = {arg.name, arg_index++};
 }
 
@@ -1101,7 +1099,7 @@ template <typename T, typename Char,
           FMT_ENABLE_IF(is_static_named_arg<T>::value)>
 FMT_CONSTEXPR void init_static_named_arg(named_arg_info<Char>* named_args,
                                          int& arg_index, int& named_arg_index) {
-  check_for_duplicate(named_args, named_arg_index, T::name);
+  check_for_duplicate<Char>(named_args, named_arg_index, T::name);
   named_args[named_arg_index++] = {T::name, arg_index++};
 }
 
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -1072,7 +1072,8 @@ template <typename Char, typename T, FMT_ENABLE_IF(is_named_arg<T>::value)>
 void init_named_arg(named_arg_info<Char>* named_args, int& arg_index,
                     int& named_arg_index, const T& arg) {
   for (auto i = 0; i < named_arg_index; ++i) {
-    if (basic_string_view<Char>(named_args[i].name) == basic_string_view<Char>(arg.name)) {
+    if (basic_string_view<Char>(named_args[i].name) ==
+        basic_string_view<Char>(arg.name)) {
       report_error("duplicate named args found");
     }
   }
@@ -1090,7 +1091,8 @@ template <typename T, typename Char,
 FMT_CONSTEXPR void init_static_named_arg(named_arg_info<Char>* named_args,
                                          int& arg_index, int& named_arg_index) {
   for (auto i = 0; i < named_arg_index; ++i) {
-    if (basic_string_view<Char>(named_args[i].name) == basic_string_view<Char>(T::name)) {
+    if (basic_string_view<Char>(named_args[i].name) ==
+        basic_string_view<Char>(T::name)) {
       report_error("duplicate named args found");
     }
   }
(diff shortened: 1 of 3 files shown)


include/fmt/base.h
```diff
@@ -1071,6 +1071,12 @@ void init_named_arg(named_arg_info<Char>*, int& arg_index, int&, const T&) {
 template <typename Char, typename T, FMT_ENABLE_IF(is_named_arg<T>::value)>
 void init_named_arg(named_arg_info<Char>* named_args, int& arg_index,
                     int& named_arg_index, const T& arg) {
+  for (auto i = 0; i < named_arg_index; ++i) {
+    if (basic_string_view<Char>(named_args[i].name) ==
+        basic_string_view<Char>(arg.name)) {
+      report_error("duplicate named args found");
+    }
+  }
```
vitaut, line 1079: Please move it to a separate function in `detail` and reuse here and below.

include/fmt/base.h
```diff
@@ -1071,6 +1071,12 @@ void init_named_arg(named_arg_info<Char>*, int& arg_index, int&, const T&) {
 template <typename Char, typename T, FMT_ENABLE_IF(is_named_arg<T>::value)>
 void init_named_arg(named_arg_info<Char>* named_args, int& arg_index,
                     int& named_arg_index, const T& arg) {
+  for (auto i = 0; i < named_arg_index; ++i) {
+    if (basic_string_view<Char>(named_args[i].name) ==
+        basic_string_view<Char>(arg.name)) {
```
vitaut, line 1076: It's better to move this out of the loop to avoid recomputing length.

test/compile-error-test/CMakeLists.txt
```diff
@@ -209,6 +209,14 @@ if (CMAKE_CXX_STANDARD GREATER_EQUAL 20)
...
     #endif
   " ERROR)
+  expect_compile(format-string-duplicate-name-error "
+    #if defined(FMT_HAS_CONSTEVAL) && FMT_USE_NONTYPE_TEMPLATE_ARGS
+      using namespace fmt::literals;
+      fmt::print(\"{bar}\", \"bar\"_a=42, \"bar\"_a=43);
+    #else
+      #error
+    #endif
+  " ERROR)
```
vitaut, line 219: These tests are expensive and we don't need to use them here, please remove.

vitaut (changes requested):
Thanks for the PR!

vitaut:
> Should I file a separate ticket to track?

Just add a comment in #4124 since it's part of general named argument support in format string compilation.

test/compile-error-test/CMakeLists.txt
dinomight: What are the situations where you'd want to add these? I figured it was important to verify that compile fails in this situation.

dinomight:
I'm not sure why the refactoring made GCC13 unhappy, but it's complaining about possibly uninitialized values being referenced. I can initialize those values in the field and make GCC13 happy, but then GCC10 with C++11 breaks because of some initializer list-style initializations. I'm going to try and add a constructor to make everybody happy and you can let me know if there is a better way.

include/fmt/base.h
```diff
@@ -1060,17 +1060,34 @@ template <typename... Args> constexpr auto count_static_named_args() -> int {
 }
 
 template <typename Char> struct named_arg_info {
+  FMT_CONSTEXPR named_arg_info() : name(nullptr), id(0) {}
+  FMT_CONSTEXPR named_arg_info(const Char* a_name, const int an_id)
+      : name(a_name), id(an_id) {}
```
vitaut, line 1065: Initialization is done in `init*named_arg`, we shouldn't duplicate it here.

include/fmt/base.h
```diff
@@ -1084,6 +1101,7 @@ template <typename T, typename Char,
           FMT_ENABLE_IF(is_static_named_arg<T>::value)>
 FMT_CONSTEXPR void init_static_named_arg(named_arg_info<Char>* named_args,
                                          int& arg_index, int& named_arg_index) {
+  check_for_duplicate(named_args, named_arg_index, T::name);
```
vitaut, line 1104: I wonder if calling `check_for_duplicate` only if `named_arg_index` is nonzero fixes the uninitialized warning.

include/fmt/base.h
```diff
@@ -1060,17 +1060,34 @@ template <typename... Args> constexpr auto count_static_named_args() -> int {
...
+      : name(a_name), id(an_id) {}
   const Char* name;
   int id;
 };
 
+template <typename Char>
+FMT_CONSTEXPR void check_for_duplicate(
+    const named_arg_info<Char>* const named_args, const int named_arg_index,
+    const Char* const arg_name) {
+  const basic_string_view<Char> arg_name_view(arg_name);
```
vitaut, line 1074: You can simplify this by passing arg_name as a string_view.

include/fmt/base.h
```diff
@@ -1060,17 +1060,34 @@ template <typename... Args> constexpr auto count_static_named_args() -> int {
...
+  FMT_CONSTEXPR named_arg_info() : name(nullptr), id(0) {}
+  FMT_CONSTEXPR named_arg_info(const Char* a_name, const int an_id)
+      : name(a_name), id(an_id) {}
   const Char* name;
   int id;
 };
 
+template <typename Char>
+FMT_CONSTEXPR void check_for_duplicate(
+    const named_arg_info<Char>* const named_args, const int named_arg_index,
```
vitaut, line 1072: Let's not overuse const - it just clutters the code without bringing much benefit when the scope is small.

vitaut (changes requested):

test/compile-error-test/CMakeLists.txt
vitaut: It is basically an integration test that checks that the internal checking machinery translates into actual compiler errors. It shouldn't be used for unit testing.

include/fmt/base.h
dinomight: I'm going to try your idea of only calling if the index is greater than zero, but if that doesn't work, I need the default construct and if I have the default construct, GCC10 will need the other one.

include/fmt/base.h
dinomight: I just want my mutable by default like rust 😄 . This is also an area I disagree with the CPP core guidelines. I'll begrudgingly update this, though.

test/compile-error-test/CMakeLists.txt
dinomight: I guess I'm a little confused because that's what this is doing. I wrote a unit test for the runtime exception version. I figured since there's also a compile time failure, a test would be required to make sure it functions, especially since the two routes are different in the code.

include/fmt/base.h
dinomight: Both functions have been removed. The if check appeased GCC13, it seems.

vitaut:
Merged, thanks. One drawback of this is quadratic complexity but this can be addressed later.

test/compile-error-test/CMakeLists.txt
vitaut: You could unit test that exception is thrown but I think it's fine as is.

dinomight: I did add a test to verify dynamic named args cause exceptions at runtime. In order to compile the above lines and verify it throws would require I only enable the test when we're _not_ using non-type template parameters. In that situation, it just devolves to the dynamic named args and we don't test the static, compile-time check machinery, right?

vitaut: Compile-time check machinery just translates exceptions to compile-time errors and it's (roughly) the same for all errors so it doesn't need to be tested multiple times. It's sufficient to test the exception path.

include/fmt/base.h
```diff
@@ -1064,13 +1064,28 @@ template <typename Char> struct named_arg_info {
   int id;
 };
 
+template <typename Char>
+FMT_CONSTEXPR void check_for_duplicate(named_arg_info<Char>* named_args,
+                                       int named_arg_index,
+                                       basic_string_view<Char> arg_name) {
+  if (named_arg_index <= 0) return;
```
vitaut, line 1071: As it turned out the actual change that worked around the gcc bug was removing const in `named_args`, not adding the check =)

dinomight: OK, that's kind of funny. Which const was it, cause I had a lot 😄 Makes me curious why the const additions made it seem like it might not be initialized to GCC. Maybe it just generated some weird assumptions in the compiler checks as part of the inlining process?

vitaut: `const named_arg_info<Char>*`, looks like https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100417

PR #4377, opened 2025-03-06
Suppress a "loop not reachable" warning in nvcc

KyleFromNVIDIA:
This compile-time optimization causes nvcc to throw a "loop not reachable" warning. Suppress it.

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -465,6 +465,9 @@ static_assert(!FMT_UNICODE || use_utf8,
 template <typename T> constexpr const char* narrow(const T*) { return nullptr; }
 constexpr FMT_ALWAYS_INLINE const char* narrow(const char* s) { return s; }
 
+#if defined(__NVCC__)
+#  pragma nv_diag_suppress 128
+#endif
 template <typename Char>
 FMT_CONSTEXPR auto compare(const Char* s1, const Char* s2, std::size_t n)
     -> int {
@@ -475,6 +478,9 @@ FMT_CONSTEXPR auto compare(const Char* s1, const Char* s2, std::size_t n)
   }
   return 0;
 }
+#if defined(__NVCC__)
+#  pragma nv_diag_default 128
+#endif
 
 namespace adl {
 using namespace std;


vyasr:
For context, this CMake script will demonstrate the problem by building [nvbench](https://github.com/NVIDIA/nvbench)
```
  cmake_minimum_required(VERSION 3.26.4 FATAL_ERROR)

  project(test_nvbench_fmt LANGUAGES CXX)

  set(RAPIDS_VERSION_MAJOR_MINOR "25.04")

  if(NOT EXISTS "${CMAKE_CURRENT_BINARY_DIR}/RAPIDS-${RAPIDS_VERSION_MAJOR_MINOR}.cmake")
    file(
      DOWNLOAD
      "https://raw.githubusercontent.com/rapidsai/rapids-cmake/branch-${RAPIDS_VERSION_MAJOR_MINOR}/RAPIDS.cmake"
      "${CMAKE_CURRENT_BINARY_DIR}/RAPIDS-${RAPIDS_VERSION_MAJOR_MINOR}.cmake"
    )
  endif()

  set(CPM_DOWNLOAD_fmt ON)
  include("${CMAKE_CURRENT_BINARY_DIR}/RAPIDS-${RAPIDS_VERSION_MAJOR_MINOR}.cmake")

  include(${rapids-cmake-dir}/cpm/nvbench.cmake)

  rapids_cpm_init()
  rapids_cpm_nvbench(BUILD_STATIC)
```

The resulting warnings look like:
```
/home/coder/local/testing/nvbench_fmt/build/_deps/fmt-src/include/fmt/base.h(471): warning #128-D: loop is not reachable
    for (; n != 0; ++s1, ++s2, --n) {
    ^
          detected during:
            instantiation of "auto fmt::v11::detail::compare(const Char *, const Char *, std::size_t)->int [with Char=char]" at line 583
            instantiation of "auto fmt::v11::basic_string_view<Char>::compare(fmt::v11::basic_string_view<Char>) const->int [with Char=char]" at line 591
            instantiation of class "fmt::v11::basic_string_view<Char> [with Char=char]" at line 1929
            instantiation of "auto fmt::v11::basic_format_args<Context>::get_id(fmt::v11::basic_string_view<Char>) const->int [with Context=fmt::v11::context, Char=char]" at line 1969

Remark: The warnings can be suppressed with "-diag-suppress <warning-number>"
```

vitaut:
Thanks for the PR but we try to avoid cluttering the code with compiler-specific pragmas. You can suppress the warning with `FMT_SYSTEM` or other means. If there is a simple workaround that doesn't require pragmas that might be acceptable too.

dinomight:
@KyleFromNVIDIA You might try tinkering with the `if` statement itself around here:

https://github.com/fmtlib/fmt/blob/6fdf225a32625389c53e99e4d2b928df24a0b124/include/fmt/base.h#L472-L477

The if statement is likely known at compile time and is causing the issue. If you can use some of the `FMT_CONSTEXPR` macros to switch it to an `if constexpr` with `else` statement, the warning may also go away.

PR #4354, opened 2025-02-20
Fix warning generated by VS17.4 related to formattable concept

dinomight:
In VS17.4, the `formattable` local variable hides a concept of the same name. This causes a C4459 warning to be triggered. Renaming the local variable fixes the warning.

Fixes #4353

PR #4378, opened 2025-03-08
Suppress `formattable hides the global definition` warning 

evanwporter:
This fixes the warning that is received on MSVC v17.3.6 with the compiler flag `/W4`. The warning is as follows: `declaration of 'formattable' hides global declaration`. It appears on every file where `fmt/core.h` is included.

~~I renamed it to `is_formattable` but this can be changed since it seems like `is_formattable` is used in other places as well.~~

EDIT: After looking through other PRs, namely #4377, I thought it may be preferred (and less breaking) to suppress the warnings via macros rather than change existing code.

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -1662,10 +1662,10 @@ template <typename T, typename Char>
 FMT_VISIBILITY("hidden")  // Suppress an ld warning on macOS (#3769).
 FMT_CONSTEXPR auto invoke_parse(parse_context<Char>& ctx) -> const Char* {
   using mapped_type = remove_cvref_t<mapped_t<T, Char>>;
-  constexpr bool formattable =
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR but we try to avoid cluttering the code with compiler-specific pragmas.

evanwporter:
@vitaut 

Idk if its possible for me to reopen but I renamed `formattable` to `is_it_formattable`. This will get rid of MSVC warning [4459](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4459?view=msvc-170). At the same time it avoids conflicts with the already named variable `is_formattable`

dinomight:
@evanwporter  See #4354 . The issue should go away in newer compilers, even if the fix for older compilers isn't significant.

PR #4383, opened 2025-03-12
Add support for incomplete types

### LINKED ISSUE #3180: Support Incomplete Types, somehow
Hi,

I'd like to come upwith a requirement to support the combination of fmt with incomplete types, that have been specialized in formatter.

To clarify the use case, an example might help
```cpp
class IncompleteType;

namespace FmtHelper
{
  fmt::appender FormatTo(fmt::appender out, IncompleteType const&); // resides in linked code
};

namespace fmt
{
   template <typename charT> 
   struct formatter<IncompleteType, charT, void>
   {
        template <typename ctxT>
        constexpr auto parse(ctxT& ctx) -> decltype(ctx.begin())
        {
            return ctx.begin();
        }

        template <typename ctxT>
        constexpr auto format(IncompleteTypeconst& x, ctxT& ctx) noexcept
        {
            return FmtHelper::FormatTo(ctx.out(), x);
        }
   };
};
```

Rationale
* `IncompleteType` injects evil headers (i.e. WinApi or others).
* `IncompleteType` is from a 3rdparty lib
* `IncompleteType` should be usable in headers in combination with fmt, without the need to have full type knowledge, thus usage is not in  implementation files only.

Limitations
* it is ok, to have only support for the standard format_context and buffer, therefore `fmt::appender` is enough and other overloads may be added, if requirements arise.

Currently
* it is not possible to do it in that way, because type reflection vi is_base_of for compile time strings and sophisticated noexcept specifier reflection dont allo to use incomplete types.
* im am fully aware, that this is not a common use case and the requirement comes more from productive environments with extremly huge and divers code bases and tech stacks, including microsoft specific extensions like c+/cli and others.

Workaround
* we currently wrap such types in another type, whcih is fully known at compile time and redirects to the implementation, but this requires client code to wrap such arguments and thats not that "nice".

[godbolt: simulation](https://godbolt.org/z/GKv3jzc4j)

LocalSpook:
This just requires replacing the `view` base with an `is_view` type trait that can be queried on incomplete types.

Fixes #3180.


vitaut (changes requested):
Thanks for the PR. Overall looks good but let's not introduce a new test binary just for this and instead add a test case to format-test (I don't think we need to test interaction with color).

LocalSpook:
Done

vitaut (changes requested):
Thinking more of it, it's quite convenient to inherit from `view`. I suggest keeping it by defaulting `is_view` to a SFINAE-friendly equivalent to `is_base_of`, something like:

```
struct view {};

template <typename T, typename Enable = void>
struct is_view : std::false_type {};

template <typename T>
struct is_view<
  T, std::void_t<decltype(static_cast<view&>(std::declval<T&>()))>>
    : std::true_type {};
```

vitaut:
Merge with tweaks to keep the opt-in simple in https://github.com/fmtlib/fmt/commit/c70913835978492f46c4f00518f7df8e5b02cc5a. Thanks!

PR #4388, opened 2025-03-13
Unreachable code & constexpr constructor bigint()

jonasbhjulstad:
Proprietary, strict compiler complains about the following:

Unreachable code: certain branches in `fmt/base.h` are unreachable.
Fix: Add branches

Constexpr constructor constructing non-constexpr member: 
Fix: make constexpr constructor default.

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -468,10 +468,19 @@ constexpr FMT_ALWAYS_INLINE const char* narrow(const char* s) { return s; }
 template <typename Char>
 FMT_CONSTEXPR auto compare(const Char* s1, const Char* s2, std::size_t n)
     -> int {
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR but fixing all possible warnings especially on compilers we don't have access to is a non-goal. I recommend suppressing the warning on your side instead.

include/fmt/base.h
```diff
@@ -542,10 +551,11 @@ template <typename Char> class basic_string_view {
       size_ = __builtin_strlen(detail::narrow(s));
       return;
     }
-#endif
+#else
```
dinomight, line 554: This is likely not a good idea because you would completely skip the `strlen` if `Char` is not a `char`, say a `wchar_t`. The better answer would involve some kind of `if constexpr` setup so that you have a fallback for wide chars.

PR #4391, opened 2025-03-17
Compilation Error with MSVC (14.43.34808) with -std:c++latest

ankurvdev:
Fixes the following error with MSVC
```
d:\vcpkg\installed\arm64-msvc-static-md\include\fmt\base.h(2302): error C2446: ':': no conversion from 'std::tuple<T1,T2>' to 'unsigned __int64'
        with
        [
            T1=fmt::v11::detail::<unnamed-enum-is_unpacked_bit>,
            T2=size_t
        ]
d:\vcpkg\installed\arm64-msvc-static-md\include\fmt\base.h(2302): note: the template instantiation context (the oldest one first) is
D:\optimlir\proto\CodeTemplates.h(103): note: see reference to function template instantiation 'std::string fmt::v11::format<fmt::v11::join_view<std::_Vector_const_iterator<std::_Vector_val<std::_Simple_types<_Ty>>>,std::_Vector_const_iterator<std::_Vector_val<std::_Simple_types<_Ty>>>,char>>(fmt::v11::fstring<fmt::v11::join_view<std::_Vector_const_iterator<std::_Vector_val<std::_Simple_types<_Ty>>>,std::_Vector_const_iterator<std::_Vector_val<std::_Simple_types<_Ty>>>,char>>,fmt::v11::join_view<std::_Vector_const_iterator<std::_Vector_val<std::_Simple_types<_Ty>>>,std::_Vector_const_iterator<std::_Vector_val<std::_Simple_types<_Ty>>>,char> &&)' being compiled
        with
        [
            _Ty=std::string
        ]
```

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -2323,7 +2323,7 @@ constexpr auto encode_types() -> unsigned long long {
 template <typename Context, typename... T, size_t NUM_ARGS = sizeof...(T)>
 constexpr auto make_descriptor() -> unsigned long long {
   return NUM_ARGS <= max_packed_args ? encode_types<Context, T...>()
-                                     : is_unpacked_bit | NUM_ARGS;
+                                     : static_cast<unsigned long long>(is_unpacked_bit) | NUM_ARGS;
 }
 
 template <typename Context, int NUM_ARGS>


vitaut:
This looks like a compiler bug since there is no `std::tuple<T1,T2>` there, please report to Microsoft. Also you were probably testing an old version since currently the underlying type is already `unsigned long long` and the cast has no effect.

ankurvdev:
@vitaut 
Apologies if i'm misinterpreting but i dont think its a compiler bug.
I think enum implicit conversion to int was deprecated and now removed in the latest compiler which is causing this issue.

https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/c5054?view=msvc-170
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1120r0.html#dx-arithmetic-conversion-on-enumerations-deprarithconvenum

dinomight:
Likely need more information like what the code looks like that's instantiating the fmt templates. Just messing around with the enum idea on godbolt [0] suggests that it only applies when using an operator with two different enums, not between an enum and a size_t, as is the case here. That goes along with your paper link as that describes enum/enum or enum/float operations as being deprecated, but not enum/int. clang 20 appears to agree with the latest version of MSVC on godbolt with how these enums are interacting with operators.

[0] https://godbolt.org/z/jGv1q8dzj

ankurvdev:
@dinomight you compiled with c++20. This is only deprecated in c++20 

You code fails to compile with c++26 

```
example.cpp
<source>(8): warning C5054: operator '|': deprecated between enumerations of different types
ASM generation compiler returned: 0
example.cpp
<source>(8): warning C5054: operator '|': deprecated between enumerations of different types
Execution build compiler returned: 0
Program returned: 32
```

https://godbolt.org/z/5brhMaGWb

dinomight:
> @dinomight you compiled with c++20. This is only deprecated in c++20

Correct, but the warnings were still generated under C++20 for both compilers (line 8). My point was that the enum/int combination (line 7) is _not_ a warning and that's the one that mirrors the enum/size_t in this PR and has me wondering how you are instantiating the fmt templates.

ankurvdev:
Thanks sorry I missed that earlier (line 7 is being successfully compiled)
I'll report this to the MSVC compiler folks and see if i can get a concise godbolt repro for the issue

PR #4401, opened 2025-04-01
Enable moo-dern C++

vitaut:
Example:

```c++
#include <fmt/format.h>

int main() {
  fmt::cowsay(
    "I'm udderly impressed by {}! Goodbye, {}, hello, moo-dern C++! 🐂",
    "{fmt}", "iostreams");
}
```

Output:

image

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -14,6 +14,7 @@
 #  include <climits>
 #  include <cmath>
 #  include <exception>
+#  include <vector>
 #endif
 
 #if defined(_WIN32) && !defined(FMT_USE_WRITE_CONSOLE)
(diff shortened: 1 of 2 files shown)


AldanTanneo:
# :cow:

include/fmt/format-inl.h
```diff
@@ -1944,6 +1945,41 @@ FMT_FUNC auto is_printable(uint32_t cp) -> bool {
 
 }  // namespace detail
 
+FMT_FUNC void vcowsay(string_view fmt, format_args args) {
+  std::string message = vformat(fmt, args);
+  std::vector<std::string> lines;
```
foonathan, line 1950: Can this be optimized? I need to have cows speak at GB/s on an embedded system, so doing an eager formatting followed by allocation and many string copies isn't acceptable. Can you do something like formatting to an output iterator that takes care of chunking?

foonathan (changes requested):

jevinskie:
Finally I can replace my Enterprise-grade Perl scripts!

include/fmt/format-inl.h
marcusmueller: I actually had that use case, hence me asking https://stackoverflow.com/questions/70280141/how-to-figure-out-the-length-of-the-result-of-fmtformat-without-running-it back in the day, to avoid having to keep a Gigacharacter of text in lines in memory, just to be able to pretty print it with a box around it. My users demand swift display for human consumption!

Instead of an vector of actual wrapped lines, the obvious solution would've been to finalize `template<T,U> fmt::ranges_v2::tokenized_view(T fmt, U delimiter_fmt, fmt_args..., delimiter_args...)` and `fmt::ranges_v2::line_breaking_view(unsigned length)` and `fmt::ranges_v2::experimental::iterative_format_view(T fmt)`, to allow C++27 to lazily consteval the operations "break lines at spaces after latest 40 characters, and then put the result in centered between vertical box drawing lines". 

It would have been the obvious and only sensible solution. Yet we get this bovine crap. I want my money back!

PR #4405, opened 2025-04-05
Update LICENSE

its-19818942118:
[x]: I have read the `CONTRIBUTING.md` file.


vitaut:
Thanks for the PR but the current license uses standard wording that doesn't need change.

its-19818942118:
> Thanks for the PR, but the current license uses standard wording, which doesn't need change.

Ah ok, sure. No problem.

PR #4407, opened 2025-04-06
Added missing FMT_STRING in fmt::println().

madmaxoft:
The macro invocation is missing, breaking anything that uses `fmt::println()` in `FMT_ENFORCE_COMPILE_STRING` mode.


vitaut (changes requested):
Thanks for the PR. The change looks reasonable but please apply clang-format.

vitaut (approved):

vitaut:
Merged, thanks!

PR #4413, opened 2025-04-14
Fix build error with MSVC v141

### LINKED ISSUE #4412: Build error with MSVC v141
## Issue

The build for `fmt 11.1.4` succeeds with MSVC v141 on GitHub Actions of this repo, but it fails in my local environment with the following error. 

```
format.cc
fmt\include\fmt\format.h(3853): error C2752: 'fmt::v11::formatter<unsigned int,char,void>': more than one partial specialization matches the template argument list
fmt\include\fmt\format.h(3853): note: could be 'fmt::v11::formatter<T,Char,std::enable_if<fmt::v11::detail::type_constant<T,Char>::value!=,void>::type>'
fmt\include\fmt\format.h(3853): note: or       'fmt::v11::formatter<T,Char,void>'
```

## System

Error:

- Visual Studio 2017 (Version 15.9.71) with MSVC v141
- Visual Studio 2022 (Version 17.13.6) with MSVC v141

OK:

- Visual Studio 2022 (Version 17.13.6) with MSVC v142 or v143

(Others conditions: x86 or x64, C++ 14 or 17, release or debug)

## Step to reproduce

Build this solution: [fmt.zip](https://github.com/user-attachments/files/19741950/fmt.zip).

## Remarks

- This error can be fixed by https://github.com/hirohira9119/fmt/commit/90b80f04a5b731720bf9072040d634510ac15fc3.
  Since there is no constant that specifies the platform toolset version, the fix targets all versions of MSVC.

hirohira9119:
Fixes #4412

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -326,8 +326,8 @@ using underlying_t = typename std::underlying_type<T>::type;
 template <typename T> using decay_t = typename std::decay<T>::type;
 using nullptr_t = decltype(nullptr);
 
-#if FMT_GCC_VERSION && FMT_GCC_VERSION < 500
-// A workaround for gcc 4.9 to make void_t work in a SFINAE context.
+#if (FMT_GCC_VERSION && FMT_GCC_VERSION < 500) || FMT_MSC_VERSION
+// A workaround for gcc 4.9 and MSVC 2017 to make void_t work in a SFINAE context.
 template <typename...> struct void_t_impl {
   using type = void;
 };
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -327,7 +327,8 @@ template <typename T> using decay_t = typename std::decay<T>::type;
 using nullptr_t = decltype(nullptr);
 
 #if (FMT_GCC_VERSION && FMT_GCC_VERSION < 500) || FMT_MSC_VERSION
-// A workaround for gcc 4.9 and MSVC 2017 to make void_t work in a SFINAE context.
+// A workaround for gcc 4.9 and MSVC v141 to make void_t work in a SFINAE
+// context.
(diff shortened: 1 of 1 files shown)


include/fmt/base.h
```diff
@@ -326,8 +326,8 @@ using underlying_t = typename std::underlying_type<T>::type;
 template <typename T> using decay_t = typename std::decay<T>::type;
 using nullptr_t = decltype(nullptr);
 
-#if FMT_GCC_VERSION && FMT_GCC_VERSION < 500
-// A workaround for gcc 4.9 to make void_t work in a SFINAE context.
+#if (FMT_GCC_VERSION && FMT_GCC_VERSION < 500) || FMT_MSC_VERSION
```
vitaut, line 329: Can we put more specific MSC version here?

hirohira9119: The MSC version depends on the Visual Studio version, but not on the platform toolset. For example:

* Visual Studio 2017 (v141 toolset) -> `_MSC_VER == 1916`
* Visual Studio 2022 (v141 toolset) -> `_MSC_VER == 1943`
* Visual Studio 2022 (v143 toolset) -> `_MSC_VER == 1943`

Because of this, we can't accurately detect the scenario where Visual Studio 2022 is used with the v141 toolset by comparing the value of `_MSC_VER`. Also, I haven't found a preprocessor definition that provides the platform toolset version.

As a result, I've applied this fix to all MSVC compilers.

hirohira9119: Thank you for the review. I've corrected the formatting. Also, I've updated the comment to be more accurate: MSVC 2017 -> MSVC v141.

vitaut:
Thank you

PR #4418, opened 2025-04-17
Add support for Apple framework builds

Treata11:
Greetings,

In order to enable iOS-derived compilations, I introduced a new flag, `FMT_FRAMEWORK`, which is set to FALSE by default.
I introduced the required properties for framework generation as described in [CMake docs](https://cmake.org/cmake/help/latest/prop_tgt/FRAMEWORK.html#prop_tgt:FRAMEWORK)

And building is successful using the following commands:
```bash
cmake -S. -B build_iOS -DFMT_TEST=OFF -DFMT_DOC=OFF -DFMT_FRAMEWORK=ON -DCMAKE_SYSTEM_NAME=iOS

sudo cmake --build build_iOS --target install --config Release
```

The framewok bundle contains the binary, metadata & the following headers
```zshell
args.h       color.h      format-inl.h ostream.h    std.h
base.h       compile.h    format.h     printf.h     xchar.h
chrono.h     core.h       os.h         ranges.h
```


vitaut:
Thanks for the PR but I am not sure if this is a common enough to warrant a new configuration option. You should be able to create a framework target separately.

Treata11:
> Thanks for the PR but I am not sure if this is a common enough to warrant a new configuration option. You should be able to create a framework target separately.


Not quite common indeed; But large-scale projects depending on FMT, would definitely need _the option_ to build for Apple Platforms. 
It's as simple as a single CMake flag, which means that the dependency-managers of other projects would be able to build for iOS devices very conveniently...

PR #4429, opened 2025-04-26
Avoiding __builtin_strlen

### LINKED ISSUE #4423: Don't use __builtin_strlen
When upgrading `{fmt}`, https://github.com/fmtlib/fmt/commit/b4d1d7f8e broke some of our usage. We had a place where we were building up a format string using a type like this:

```cpp
template <size_t N>
struct fixed_string {
    char data[N] = {}; 

    constexpr fixed_string(char const (&m)[N]) {
        for (size_t i = 0; i != N; ++i) {
            data[i] = m[i];
        }
    }   
};
```

And trying to do something like:
```cpp
static constexpr auto f = fixed_string("x={}");
fmt::print(f.data, 42);
```

This works on fmt 10.2.1 and 11.0, [fails on trunk](https://compiler-explorer.com/z/MvM86dd5j) — after that commit, with:

```cpp
/opt/compiler-explorer/libs/fmt/trunk/include/fmt/base.h:543:31: error: '__builtin_strlen(((const char*)(& f.fixed_string<5>::data)))' is not a constant expression
  543 |       size_ = __builtin_strlen(detail::narrow(s));
      |               ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~
```

[gcc says](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=119866) that `__builtin_strlen` isn't usable like this. Which... if they say so. But seems unlikely this will change, so I think `{fmt}` should also revert to previous usage.

brevzin:
The added test fails on gcc 14 because of the call to `__builtin_strlen`, so just avoiding it during constant evaluation. Fixed #4423.

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -539,7 +539,7 @@ template <typename Char> class basic_string_view {
 #endif
   FMT_CONSTEXPR20 basic_string_view(const Char* s) : data_(s) {
 #if FMT_HAS_BUILTIN(__builtin_strlen) || FMT_GCC_VERSION || FMT_CLANG_VERSION
-    if (std::is_same<Char, char>::value) {
+    if (std::is_same<Char, char>::value && !detail::is_constant_evaluated()) {
       size_ = __builtin_strlen(detail::narrow(s));
       return;
     }
--- a/test/format-test.cc
+++ b/test/format-test.cc
@@ -2564,6 +2564,7 @@ auto fmt::formatter<incomplete_type>::format(const incomplete_type&,
   return formatter<int>::format(42, ctx);
 }
 
+#if FMT_CPLUSPLUS >= 201703L
 namespace {
 
 template <size_t N>
@@ -2583,3 +2584,4 @@ TEST(format_test, fixed_string_constant) {
   static constexpr auto f = fixed_string<5>("x={}");
   EXPECT_EQ(fmt::format(f.data, 42), "x=42");
 }
+#endif
--- a/test/format-test.cc
+++ b/test/format-test.cc
@@ -2580,8 +2580,9 @@ struct fixed_string {
 
 }
 
-TEST(format_test, fixed_string_constant) {
-  static constexpr auto f = fixed_string<5>("x={}");
-  EXPECT_EQ(fmt::format(f.data, 42), "x=42");
+TEST(base_test, from_constexpr_fixed_string) {
+  static constexpr auto fs = fixed_string<5>("x={}");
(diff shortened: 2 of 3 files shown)


vitaut (changes requested):
Thanks for the PR! The change LGTM but please move the test to `base-test` and check the actual API being changed, namely `basic_string_view`'s ctor.

brevzin:
> Thanks for the PR! The change LGTM but please move the test to `base-test` and check the actual API being changed, namely `basic_string_view`'s ctor.

Makes sense, done.

include/fmt/base.h
```diff
@@ -539,7 +539,7 @@ template <typename Char> class basic_string_view {
 #endif
   FMT_CONSTEXPR20 basic_string_view(const Char* s) : data_(s) {
 #if FMT_HAS_BUILTIN(__builtin_strlen) || FMT_GCC_VERSION || FMT_CLANG_VERSION
-    if (std::is_same<Char, char>::value) {
+    if (std::is_same<Char, char>::value && !detail::is_constant_evaluated()) {
       size_ = __builtin_strlen(detail::narrow(s));
```
mwinterb, line 543: What's the advantage of still using `__builtin_strlen`? Will gcc or clang still end up producing a constant in some scenarios that they wouldn't if just `strlen` was used?

mwinterb: Well... libstdc++ __builtin_strlen in char_traits, so I guess there's that.

test/format-test.cc
```diff
@@ -2563,3 +2563,26 @@ auto fmt::formatter<incomplete_type>::format(const incomplete_type&,
...
+};
+
+}
+
+TEST(base_test, from_constexpr_fixed_string) {
+  static constexpr auto fs = fixed_string<5>("x={}");
+  static constexpr auto fmt = fmt::string_view(fs.data);
+  EXPECT_EQ(fmt, "x={}");
+}
+#endif
```
vitaut, line 2588: This should go to `base-test.cc`, next to other `fmt::string_view` tests: https://github.com/fmtlib/fmt/blob/master/test/base-test.cc#L40-L93

include/fmt/base.h
vitaut: There was some benefit to the debug codegen at some point but probably no longer matters on newer compilers.

vitaut:
Merged, thanks!

include/fmt/base.h
vitaut: More importantly, `strlen` is not constexpr.

PR #4434, opened 2025-05-03
Fix reference_wrapper ambiguity with format_as

### LINKED ISSUE #4424: Formatter for `std::reference_wrapper<T>` results in an ambiguous template when `T` can be used with `format_as`
Fmt 11.1.0 added a formatter specialization for `std::reference_wrapper<T>`.

https://github.com/fmtlib/fmt/blob/c7925241c76576dd610915f8120a055eedbe0ad7/include/fmt/std.h#L698-L707

This can result in errors due to ambiguous templates when `T` can be used with `format_as`:

```cpp
#include <fmt/format.h>
#include <fmt/std.h>

#include <functional>

struct S {
    int x;
};

int format_as(S);

int main() {
    S s;
    fmt::format("{}", std::ref(s));
}
```

```
<source>:14:16:   in 'constexpr' expansion of 'fmt::v11::fstring<std::reference_wrapper<S> >("{}")'
/opt/compiler-explorer/libs/fmt/trunk/include/fmt/base.h:1119:52: error: ambiguous template instantiation for 'struct fmt::v11::formatter<std::reference_wrapper<S>, char, void>'
 1119 |     remove_cvref_t<decltype(formatter<T>::format_as(std::declval<const T&>()))>;
      |                             ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
/opt/compiler-explorer/libs/fmt/trunk/include/fmt/format.h:3810:8: note: candidates are: 'template<class T, class Char> struct fmt::v11::formatter<T, Char, fmt::v11::void_t<typename std::remove_cv<typename std::remove_reference<decltype (format_as(declval<const T&>()))>::type>::type> > [with T = std::reference_wrapper<S>; Char = char]'
 3810 | struct formatter<T, Char, void_t<detail::format_as_result<T>>>
      |        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from <source>:2:
/opt/compiler-explorer/libs/fmt/trunk/include/fmt/std.h:699:8: note:                 'template<class T, class Char> struct fmt::v11::formatter<std::reference_wrapper<_Tp>, Char, typename std::enable_if<std::integral_constant<bool, (! std::is_same<decltype (fmt::v11::detail::type_mapper<Char>::map(declval<typename std::conditional<std::is_void<typename std::remove_cv<typename std::remove_reference<_Tp>::type>::type>::value, int*, typename std::remove_cv<typename std::remove_reference<_Tp>::type>::type>::type&>())), void>::value)>::value, void>::type> [with T = S; Char = char]'
  699 | struct formatter<std::reference_wrapper<T>, Char,
      |        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  700 |                  enable_if_t<is_formattable<remove_cvref_t<T>, Char>::value>>
      |                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

https://godbolt.org/z/r71vbP4KG

This is an unfortunate consequence of `std::reference_wrapper` having a non-explicit `operator T&` which allows a call like `format_as(std::ref(s))` to be valid.

Repro without `fmt/std.h`, manually providing the specialization: https://godbolt.org/z/TeeP49x95.

I can think of two possible solutions, I'm sure more exist:

One possible solution is to embrace `format_as` instead of specializing `fmt::formatter` for this:

```cpp
template <typename T>
T& format_as(std::reference_wrapper<T>);
```

This solves the problem as naturally `int format_as(S);` is a better match https://godbolt.org/z/h173rEa6q

Another possible solution is to constrain `fmt::formatter<std::reference_wrapper<T>>` to types `T` which are not `format_as`-ible https://godbolt.org/z/j66Ko5Tzq.

jeremy-rifkin:
This PR fixes #4424. It was unfortunately not possible to just implement this with `template <typename T> T& format_as(std::reference_wrapper<T>);` in the general case so I've just guarded the formatter specialization for format_as.

Happy to adjust the changes here if needed!

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -713,6 +713,7 @@ template <typename T, typename Char> struct formatter<std::complex<T>, Char> {
   }
 };
 
+namespace detail {
 template <typename T, typename Enable = void>
 struct has_format_as : std::false_type {};
 template <typename T>
@@ -725,15 +726,15 @@ template <typename T>
 struct has_format_as_member<
     T, void_t<decltype(formatter<T>::format_as(std::declval<const T&>()))>>
     : std::true_type {};
+}  // namespace detail
 
 // Guard against format_as because reference_wrappers are implicitly convertible
 // to T&
-
 template <typename T, typename Char>
-struct formatter<
-    std::reference_wrapper<T>, Char,
-    enable_if_t<is_formattable<remove_cvref_t<T>, Char>::value &&
-                !has_format_as<T>::value && !has_format_as_member<T>::value>>
+struct formatter<std::reference_wrapper<T>, Char,
+                 enable_if_t<is_formattable<remove_cvref_t<T>, Char>::value &&
+                             !detail::has_format_as<T>::value &&
+                             !detail::has_format_as_member<T>::value>>
     : formatter<remove_cvref_t<T>, Char> {
   template <typename FormatContext>
   auto format(std::reference_wrapper<T> ref, FormatContext& ctx) const
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -713,9 +713,27 @@ template <typename T, typename Char> struct formatter<std::complex<T>, Char> {
   }
 };
 
+template <typename T, typename Enable = void>
+struct has_format_as : std::false_type {};
+template <typename T>
+struct has_format_as<T, void_t<decltype(format_as(std::declval<const T&>()))>>
+    : std::true_type {};
+
+template <typename T, typename Enable = void>
+struct has_format_as_member : std::false_type {};
+template <typename T>
+struct has_format_as_member<
+    T, void_t<decltype(formatter<T>::format_as(std::declval<const T&>()))>>
+    : std::true_type {};
+
+// Guard against format_as because reference_wrappers are implicitly convertible
+// to T&
+
 template <typename T, typename Char>
-struct formatter<std::reference_wrapper<T>, Char,
-                 enable_if_t<is_formattable<remove_cvref_t<T>, Char>::value>>
+struct formatter<
+    std::reference_wrapper<T>, Char,
+    enable_if_t<is_formattable<remove_cvref_t<T>, Char>::value &&
+                !has_format_as<T>::value && !has_format_as_member<T>::value>>
     : formatter<remove_cvref_t<T>, Char> {
   template <typename FormatContext>
   auto format(std::reference_wrapper<T> ref, FormatContext& ctx) const
--- a/test/std-test.cc
+++ b/test/std-test.cc
@@ -414,4 +414,22 @@ TEST(std_test, format_shared_ptr) {
 TEST(std_test, format_reference_wrapper) {
   int num = 35;
   EXPECT_EQ("35", fmt::to_string(std::cref(num)));
+  EXPECT_EQ("35", fmt::to_string(std::ref(num)));
+  EXPECT_EQ("35", fmt::format("{}", std::cref(num)));
+  EXPECT_EQ("35", fmt::format("{}", std::ref(num)));
+}
+
+// Regression test for https://github.com/fmtlib/fmt/issues/4424
+struct type_with_format_as {
+  int x;
+};
+
+int format_as(const type_with_format_as& t) { return t.x; }
+
+TEST(std_test, format_reference_wrapper_with_format_as) {
+  type_with_format_as t{20};
+  EXPECT_EQ("20", fmt::to_string(std::cref(t)));
+  EXPECT_EQ("20", fmt::to_string(std::ref(t)));
+  EXPECT_EQ("20", fmt::format("{}", std::cref(t)));
+  EXPECT_EQ("20", fmt::format("{}", std::ref(t)));
 }
--- a/test/std-test.cc
+++ b/test/std-test.cc
@@ -433,3 +433,17 @@ TEST(std_test, format_reference_wrapper_with_format_as) {
   EXPECT_EQ("20", fmt::format("{}", std::cref(t)));
   EXPECT_EQ("20", fmt::format("{}", std::ref(t)));
 }
+
+struct type_with_format_as_string {
+    std::string str;
+};
+
+std::string format_as(const type_with_format_as_string& t) { return t.str; }
+
+TEST(std_test, format_reference_wrapper_with_format_as_string) {
+  type_with_format_as_string t{"foo"};
+  EXPECT_EQ("foo", fmt::to_string(std::cref(t)));
+  EXPECT_EQ("foo", fmt::to_string(std::ref(t)));
+  EXPECT_EQ("foo", fmt::format("{}", std::cref(t)));
+  EXPECT_EQ("foo", fmt::format("{}", std::ref(t)));
+}
--- a/test/std-test.cc
+++ b/test/std-test.cc
@@ -435,7 +435,7 @@ TEST(std_test, format_reference_wrapper_with_format_as) {
 }
 
 struct type_with_format_as_string {
-    std::string str;
+  std::string str;
 };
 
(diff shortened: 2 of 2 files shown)


vitaut:
> It was unfortunately not possible to just implement this with template <typename T> T& format_as(std::reference_wrapper<T>);`

Could you elaborate?

jeremy-rifkin:
When I attempted the approach of
```cpp
template <typename T>
T& format_as(std::reference_wrapper<T> ref) { return ref.get(); }
```

I got the following error for the test line `EXPECT_EQ("35", fmt::format("{}", std::ref(num)));`

```cpp
/mnt/c/Users/rifkin/home/projects/open-source/fmt/include/fmt/base.h: In instantiation of ‘constexpr fmt::v11::detail::value<Context>::value(const T&, fmt::v11::detail::custom_tag) [with T = std::reference_wrapper<int>; typename std::enable_if<(! has_formatter<T, typename Context::char_type>()), int>::type <anonymous> = 0; Context = fmt::v11::context]’:
/mnt/c/Users/rifkin/home/projects/open-source/fmt/include/fmt/base.h:2235:65:   required from ‘fmt::v11::detail::value<Context>::value(T&) [with T = std::reference_wrapper<int>; typename std::enable_if<(std::integral_constant<bool, (((((((std::is_class<T>::value || std::is_enum<T>::value) || std::is_union<T>::value) || std::is_array<T>::value) && (! fmt::v11::detail::has_to_string_view<T, void>::value)) && (! fmt::v11::detail::is_named_arg<T>::value)) && (! fmt::v11::detail::use_format_as<T>::value)) && (! fmt::v11::detail::use_format_as_member<typename std::remove_const<T>::type, std::integral_constant<bool, true> >::value))>::value || (!1)), int>::type <anonymous> = 0; Context = fmt::v11::context]’
/mnt/c/Users/rifkin/home/projects/open-source/fmt/include/fmt/format.h:4200:17:   required from ‘std::string fmt::v11::format(format_string<T ...>, T&& ...) [with T = {std::reference_wrapper<int>}; std::string = std::__cxx11::basic_string<char>; format_string<T ...> = fstring<std::reference_wrapper<int> >]’
/mnt/c/Users/rifkin/home/projects/open-source/fmt/test/std-test.cc:419:3:   required from here
/mnt/c/Users/rifkin/home/projects/open-source/fmt/include/fmt/base.h:2262:45: error: ‘fmt::v11::detail::type_is_unformattable_for<std::reference_wrapper<int>, char> _’ has incomplete type
 2262 |     type_is_unformattable_for<T, char_type> _;
```

And similar for `fmt::to_string`.

Adding the line `fmt::detail::format_as_result<std::reference_wrapper<int>> x;` to debug
```cpp
/mnt/c/Users/rifkin/home/projects/open-source/fmt/include/fmt/base.h: In substitution of ‘template<class T> using fmt::v11::detail::format_as_result = fmt::v11::remove_cvref_t<decltype (format_as(declval<const T&>()))> [with T = std::reference_wrapper<int>]’:
/mnt/c/Users/rifkin/home/projects/open-source/fmt/test/std-test.cc:421:61:   required from here
/mnt/c/Users/rifkin/home/projects/open-source/fmt/include/fmt/base.h:1115:38: error: ‘format_as’ was not declared in this scope, and no declarations were found by argument-dependent lookup at the point of instantiation [-fpermissive]
 1115 |     remove_cvref_t<decltype(format_as(std::declval<const T&>()))>;
      |                             ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
/mnt/c/Users/rifkin/home/projects/open-source/fmt/include/fmt/std.h:750:4: note: ‘template<class T> T& format_as(std::reference_wrapper<_Tp>)’ declared here, later in the translation unit
  750 | T& format_as(std::reference_wrapper<T> ref) { return ref.get(); }
      |    ^~~~~~~~~
```

I figured it'd be easiest to just update the formatter to check for existing format_as / format_as_member ability

include/fmt/std.h
```diff
@@ -713,9 +713,27 @@ template <typename T, typename Char> struct formatter<std::complex<T>, Char> {
...
+// Guard against format_as because reference_wrappers are implicitly convertible
+// to T&
+
 template <typename T, typename Char>
-struct formatter<std::reference_wrapper<T>, Char,
-                 enable_if_t<is_formattable<remove_cvref_t<T>, Char>::value>>
+struct formatter<
+    std::reference_wrapper<T>, Char,
+    enable_if_t<is_formattable<remove_cvref_t<T>, Char>::value &&
+                !has_format_as<T>::value && !has_format_as_member<T>::value>>
```
phprus, line 736: Is it possible to reuse ``use_format_as`` and ``use_format_as_member``?

https://github.com/fmtlib/fmt/blob/59259a5fde0d555da6faa1f300e83cc2eba61193/include/fmt/base.h#L1120-L1141

vitaut (changes requested):
I suggest using `format_as` but make it a member (since we can't inject it into the `std` namespace`).

include/fmt/std.h
jeremy-rifkin: Unfortunately this doesn't work because of the `std::is_arithmetic`. Consider, e.g.,
```cpp
struct type_with_format_as_string {
    std::string str;
};

std::string format_as(const type_with_format_as_string& t) { return t.str; }
```

`use_format_as<type_with_format_as_string >::value` is false however `fmt::format("{}", type_with_format_as_string{"foo"})` works as it should. As such, attempting to use `use_format_as` for the check results in the same ambiguities.

jeremy-rifkin:
Adding the following to the formatter causes the most bizarre thing to happen
```cpp
  static auto format_as(std::reference_wrapper<T> ref) -> T& {
    return ref.get();
  }
```
Suddenly the to_string here evaluates to `"#"`
```cpp
  int num = 35;
  fmt::to_string(std::cref(num));
```

This seems to be because the chosen `to_string` is
```cpp
template <typename T, FMT_ENABLE_IF(!std::is_integral<T>::value &&
                                    !detail::use_format_as<T>::value)>
FMT_NODISCARD auto to_string(const T& value) -> std::string {
  auto buffer = memory_buffer();
  detail::write<char>(appender(buffer), value);
  return {buffer.data(), buffer.size()};
}
```
which writes the int value as a char. Presumably this is because there is no `to_string` using member format_as, yet adding such an overload causes more issues.

Since any use of member format_as here will depend on the additional constraints for the reference_wrapper formatter specialization this PR adds and there appear to be more involved changes required to get member format_as working, could this PR be merged as-is with the other changes deferred to another PR?

include/fmt/std.h
```diff
@@ -713,9 +713,27 @@ template <typename T, typename Char> struct formatter<std::complex<T>, Char> {
...
+template <typename T>
+struct has_format_as<T, void_t<decltype(format_as(std::declval<const T&>()))>>
+    : std::true_type {};
+
+template <typename T, typename Enable = void>
+struct has_format_as_member : std::false_type {};
+template <typename T>
+struct has_format_as_member<
+    T, void_t<decltype(formatter<T>::format_as(std::declval<const T&>()))>>
+    : std::true_type {};
```
vitaut, line 727: This should go to the `detail` namespace.

vitaut:
> could this PR be merged as-is with the other changes deferred to another PR?

OK but please move `has_format_as*` to `detail`.

include/fmt/std.h
jeremy-rifkin: Thanks, will do

jeremy-rifkin:
Thanks, taken care of

vitaut:
Thank you!

PR #4440, opened 2025-05-08
add version macroses for more user friendly usage

IRainman:
Simple version macroses for user-friendly experience.


vitaut:
Thanks for the PR but I think we should keep it simple. Individual components are easy to extract from the full version.

PR #4442, opened 2025-05-09
Change component prefix for NSIS compatibility

### LINKED ISSUE #4441: CPack error on Windows with NSIS due to invalid character in component name 'fmt-core'
> The header-only mode is not recommended and we shouldn't change install behavior based on it but a PR to rename the component to `fmt-core` is welcome (fmtlib is the name of the github org, not the project). I think it can be done unconditionally since the name is global. 

 _Originally posted by @vitaut in [#4283](https://github.com/fmtlib/fmt/issues/4283#issuecomment-2568124123)_

The rename of the "core" package to "fmt-core" causes CPack packaging with NSIS on Windows to fail with the message
`Error: invalid characters in variable name "fmt-core_selected", use only characters [a-z][A-Z][0-9], '.' and '_'` as NSIS cannot have the `-` character in variable names.

n-stein:
Fixes #4441.

NSIS does not accept the `-` character in variable names, so CPack on Windows fails with the current component names `fmt-core` and `fmt-doc`.

Switching to `fmt_core` and `fmt_doc` resolves the issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -431,7 +431,7 @@ if (FMT_INSTALL)
 
   # Install the library and headers.
   install(TARGETS ${INSTALL_TARGETS}
-          COMPONENT fmt-core
+          COMPONENT fmt_core
           EXPORT ${targets_export_name}
           LIBRARY DESTINATION ${FMT_LIB_DIR}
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks

PR #4445, opened 2025-05-12
Remove some redundant `const`s and `inline`s

LocalSpook:
`constexpr` variables are implicitly `const`, and member functions defined within their class definition are implicitly `inline`.

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -326,7 +326,7 @@ inline auto get_classic_locale() -> const std::locale& {
 }
 
 template <typename CodeUnit> struct codecvt_result {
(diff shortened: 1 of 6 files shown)


vitaut:
Thanks for the PR. `inline` is not redundant with modules: #4169, so that part needs to be reverted.

vitaut (changes requested):
Please revert the `inline` changes.

LocalSpook:
Wow, I was unaware of the module changes. I've read up on them now.

Sorry for the delay, and sorry for the force-push; I couldn't wrangle git into reverting those lines in a new commit.

vitaut:
Thank you!

PR #4447, opened 2025-05-14
Remove some implicit conversions

dodomorandi:
This fixes some implicit conversions that can cause noise when compiling other projects if conversions warnings (`-Wconversion -Wsign-conversion` in GCC) are enabled.


vitaut (changes requested):
Please provide a godbolt repro demonstrating the warnings.

PR #4451, opened 2025-05-18
avoid include locale inline if c++20 modules are enabled

arBmind:
MSVC hints with:
```
fmt\include\fmt\format-inl.h(26): warning C5244: '#include <locale>' in the purview of module 'fmt' appears erroneous.  Consider moving that directive before the module declaration, or replace the textual inclusion with 'import <locale>;'.
```

Then fails the build with `type redefinition`.

See https://github.com/arBmind/fmtlib-module-experiment for a full example.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -22,7 +22,7 @@
 
 #include "format.h"
 
-#if FMT_USE_LOCALE
+#if FMT_USE_LOCALE && !defined(FMT_MODULE)
 #  include <locale>
 #endif
 


vitaut (changes requested):
Please include the error in the PR summary and revert unrelated formatting changes.

include/fmt/format-inl.h
```diff
@@ -22,10 +25,6 @@
 
 #include "format.h"
 
-#if FMT_USE_LOCALE
-#  include <locale>
-#endif
```
vitaut, line 27: I think the correct fix is to keep the include here but change the ifdef to

```
#if !defined(FMT_MODULE) && FMT_USE_LOCALE
```

vitaut:
Merged, thanks!

PR #4453, opened 2025-05-20
Give useful error when misusing fmt::ptr.

TobiSchluter:
static_assert(bla, "") prints an empty message but not the condition with  MSVC. Add an informative message.

E.g.
`int i; fmt::print("{}", fmt::ptr(i))`
will print the empty message from the static_assert and the code location, but not the static_assert's condition. This patch makes it print a useful message.

I audited all static_asserts, and this appears to be the only user-facing static_assert that didn't contain a message.


vitaut:
Merged, thanks

PR #4456, opened 2025-05-26
Enable constexpr support for fmt::format (fmtlib#3403)

msvetkin:
This PR allows to use fmt::format at compile-time by:                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                

- Adding FMT_CONSTEXPR_STRING macro
- Mark format/to_string with FMT_CONSTEXPR_STRING
- Add a few fmt::format compile-time test-cases

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -117,6 +117,33 @@
 #  define FMT_NOINLINE
 #endif
 
+// Detect constexpr std::string.
+#if !FMT_USE_CONSTEVAL
+#  define FMT_USE_CONSTEXPR_STRING 0
+#elif defined(__cpp_lib_constexpr_string) && \
+    __cpp_lib_constexpr_string >= 201907L
+#  if FMT_CLANG_VERSION && FMT_GLIBCXX_RELEASE
+// clang + libstdc++ are able to work only starting with gcc13.3
+// https://gcc.gnu.org/bugzilla/show_bug.cgi?id=113294
+#    if FMT_GLIBCXX_RELEASE < 13
+#      define FMT_USE_CONSTEXPR_STRING 0
+#    elif FMT_GLIBCXX_RELEASE == 13 && __GLIBCXX__ < 20240521
+#      define FMT_USE_CONSTEXPR_STRING 0
+#    else
+#      define FMT_USE_CONSTEXPR_STRING 1
+#    endif
+#  else
+#    define FMT_USE_CONSTEXPR_STRING 1
+#  endif
+#else
+#  define FMT_USE_CONSTEXPR_STRING 0
+#endif
+#if FMT_USE_CONSTEXPR_STRING
+#  define FMT_CONSTEXPR_STRING constexpr
+#else
+#  define FMT_CONSTEXPR_STRING
+#endif
+
 // GCC 4.9 doesn't support qualified names in specializations.
 namespace std {
 template <typename T> struct iterator_traits<fmt::basic_appender<T>> {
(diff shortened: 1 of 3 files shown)


include/fmt/format.h
```diff
@@ -117,6 +117,33 @@
...
+// Detect constexpr std::string.
+#if !FMT_USE_CONSTEVAL
+#  define FMT_USE_CONSTEXPR_STRING 0
+#elif !defined(__cpp_lib_constexpr_string)
+#  define FMT_USE_CONSTEXPR_STRING 0
+#elif defined(__cpp_lib_constexpr_string) && \
+    __cpp_lib_constexpr_string < 201907L
+#  define FMT_USE_CONSTEXPR_STRING 0
+#elif FMT_CLANG_VERSION && FMT_GLIBCXX_RELEASE
+// clang + libstdc++ is able to work only start with gcc13.3
```
vitaut, line 129: nit: is -> are, start -> starting

```
// clang + libstdc++ are able to work only starting with gcc13.3
```

test/compile-test.cc
```diff
@@ -421,3 +421,66 @@ TEST(compile_time_formatting_test, multibyte_fill) {
...
 }
 #endif
+
+#if FMT_USE_CONSTEXPR_STRING
+
+TEST(compile_test, constexpr_format) {
+  {
+    constexpr auto result = []() {
+      return fmt::format(FMT_COMPILE("{}"), true) == "true";
+    }();
```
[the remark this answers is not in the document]
msvetkin: I have replaced boolean test cases with `int` and drop test case for a type with user provided formatter.

include/fmt/format.h
```diff
@@ -117,6 +117,33 @@
...
+#if !FMT_USE_CONSTEVAL
+#  define FMT_USE_CONSTEXPR_STRING 0
+#elif defined(__cpp_lib_constexpr_string) && \
+    __cpp_lib_constexpr_string >= 201907L
+#  if FMT_CLANG_VERSION && FMT_GLIBCXX_RELEASE
+// clang + libstdc++ are able to work only starting with gcc13.3
+// https://gcc.gnu.org/bugzilla/show_bug.cgi?id=113294
+#    if FMT_GLIBCXX_RELEASE < 13
+#      define FMT_USE_CONSTEXPR_STRING 0
+#    elif defined(__GLIBCXX__) && __GLIBCXX__ < 20240521
```
phprus: The ``__GLIBCXX__`` check is redundant and incorrect.

```
libstdc++ 13.3: __GLIBCXX__  == 20240521
libstdc++ 14.1: __GLIBCXX__  == 20240507
```

This condition is ``true`` for GCC-14.

The order of ``__GLIBCXX__`` is not equivalent to the order of gcc major versions.

~~Only the ``FMT_GLIBCXX_RELEASE`` value needs to be checked.~~

test/compile-test.cc
```diff
@@ -421,3 +421,44 @@ TEST(compile_time_formatting_test, multibyte_fill) {
   EXPECT_EQ("жж42", test_format<8>(FMT_COMPILE("{:ж>4}"), 42));
 }
 #endif
+
+#if FMT_USE_CONSTEXPR_STRING
+
+TEST(compile_test, constexpr_format) {
```
phprus, line 427: I think the long string formatting should be checked, since due to SSO short string tests don't use dynamic allocation in strings.

phprus (changes requested):

include/fmt/format.h
phprus: My suggestion:

```c++
#    if FMT_GLIBCXX_RELEASE < 13
#      define FMT_USE_CONSTEXPR_STRING 0
#    elif FMT_GLIBCXX_RELEASE == 13 && __GLIBCXX__ < 20240521
#      define FMT_USE_CONSTEXPR_STRING 0
#    else
#      define FMT_USE_CONSTEXPR_STRING 1
#    endif
```

msvetkin: Thank you!

vitaut (approved):

vitaut:
Merged, thanks!

Issue #4452, opened 2025-05-19
Override/Replace/Customize formatter for std::optional


robertfalkenberg:
The current version of `fmt` provides a [powerful generic formatter](https://github.com/fmtlib/fmt/blob/master/include/fmt/std.h#L355) for `std::optional<T>`, which automatically invokes the formatter of `T`.

The formatted output is as follows (assuming `T == int`):

* `"optional(42)"` if the optional was set to `42`;
* `"none"` if the optional was not set (`std::nullopt`).

For our software project, however, the formatter for `std::optional` should to be adapted, e.g. to obtain the following compact outputs:

* `"42"` if the optional was set to `42`;
* `"null"` if the optional was not set (`std::nullopt`).

Obviously redefining the formatter for the generic type (or an explicit type) makes the compiler complain about ambiguous definition;

Although I see the formatter is wrapped in
```
#ifdef __cpp_lib_optional
...
#endif  // __cpp_lib_optional
```
this is not meant as a user-defined switch to disable the existing definition (and allow/enforce the user to provide a custom definition at another place) - instead this wrapper only checks whether `std::optional` is available in the compiler.

What is the best way to achieve a custom generic formatter for `std::optional`, without patching the `fmt` library (so that fmt can be updated without re-patching it again)?

The same approach could possibly applied for `std::expected` and `std::variant`.

sergiud:
You can use a proxy for this:
```cpp
#include <algorithm>
#include <optional>
#include <string_view>

#include <fmt/format.h>

template<class T>
struct CustomFormatterHolder
{
    T inner;
};

template<class T>
struct fmt::formatter<CustomFormatterHolder<std::optional<T>>> : fmt::formatter<fmt::string_view>
{
    template<class FormatContext>
    decltype(auto) format(const CustomFormatterHolder<std::optional<T>>& value, FormatContext& ctx) const
    {
        if (value.inner) {
            return fmt::format_to(ctx.out(), "{}", *value.inner);
        }

        constexpr std::string_view null = "null";
        return std::copy(null.begin(), null.end(), ctx.out());
    }
};

int main()
{
    fmt::println("{}", CustomFormatterHolder<std::optional<int>>{42});
    fmt::println("{}", CustomFormatterHolder<std::optional<int>>{std::nullopt});
}
```
Output:
```
42
null
```
https://godbolt.org/z/x1qf6W14K

robertfalkenberg:
[...]

PR #4457, opened 2025-05-30
Support to disable formatters for std::optional and std::expected

robertfalkenberg:
This MR adds support to disable formatters for `std::optional` and `std::expected`, for example, if a custom standard definition is preferred.

The MR follows the same approach as for `std::variant` in https://github.com/fmtlib/fmt/pull/3716.

To disable these formatters, define the respective macros

```c++
#define FMT_CPP_LIB_OPTIONAL 0
#define FMT_CPP_LIB_EXPECTED 0
```
before loading the fmt library (or via command line to the compiler).

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -77,6 +77,22 @@
 #  endif
 #endif
 
+#ifndef FMT_CPP_LIB_OPTIONAL
+#  ifdef __cpp_lib_optional
+#    define FMT_CPP_LIB_OPTIONAL __cpp_lib_optional
+#  else
+#    define FMT_CPP_LIB_OPTIONAL 0
+#  endif
+#endif
+
+#ifndef FMT_CPP_LIB_EXPECTED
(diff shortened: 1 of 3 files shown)


vitaut:
Thanks for the PR but you should only provide formatters for your own types. The macro to disable `std::variant` formatter is a workaround for old libc++ on macOS and will be removed.

robertfalkenberg:
Thank you for your time reviewing the MR.

> The macro to disable std::variant formatter is a workaround for old libc++ on macOS and will be removed.

OK, I agree - if the plan is to remove those workarounds, then there shouldn't be any extensions towards the opposite direction.

Do you see any other way(s) to provide a project-wide custom version for default `std::optional(ANY_TYPE)` that does not require
* patching of the imported `fmt` library and re-patching it on each update;
* wrapping each `optional` into a custom type with custom formatter (as proposed in #4452).

PR #4459, opened 2025-05-31
Export fmt::dynamic_format_arg_store

tkhyn:
When consuming `fmt` as a C++20 module I could not access `fmt::dynamic_format_arg_store`.

I believe it should be exported and this PR fixes that!


phprus:
@vitaut, fuzzing found a error in the implementation of ``write_float``.

vitaut:
> @vitaut, fuzzing found a error in the implementation of `write_float`.

Yeah, I saw it too. Looks like there is an integer overflow for a very large precision.

vitaut:
Merged, thanks!

PR #4471, opened 2025-06-18
Replace memset with constexpr fill_n in bigint::align and fix memset count to account for sizeof T

Tomek-Stolarczyk:
Had an issue with a particularly pedantic compiler that flagged the use of memset (a non-constexpr labeled function) in the use of bigint::align when it's marked FMT_CONSTEXPR=constexpr.
Using the internal fill_n allows the function to stay constexpr and use memset when it isn't.

Also noticed that when memset is used in fill_n, it doesn't account for sizeof(T) so when using fill_n to clear a large buffer, memset will leave a portion of the buffer unmodified.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -553,7 +553,7 @@ FMT_CONSTEXPR auto fill_n(OutputIt out, Size count, const T& value)
 template <typename T, typename Size>
 FMT_CONSTEXPR20 auto fill_n(T* out, Size count, char value) -> T* {
   if (is_constant_evaluated()) return fill_n<T*, Size, T>(out, count, value);
-  std::memset(out, value, to_unsigned(count));
+  std::memset(out, value, to_unsigned(count) * sizeof(T));
   return out + count;
 }
 
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -2799,7 +2799,7 @@ class bigint {
     bigits_.resize(to_unsigned(num_bigits + exp_difference));
     for (int i = num_bigits - 1, j = i + exp_difference; i >= 0; --i, --j)
       bigits_[j] = bigits_[i];
-    memset(bigits_.data(), 0, to_unsigned(exp_difference) * sizeof(bigit));
+    fill_n(bigits_.data(), to_unsigned(exp_difference), 0);
     exp_ -= exp_difference;
   }
 


vitaut:
What compiler/version are you referring to and what is the full error?

include/fmt/format.h
```diff
@@ -553,7 +553,7 @@ FMT_CONSTEXPR auto fill_n(OutputIt out, Size count, const T& value)
 template <typename T, typename Size>
 FMT_CONSTEXPR20 auto fill_n(T* out, Size count, char value) -> T* {
   if (is_constant_evaluated()) return fill_n<T*, Size, T>(out, count, value);
-  std::memset(out, value, to_unsigned(count));
+  std::memset(out, value, to_unsigned(count) * sizeof(T));
```
mwinterb: Isn't using `memset` here invalid if `sizeof(T)` isn't 1? (It's probably only actually valid if `sizeof(T)` is 1 and `T` is trivially copy assignable.)
i.e. In [here in chrono](https://github.com/fmtlib/fmt/blob/730fd4d9a7a9f5973a47c5c540becc71b62b8387/include/fmt/chrono.h#L1083) if `OutputIt` is `wchar_t*` it's going to store `L'\u3030'` (or `L'\U30303030'`), not `L'\u30'`.

This is already a problem...
https://gcc.godbolt.org/z/61Tsazxb5

Tomek-Stolarczyk:
> What compiler/version are you referring to and what is the full error?

I couldn't reproduce this on public compilers but the error message recieved was
`deps/fmt/11.2.0/include/fmt/format.h: In member function 'constexpr void fmt::v11::detail::bigint::align(const fmt::v11::detail::bigint&)':
deps/fmt/11.2.0/include/fmt/format.h:2731:11: error: call to non-'constexpr' function 'void* memset(void*, int, size_t)'
     memset(bigits_.data(), 0, to_unsigned(exp_difference) * sizeof(bigit));`

> Thanks for the PR but please provide more details per my earlier comment and address warnings reported in CI:

Will work on getting the builds green, thanks

include/fmt/format.h
Tomek-Stolarczyk: I agree, I think fill_n should also be `auto fill_n(T* out, Size count, const T& value) `, similar to how the standard has defined it. `char value` -> `return fill_n<T*, Size, T>(out, count, value);` is performing a decimating cast on `value`. If sizeof(T) > 1, we're bound to lose data.

vitaut: `char` is fine here because this is a version specifically optimized for it. All we need is to constrain on `sizeof(T) == 1`.

Tomek-Stolarczyk: Theres a fundamental problem with using std::memset on a T* and using `char value` to initialize the elements

include/fmt/format.h
```diff
@@ -553,6 +553,7 @@ FMT_CONSTEXPR auto fill_n(OutputIt out, Size count, const T& value)
 template <typename T, typename Size>
 FMT_CONSTEXPR20 auto fill_n(T* out, Size count, char value) -> T* {
   if (is_constant_evaluated()) return fill_n<T*, Size, T>(out, count, value);
+  static_assert(1 ==  sizeof(T), "sizeof(T) must be 1 to use char for initialization");
```
vitaut: nit: extra space after `==` and normally the literal is on the right-hand side: `sizeof(T) == 1`.

include/fmt/format.h
```diff
@@ -2799,7 +2800,7 @@ class bigint {
     bigits_.resize(to_unsigned(num_bigits + exp_difference));
     for (int i = num_bigits - 1, j = i + exp_difference; i >= 0; --i, --j)
       bigits_[j] = bigits_[i];
-    memset(bigits_.data(), 0, to_unsigned(exp_difference) * sizeof(bigit));
+    fill_n(bigits_.data(), to_unsigned(exp_difference), 0U);
```
vitaut, line 2804: Why `U`?

Tomek-Stolarczyk: bigit is a uint32_t so bigits_.data() returns a uint32_t*. 0U is the unsigned literal for 0. I could use `fill_n(bigits_.data(), to_unsigned(exp_difference), to_unsigned(0));` if that's better but thought using the literal `U` would convey the same meaning. LMK if you prefer I change it to something else. The 0 is by default signed so I got a signed/unsigned warning.

vitaut:
Merged, thanks!

PR #4475, opened 2025-06-20
Avoid redundant work when processing UTF-8 strings

localspook:
`convert_precision_to_size` calculates how many *bytes* of a UTF-8 string we can print before we exceed a display width limit, but does it using a weird two-step:

1. Walk the string, calculating the number of *code points* we can print, *and throw away the position we stopped at*.
2. Walk the string *again* to convert the number of code points into a number of bytes (i.e. recalculate the position we stopped at).

Put like that, the fix is clear. Also, this is the only place `code_point_index`, the function that does step 2, is used, so we delete it.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -696,26 +696,6 @@ FMT_CONSTEXPR inline auto compute_width(string_view s) -> size_t {
   return num_code_points;
 }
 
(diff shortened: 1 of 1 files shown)


vitaut:
Nice improvement, thanks!

PR #4476, opened 2025-06-20
Add FMT_EXPORT on ranges.h customization points

Mishura4:
Title, `fmt/ranges.h` has a bunch of missing exports, in particular I needed to specialize `fmt::is_tuple_formattable` to `std::false_type`.

### DIFF
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -279,11 +279,13 @@ template <typename FormatContext> struct format_tuple_element {
 
 }  // namespace detail
 
+FMT_EXPORT
 template <typename T> struct is_tuple_like {
   static constexpr bool value =
       detail::is_tuple_like_<T>::value && !detail::is_range_<T>::value;
 };
 
+FMT_EXPORT
 template <typename T, typename C> struct is_tuple_formattable {
   static constexpr bool value = detail::is_tuple_formattable_<T, C>::value;
 };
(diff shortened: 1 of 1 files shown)


include/fmt/ranges.h
```diff
@@ -279,11 +279,13 @@ template <typename FormatContext> struct format_tuple_element {
...
 }  // namespace detail
 
+FMT_EXPORT
 template <typename T> struct is_tuple_like {
   static constexpr bool value =
       detail::is_tuple_like_<T>::value && !detail::is_range_<T>::value;
 };
 
+FMT_EXPORT
 template <typename T, typename C> struct is_tuple_formattable {
```
vitaut, line 289: I think only `is_tuple_like` should be public/specializable for consistency with ranges.

include/fmt/ranges.h
```diff
@@ -359,12 +362,14 @@ struct is_formattable_delayed
     : is_formattable<uncvref_type<maybe_const_range<R>>, Char> {};
 }  // namespace detail
 
+FMT_EXPORT
 template <typename...> struct conjunction : std::true_type {};
```
vitaut: This doesn't need to be public, it's a mostly internal replacement for `std::conjunction`.

include/fmt/ranges.h
vitaut: On a second thought, `is_tuple_like` is more general since it also controls `join` so exporting both makes sense.

vitaut:
Merged, thanks!

PR #4477, opened 2025-06-20
Fix undeclared malloc/free error with libc++-21

remyjette:
`<cstdlib>` was not being included, so malloc and free were only declared via transitive includes. Some includes changed in the latest libc++-21 build which broke fmt: https://godbolt.org/z/vErfGTh1d

Also changed `malloc`/`free` to `std::malloc` and `std::free`, as putting those symbols in the global namespace is optional for the implementation when including `<cstdlib>`.


vitaut:
Thank you for the fix!

PR #4478, opened 2025-06-23
Fix interaction between debug presentation, precision, and width for strings

localspook:
Consider our function that formats strings. It has 2 issues:
```cpp
FMT_CONSTEXPR auto write(OutputIt out, basic_string_view<Char> s,
                         const format_specs& specs) -> OutputIt {
  auto data = s.data();
  auto size = s.size();
  if (specs.precision >= 0 && to_unsigned(specs.precision) < size)
    size = convert_precision_to_size(s, to_unsigned(specs.precision)); // Issue 1: we truncate the string
                                                                       // to fit within the precision...
  bool is_debug = specs.type() == presentation_type::debug;            //
  if (is_debug) {                                                      //
    auto buf = counting_buffer<Char>();                                //
    write_escaped_string(basic_appender<Char>(buf), s); // ...but don't account for the fact that it
    size = buf.count();                                 // can expand when escaped.
  } // ^^^^^^^^^^^^^^^^──────┐
                          // │ Issue 2: we reinterpret size in bytes as 
  size_t width = 0;       // │ display width. We can't do that, because
  if (specs.width != 0) { // │ debug presentation doesn't mean "just ASCII".
    width =     // vvvv──────┘
        is_debug ? size : compute_width(basic_string_view<Char>(data, size));
  }
  return write_padded<Char>(
      out, specs, size, width, [=](reserve_iterator<OutputIt> it) {
        return is_debug ? write_escaped_string(it, s)
                        : copy<Char>(data, data + size, it);
      });
}
```
Issue 1 manifests like this:

```cpp
fmt::println("{:.2?}", "\n"); // Prints "\n", but should print "\
```

And Issue 2 like this:

```cpp
fmt::println("{:*<5?}", "щ"); // Prints "щ"*, but should print "щ"** 
                              // (щ is two bytes, so {fmt} thinks it takes two terminal cells)
```

[Godbolt link](https://godbolt.org/z/Mn81hrqTK). This PR adds the logic to handle such cases.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -657,21 +657,9 @@ FMT_CONSTEXPR void for_each_codepoint(string_view s, F f) {
   } while (buf_ptr < buf + num_chars_left);
 }
 
-template <typename Char>
-inline auto compute_width(basic_string_view<Char> s) -> size_t {
-  return s.size();
-}
-
-// Computes approximate display width of a UTF-8 string.
-FMT_CONSTEXPR inline auto compute_width(string_view s) -> size_t {
-  size_t num_code_points = 0;
-  // It is not a lambda for compatibility with C++14.
-  struct count_code_points {
-    size_t* count;
-    FMT_CONSTEXPR auto operator()(uint32_t cp, string_view) const -> bool {
-      *count += to_unsigned(
-          1 +
-          (cp >= 0x1100 &&
+FMT_CONSTEXPR inline auto display_width_of(uint32_t cp) noexcept -> size_t {
+  return to_unsigned(
+      1 + (cp >= 0x1100 &&
            (cp <= 0x115f ||  // Hangul Jamo init. consonants
             cp == 0x2329 ||  // LEFT-POINTING ANGLE BRACKET
             cp == 0x232a ||  // RIGHT-POINTING ANGLE BRACKET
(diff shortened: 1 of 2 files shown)


localspook:
Tangent, but: while working in this area of the code, I noticed we escape characters differently than the standard: we use `\uXXXX` and `\UXXXXXXXX`, while the standard uses `\u{...}`... is this intentional?

include/fmt/format.h
```diff
@@ -2165,22 +2210,19 @@ FMT_CONSTEXPR auto write(OutputIt out, basic_string_view<Char> s,
...
   }
 
-  size_t width = 0;
-  if (specs.width != 0) {
-    width =
-        is_debug ? size : compute_width(basic_string_view<Char>(data, size));
-  }
   return write_padded<Char>(
-      out, specs, size, width, [=](reserve_iterator<OutputIt> it) {
+      out, specs, size, [&](reserve_iterator<OutputIt> it) {
```
vitaut, line 2214: Why capture by reference?

vitaut:
> Tangent, but: while working in this area of the code, I noticed we escape characters differently than the standard: we use `\uXXXX` and `\UXXXXXXXX`, while the standard uses `\u{...}`... is this intentional?

IIRC the `\u{...}` syntax was added recently. I don't think it matters much but we could switch to it for consistency (not in this PR obviously).

include/fmt/format.h
localspook: I think I instinctively write `[&]` to avoid copies. In this case the semantics are the same either way.

vitaut:
Merged, thank you!

PR #4482, opened 2025-06-27
compatible for c++11 but GCC < 5

aengusjiang:

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -73,10 +73,10 @@
 #  endif
 #endif  // FMT_MODULE
 
-#if (defined(__GNUC__) && __GNUC__ < 5) || defined(_MSC_VER)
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR but please provide more details about the issue you are trying to solve, ideally a godbolt repro.

aengusjiang:
before modify, it will compiler error:
https://godbolt.org/z/jGndjsKcP

phprus:
As far as I see, support for gcc 4.8 has been removed since commit https://github.com/fmtlib/fmt/commit/6797f0c39a4ef13061cbc3bb850c35af7428fdc4

vitaut:
The minimum supported gcc version is 4.9 so you'll need to either upgrade your compiler or use an earlier {fmt} release.

PR #4488, opened 2025-07-06
Fix `import std` in clang++, and add options for `import std.compat`

anonymouspc:
When `#define FMT_IMPORT_STD`, `{fmt}` will do `import std` instead of `#include <std...>`. However this is currently broken on clang++. This PR fixes:
1. clang++ cannot find `memcpy`, `memcmp`, `abort` when importing std. Fix it by adding `#include <string.h>` and `#include <stdlib.h>` in global module.
2. `include/fmt/std.h` uses `std::__bit_const_reference` which is not exported by libc++ and cannot be found. Guard it by macros.

BTW, we can import C std symbols by `import std.compat`, and in this case we do not need the `<limits.h>`, `<stdint.h>`... as well. So that I add an optional  `FMT_IMPORT_STD_COMPAT` macro to tell `{fmt}` whether `import std.compat`.

**Thank you!**

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -225,7 +225,7 @@ template <typename T> struct is_bit_reference_like {
 
 // Workaround for libc++ incompatibility with C++ standard.
 // According to the Standard, `bitset::operator[] const` returns bool.
-#ifdef _LIBCPP_VERSION
+#if defined(_LIBCPP_VERSION) && !defined(FMT_IMPORT_STD)
 template <typename C>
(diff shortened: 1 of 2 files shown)


shyeyian:
> Thanks for the PR. The fixes LGTM but I don't think we need to add another configuration macro (`FMT_IMPORT_STD_COMPAT`).

Okay, I've reverted that :)

vitaut:
Merged, thanks!

PR #4497, opened 2025-07-14
Make compatible with WASI

### LINKED ISSUE #4496: Does not build on WebAssembly WASI due to lack of `dup`, `dup2`, and `pipe` system calls
These are not available architecturally.

whitequark:
WASI is a POSIX subset that doesn't have `dup`, `dup2`, or `pipe` system calls.

Fixes #4496.

It is not completely clear to me that this is the right fix, nor is it clear to me why fmt needs these syscalls in first place. But it seems to fix the build.


vitaut:
Merged, thanks!

PR #4505, opened 2025-07-29
Add FMT_USER_PROVIDED_ASSERT_FAIL

HazardyKnusperkeks:
That way one can provide ones own implementation for details::assert_fail.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -33,12 +33,14 @@
 FMT_BEGIN_NAMESPACE
 namespace detail {
 
+#ifndef FMT_USER_PROVIDED_ASSERT_FAIL
 FMT_FUNC void assert_fail(const char* file, int line, const char* message) {
   // Use unchecked std::fprintf to avoid triggering another assertion when
   // writing to stderr fails.
   fprintf(stderr, "%s:%d: assertion failed: %s", file, line, message);
   abort();
 }
+#endif
 
 FMT_FUNC void format_error_code(detail::buffer<char>& out, int error_code,
                                 string_view message) noexcept {


HazardyKnusperkeks:
I want to use libfmt in a bare metal project, with exceptions disabled and I don't want to pull in any of the printf-familiy, so I need my own implementation. Instead of a weak symbol one could hide the implementation behind an `#if` (e.g. `FMT_USER_PROVIDED_ASSERT_FAIL`), if that's more to your liking.

HazardyKnusperkeks:
> I want to use libfmt in a bare metal project, with exceptions disabled and I don't want to pull in any of the printf-familiy, so I need my own implementation. Instead of a weak symbol one could hide the implementation behind an `#if` (e.g. `FMT_USER_PROVIDED_ASSERT_FAIL`), if that's more to your liking.

Seeing the failing tests, that seems to be needed?

vitaut:
Thanks for the PR but I don't think we should add yet another configuration macro. You can already control this via `FMT_ASSERT` / `FMT_THROW`.

HazardyKnusperkeks:
I actually can not really. Because the functions I'd like to be called is not yet in scope, when `fmt/format.h` is included, even if using the classic `assert`, obviously the same for my own functions.

```
/opt/bin/g++ -g -O0 -Wall -Wextra -m64 -pipe -fexceptions -fvisibility=default -fPIC -DFMT_USE_NOEXCEPT '-DFMT_ASSERT(c,t)=assert(c)' -I.../libs/3rdParty/fmt/include -std=c++2b -o.../build/Desktop-Debug/Debug_Desktop_038b678e9426a45b/fmt.875c51b4/f5439ee4c1a5e132/os.cc.o -c .../libs/3rdParty/fmt/src/os.cc
.../libs/3rdParty/fmt/include/fmt/format.h: In function 'constexpr fmt::v11::detail::uint128_fallback fmt::v11::detail::operator*(const uint128_fallback&, uint32_t)':
<command-line>: error: 'assert' was not declared in this scope
.../libs/3rdParty/fmt/include/fmt/format.h:325:5: note: in expansion of macro 'FMT_ASSERT'
  325 |     FMT_ASSERT(lhs.hi_ == 0, "");
      |     ^~~~~~~~~~
In file included from .../libs/3rdParty/fmt/include/fmt/os.h:11,
                 from .../libs/3rdParty/fmt/src/os.cc:13:
```

But declaring it as weak, if and only if FMT_HEADER_ONLY is not defined would also work for me.

vitaut:
Good point, let's reopen this.

vitaut:
Have you considered wrapping {fmt} headers such that your custom assertion function is visible at the point of use?

HazardyKnusperkeks:
I have not. But that would be hard for other libraries which also use fmt (which I don't have yet), and fmt's own source files. I did hope to land some form in the official code.

I understand your concern, and if we don't find an agreement here, I'll probably go with a fork.

vitaut (changes requested):
I think a weak symbol is slightly better than introducing a configuration macro. Also if we are making it overridable the symbol should go from `detail` to the `fmt` (with version) namespace. For ABI compatibility we should also have `detail::assert_fail` that can forward to the public API version.

include/fmt/base.h
```diff
@@ -270,6 +270,16 @@ FMT_PRAGMA_CLANG(diagnostic push)
...
 
+#if (FMT_GCC_VERSION || FMT_CLANG_VERSION) && !defined(FMT_HEADER_ONLY) && \
+    !FMT_WIN32
+// Header only (which implies everything is inline) and weak symbols do not work
+// together. And on Windows weak symbols seem not to work properly.
+#  define FMT_WEAK __attribute__((weak))
+#endif
+#ifndef FMT_WEAK
+#  define FMT_WEAK
+#endif
```
vitaut: Looking at all the caveats, I am now leaning towards your previous version with a configuration macro.

Also what's up with the test failures?

vitaut: Naming-wise I suggest `FMT_CUSTOM_ASSERT_FAIL`.

vitaut (changes requested):

include/fmt/base.h
HazardyKnusperkeks: I had some update trouble with my Linux machine to investigate, my gut feeling is, that the assert tries to jump to an invalid address. I had something like on my Windows machine too, but thought that was because (as I've been told), that weak symbols currently do not work with binutils.

include/fmt/base.h
```diff
@@ -348,6 +348,11 @@ template <typename T> constexpr auto max_of(T a, T b) -> T {
   return a > b ? a : b;
 }
 
+FMT_BEGIN_EXPORT
+FMT_NORETURN FMT_API void assert_fail(const char* file, int line,
+                                      const char* message);
+FMT_END_EXPORT
```
vitaut: I don't think it needs to be exported from the module. The reason why it is in the `fmt` namespace is that it can be replaced by users, but it's not supposed to be called directly.

vitaut:
Overall looks good but please address the remaining comment and rebase.

HazardyKnusperkeks:
Done, sorry for the noise.

vitaut (approved):

vitaut:
Merged, thanks!

PR #4511, opened 2025-08-05
Refactor `ansi_color_escape` to track size instead of using a null terminator

localspook:
It's a small thing, but I think the code is clearer this way, and it avoids throwing away the size in the constructor and then recalculating it in the `end` function.


vitaut:
Merged, thank you!

PR #4512, opened 2025-08-05
Ignore some more files in `.gitignore`

localspook:
`.cache` is generated by the clangd language server, `CMakeUserPresets.json` are local presets that shouldn't be checked in.


vitaut:
Sure, why not

PR #4515, opened 2025-08-14
Silence unreachable code warnings in MSVC when FMT_USE_EXCEPTIONS are…

mattiasljungstrom:
Removes warnings about unreachable code in VS 2022, 17.14.12 when exceptions are disabled.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -138,8 +138,12 @@ FMT_FUNC void report_error(const char* message) {
   // from MSVC.
   FMT_THROW(format_error(message));
 #else
-  fputs(message, stderr);
-  abort();
+  // Silence unreachable code warnings in MSVC.
+  volatile bool b = true;
+  if (b) {
+    fputs(message, stderr);
+    abort();
+  }
 #endif
 }
 


vitaut:
Please post the full warning.

include/fmt/format-inl.h
```diff
@@ -138,8 +138,12 @@ FMT_FUNC void report_error(const char* message) {
   // from MSVC.
   FMT_THROW(format_error(message));
 #else
-  fputs(message, stderr);
-  abort();
+  // Silence unreachable code warnings in MSVC.
+  volatile bool b = true;
+  if (b) {
```
vitaut, line 143: I wonder if this can be

```
if (const_check(true)) {
```

mattiasljungstrom:
> Please post the full warning.

Basically every `return` that happens after the call to `report_error()`:

```
fmt\include\fmt\format.h(3522): warning C4702: unreachable code
fmt\include\fmt\format.h(2192): warning C4702: unreachable code
fmt\include\fmt\base.h(901): warning C4702: unreachable code
fmt\include\fmt\base.h(901): warning C4702: unreachable code
fmt\include\fmt\format.h(3439): warning C4702: unreachable code
fmt\include\fmt\base.h(1433): warning C4702: unreachable code
fmt\include\fmt\base.h(1563): warning C4702: unreachable code
fmt\include\fmt\base.h(1567): warning C4702: unreachable code
fmt\include\fmt\base.h(1349): warning C4702: unreachable code
fmt\include\fmt\base.h(1590): warning C4702: unreachable code
fmt\include\fmt\base.h(1618): warning C4702: unreachable code
fmt\include\fmt\base.h(1625): warning C4702: unreachable code
```

include/fmt/format-inl.h
mattiasljungstrom: Tried this version and it unfortunately doesn't stop the warnings: 
```
FMT_FUNC void report_error(const char* message) {
#if FMT_USE_EXCEPTIONS
  // Use FMT_THROW instead of throw to avoid bogus unreachable code warnings
  // from MSVC.
  FMT_THROW(format_error(message));
#else
  if (detail::const_check(true)) {
    fputs(message, stderr);
    abort();
  }
#endif
}
```

vitaut:
Could you provide a gobolt repro? I tried to repro in https://www.godbolt.org/z/Mjb6xY4vT but to no avail.

mattiasljungstrom:
> Could you provide a gobolt repro? I tried to repro in https://www.godbolt.org/z/Mjb6xY4vT but to no avail.

It seems to only trigger in release mode, and with /W4 /O2

https://www.godbolt.org/z/YfPYK6334

vitaut:
Merged, thanks.

PR #4521, opened 2025-08-29
Use FMT_THROW in report_error

HazardyKnusperkeks:
With exceptions: No change (if FMT_THROW is not user provided)

Without exceptions: Now also states where it comes from (report_error) and uses fprintf instead of fputs.

My motivation is, now that I have my own fmt::assert_failed, also to get rid of fputs for my embedded build.


vitaut:
Looks reasonable, merged.

PR #4527, opened 2025-09-02
Improve Bazel support

mering:
Bazel requires the MODULE.bazel and BUILD.bazel file to be in the root of the repository. With the migration from WORKSPACE to MODULE.bazel, the old approach of moving/renaming files is no longer supported.


vitaut:
Thanks for the PR but as discussed previously, we shouldn't clutter the top-level directory. If you absolutely have to put it in the top-level directory, I suggest creating a separate repo that pulls in {fmt} and provides a Bazel config. We could potentially put this repo under the fmtlib org on github to make it more "official".

PR #4528, opened 2025-09-03
Combined emphases into a single escape sequence

Anas-Hamdane:
Instead of generating separate escape sequences for each emphasis (e.g. `\x1b[1m\x1b[3m` for bold and italic), combine them into a single escape sequence by separating the codes with semicolons (e.g. `\x1b[1;3m`).


vitaut:
Thank you!

PR #4530, opened 2025-09-03
Delete Bazel files

mering:
Bazel requires `BUILD.bazel` files to be in-tree and a `MODULE.bazel` file to mark the root folder of the module. Moving the Bazel support files to the appropriate place was denied in #4527.
So the Bazel related files in the current folder are useless. Worse, they even actively break downstream packaging (for example when `BUILD.bazel` loads from repositories not specified in the downstream `MODULE.bazel` file or when the paths to source files relative to the `BUILD.bazel` file don't exist.
Therefore, @Vertexwahn and myself [agreed](https://github.com/bazelbuild/bazel-central-registry/pull/5736#issuecomment-3246719612) that it would be best to delete these files from this repository.


Vertexwahn (approved):

vitaut:
Thanks!

Issue #4533, opened 2025-09-07
Missing include in gtest-extra.h?


FatihBAKIR:
Building fmt with clang-cl currently fails for me with this error:

```In file included from test/format-test.cc:37:
test/gtest-extra.h(128,49): error: use of undeclared identifier '_CRT_ASSERT'
  128 |         original_report_mode_(_CrtSetReportMode(_CRT_ASSERT, 0)) {}
      |                                                 ^
test/gtest-extra.h(131,23): error: use of undeclared identifier '_CRT_ASSERT'
  131 |     _CrtSetReportMode(_CRT_ASSERT, original_report_mode_);
      |                       ^
2 errors generated.
```

I see we have
```
#ifdef _MSC_VER
#  include <crtdbg.h>
#endif
```
in test/test-main.cc but no such include in format-test.cc.

Adding the include to gtest-extra.h fixes the problem

PR #4534, opened 2025-09-07
Add missing include for crtdbg.h

FatihBAKIR:
This header uses _CRT_ASSERT, which is defined in crtdbg.h but does not include it, causing a build error in format-test.cc

Fixes the issue by including the header

Fixes https://github.com/fmtlib/fmt/issues/4533


vitaut:
Thanks!

PR #4535, opened 2025-09-08
Add support for println with compiled format string.

TobiSchluter:
Using compiled format strings with `println` doesn't work because the corresponding overloads aren't defined. E.g., the following won't compile
```
#include <fmt/compile.h>
#include <fmt/base.h>
int main() { 
    fmt::println(FMT_COMPILE("{}"), "doesn't work");
}
```
The same works with `print`, see e.g. https://godbolt.org/z/8oMj6ofE3

This adds the missing overloads together with tests.

### DIFF
--- a/include/fmt/compile.h
+++ b/include/fmt/compile.h
@@ -545,6 +545,21 @@ void print(const S& fmt, T&&... args) {
   print(stdout, fmt, std::forward<T>(args)...);
 }
 
+template <typename S, typename... T,
+          FMT_ENABLE_IF(is_compiled_string<S>::value)>
+void println(std::FILE* f, const S& fmt, T&&... args) {
+  auto buf = memory_buffer();
+  fmt::format_to(appender(buf), fmt, std::forward<T>(args)...);
+  buf.push_back('\n');
+  detail::print(f, {buf.data(), buf.size()});
+}
+
+template <typename S, typename... T,
+          FMT_ENABLE_IF(is_compiled_string<S>::value)>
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR but the compile-time API is just an optimization and intentionally provides a subset of the API where one could expect performance improvements (at least in principle). `print` can be made more efficient than the equivalent `println` code so I don't think we should provide a compile-time API for the latter. You can always use `print` or the runtime API.

TobiSchluter:
Thanks, I'm only responding now because the message slipped past me while on vacation. I mostly implemented this because trying to use `FMT_COMPILE` with MSVC gave a fairly incomprehensible error message, and it seemed like an innocent enough addition. If we're talking "in principle", the format string in the `println` case could be compiled to include the `\n`, yielding the exact same performance for both :-)

That said, there is an actual gap or inconsistency that this PR fixes: if you use an antiquated C++ standard (e.g. C++14), `FMT_COMPILE` doesn't do anything. So, without my patch, the same code may work or not work depending on the value of `--std=`
See here https://godbolt.org/z/TcazEvP5d , where MSVC succeeds with default flags, but fails with `/std:c++latest`, whereas clang needs `--std=c++14` to succeed but fails with defaults

vitaut:
It is expected that `FMT_COMPILE` has different behavior on older versions, so if you are using it you should definitely test on a compatible version (C++17 or later). I don't think it's worth duplicating APIs to try diagnose misuse on older versions.

PR #4536, opened 2025-09-08
[cmake] fix `MASTER_PROJECT` heuristic; only enable install if root

crueter:
SOURCE vs CURRENT_SOURCE is unreliable. `DEFINED PROJECT_NAME` is a much more portable way of determination.

Also, installation shouldn't default to on if it is a subproject.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,9 @@ endif ()
 # or if it is the master project.
 if (NOT DEFINED FMT_MASTER_PROJECT)
   set(FMT_MASTER_PROJECT OFF)
-  if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR but could you elaborate why the current check is unreliable?

crueter:
> Thanks for the PR but could you elaborate why the current check is unreliable?

With CPM.cmake and I believe FetchContent, FMT_MASTER_PROJECT would always evaluate to ON since those set CMAKE_SOURCE_DIR to CMAKE_CURRENT_SOURCE_DIR for this subdirectory. I remember there being some good reason for this but I can't recall exactly what it is; regardless, with this change, FMT_MASTER_PROJECT now works in all scenarios I tried: CPM, FetchContent, Git submodules, and as a standalone project.

vitaut:
Merged, thanks!

PR #4554, opened 2025-09-28
feat: Allow fmt::join to accept rvalue tuples

teruyamato0731:
### Summary

This PR updates `fmt::join` and its underlying `tuple_join_view` to accept rvalue tuples. This enhances usability and fixes a critical lifetime issue, particularly when using `fmt::join` within `format_as` specializations.

### The Problem

Currently, `fmt::join` only accepts an lvalue reference to a tuple. This prevents passing a temporary tuple, such as one returned from `std::make_tuple` or `std::tie`, directly to the function.

This limitation is particularly dangerous when implementing a `format_as` overload. For example:

```cpp
struct SocketAddr {
  std::string addr;
  int port;
};

// This is unsafe with the current implementation.
auto format_as(const SocketAddr& s) {
  // std::tie returns a temporary (rvalue) tuple of references.
  return fmt::join(std::tie(s.addr, s.port), ":");
}
```

In the code above, the `tuple_join_view` returned by `fmt::join` holds a **dangling reference** to the temporary tuple created by `std::tie`, which is destroyed at the end of the `format_as` function. This leads to undefined behavior when the view is later used for formatting.

### The Solution

This PR resolves the issue by changing `fmt::join` to accept tuples via a forwarding reference (`T&&`).

- If an lvalue tuple is passed, it stores a reference as before.
- If an rvalue tuple is passed, it is moved or copied into the `tuple_join_view`, extending its lifetime and preventing dangling references.

With this change, the `format_as` implementation shown above becomes safe and works as intuitively expected. This makes `fmt::join` a much more powerful and safe tool for creating custom formatters for aggregate types.

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -1598,7 +1598,7 @@ class get_locale {
     ignore_unused(loc);
     ::new (&locale_) std::locale(
 #if FMT_USE_LOCALE
-      loc.template get<std::locale>()
+        loc.template get<std::locale>()
 #endif
     );
   }
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -113,12 +113,13 @@ struct is_range_<T, void>
 
 // tuple_size and tuple_element check.
 template <typename T> class is_tuple_like_ {
-  template <typename U, typename V = remove_cvref_t<U>>
-  static auto check(int) -> decltype(std::tuple_size<V>::value, 0);
+  template <typename U, typename V = typename std::remove_cv<U>::type>
+  static auto check(U* p) -> decltype(std::tuple_size<V>::value, 0);
   template <typename> static void check(...);
 
  public:
-  static constexpr bool value = !std::is_void<decltype(check<T>(0))>::value;
+  static constexpr bool value =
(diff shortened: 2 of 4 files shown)


vitaut:
Thanks for the PR but this handles only one case of dangling references and introduces a major inconsistency in handling rvalue and lvalue references. Therefore I don't think we should go this route and instead we should try to prevent dangling references for both rvalue and lvalue references in a similar way. https://github.com/fmtlib/fmt/commit/b5b9317a3c1b794b0cd1087a4b4076fc2c278dee implements the first step in this direction by adding a warning on unsafe use for clang 12+:

```c++
auto format_as(const SocketAddr& s) {
  // std::tie returns a temporary (rvalue) tuple of references.
  return fmt::join(std::tie(s.addr, s.port), ":");
}
```

```
warning: returning address of local temporary object [-Wreturn-stack-address]
  return fmt::join(std::tie(s.addr, s.port), ":");
                   ^~~~~~~~~~~~~~~~~~~~~~~~
```

https://www.godbolt.org/z/4zTnWcKvc

We could extend this to other compilers and turn this warning into an error by default.

Issue #4542, opened 2025-09-15
std-test fails on clang-cl + libc++


FatihBAKIR:
```
test/std-test.cc(333): error: Value of: fmt::format("{:t}", ex)
Expected: starts with "std::system_error: "
  Actual: "std::__1::system_error: message"
test/std-test.cc(343): error: Value of: fmt::format("{:t}", ex)
Expected: starts with "std::filesystem::filesystem_error: "
  Actual: "std::__1::__fs::filesystem::filesystem_error: filesystem error: message"
```

This happens when building with libc++ instead of the microsoft standard library for windows using clang-cl. The demangling of the type name when `FMT_HAS_ABI_CXA_DEMANGLE` is not defined (which is the case when building for MSVC targets) does not consider the inline namespaces, and the test fails.

FatihBAKIR:
Fixing this by copying the name to a temporary string and doing the same transformation in the `FMT_HAS_ABI_CXA_DEMANGLE` is pretty trivial, but I suspect we want something that avoids allocations here.

vitaut:
> Fixing this by copying the name to a temporary string and doing the same transformation in the FMT_HAS_ABI_CXA_DEMANGLE is pretty trivial ...

Could you submit a PR? I think allocation is fine - it can always be optimized later if anyone is interested.

PR #4557, opened 2025-09-30
style: Correct indentation in locale initialization

teruyamato0731:
Commit 5f66e07 introduced a linting error due to incorrect indentation in the locale initialization code.

This PR corrects the indentation to resolve the issue.

PR #4560, opened 2025-10-01
write_demangled_name supports libc++ + clang-cl

FatihBAKIR:
The current implementation assumes whenever we're on an FMT_MSC_VERSION compiler, the standard library is MSVC's STL. However, with clang-cl we have the possibility of using LLVM libc++ instead of MSVC STL. In that scenario, the previous implementation produced the wrong demangled names for RTTI types.

This patch detects the different combinations, and combines the existing demangling implementations to produce the correct names and make all tests pass on libc++ + clang-cl.

Fixes https://github.com/fmtlib/fmt/issues/4542

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -139,50 +139,39 @@ template <typename Variant, typename Char> class is_variant_formattable {
 #endif  // FMT_CPP_LIB_VARIANT
 
 #if FMT_USE_RTTI
-
-template <typename OutputIt>
-auto write_demangled_name(OutputIt out, const std::type_info& ti) -> OutputIt {
-#  ifdef FMT_HAS_ABI_CXA_DEMANGLE
-  int status = 0;
-  size_t size = 0;
-  std::unique_ptr<char, void (*)(void*)> demangled_name_ptr(
-      abi::__cxa_demangle(ti.name(), nullptr, &size, &status), &std::free);
-
-  string_view demangled_name_view;
-  if (demangled_name_ptr) {
-    demangled_name_view = demangled_name_ptr.get();
-
-    // Normalization of stdlib inline namespace names.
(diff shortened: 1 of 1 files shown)


FatihBAKIR:
@vitaut the lint/format_code task is failing but it also points out a file I did not touch. I also formatted the files locally with clang-format using the `.clang-format` file in the repo, but it still complains in CI. What's the procedure here? Should I manually apply what the CI wants including the unrelated file?

vitaut:
The formatting issue in `chrono.h` has been fixed in #4557. Please rebase and it should go away.

vitaut:
As for the formatting issue in `std.h`, it might be caused by a different version of clang-format. I suggest applying the change from CI manually:

```diff
-auto normalize_msvc_abi_name(string_view abi_name_view,
-                             OutputIt out) -> OutputIt {
+auto normalize_msvc_abi_name(string_view abi_name_view, OutputIt out)
+    -> OutputIt {
```

vitaut:
Thank you for the fix!

include/fmt/std.h
```diff
@@ -139,50 +139,39 @@ template <typename Variant, typename Char> class is_variant_formattable {
...
-        // This is safe, because demangled_name is NUL-terminated.
-        if (from[0] == '_' && from[1] == '_') {
-          char* next = from + 1;
-          while (next < end && *next != ':') next++;
-          if (next[0] == ':' && next[1] == ':') {
-            from = next + 2;
-            continue;
-          }
+string_view normalize_libcxx_inline_namespaces(string_view demangled_name_view,
+                                               char* begin) {
```
vitaut, line 143: This function should have been inline because it is defined in the header.

FatihBAKIR: Oops, mb

PR #4562, opened 2025-10-02
support `std::optional` holding cv-qualified types

### LINKED ISSUE #4561: `std::optional<const T>` doesn't work
`fmt::print("{}", std::optional<const int>{});` doesn't compile. Below is my initial analysis (I'm not really familiar with `fmt` internals):
- `is_formattable<std::optional<const int>>` is `false` because
- `has_formatter` is `false` because
- `decltype(formatter<U, Char>().format(*p, *ctx), std::true_type())` is SFINAEd out because
- `formatter<U, Char>()` is invalid because
- default ctor for `formatter<std::optional<const int>>` is deleted because
- default ctor for `formatter<T, Char> underlying_` [here](https://github.com/fmtlib/fmt/blob/master/include/fmt/std.h#L356) is deleted because
- default ctor for `formatter<const int>` is deleted [here](https://github.com/fmtlib/fmt/blob/4801f54e59d4478a2fe150b7e3ee8327541431aa/include/fmt/base.h#L652) *probably* because
- `native_formatter` is not selected because `type_constant<const int>::value` is `custom_type` instead of `int_type`

I can only guess that `remove_cv_t` is required somewhere in that chain.

OleksandrKvl:
Fixes #4561

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -353,7 +353,7 @@ template <typename T, typename Char>
 struct formatter<std::optional<T>, Char,
                  std::enable_if_t<is_formattable<T, Char>::value>> {
  private:
-  formatter<T, Char> underlying_;
+  formatter<std::remove_cv_t<T>, Char> underlying_;
   static constexpr basic_string_view<Char> optional =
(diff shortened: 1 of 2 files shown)


vitaut:
Thank you!

PR #4563, opened 2025-10-02
Add C++26-style println overloads for convenience

teruyamato0731:
Summary

- Add a set of convenience overloads for fmt::println to make common use-cases more concise and to align with C++26-style usability expectations.

What changed

- New overloads of println were added:
  - `void println()` — writes a newline to stdout.
  - `void println(FILE* f)` — writes a newline to a specified `FILE*`.
- These overloads forward to the existing `print` implementation.

Examples

- No-arg newline:
  - `fmt::println();`
- writes newline to a `FILE*` (stderr in this example):
  - `fmt::println(stderr);`


vitaut:
Thanks for the PR but considering that these functions add very little value and C++26 hasn't even been finalized yet, I don't think we should add them at this point.

PR #4567, opened 2025-10-06
docs: range formatter grammar fix

### LINKED ISSUE #4566: Range formatter grammar
I was trying to understand the grammar of the range formatters here:
https://fmt.dev/latest/syntax/#range-format-specifications

it is stated to be: 
```range_format_spec ::= ["n"][range_type][range_underlying_spec]```
is supposed to be
```range_format_spec ::= ["n"][range_type][":" range_underlying_spec]```

I assume that the `range_underlying_spec` is the format_spec fot the the underlaying type, like for example:
https://fmt.dev/latest/syntax/#format-spec](https://fmt.dev/latest/syntax/#format-spec

which does not include the ":" 

Or am I missing something? 

would be great with an example that specifies both the (n|range_type) and the range_underlying_spec at the same time

petersteneteg:
fixes #4566

### DIFF
--- a/doc/syntax.md
+++ b/doc/syntax.md
@@ -718,7 +718,7 @@ These modifiers are only supported for the `'H'`, `'I'`, `'M'`, `'S'`, `'U'`,
 Format specifications for range types have the following syntax:
 
 <pre><code class="language-json"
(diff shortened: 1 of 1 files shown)


vitaut:
Merged, thanks!

PR #4570, opened 2025-10-10
Revert std::malloc/std::free to global malloc/free (#4569)

### LINKED ISSUE #4569: Usage of std::malloc/free prevents global memory allocator cusomization
With https://github.com/fmtlib/fmt/pull/4477 the usage of malloc and free was changed to std::malloc and std::free.
While this is most likely the correct approach, it breaks global allocator customization since the std:: variants can not be overriden.
One example would be [mimalloc](https://github.com/microsoft/mimalloc), which redefines these globally.

When such a library or global override is used, you end up with a compilation failure since the "free" in "std::free" is redefined to something else.

An alternative to reverting to global malloc/free could also be the introduction of defines that can be configured along the lines of this:
```
#ifndef FMT_MALLOC
#define FMT_MALLOC std::malloc
#endif

#ifndef FMT_FREE
#define FMT_FREE std::free
#endif
 ```

### PR DESCRIPTION
**Pull request by rohitsutreja**
Fixes #4569.

This reverts the change from malloc/free to std::malloc/std::free introduced in [#4477](https://github.com/fmtlib/fmt/pull/4477).
Using the std:: versions  broke compatibility with global allocator overrides (e.g. mimalloc), since std::malloc can’t be intercepted.

Fix

Switched back to global malloc / free.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -490,8 +490,8 @@ template <typename OutputIt,
 #if FMT_CLANG_VERSION >= 307 && !FMT_ICC_VERSION
 __attribute__((no_sanitize("undefined")))
 #endif
-FMT_CONSTEXPR20 inline auto reserve(OutputIt it, size_t n) ->
-    typename OutputIt::value_type* {
+FMT_CONSTEXPR20 inline auto
+reserve(OutputIt it, size_t n) -> typename OutputIt::value_type* {
   auto& c = get_container(it);
   size_t size = c.size();
   c.resize(size + n);

### CODE REVIEW

File: include/fmt/format.h [hunk 1]
```diff
@@ -44,7 +44,7 @@
 #  include <cmath>    // std::signbit
 #  include <cstddef>  // std::byte
 #  include <cstdint>  // uint32_t
-#  include <cstdlib>  // std::malloc, std::free
+#  include <cstdlib>  // malloc, free
```
vitaut: Since we are using non-std variants let's include `<stdlib.h>`.

rohitsutreja: > Since we are using non-std variants let's include `<stdlib.h>`.

done.

include/fmt/format.h
```diff
@@ -489,8 +490,8 @@ template <typename OutputIt,
 #if FMT_CLANG_VERSION >= 307 && !FMT_ICC_VERSION
 __attribute__((no_sanitize("undefined")))
 #endif
-FMT_CONSTEXPR20 inline auto
-reserve(OutputIt it, size_t n) -> typename OutputIt::value_type* {
+FMT_CONSTEXPR20 inline auto reserve(OutputIt it, size_t n) ->
+    typename OutputIt::value_type* {
```
vitaut: clang-format is unhappy about formatting here (might be due to version mismatch): https://github.com/fmtlib/fmt/actions/runs/18409776782/job/52458935547?pr=4570

rohitsutreja: right, it was due to version mismatch. fixed it.

vitaut:
Thank you!

Issue #4565, opened 2025-10-06
When using MSVC to compile v12.0.0, many compilation warnings are generated


esoft-com:
`size_t` and `unsigned __int64` do not match when compiling with x86, resulting in the compiler generating a lot of compilation information
```
fmt\format.h(2509,32): warning C4244: 'argument': conversion from 'unsigned __int64' to 'size_t', possible loss of data [app.vcxproj]
fmt\format.h(3505): message : see reference to function template instantiation'OutputIt fmt::v12::detail::write_fixed<Char,fmt::v12::detail::fallback_digit_grouping<Char>,OutputIt,fmt::v12::detail::dragonbox::decimal_fp<float>>(OutputIt,const DecimalFP &,int,Char,const fmt::v12::format_specs &,fmt::v12::sign,fmt::v12::locale_ref)' being compiled [app.vcxproj]
         with
         [
             OutputIt=fmt::v12::basic_appender<char>,
             Char=char,
             DecimalFP=fmt::v12::detail::dragonbox::decimal_fp<float>
         ]
fmt\format.h(3649): message : see reference to function template instantiation 'OutputIt fmt::v12::detail::write<Char,fmt::v12::basic_appender<char>,T,0>(OutputIt,T)' being compiled [app.vcxproj]
         with
         [
             OutputIt=fmt::v12::basic_appender<char>,
             Char=char,
             T=float
         ]
fmt\base.h(2541): message : see reference to function template instantiation 'void fmt::v12::detail::default_arg_formatter<char>::operator ()<float,0>(T)' being compiled [app.vcxproj]
         with
         [
             T=float
         ]
fmt\base.h(2541): message : see reference to function template instantiation 'void fmt::v12::detail::default_arg_formatter<char>::operator ()<float,0>(T)' being compiled [app.vcxproj]
         with
         [
             T=float
         ]
fmt\format-inl.h(1460): message : see reference to function template instantiation 'void fmt::v12::basic_format_arg<Context>::visit<fmt::v12::detail::default_arg_formatter<char>>(Visitor &&)const' being compiled [app.vcxproj]
         with
         [
             Context=fmt::v12::context,
             Visitor=fmt::v12::detail::default_arg_formatter<char>
         ]
fmt\format-inl.h(1460): message : see reference to function template instantiation 'void fmt::v12::basic_format_arg<Context>::visit<fmt::v12::detail::default_arg_formatter<char>>(Visitor &&)const' being compiled [app.vcxproj]
         with
[...]

PR #4572, opened 2025-10-11
[fix] #4565 When using MSVC to compile v12.0.0, many compilation warn…

LiangHuDream:
@esoft-com This warning occurs because on x86 Windows, size_t is a 32-bit unsigned integer, while unsigned __int64 is a 64-bit unsigned integer. Implicitly converting from unsigned __int64 to size_t can cause data loss, so MSVC emits warning C4244.
To resolve this, we should explicitly cast the value to size_t wherever the conversion happens, ideally at the point where the warning is triggered. thank you for checking the code

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -41,10 +41,11 @@
 #include "base.h"
 
 #ifndef FMT_MODULE
+#  include <stdlib.h>  // malloc, free
+
 #  include <cmath>    // std::signbit
 #  include <cstddef>  // std::byte
 #  include <cstdint>  // uint32_t
-#  include <cstdlib>  // std::malloc, std::free
 #  include <cstring>  // std::memcpy
 #  include <limits>   // std::numeric_limits
 #  include <new>      // std::bad_alloc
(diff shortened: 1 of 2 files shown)


LiangHuDream:
@vitaut 
This warning occurs because on x86 Windows, size_t is a 32-bit unsigned integer, while unsigned __int64 is a 64-bit unsigned integer. Implicitly converting from unsigned __int64 to size_t can cause data loss, so MSVC emits warning C4244.
To resolve this, we should explicitly cast the value to size_t wherever the conversion happens, ideally at the point where the warning is triggered. thank you for checking the code

include/fmt/format.h
```diff
@@ -2506,7 +2506,7 @@ FMT_CONSTEXPR20 auto write_fixed(OutputIt out, const DecimalFP& f,
     auto grouping = Grouping(loc, specs.localized());
     size += grouping.count_separators(exp);
     return write_padded<Char, align::right>(
-        out, specs, to_unsigned(size), [&](iterator it) {
+        out, specs, to_unsigned(static_cast<size_t>(size)), [&](iterator it) {
```
vitaut: Let's remove `to_unsigned` since with the cast the argument is already unsigned.

include/fmt/format.h
```diff
@@ -2506,7 +2506,7 @@ FMT_CONSTEXPR20 auto write_fixed(OutputIt out, const DecimalFP& f,
     auto grouping = Grouping(loc, specs.localized());
     size += grouping.count_separators(exp);
     return write_padded<Char, align::right>(
-        out, specs, to_unsigned(size), [&](iterator it) {
+        out, specs, tatic_cast<size_t>(size), [&](iterator it) {
```
SteveRussell33: Typo: should be ``static_cast``

include/fmt/format.h
LiangHuDream: thank you for checking my code.

LiangHuDream: > Let's remove `to_unsigned` since with the cast the argument is already unsigned.

I modified it .

include/fmt/format.h
LiangHuDream: > Typo: should be `static_cast`

thank you, I modified it,  and I run it successfully.

vitaut:
Merged, thanks.

PR #4577, opened 2025-10-15
Fix module fmt::join<string_view>

### LINKED ISSUE #4379: Module version can't compile fmt::join<std::string_view>
Minimal repro:

Building with:
* clang 19.1.7. Also reproduces with gcc 14.2.1
* ninja 1.12.1
* cmake 3.31.6

Minimal cmake project to reproduce: [proj.zip](https://github.com/user-attachments/files/19166317/proj.zip)

Files inside (just in case):
CMakeLists.txt:
```
cmake_minimum_required(VERSION 3.30 FATAL_ERROR)

set(CMAKE_CXX_STANDARD 23)

project(main LANGUAGES CXX)

add_executable(main main.cpp)

include(FetchContent)
FetchContent_Declare(
        fmt
        GIT_REPOSITORY https://github.com/fmtlib/fmt
        GIT_TAG        123913715afeb8a437e6388b4473fcc4753e1c9a) # 11.1.4


# Comment to disable fmt module

target_sources(main
        PUBLIC FILE_SET CXX_MODULES
        FILES
        ${fmt_SOURCE_DIR}/src/fmt.cc
)
target_include_directories(main PRIVATE ${fmt_SOURCE_DIR}/include)


# Uncomment for header only
# target_link_libraries(main PRIVATE fmt::fmt-header-only)
```

```main.cpp
#include <string>
#include <vector>

import fmt;
// #include <fmt/format.h>
// #include <fmt/ranges.h>


int main(int argc, char* argv[])
{
    std::vector<std::string> strings { "AAA", "BBB", "CCC" };
    std::vector<std::string_view> svs;

    for (size_t i = 0; i < strings.size(); ++i)
        svs.emplace_back(strings[i]);

    fmt::print("{}\n", fmt::join(svs, ", "));
    return 0;
}

```

```
$ mkdir build
$ cd build/
$ CC=clang CXX=clang++ cmake .. -DCMAKE_GENERATOR=Ninja
-- The CXX compiler identification is Clang 19.1.7
-- Check for working CXX compiler: /usr/lib/ccache/bin/clang++ - skipped
Cloning into 'fmt-src'...
HEAD is now at 12391371 Update version
-- {fmt} version: 11.1.4
-- Build type: 
-- Performing Test HAS_NULLPTR_WARNING - Success
-- Configuring done (4.4s)
-- Generating done (0.0s)
-- Build files have been written to: /tmp/other/proj/build
```

```
$ ninja
[8/9] Building CXX object CMakeFiles/main.dir/main.cpp.o
FAILED: CMakeFiles/main.dir/main.cpp.o 
/usr/lib/ccache/bin/clang++  -I/tmp/other/proj/build/_deps/fmt-src/include -std=gnu++23 -MD -MT CMakeFiles/main.dir/main.cpp.o -MF CMakeFiles/main.dir/main.cpp.o.d @CMakeFiles/main.dir/main.cpp.o.modmap -o CMakeFiles/main.dir/main.cpp.o -c /tmp/other/proj/main.cpp
In module 'fmt' imported from /tmp/other/proj/main.cpp:4:
/tmp/other/proj/build/_deps/fmt-src/include/fmt/base.h:2235:45: error: implicit instantiation of undefined template 'fmt::detail::type_is_unformattable_for<fmt::join_view<__gnu_cxx::__normal_iterator<std::basic_string_view<char> *, std::vector<std::basic_string_view<char>>>, __gnu_cxx::__normal_iterator<std::basic_string_view<char> *, std::vector<std::basic_string_view<char>>>>, char>'
 2235 |     type_is_unformattable_for<T, char_type> _;
/tmp/other/proj/build/_deps/fmt-src/include/fmt/base.h:2208:44: note: in instantiation of function template specialization 'fmt::detail::value<fmt::context>::value<fmt::join_view<__gnu_cxx::__normal_iterator<std::basic_string_view<char> *, std::vector<std::basic_string_view<char>>>, __gnu_cxx::__normal_iterator<std::basic_string_view<char> *, std::vector<std::basic_string_view<char>>>>, 0>' requested here
 2208 |   FMT_CONSTEXPR20 FMT_INLINE value(T& x) : value(x, custom_tag()) {}
      |                                            ^
/tmp/other/proj/build/_deps/fmt-src/include/fmt/base.h:2913:22: note: in instantiation of function template specialization 'fmt::detail::value<fmt::context>::value<fmt::join_view<__gnu_cxx::__normal_iterator<std::basic_string_view<char> *, std::vector<std::basic_string_view<char>>>, __gnu_cxx::__normal_iterator<std::basic_string_view<char> *, std::vector<std::basic_string_view<char>>>>, 0>' requested here
 2913 |   vargs<T...> va = {{args...}};
      |                      ^
/tmp/other/proj/main.cpp:17:10: note: in instantiation of function template specialization 'fmt::print<fmt::join_view<__gnu_cxx::__normal_iterator<std::basic_string_view<char> *, std::vector<std::basic_string_view<char>>>, __gnu_cxx::__normal_iterator<std::basic_string_view<char> *, std::vector<std::basic_string_view<char>>>>>' requested here
   17 |     fmt::print("{}\n", fmt::join(svs, ", "));
      |          ^
/tmp/other/proj/build/_deps/fmt-src/include/fmt/base.h:2075:45: note: template is declared here
 2075 | template <typename T, typename Char> struct type_is_unformattable_for;
1 error generated.
ninja: build stopped: subcommand failed.
```

Note that this works fine if:
* fmt::join is used with std::string instead of std::string_view (replace `fmt::join(svs, ", ")` with `fmt::join(strings, ", ")`)
* The non-module version (header only or library) is used (edit  `CMakeLists.txt` and `main.cpp` to use the default build instead of a module).

That is, it only fails when using **both** std::string_view and the module build. It happens with other range iterators, like std::unordered_map and so on.

I though it might be some missing export but I can't find what's missing.

Simple patch to change from module to non-module version:
```patch
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 60cf589..69c6acb 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,13 +16,13 @@ FetchContent_MakeAvailable(fmt)
 
 # Comment to disable fmt module
 
-target_sources(main
-        PUBLIC FILE_SET CXX_MODULES
-        FILES
-        ${fmt_SOURCE_DIR}/src/fmt.cc
-)
-target_include_directories(main PRIVATE ${fmt_SOURCE_DIR}/include)
+# target_sources(main
+#         PUBLIC FILE_SET CXX_MODULES
+#         FILES
+#         ${fmt_SOURCE_DIR}/src/fmt.cc
+# )
+# target_include_directories(main PRIVATE ${fmt_SOURCE_DIR}/include)
 
 
 # Uncomment for header only
-# target_link_libraries(main PRIVATE fmt::fmt-header-only)
+target_link_libraries(main PRIVATE fmt::fmt-header-only)
diff --git a/main.cpp b/main.cpp
index 6b5fb4e..4c06e60 100644
--- a/main.cpp
+++ b/main.cpp
@@ -1,9 +1,9 @@
 #include <string>
 #include <vector>
 
-import fmt;
-// #include <fmt/format.h>
-// #include <fmt/ranges.h>
+// import fmt;
+#include <fmt/format.h>
+#include <fmt/ranges.h>
 
 
 int main(int argc, char* argv[])
```

Arghnews:
`FMT_USE_STRING_VIEW` define was incorrectly not being defined when using modules, so there was no appropriate formatter specialization for std::string_view. Fix is to move define out of `#ifndef FMT_MODULE` block

Fixes #4379

I'm using linux/clang/ninja, couldn't see how to compile module-test.cc from a quick look, happy to add a test in there if you can give me a pointer how I should build it and you prefer.

Test would just be like the original bug filed, eg.:
```cpp
    std::vector<std::string_view> svs;
    fmt::format("{}\n", fmt::join(svs, ", ")); // Actual values
```

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -40,6 +40,12 @@
 
 #include "base.h"
 
+// libc++ supports string_view in pre-c++17.
+#if FMT_HAS_INCLUDE(<string_view>) && \
+    (FMT_CPLUSPLUS >= 201703L || defined(_LIBCPP_VERSION))
+#  define FMT_USE_STRING_VIEW
+#endif
+
 #ifndef FMT_MODULE
 #  include <stdlib.h>  // malloc, free
 
@@ -62,11 +68,8 @@
 #    include <bit>  // std::bit_cast
 #  endif
 
-// libc++ supports string_view in pre-c++17.
-#  if FMT_HAS_INCLUDE(<string_view>) && \
-      (FMT_CPLUSPLUS >= 201703L || defined(_LIBCPP_VERSION))
+#  if defined(FMT_USE_STRING_VIEW)
 #    include <string_view>
-#    define FMT_USE_STRING_VIEW
 #  endif
 
 #  if FMT_MSC_VERSION


include/fmt/format.h
```diff
@@ -62,18 +62,18 @@
...
 #    include <intrin.h>  // _BitScanReverse[64], _umul128
 #  endif
 #endif  // FMT_MODULE
 
+// libc++ supports string_view in pre-c++17.
+#if FMT_HAS_INCLUDE(<string_view>) && \
+    (FMT_CPLUSPLUS >= 201703L || defined(_LIBCPP_VERSION))
+#  include <string_view>
+#  define FMT_USE_STRING_VIEW
+#endif
```
vitaut: This should be moved to before the `FMT_MODULE` check and the include itself remain conditional on `FMT_MODULE` (and `FMT_USE_STRING_VIEW`):

```c++
#  if FMT_USE_STRING_VIEW
#    include <string_view>
#  endif
```

Arghnews: Have made this change (assuming I understood what you meant, since <string_view> is already included in the fmt.cc global module fragment so we don't want to include it here when FMT_MODULE is defined)

vitaut:
Thank you!

PR #4578, opened 2025-10-16
docs: format two unescaped printf references with backticks

zimmerman-dev:
This PR wraps the remaining two unformatted instances of `printf` in backticks so they match how it is styled in the rest of the documentation. No content or wording changes. Just formatting for consistency.

Also removes a small amount of trailing whitespace on an unrelated line (automatically stripped by the editor).


vitaut:
thanks

Issue #4279, opened 2024-12-28
error: module file CMakeFiles/__cmake_cxx23.dir/std.pcm cannot be loaded due to a configuration mismatch with the current compilation 


ClausKlein:
```
bash-5.2$ make
Executing workflow step 1 of 4: configure preset "dev"

Preset CMake variables:

  CMAKE_BUILD_TYPE="Debug"
  CMAKE_CXX_EXTENSIONS="ON"
  CMAKE_CXX_FLAGS="-fstack-protector-strong -Wall -Wextra -Wpedantic -Wno-conversion -Wno-sign-conversion -Wcast-qual -Wformat=2 -Wundef -Wfloat-equal -Wshadow -Wcast-align -Wunused -Wnull-dereference -Wdouble-promotion -Wno-implicit-fallthrough -Wextra-semi -Woverloaded-virtual -Wnon-virtual-dtor -Wno-old-style-cast"
  CMAKE_CXX_STANDARD="23"
  CMAKE_CXX_STANDARD_REQUIRED="ON"
  CMAKE_EXPORT_COMPILE_COMMANDS="ON"
  CMAKE_INSTALL_PREFIX:PATH="/Users/clausklein/Workspace/cpp/cxx20/fmt-module/stagedir"
  CMAKE_PREFIX_PATH:STRING="/Users/clausklein/Workspace/cpp/cxx20/fmt-module/stagedir"
  FMT_DEVELOPER_MODE="ON"

-- use ccache
-- The CXX compiler identification is Clang 19.1.6
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/local/Cellar/llvm/19.1.6/bin/clang++ - skipped
-- Detecting CXX compile features
CMake Warning (dev) at /usr/local/share/cmake-3.31/Modules/Compiler/CMakeCommonCompilerMacros.cmake:248 (cmake_language):
  CMake's support for `import std;` in C++23 and newer is experimental.  It
  is meant only for experimentation and feedback to CMake developers.
Call Stack (most recent call first):
  /usr/local/share/cmake-3.31/Modules/CMakeDetermineCompilerSupport.cmake:113 (cmake_create_cxx_import_std)
  /usr/local/share/cmake-3.31/Modules/CMakeTestCXXCompiler.cmake:83 (CMAKE_DETERMINE_COMPILER_SUPPORT)
  CMakeLists.txt:10 (project)
[...]

PR #4580, opened 2025-10-17
[fix] error: module file CMakeFiles/__cmake_cxx23.dir/std.pcm cannot be loaded due to a configuration mismatch with the current compilation  #4279

LiangHuDream:
[fix] error: module file CMakeFiles/__cmake_cxx23.dir/std.pcm cannot be loaded due to a configuration mismatch with the current compilation  #4279
What I changed

Edited CMakeLists.txt in the add_module_library logic:
Compute a std_flag that honors CMAKE_CXX_EXTENSIONS:
If CMAKE_CXX_EXTENSIONS is ON, use -std=gnu++<ver>.
Otherwise use -std=c++<ver>.
Use ${std_flag} when invoking the compiler to precompile modules:
The precompile command previously used -std=c++<ver> unconditionally; now it uses ${std_flag}.
Use ${std_flag} (and the same include dirs) when compiling the .o from the .pcm to ensure a consistent compile configuration.

Why this helps

The "configuration mismatch" (PCM/PCH load failure) is typically caused by the PCM being produced with a different -std variant (e.g., -std=gnu++23) than the compile unit loading it (e.g., -std=c++23). On Clang the difference between gnu++ and c++ influences whether GNU extensions are enabled; that triggers the "GNU extensions was enabled in PCH file but is currently disabled" diagnostic.
Making the precompile and compile use the same -std avoids that mismatch.
The errors showing that types from std (enable_if, conditional, etc.) are missing are consistent with the compiler being in a different configuration or precompiled header context (modules/PCH broken), so fixing the PCM config mismatch should eliminate those follow-on errors.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,6 +81,15 @@ function(add_module_library name)
     # `std` is affected by CMake options and may be higher than C++20.
     get_target_property(std ${name} CXX_STANDARD)
 
+    # Respect CMAKE_CXX_EXTENSIONS when forming the -std flag used to
+    # precompile modules. A mismatch between -std=gnu++<ver> and -std=c++<ver>
+    # can cause "configuration mismatch" errors when loading .pcm files.
+    if (CMAKE_CXX_EXTENSIONS)
+      set(std_flag "-std=gnu++${std}")
+    else()
+      set(std_flag "-std=c++${std}")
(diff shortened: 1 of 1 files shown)


LiangHuDream:
the format is not entirely right.

PR #4582, opened 2025-10-17
[fix] error: module file CMakeFiles/__cmake_cxx23.dir/std.pcm cannot be loaded due to a configuration mismatch with the current compilation   #4279

LiangHuDream:
[fix] error: module file CMakeFiles/__cmake_cxx23.dir/std.pcm cannot be loaded due to a configuration mismatch with the current compilation https://github.com/fmtlib/fmt/issues/4279
What I changed

The modifications I made (specific and traceable) 
Edit file: CMakeLists.txt
In the manual pre-compilation module path (CMake does not use FILE_SET to manage modules), ensure that CMake's automatic module scanning is not triggered - to avoid build failures caused by CMake requiring the compiler to support import-graph scanning.
In scenarios where it is not safe to use CMake's FILE_SET module support (such as GCC + Ninja), FMT_USE_CMAKE_MODULES is disabled, causing the build to revert to manual pre-compilation/non-FILE_SET paths.
In environments where module support cannot be reliably provided (GCC's several versions), if the user requests -DFMT_MODULE=ON, FMT_MODULE will be automatically set back to OFF (and a warning will be printed) to avoid compiler internal errors (cc1plus segment errors) caused by -fmodules-ts.
In the manual module branch as the target, CXX_SCAN_FOR_MODULES (target property) is turned off, and CMAKE_CXX_SCAN_FOR_MODULES (cache) is globally turned off under GCC to ensure CMake no longer attempts to automatically scan.
Add diagnostic output to enable seeing the decision of FMT_USE_CMAKE_MODULES during configure (helpful for subsequent debugging).
(Notes) I did not modify the public API or source code functionality; only adjusted the CMake configuration logic, with minor and low-risk modifications.

image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,15 @@ if (${CMAKE_VERSION} VERSION_LESS 3.12)
   cmake_policy(VERSION ${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION})
 endif ()
 
+set(CMAKE_CXX_EXTENSIONS OFF)
+set(CMAKE_CXX_STANDARD 23)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    add_compile_options(-stdlib=libc++)
+    add_link_options(-stdlib=libc++)
+endif()
+
 # Determine if fmt is built as a subproject (using add_subdirectory)
 # or if it is the master project.
 if (NOT DEFINED FMT_MASTER_PROJECT)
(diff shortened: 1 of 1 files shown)


vitaut:
This looks similar to #4580. In the future please update the PR instead of closing and creating a new one.

LiangHuDream:
> This looks similar to #4580. In the future please update the PR instead of closing and creating a new one.

OK, thank you for your suggestion.

vitaut:
Could you provide an example demonstrating the problem?

LiangHuDream:
> Could you provide an example demonstrating the problem?

It's ok, thank you for checking the code.

CMakeLists.txt
```diff
@@ -209,13 +248,33 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/support/cmake")
...
+# GCC (GCC lacks import-graph discovery in many setups). This check must run
+# after project() so CMAKE_CXX_COMPILER_ID is available.
+if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.28 AND
+    CMAKE_GENERATOR STREQUAL "Ninja" AND
+    NOT CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+  set(FMT_USE_CMAKE_MODULES TRUE)
+else()
+  set(FMT_USE_CMAKE_MODULES FALSE)
+endif()
+message(STATUS "FMT_USE_CMAKE_MODULES = ${FMT_USE_CMAKE_MODULES}")
```
vitaut: I think it's better not to split the logic to set `FMT_USE_CMAKE_MODULES`.

LiangHuDream: OK

LiangHuDream:
closed, this way can't solve the bug.

Issue #4576, opened 2025-10-14
Odd detail::buffer<char> related dynamic linking issue on Windows (clang-cl)


FatihBAKIR:
I'm running into an issue with my clang-cl based DLL build:

```
LINK: command "lld-link /nologo test/CMakeFiles/printf-test.dir/printf-test.cc.obj /out:bin/printf-test.exe /implib:test/printf-test.lib /pdb:bin/printf-test.pdb /version:0.0 [...] c++.lib ws2_32.lib bcrypt.lib userenv.lib msvcrt.lib msvcprt.lib /machine:x64 /INCREMENTAL:NO /subsystem:console test/test-main.lib fmt.lib test/gtest/gtest.lib kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib /MANIFEST:EMBED,ID=1" failed (exit code 1) with the following output:
lld-link: error: undefined symbol: __declspec(dllimport) public: void __cdecl fmt::v12::detail::buffer<char>::try_resize(unsigned __int64)
>>> referenced by test/CMakeFiles/printf-test.dir/printf-test.cc.obj:(int __cdecl fmt::v12::detail::format_float<double>(double, int, struct fmt::v12::format_specs const &, bool, class fmt::v12::detail::buffer<char> &))
>>> referenced by test/CMakeFiles/printf-test.dir/printf-test.cc.obj:(int __cdecl fmt::v12::detail::format_float<double>(double, int, struct fmt::v12::format_specs const &, bool, class fmt::v12::detail::buffer<char> &))
>>> referenced by test/CMakeFiles/printf-test.dir/printf-test.cc.obj:(int __cdecl fmt::v12::detail::format_float<double>(double, int, struct fmt::v12::format_specs const &, bool, class fmt::v12::detail::buffer<char> &))
>>> referenced 9 more times
```

Which I finally tracked down to this definition in os.h:

```
class FMT_API ostream : private detail::buffer<char> {
```

Since `ostream` has `FMT_API` in it, MSVC ABI considers the inline functions in `detail::buffer<char>` to be `dllimport` when used in `printf-test.cc`, and even if the definition is available in the header, the compiler expects to find the function to be defined in the dll and just emits a relocation. However `fmt.dll` does not have this symbol.

[...]

PR #4584, opened 2025-10-19
Remove FMT_API from ostream class to members

FatihBAKIR:
Putting FMT_API on the class definition propagates it to the base class detail::buffer<char>'s members. However, MSVC not emit definitions for inline members unless it sees the symbols as FMT_API when compiling.

This fix removes the FMT_API declaration from the class itself and marks individual non-inline members as FMT_API to address the issue.

Fixes https://github.com/fmtlib/fmt/issues/4576


vitaut:
It was not just formatting but a build breakage due to a botched merge, please rebase past https://github.com/fmtlib/fmt/commit/a2289b8593d22e2b791f137c996cd575238ca5db.

vitaut:
Thank you!

PR #4587, opened 2025-10-22
添加 format 函数及测试用例

adsdas121:


vitaut:
Thanks for the PR but it's better to use `format_to_n` when formatting to a fixed-size buffer.

PR #4591, opened 2025-10-27
Add missing FMT_CONSTEXPR specifier on static_format_result members.

17steen:
```cpp
file.cpp:36:27: error: constexpr variable 'string_view' must be initialized by a constant expression
   36 |     constexpr static auto string_view = str.str();
      |                           ^             ~~~~~~~~~
```

This error goes away when annotating the member function with FMT_CONSTEXPR


vitaut:
Thank you!

PR #4594, opened 2025-10-30
MSVC x86 on v12.0.0 or v12.1.0 gives compiler warnings

blizzard4591:
Using MSVC x86 to compile fmt v12.0.0 or v12.1.0 gives compiler warnings about __int64 to a 32bit unsigned int conversions.

This is a follow-up for PR #4572.


vitaut:
Merged, thanks.

PR #4595, opened 2025-10-30
Vulnerability fix: uninitialized value access in basic_format_args::get()

BhavikUbhadiya:
## Summary
Fixes uninitialized variable usage flagged by Polaris in `basic_format_args::get()` method.

## Problem
- **Details:**  `High vulnerability` due to use of uninitialized variable
- **Reported by:** Polaris
- **Description:** `Using uninitialized variable 'arg'. Field 'arg.value_' is uninitialized`
- **Location in code:** Inside `basic_format_args::get()` method in `base.h`
- **Reason:** When creating a default `format_arg` object, the union member `value_` was not explicitly initialized.

## Scan snapshot
image

## Solution
Added explicit initialization of `value_.no_value` after creating the default `format_arg` object:

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -2645,6 +2645,7 @@ template <typename Context> class basic_format_args {
   /// Returns the argument with the specified id.
   FMT_CONSTEXPR auto get(int id) const -> format_arg {
     auto arg = format_arg();
+    arg.value_.no_value = {};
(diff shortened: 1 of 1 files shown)


vitaut:
This is a false positive: the value in question is neither accessed nor it has any internal state (it's an empty class).

PR #4601, opened 2025-11-02
Add IWYU pragma to base.h

HazardyKnusperkeks:
So clang-tidy is happy if you include <fmt/format.h> for fmt::formatter<T>.

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -8,6 +8,8 @@
 #ifndef FMT_BASE_H_
 #define FMT_BASE_H_
 
+// IWYU pragma: private, include "fmt/format.h"
+
 #if defined(FMT_IMPORT_STD) && !defined(FMT_MODULE)
 #  define FMT_MODULE
 #endif


vitaut:
Thanks for the PR but I don't think it is correct to mark `fmt/base.h` as a private header in general. Even you consider just `fmt::formatter`, `fmt/base.h` is the header that defines it. You only need to include `fmt/format.h` or other headers like `fmt/std.h` if you need specializations defined in those headers.

HazardyKnusperkeks:
I never considered `fmt/base.h` to be part of the API. But if it is, then everything is fine. Thanks.

mwinterberg-dtek:
https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md#iwyu-pragma-export

would annotating the "base.h" include with export make everyone happy?

HazardyKnusperkeks:
I'm happy. ;) If clang-tidy would be happy, I don't know.

vitaut:
`fmt/format.h` is not a facade for `fmt/base.h` so I don't think the include should be marked with export.

PR #4626, opened 2025-12-08
Make it work on newer mkdocstrings, fix deprecation warnings

mochaaP:
Tested on mkdocs 1.6.1 + mkdocstrings 1.0.0 + Python 3.14.0.
Ruff check target = py3.8

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -24,7 +24,7 @@ jobs:
       run: |
         sudo apt update
         sudo apt install doxygen
-        pip install mkdocs-material==9.5.25 mkdocstrings==0.26.1 mike==2.1.1
+        pip install mkdocs-material==9.7.0 mkdocstrings==1.0.0 mike==2.1.3 typing_extensions==4.15.0
         cmake -E make_directory ${{runner.workspace}}/build
         # Workaround https://github.com/actions/checkout/issues/13:
         git config --global user.name "$(git --no-pager log --format=format:'%an' -n 1)"


support/python/mkdocstrings_handlers/cxx/__init__.py
```diff
@@ -2,206 +2,273 @@
...
 
+# pyright: strict
+
 import os
-import xml.etree.ElementTree as ElementTree
+import xml.etree.ElementTree as ET
 from pathlib import Path
 from subprocess import PIPE, STDOUT, CalledProcessError, Popen
-from typing import Any, List, Mapping, Optional
+from typing import Any, ClassVar, TYPE_CHECKING, final, override
```
vitaut: CI complains about `override`: 

```
ImportError: cannot import name 'override' from 'typing' (/usr/lib/python3.10/typing.py)
```
https://github.com/fmtlib/fmt/actions/runs/20028626137/job/57455771194

mochaaP: fixed

mochaaP:
it passes now: https://github.com/mcha-forks/fmt/actions/runs/20040662950/job/57473549407

.github/workflows/doc.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       run: |
         sudo apt update
         sudo apt install doxygen
-        pip install mkdocs-material==9.5.25 mkdocstrings==0.26.1 mike==2.1.1
+        pip install mkdocs-material==9.7.0 mkdocstrings==1.0.0 mike==2.1.3 typing_extensions==4.15.0
```
vitaut, line 27: Are typing_extension required? In general it's good to keep the number of deps small.

mochaaP: otherwise we add a shim inline:

```python
def override(method, /):
    try:
        method.__override__ = True
    except (AttributeError, TypeError):
        pass
    return method
```

mochaaP: this is the backported runtime behavior for builtin typing. it's [maintained by the python team](https://github.com/python/typing_extensions) and only has one file so i'm okay with it.

vitaut:
Merged, thank you!

PR #4632, opened 2025-12-18
Add test cases to extend bit_test coverage to signed integers

KeshavSM10:


vitaut:
Thanks for the PR but it seems that what this test is actually testing is whether integers are two's complement which only recently became guaranteed. Since we don't rely on this, I don't think we need to test it specifically.

KeshavSM10:
Hey, thanks for the review and the constructive comments. This was my first experience contributing to open source and I learned a lot. I’ll keep this in mind for future contributions.

PR #4638, opened 2025-12-30
Fix missing named-arg validation for compiled format strings

jaeheonshim:
Added `DYNAMIC_NAMES` type parameter to compile_format_string, which is set to true when runtime named arguments are present. Fixed missing named-arg validation by asserting that if DYNAMIC_NAMES is false, the `arg_index` produced by `get_arg_index_by_name(arg_id_result.arg_id.name, Args{})` in the compile time parser must not be -1.

Also added a few tests under `test/compile-error-test` to verify intended behavior.

#4124

### DIFF
--- a/test/compile-error-test/CMakeLists.txt
+++ b/test/compile-error-test/CMakeLists.txt
@@ -210,34 +210,6 @@ if (CMAKE_CXX_STANDARD GREATER_EQUAL 20)
       #error
     #endif
   " ERROR)
-
-  # Format string compilation tests
-  expect_compile(compiled-format-string-missing-argument-error "
-    #if defined(FMT_HAS_CONSTEVAL) && FMT_USE_NONTYPE_TEMPLATE_ARGS
-      using namespace fmt::literals;
-      fmt::format(FMT_COMPILE(\"{x}\"));
-    #else
-      #error
-    #endif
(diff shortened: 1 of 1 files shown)


test/compile-error-test/CMakeLists.txt
```diff
@@ -209,6 +210,34 @@ if (CMAKE_CXX_STANDARD GREATER_EQUAL 20)
       #error
     #endif
   " ERROR)
+
+  # Format string compilation tests
+  expect_compile(compiled-format-string-missing-argument-error "
+    #if defined(FMT_HAS_CONSTEVAL) && FMT_USE_NONTYPE_TEMPLATE_ARGS
+      using namespace fmt::literals;
```
vitaut: This is unused both here and below.

include/fmt/compile.h
```diff
@@ -585,4 +598,4 @@ template <size_t N> class static_format_result {
 FMT_END_EXPORT
 FMT_END_NAMESPACE
 
-#endif  // FMT_COMPILE_H_
+#endif  // FMT_COMPILE_H_
```
vitaut: Missing newline

vitaut:
Merged, thank you!

PR #4639, opened 2025-12-30
Make digits2_i noexcept to match digits2 function

Skylion007:
Add noexcept specifier to digits2_i too make it match https://github.com/fmtlib/fmt/commit/3269c1cea53cb84ab971b36486627367cf73cadc


vitaut:
Thanks!

Issue #4636, opened 2025-12-28
`printf` `*` precision spec issue with positional arguments


akrieger:
I understand the `printf` api is not recommended for new usage. I'm migrating an older codebase from a custom formatting utility over printf to using fmtlib for binary size wins. There is one test case which correctly identifies a failing case in fmtlib which I am reporting in case it is interesting or useful.

`fmt::printf("%6$-*5$.*4$f%3$s%2$s%1$s", "", "", "", 7, 4, 100.44);` returns `width is not integer`.
`::printf("%6$-*5$.*4$f%3$s%2$s%1$s", "", "", "", 7, 4, 100.44)` prints `100.4400000`.

This appears to be because when fmt handles the '*5%' specifier it does not recognize the positional argument and so tries to fetch the -1st arg, which is out of bounds and causes a wild read.

https://godbolt.org/z/YYj7WTqb5

vitaut:
Wow, I don't even know how to parse `%6$-*5$.*4$f` myself =).

akrieger:
I *think* it means, print the 6th arg, left justified, padded with spaces, with a minimum total width of the 5th argument, and a minimum decimal precision of the 4th argument, as a float.

vitaut:
Neither width nor precision support positional arguments because I didn't know it's a thing: https://github.com/fmtlib/fmt/blob/605dd2a1d99c78e2d7f9662beab60c4a25af6941/include/fmt/printf.h#L353-L363

A PR to add such support would be welcome. For width it's trivial - just replacing `-1` with `arg_index` but making it work in all three places requires a bit more work.

vitaut:
Fixed by #4643. Thanks @KareemOtoum!

PR #4643, opened 2026-01-01
add support for positional arguments as width and precision specifiers

KareemOtoum:
this commit adds support for using positional arguments as width and precision specifiers in ```printf.h``` format strings (including as width/precision for float formats) as pointed out by issue  https://github.com/fmtlib/fmt/issues/4636.
Tests have been added to ```printf-test.cc```


vitaut:
Thank you!

PR #4644, opened 2026-01-01
test(scan): add basic float scan coverage

togunchan:
Adds basic test coverage for the experimental scan API’s floating-point handling, including simple float/double parsing and surrounding whitespace.


vitaut:
Thanks

PR #4647, opened 2026-01-04
test(scan): document current exponent parsing behavior

togunchan:
Documents the current behavior of the experimental scan API when parsing floating-point inputs containing exponent markers (e/E) and a leading '+' sign by asserting both the parsed value and the stop position using scan_to. No implementation changes.


vitaut:
Thanks for the PR but I don't think we need to test incompleteness of internal experimental methods.

PR #4649, opened 2026-01-08
zoned_time

LukashonakV:
Hi @vitaut , this PR provides zoned_time formatter.

----

```c++
#include <fmt/chrono.h>  
#include <spdlog/spdlog.h>  

using namespace std::chrono;  

int main(int argc, char* argv[]) {
  const std::chrono::zoned_time now{locate_zone("America/Chicago"),
                       floor<seconds>(system_clock::now())};
  std::string fmt_result{fmt::format("{:%H:%M %Z %z}", now)};
  std::string std_result{std::format("{:%H:%M %Z %z}", now)};

  spdlog::info("\nfmt_result: {}\nstd_result: {}", fmt_result, std_result);
  return 0;
}
```
With the result:
```sh
[2026-01-09 02:09:56.643] [info] 
fmt_result: 17:09 CST -0600
std_result: 17:09 CST -0600
```

### DIFF
--- a/include/fmt/chrono.h
+++ b/include/fmt/chrono.h
@@ -265,6 +265,15 @@ using utc_time = std::chrono::time_point<detail::utc_clock, Duration>;
 template <class Duration>
 using local_time = std::chrono::time_point<detail::local_t, Duration>;
 
+// Check if std::chrono::zoned_time is available.
+#ifdef FMT_HAVE_STD_ZONED_TIME
+// Use the provided definition.
+#elif defined(__cpp_lib_chrono)
+#  define FMT_HAVE_STD_ZONED_TIME (__cpp_lib_chrono >= 201907L)
+#else
+#  define FMT_HAVE_STD_ZONED_TIME 0
+#endif
+
 namespace detail {
 
 // Prevents expansion of a preceding token as a function-style macro.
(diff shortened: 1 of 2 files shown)


vitaut (changes requested):
The formatter should be conditionally defined since `zoned_time` is a C++20 feature.

LukashonakV:
Hi @vitaut , yes I've checked CI logs and is trying to solve it

CMakeLists.txt
```diff
@@ -265,6 +265,65 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   set(WERROR_FLAG -Werror)
 endif ()
 
+include(CheckCXXSourceCompiles)
```
mwinterb: Is checking `__cpp_lib_chrono` for `zoned_time` not reliable enough as it (seemingly) is for `utc_time` and `local_time`?

LukashonakV: O gosh. Thank you a lot. Looks like I'm looking in wrong way :)

LukashonakV:
Now it should be fine.

LukashonakV:
Hi @vitaut , can you please review once again, when you have a time of course.  Thanks)

include/fmt/chrono.h
```diff
@@ -265,6 +265,20 @@ using utc_time = std::chrono::time_point<detail::utc_clock, Duration>;
...
+#elif defined(__cpp_lib_chrono)
+#  define FMT_HAVE_STD_ZONED_TIME (__cpp_lib_chrono >= 201907L)
+#else
+#  define FMT_HAVE_STD_ZONED_TIME 0
+#endif
+
+#if FMT_HAVE_STD_ZONED_TIME
+template <typename Duration, typename TimeZonePtr>
+using zoned_time = std::chrono::zoned_time<Duration, TimeZonePtr>;
+#endif
```
vitaut: Let's not introduce this alias. We only do it as a fallback mechanism which is not needed here.

LukashonakV: It's done

include/fmt/chrono.h
```diff
@@ -2240,6 +2249,42 @@ struct formatter<local_time<Duration>, Char>
...
+
+  template <typename FormatContext>
+  auto format(const std::chrono::zoned_time<Duration, TimeZonePtr>& val,
+              FormatContext& ctx) const -> decltype(ctx.out()) {
+    auto time_info = val.get_info();
+    auto time_since_epoch = val.get_local_time().time_since_epoch();
+    auto seconds_since_epoch =
+        detail::duration_cast<std::chrono::seconds>(time_since_epoch);
+    // Use gmtime to prevent time zone conversion since local_time has an
+    // unspecified time zone.
```
vitaut: Looks like this comment was copied from `local_time` formatter but I don't think it applies here because we are formatting `zoned_time` which does have a specified time zone. I suggest removing the comment.

LukashonakV: Correct, I used local_time formatter as a base for this PR. It's done

vitaut (changes requested):
Overall looks good but please add a unit test to `chrono-test.cc`.

LukashonakV:
> Overall looks good but please add a unit test to `chrono-test.cc`.
Sure, let me do it today.  Thanks

LukashonakV:
Hi @vitaut , tests are added. Hope they are ok.

LukashonakV:
> Looks like there are some CI failures.

Yeah, had not time to check . Will try propably tomorrow

LukashonakV:
> Looks like there are some CI failures.

Yeah, had not time to check . Will try propably tomorrow

include/fmt/chrono.h
```diff
@@ -2240,6 +2249,40 @@ struct formatter<local_time<Duration>, Char>
...
+    auto time_info = val.get_info();
+    auto time_since_epoch = val.get_local_time().time_since_epoch();
+    auto seconds_since_epoch =
+        detail::duration_cast<std::chrono::seconds>(time_since_epoch);
+    std::tm t = gmtime(seconds_since_epoch.count());
+    // Create a custom tm with timezone info if supported
+    if constexpr (detail::has_tm_zone<std::tm>::value) {
+      t.tm_zone = time_info.abbrev.c_str();
+      t.tm_gmtoff = time_info.offset.count();
+    }
```
vitaut, line 2273: I think we should only provide this specialization if `tm` supports time zones. Otherwise the timezone information will be ignored. This is basically why the tests are failing.

vitaut:
@LukashonakV do you still plan to update this PR?

LukashonakV:
Hi @vitaut , sure. Apologies for the delay.  A bit busy with main activities.  Hope next week will be able to check.  Actually I saw the problem and was wondering about the solution..

vitaut:
No hurry, just wanted to check if we should keep this PR open. Since timezone names are system-specific I think it would be fine to limit tests to Linux (and we probably don't need to test so many of them, one or two should be sufficient).

vitaut:
I'll close it for now - we can reopen once this PR is updated.

PR #4653, opened 2026-01-15
Fix C++20 concept detection

tearfur:
Fix compilation on GCC < 8.4 with `-fconcepts`.

~From GCC 8.4 onwards, ISO-19217 concepts `-fconcepts` are compatible with C++20 concepts thanks to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=88075.~ Edit: Irrelevant

Repro of what I'm trying to fix: https://godbolt.org/z/vv179vWGe

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -2803,7 +2803,7 @@ template <typename T, typename Char = char>
 using is_formattable = bool_constant<!std::is_same<
     detail::mapped_t<conditional_t<std::is_void<T>::value, int*, T>, Char>,
     void>::value>;
-#ifdef __cpp_concepts
+#if defined(__cpp_concepts) && __cpp_concepts >= 201907L
 template <typename T, typename Char = char>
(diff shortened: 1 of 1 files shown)


include/fmt/base.h
```diff
@@ -2805,7 +2814,7 @@ using is_formattable = bool_constant<!std::is_same<
     void>::value>;
 #ifdef __cpp_concepts
```
vitaut: To elaborate: we should only check `__cpp_concepts >= 201907L` here.

tearfur: 13c1cbc02289d62c351299ff116646f92962c371 Updated.

tearfur: 9ab39bada8b9af2ea96a40dd420afe74baf9d46e Updated again to fix `-Wundef` warning.

vitaut (approved):

vitaut:
Merged, thanks for the fix!

PR #4659, opened 2026-01-22
Add `constexpr` to argument `value` constructors

elbeno:
I'm having issues with using libfmt at compile time which come down to some value constructors not being marked `constexpr`. I think `constexpr` is the right qualifier to use here rather than `FMT_CONSTEXPR`?


vitaut:
Thank you! `constexpr` is indeed preferred if it's compatible with C++11.

PR #4660, opened 2026-01-22
Fix range suppresor formatter

### LINKED ISSUE #4123: Cannot suppress range formatter when type has a container_type member type.
The enabling condition of the adaptor formatter around line 838 of ranges.h is:

    enable_if_t<conjunction<detail::is_container_adaptor_like<T>,
                            bool_constant<range_format_kind<T, Char>::value ==
                                          range_format::disabled>>::value>>

This is probably supposed to be:

    enable_if_t<conjunction<detail::is_container_adaptor_like<T>,
                            bool_constant<range_format_kind<T, Char>::value !=
                                          range_format::disabled>>::value>>

So that the formatter may be disabled via a is_range specialization.

See compilation error in https://godbolt.org/z/Kq33exr6x

soumik15630m:
Fixes #4123.

### Problem
The `formatter` specialization for container adaptors (like `std::stack`, `std::queue`, `std::priority_queue`) was overly aggressive. It matched any type with a `container_type` member alias, causing ambiguous partial specializations when someone tried to define a custom formatter for their own types that happened to have a `container_type`.

### Solution
added a `fmt::is_container_adaptor<T>` trait. This defaults to `std::true_type` but allows users to specialize it to `std::false_type` for their custom types. This removes the library's adaptor formatter from the overload set via SFINAE, resolving the ambiguity and allowing custom formatters to work.

### Verification
Added a regression test in `test/ranges-test.cc` (`container_adaptor_partial_specialization`) to test the fix and all 

thanks .. please let me know if anything need to be changed

### DIFF
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -347,11 +347,6 @@ template <typename T, typename Char> struct is_range {
       detail::is_range_<T>::value && !detail::has_to_string_view<T>::value;
 };
 
-/**
- * returns "true" if "T" is a container adaptor (like "std::stack")
- * that should be formatted by iterating over the underlying container.
- * */
-
 namespace detail {
 
 template <typename Char, typename Element>
@@ -769,6 +764,12 @@ template <typename Container> struct all {
   auto end() const -> typename Container::const_iterator { return c.end(); }
 };
 }  // namespace detail
+
+/**
+ * returns "true" if "T" is a container adaptor (like "std::stack")
+ * that should be formatted by iterating over the underlying container.
+ * */
+
 FMT_EXPORT
 template <typename T>
 struct is_container_adaptor : detail::is_container_adaptor_like<T> {};


soumik15630m:
@vitaut Tried to fix the problem, check once and please let me know if I need to update anything 
Thanks ..

soumik15630m:
Ping! 
Friendly ping — this PR has all ci  checks passed.
Let me know if there’s anything else needed.
Thanks 👍

soumik15630m:
> Thanks for the PR! I don't think we should introduce another trait for this. Have you tried applying the suggestion from https://github.com/fmtlib/fmt/issues/4123#issue-2470353471?

I considered the approach from #4123, but it would disable formatting for standard adaptors like std::stack, std::queue, and std::priority_queue.
These types are not ranges and resolve to range_format::disabled, so checking != disabled would incorrectly exclude them. The current == disabled logic is what allows the adaptor formatter to trigger correctly.

I introduced the `is_container_adaptor` trait to distinguish between "standard adaptors (which are disabled ranges)" and "user types that explicit opt-out via `is_range` specialization."

Let me know if I missed something or if you'd like me to proceed differently!

vitaut:
I looked into it in more details and you are absolutely right that we cannot reuse the `is_range` trait here. So it makes sense to introduce `is_container_adaptor`. However, I suggest merging `is_container_adaptor_like` into it.

test/ranges-test.cc
```diff
@@ -17,6 +17,10 @@
 #include <utility>
 #include <vector>
 
+#include "fmt/base.h"
+#include "fmt/ranges.h"
+#include "posix-mock.h"
```
vitaut: I don't think we need these. `fmt/ranges.h` is included above and is effectively a superset of `fmt/base.h`.

test/ranges-test.cc
```diff
@@ -797,3 +801,39 @@ struct not_range {
...
+    std::vector<T> c = {1, 2, 3};
+
+    typename std::vector<T>::const_iterator begin() const { return c.begin(); }
+    typename std::vector<T>::const_iterator end() const { return c.end(); }
+  };
+}  // namespace test_detail
+
+namespace fmt {
+  template <typename T>
+  struct is_range<test_detail::partial_opt_out_wrapper<T>, char> : std::false_type {};
```
vitaut: I think we should drop this and test that `is_container_adaptor` is sufficient for the opt out.

soumik15630m:
Makes sense. I’ll drop this specialization and use is_container_adaptor for the opt-out. I’ll update the tests accordingly and push the changes.

soumik15630m:
Updated ....
Let me know if there's anything else...

include/fmt/ranges.h
```diff
@@ -347,6 +347,11 @@ template <typename T, typename Char> struct is_range {
       detail::is_range_<T>::value && !detail::has_to_string_view<T>::value;
 };
 
+/**
+ * returns "true" if "T" is a container adaptor (like "std::stack")
+ * that should be formatted by iterating over the underlying container.
+ * */
```
vitaut: This needs to be moved to the definition of is_container_adaptor.

vitaut (changes requested):

include/fmt/ranges.h
soumik15630m: Sure .... it's updated now 

My bad i overlooked it 
thanks

vitaut:
Merged, thank you!

soumik15630m:
Thanks 👍

Issue #4646, opened 2026-01-03
Potential Regression on #4189: Incorrectly attempting to link `flockfile`


mvastola:
Hi,
So I'm running into the same issue identified in #4189. I'm trying to compile fmt for use on a Raspberry Pi Pico 2 (with the official [`pico-sdk`](https://github.com/raspberrypi/pico-sdk)). I'm using the latest release (12.1.0).

I'm assuming the issue has to do with the function being defined in the headers but not any of the linked libraries (as this is for a micro-controller without an OS or, thus, any filesystems).

If there isn't a way to properly disable this using existing compile flags/checks, can one be added? 
As a workaround I'm currently just commenting out these lines:
https://github.com/fmtlib/fmt/blob/407c905e45ad75fc29bf0f9bb7c5c2fd3475976f/include/fmt/format-inl.h#L1481-L1483

----

As a potential aside (not sure if it's relevant), I'm having a bit of difficulty fully debugging the issue because I can see the `enable_if`-style guard for `has_flockfile` checks for the existence of the `f(un)lockfile` functions, but functions of this name are literally defined within `fmt`, so I'm honestly not sure if it's the standard library headers or `fmt` itself that is creating the belief the `f(un)lockfile` functions exist. 

I also can't make heads or tails of the functions within `fmt` of that name because they depend on functions named `_(un)lock_file`, and I wasn't able to figure out where they are supposed to be defined.

https://github.com/fmtlib/fmt/blob/407c905e45ad75fc29bf0f9bb7c5c2fd3475976f/include/fmt/format-inl.h#L1465-L1470

vitaut:
Please post the full error that you get.

vitaut:
Most likely your system defines `flockfile` but doesn't provide an implementation - I think we've seen a broken system like this in the past. You can disable the file locking functionality by compiling with the `FMT_USE_FALLBACK_FILE` macro defined to 1.

vitaut:
Closing since it doesn't look like a bug in {fmt} and there is a way to workaround it on problematic systems.

mvastola:
@vitaut,
Sorry for the delay, but this is the error I'm getting. 

```
/usr/lib/gcc/arm-none-eabi/14.2.1/../../../arm-none-eabi/bin/ld: ../_deps/fmt-build/libfmtd.a(format.cc.o): in function `fmt::v12::detail::file_print_buffer<__sFILE, void>::file_print_buffer(__sFILE*)':
[...]

PR #4666, opened 2026-02-03
Add FMT_USE_FLOCKFILE macro to force has_flockfile trait to false

mvastola:
Fixes fmtlib/fmt#4646

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -1489,12 +1489,16 @@ template <typename F> auto getc_unlocked(F* f) -> decltype(_fgetc_nolock(f)) {
 }
 #endif
 
+#ifndef FMT_USE_FLOCKFILE
+#  define FMT_USE_FLOCKFILE 1
+#endif
+
 template <typename F = FILE, typename Enable = void>
 struct has_flockfile : std::false_type {};
 
 template <typename F>
 struct has_flockfile<F, void_t<decltype(flockfile(&std::declval<F&>()))>>
-    : std::true_type {};
+    : bool_constant<FMT_USE_FLOCKFILE != 0> {};
 
 // A FILE wrapper. F is FILE defined as a template parameter to make system API
 // detection work.


include/fmt/format-inl.h
```diff
@@ -1489,12 +1489,18 @@ template <typename F> auto getc_unlocked(F* f) -> decltype(_fgetc_nolock(f)) {
...
+#  define FMT_USE_FLOCKFILE 1
+#endif
+
 template <typename F = FILE, typename Enable = void>
 struct has_flockfile : std::false_type {};
 
+#if !FMT_USE_FLOCKFILE
 template <typename F>
 struct has_flockfile<F, void_t<decltype(flockfile(&std::declval<F&>()))>>
     : std::true_type {};
```
vitaut: Let's change `std::true_type` to `bool_constant<FMT_USE_FLOCKFILE != 0>` and remove conditional compilation.

mvastola: Sounds good. Was wondering if you had a preferred way to implement the check.

include/fmt/format-inl.h
```diff
@@ -1489,12 +1489,16 @@ template <typename F> auto getc_unlocked(F* f) -> decltype(_fgetc_nolock(f)) {
...
+#  define FMT_USE_FLOCKFILE 1
+#endif
+
 template <typename F = FILE, typename Enable = void>
 struct has_flockfile : std::false_type {};
 
 template <typename F>
 struct has_flockfile<F, void_t<decltype(flockfile(&std::declval<F&>()))>>
-    : std::true_type {};
+    : std::bool_constant<FMT_USE_FLOCKFILE != 0> {};
```
vitaut: `std::` should go away because we support C++11 and have our own `bool_constant` fallback.

include/fmt/format-inl.h
mvastola: Changed it, but `std::bool_constant` is c++17, so CI is failing now. Should I revert to conditional compilation? Or would you rather I define a
```cpp
template< bool B >
using bool_constant = std::integral_constant<bool, B>;
```
in the `fmt` namespace somewhere? (Or I could just directly use `std::integral_constant`. Your call.)

mvastola: (Source: https://isocpp.org/files/papers/p0636r3.html -- search for `N4389`)

include/fmt/format-inl.h
mvastola: Ah. Gotcha

vitaut:
Thank you!

Issue #4663, opened 2026-01-30
Consider adding a C API


vitaut:
There is a well-defined binary interface between top-level functions such as `fmt::format` and their type-erased counterparts like `fmt::vformat`. This interface relies only on basic C++ features and should be compatible with C (except for the overload returning `std::string` obviously). As a result, it may be possible to provide a C API that builds argument lists using `_Generic` and invokes the type-erased functions.

Unlike the `printf` family of functions, such an API could be fully type-safe, including for runtime format strings, and would not require encoding type information in the format string itself.

This issue is opened to gauge interest. React with a 👍 or another emoji if you would use this.

dimpase:
might be useful for various FFIs.

soumik15630m:
Hey I was curious so i prototyped   C11 `_Generic`-based wrapper to check for the type-safe C API.

The design uses a tagged union + `_Generic` macro to map native C types to a `fmt::dynamic_format_arg_store` backend. It seems to work well without requiring the user to manually specify types in the format string.

Prototype (C11):
``` c
FmtArg args[] = { FMT_MAKE_ARG(42), FMT_MAKE_ARG("World") };
fmt_c_format(buffer, 256, "Hello {} from C! Value: {}", args, 2);
```

This correctly dispatches int, double, and char* at runtime while preserving buffer safety.......the tagged union is ABI-stable and keeps the C header logic-free- no type leaking

If there is interest, I can clean this up into a PR. It would require a minimal extern "C" bridge TU, without expanding the public C++ API surface.

vitaut:
> If there is interest, I can clean this up into a PR.

Sure, a PR is welcome!

soumik15630m:
Thanks for welcoming a PR!

I've implemented a working C API using `_Generic` for type-safe arguments. Before finalizing, I wanted to confirm the architecture since I hit issues trying to use `basic_format_arg` directly.

## Current Implementation (Working)
```c
// C-compatible tagged union
typedef struct {
    FmtType type;  // FMT_INT, FMT_DOUBLE, FMT_STRING, etc.
    union { int64_t i64; double f64; const char* str; ... } value;
} FmtArg;

// C++ bridge converts to fmt
int fmt_c_format(..., const FmtArg* args, size_t count) {
    fmt::dynamic_format_arg_store<fmt::format_context> store;
    for (size_t i = 0; i < count; ++i) {
        switch (args[i].type) {
            case FMT_INT: store.push_back(args[i].value.i64); break;
[...]

PR #4671, opened 2026-02-07
Add C11 API with type-safe formatting (#4663)

soumik15630m:
Issues #4663

## Summary
Implements a type-safe C11 API for {fmt} using `_Generic` macros and a fixed-size argument array as discussed in #4663.

## Design
- **Type safety**: `_Generic` macros (`FMT_MAKE_ARG`) automatically map C types to tagged union
- **ABI stability**: Explicit struct padding, version tracking, stable binary interface
- **Performance**: Fixed-size thread-local array (16 args max), zero allocations for common case
- **Error handling**: Thread-local error strings, negative return codes
- **Custom types**: Callback-based custom formatters with automatic buffer resizing

Architecture: `C type → FmtArg → fixed array → basic_format_arg → vformat`

## Features
- All C primitive types (int, float, double, long double, bool, char, pointers)
- Format specifiers (padding, precision, hex, alignment, positional args)
- Custom formatters via callbacks
- Printf-style macros (`fmt_printf`, `fmt_fprintf`, `fmt_snprintf`)
- Comprehensive error reporting

## Testing
40+ test cases covering:
- All primitive types and format specifiers
- Buffer overflow safety
- Custom formatters
- Error conditions
- Multi-threading safety (thread-local storage)

## Platform Support
- C11 required
- MSVC 2019+ with `/Zc:preprocessor` flag (compile-time check included)
- GCC/Clang with C11 support
- Cross-platform DLL export macros

### DIFF
--- a/include/fmt/fmt-c.h
+++ b/include/fmt/fmt-c.h
@@ -198,4 +198,3 @@ static inline fmt_arg fmt_from_char(int x) {
 #endif  // __cplusplus
 
 #endif  // FMT_C_H
-
--- a/src/fmt-c.cc
+++ b/src/fmt-c.cc
@@ -53,4 +53,3 @@ int fmt_vformat(char* buffer, size_t capacity, const char* format_str,
 }
 
 }  // extern "C"
-
--- a/test/test_c.c
+++ b/test/test_c.c
@@ -200,4 +200,3 @@ int main(void) {
   printf("\n=== All tests passed! ===\n");
   return 0;
 }
-


include/fmt/c.h
```diff
@@ -0,0 +1,276 @@
...
+#define FMT_C_MAX_ARGS 16
+
+#define FMT_OK 0
+#define FMT_ERR_NULL_FORMAT -1
+#define FMT_ERR_EXCEPTION -2
+#define FMT_ERR_MEMORY -3
+
+#ifdef __cplusplus
+#  include <cstddef>
+#  include <cstdint>
```
vitaut: This is not needed, we should always use C headers here.

include/fmt/fmt-c.h
```diff
@@ -0,0 +1,276 @@
...
+}
+#endif
+
+#ifndef __cplusplus
+
+// Require modern MSVC with conformant preprocessor
+#  if defined(_MSC_VER) && (!defined(_MSVC_TRADITIONAL) || _MSVC_TRADITIONAL)
+#    error \
+        "C API requires MSVC 2019+ with /Zc:preprocessor flag. Add /Zc:preprocessor to your compiler flags."
+#  endif
```
vitaut, line 124: What is this for?

src/c.cc
```diff
@@ -0,0 +1,241 @@
...
+#include "fmt/c.h"
+
+#include <fmt/core.h>
+
+#include <array>
+#include <exception>
+#include <string>
+#include <vector>
+
+static const size_t MAX_PACKED_ARGS = FMT_C_MAX_ARGS;
```
vitaut: This is not needed because `FMT_C_MAX_ARGS` can be used directly.

src/c.cc
```diff
@@ -0,0 +1,241 @@
...
+
+extern "C" {
+static thread_local std::string g_last_error;
```
vitaut: We shouldn't use mutable globals.

include/fmt/c.h
```diff
@@ -0,0 +1,276 @@
...
+    f(a), f(b), f(c), f(d), f(e), f(g), f(h), f(i), f(j), f(k), f(l), f(m), \
+        f(n), f(o), f(p)
+#  define FMT_MAP_16(f, a, b, c, d, e, g, h, i, j, k, l, m, n, o, p, q)     \
+    f(a), f(b), f(c), f(d), f(e), f(g), f(h), f(i), f(j), f(k), f(l), f(m), \
+        f(n), f(o), f(p), f(q)
+
+#  define FMT_MAP(f, ...) \
+    FMT_CAT(FMT_MAP_, FMT_NARG(__VA_ARGS__))(f, ##__VA_ARGS__)
+
+#  define fmt_snprintf(buf, cap, fmt, ...)                               \
```
vitaut: Please rename to `fmt_format` since it uses a different syntax from `printf`.

include/fmt/c.h
```diff
@@ -0,0 +1,276 @@
...
+        (FmtArg[]){{FMT_INT}, FMT_MAP(FMT_MAKE_ARG, ##__VA_ARGS__)} + 1, \
+        FMT_NARG(__VA_ARGS__))
+
+#  define fmt_fprintf(f, fmt, ...)                                       \
+    fmt_c_print(                                                         \
+        f, fmt,                                                          \
+        (FmtArg[]){{FMT_INT}, FMT_MAP(FMT_MAKE_ARG, ##__VA_ARGS__)} + 1, \
+        FMT_NARG(__VA_ARGS__))
+
+#  define fmt_printf(fmt, ...) fmt_fprintf(stdout, fmt, ##__VA_ARGS__)
```
vitaut: Let's not include these in the initial implementation and just focus on formatting to a memory buffer.

include/fmt/fmt-c.h
soumik15630m: becuase i used ' __VA_ARGS__ ' and it was causing problems while compiling with legacy compiler so i tested with zc:preprocessor and it worked ...
it need __VA_ARGS__ becuz of that FMT_MAP uses this to split and apply FMT_MAKE_ARG to each one

soumik15630m:
sure , i will implement the necessary changes

include/fmt/fmt-c.h
vitaut: We don't need this check because the compiler will already give an error if `VA_ARGS` are not supported. Let's remove.

src/c.cc
soumik15630m: I used it - so that fmt_get_c_error can retrieve it ......

but sure , i will change it 


just a quick question to remove it should i go with a optional buffer in format functions or just error codes will be enogh

include/fmt/fmt-c.h
soumik15630m: sure , i'll remove it. 
Just to clarify - the legacy preprocessor technically 'supports' __VA_ARGS__ but expands them incorrectly...effectively as a single token , which causes cryptic syntax errors downstream rather than a clear 'unsupported' error.

src/c.cc
vitaut: Error codes should be enough.

include/fmt/fmt-c.h
vitaut: Interesting. Could you give a godbolt example illustrating the cryptic error?

soumik15630m: https://godbolt.org/z/dcd83Goz3

but if we remove the compiler flags the count comes 1

include/fmt/c.h
soumik15630m: I’ve implemented the requested changes.

For context, these features were already functioning correctly before the update.

Let me know if there’s anything else you’d like me to revise.
Thanks.

include/fmt/fmt-c.h
soumik15630m: Hi @vitaut 
Just a friendly knudge...... What to do with this error message - i refactored every other thing according to the reviews.... Just this is remaining 

Thanks 👍

include/fmt/fmt-c.h
```diff
@@ -0,0 +1,263 @@
+#ifndef FMT_C_API_H
```
vitaut: Macro guards should match the file name, i.e. `FMT_C_H_`

include/fmt/fmt-c.h
```diff
@@ -0,0 +1,263 @@
...
+#define FMT_C_API_H
+
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+
+#define FMT_C_ABI_VERSION 1
```
vitaut: I don't think we need this.

include/fmt/fmt-c.h
```diff
@@ -0,0 +1,263 @@
...
+#define FMT_C_MAX_ARGS 16
+
+#define FMT_OK 0
+#define FMT_ERR_NULL_FORMAT -1
+#define FMT_ERR_EXCEPTION -2
+#define FMT_ERR_MEMORY -3
+#define FMT_ERR_INVALID_ARG -4
```
vitaut: Let's make this an enum.

include/fmt/fmt-c.h
```diff
@@ -0,0 +1,263 @@
...
+
+#if defined(_WIN32) && !defined(FMT_C_STATIC)
+#  ifdef FMT_C_EXPORT
+#    define FMT_C_API __declspec(dllexport)
+#  else
+#    define FMT_C_API __declspec(dllimport)
+#  endif
+#else
+#  define FMT_C_API
+#endif
```
vitaut: Static library should be sufficient for now, let's remove.

include/fmt/fmt-c.h
```diff
@@ -0,0 +1,263 @@
...
+
+// Custom formatter callback
+// Returns number of bytes written (excluding null terminator), or -1 on error
+typedef int (*FmtCustomFn)(char* buf, size_t cap, const void* data);
```
vitaut: Let's remove this too.

include/fmt/fmt-c.h
```diff
@@ -0,0 +1,263 @@
...
+  FMT_UINT,
+  FMT_FLOAT,
+  FMT_DOUBLE,
+  FMT_LONG_DOUBLE,
+  FMT_STRING,
+  FMT_PTR,
+  FMT_BOOL,
+  FMT_CHAR,
+  FMT_CUSTOM
+} FmtType;
```
vitaut: fmt uses snake_case.

include/fmt/fmt-c.h
```diff
@@ -0,0 +1,263 @@
...
+typedef struct {
+  FmtType type;
+
+  // Explicit padding for ABI stability
+  //   - type:      4 bytes (enum)
+  //   - _padding:  4 bytes (explicit alignment)
+  //   - value:    16 bytes (union, sized by long double)
+  //   - custom_fn: 8 bytes (function pointer)
+  // Ensures consistent struct size across compilers (..* 24 bytes in MSVC)
+  int32_t _padding;
```
vitaut: We shouldn't do any heroics here, let's remove.

include/fmt/fmt-c.h
```diff
@@ -0,0 +1,263 @@
...
+    double f64;
+    long double f128;
+    const char* str;
+    const void* ptr;  // Used for FMT_PTR and custom data
+    int bool_val;
+    int char_val;
+  } value;
+
+  // FMT_CUSTOM type only
+  FmtCustomFn custom_fn;
```
vitaut: Let's make it `void*` for now.

include/fmt/fmt-c.h
```diff
@@ -0,0 +1,263 @@
...
+} FmtArg;
+
+FMT_C_API int fmt_c_format(char* buffer, size_t capacity,
```
vitaut: Let's rename it to `fmt_vformat` for consistency with the C++ API.

include/fmt/fmt-c.h
```diff
@@ -0,0 +1,263 @@
...
+                           const char* format_str, const FmtArg* args,
+                           size_t arg_count);
+
+FMT_C_API int fmt_c_get_version(void);
```
vitaut: Let's remove this.

include/fmt/fmt-c.h
```diff
@@ -0,0 +1,263 @@
...
+
+static inline FmtArg fmt_from_int(int64_t x) {
+  FmtArg a;
+  a.type = FMT_INT;
+  a._padding = 0;
+  a.value.i64 = x;
+  a.custom_fn = NULL;
+  return a;
```
vitaut: This can be simplified with designated initializers.

include/fmt/fmt-c.h
```diff
@@ -0,0 +1,263 @@
...
+static inline FmtArg fmt_from_custom(const void* data, FmtCustomFn func) {
+  FmtArg a;
+  a.type = FMT_CUSTOM;
+  a._padding = 0;
+  a.value.ptr = data;
+  a.custom_fn = func;
+  return a;
+}
+
+static inline FmtArg fmt_identity(FmtArg x) { return x; }
```
vitaut: I don't think we need this.

include/fmt/fmt-c.h
vitaut: Let's keep this but I would suggest making the error message shorter, just

```
C API requires MSVC 2019+ with /Zc:preprocessor flag.
```

include/fmt/fmt-c.h
```diff
@@ -0,0 +1,263 @@
...
+
+#  define fmt_format(buf, cap, fmt, ...)                                 \
+    fmt_c_format(                                                        \
+        buf, cap, fmt,                                                   \
+        (FmtArg[]){{FMT_INT}, FMT_MAP(FMT_MAKE_ARG, ##__VA_ARGS__)} + 1, \
+        FMT_NARG(__VA_ARGS__))
+
+#endif  // !__cplusplus
+
+#endif  // FMT_C_API_H
```
vitaut: Here and elsewhere: missing newline at the end of the file.

src/fmt-c.cc
```diff
@@ -0,0 +1,133 @@
+#undef FMT_C_EXPORT
+#define FMT_C_EXPORT
```
vitaut: This is not needed.

src/fmt-c.cc
```diff
@@ -0,0 +1,133 @@
...
+
+extern "C" {
+
+int fmt_c_get_version(void) { return FMT_C_ABI_VERSION; }
+
+using Context = fmt::format_context;
+
+// Fixed-size array for type-erased format arguments
+static thread_local std::array<fmt::basic_format_arg<Context>, FMT_C_MAX_ARGS>
+    g_fixed_store;
```
vitaut: This should be stack-allocated, not global. And we don't need std::array.

src/fmt-c.cc
```diff
@@ -0,0 +1,133 @@
...
+  }
+  return true;
+}
+int fmt_c_format(char* buffer, size_t capacity, const char* format_str,
+                 const FmtArg* args, size_t arg_count) {
+  if (!format_str) return FMT_ERR_NULL_FORMAT;
+  if (arg_count > FMT_C_MAX_ARGS) return FMT_ERR_INVALID_ARG;
+
+  try {
+    std::vector<std::string> custom_buffers;
```
vitaut: What is this for?

src/fmt-c.cc
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    default: return false;
+    }
+  }
+  return true;
+}
+int fmt_c_format(char* buffer, size_t capacity, const char* format_str,
+                 const FmtArg* args, size_t arg_count) {
+  if (!format_str) return FMT_ERR_NULL_FORMAT;
```
vitaut: This should be an assert.

src/fmt-c.cc
```diff
@@ -0,0 +1,133 @@
...
+  if (arg_count > FMT_C_MAX_ARGS) return FMT_ERR_INVALID_ARG;
+
+  try {
+    std::vector<std::string> custom_buffers;
+    if (arg_count > 0) {
+      if (!args) return FMT_ERR_INVALID_ARG;
```
vitaut: also an assert

src/fmt-c.cc
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    auto format_args_view = fmt::basic_format_args<Context>(
+        g_fixed_store.data(), static_cast<int>(arg_count));
+
+    if (!buffer || capacity == 0) {
+      char tmp[1];
+      auto result = fmt::vformat_to_n(tmp, 0, format_str, format_args_view);
+      return static_cast<int>(result.size);
+    }
```
vitaut: We shouldn't special case this.

src/fmt-c.cc
```diff
@@ -0,0 +1,133 @@
...
+
+    auto result =
+        fmt::vformat_to_n(buffer, capacity - 1, format_str, format_args_view);
+    *result.out = '\0';
```
vitaut: Let's keep the same behavior as the C++ API and not null terminate.

test/test_c.c
```diff
@@ -0,0 +1,375 @@
+/* Test suite for fmt C API */
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "fmt/fmt-c.h"
+
+#define TEST(name)                          \
```
vitaut: I don't think we need this macro, let's just use normal functions.

test/test_c.c
```diff
@@ -0,0 +1,375 @@
...
+TEST(mixed_types) {
+  char buf[100];
+  fmt_format(buf, sizeof(buf), "{} {} {} {}", 42, 3.14, "text", (bool)true);
+  ASSERT_TRUE(strstr(buf, "42") != NULL);
+  ASSERT_TRUE(strstr(buf, "3.14") != NULL);
+  ASSERT_TRUE(strstr(buf, "text") != NULL);
+  ASSERT_TRUE(strstr(buf, "true") != NULL);
+}
+
+TEST(format_zero_padding) {
```
vitaut: Since the C API is just a wrapper, we don't need to re-test formatting features like padding. The only thing we need to check is that arguments are correctly forwarded.

CMakeLists.txt
```diff
@@ -532,3 +532,68 @@ if (FMT_MASTER_PROJECT AND EXISTS ${gitignore})
   set(CPACK_RESOURCE_FILE_README ${PROJECT_SOURCE_DIR}/README.md)
   include(CPack)
 endif ()
+
+# C API Wrapper
+
+option(FMT_C_API "Build C API wrapper" ON)
+
+if(FMT_C_API)
+  message(STATUS "Building C API wrapper (fmt::fmt_c)")
```
vitaut: Let's remove this.

CMakeLists.txt
```diff
@@ -532,3 +532,68 @@ if (FMT_MASTER_PROJECT AND EXISTS ${gitignore})
...
+
+  enable_language(C)
```
vitaut, line 567: This should be moved to the test section because the C API is implemented in C++.

CMakeLists.txt
```diff
@@ -532,3 +532,68 @@ if (FMT_MASTER_PROJECT AND EXISTS ${gitignore})
...
+          CXX_VISIBILITY_PRESET hidden
+  )
+
+  add_library(fmt::fmt_c ALIAS fmt_c)
+  if(FMT_INSTALL)
+    install(TARGETS fmt_c
+            EXPORT fmt-targets
+            LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+            ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+            RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
```
vitaut: I don't think we need all of these for static library.

CMakeLists.txt
```diff
@@ -532,3 +532,68 @@ if (FMT_MASTER_PROJECT AND EXISTS ${gitignore})
...
+    )
+    install(FILES include/fmt/fmt-c.h
+            DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/fmt
+    )
+  endif()
+
+  if(FMT_TEST AND EXISTS ${PROJECT_SOURCE_DIR}/test/test_c.c)
```
vitaut: EXISTS is not needed.

CMakeLists.txt
```diff
@@ -532,3 +532,68 @@ if (FMT_MASTER_PROJECT AND EXISTS ${gitignore})
...
+    message(STATUS "Adding C API test executable")
+
+    add_executable(test-c-api test/test_c.c)
+
+    set_source_files_properties(test/test_c.c PROPERTIES LANGUAGE C)
```
vitaut: This seems unnecessary.

CMakeLists.txt
```diff
@@ -532,3 +532,68 @@ if (FMT_MASTER_PROJECT AND EXISTS ${gitignore})
...
+
+    target_link_libraries(test-c-api PRIVATE fmt::fmt_c)
+#needed for c11(_generic)
+    if(MSVC)
+      target_compile_options(test-c-api PRIVATE /std:c11 /Zc:preprocessor)
+    else()
+      set_target_properties(test-c-api PROPERTIES
+              C_STANDARD 11
+              C_STANDARD_REQUIRED ON
+              C_EXTENSIONS OFF
```
vitaut: We can keep extensions.

vitaut (changes requested):

soumik15630m:
Thanks for the detailed review. I’ll  go thorugh the reviews and address these changes shortly....and push the changes

include/fmt/fmt-c.h
vitaut: To elaborate: it might be worth adding custom formatter support but to keep things manageable let's not do it in this PR.

include/fmt/fmt-c.h
soumik15630m: Done

include/fmt/fmt-c.h
soumik15630m: Makes sense....
will update

include/fmt/fmt-c.h
soumik15630m: as per your updated comment i removed the custom_fn

include/fmt/fmt-c.h
soumik15630m: sure

src/fmt-c.cc
soumik15630m: it was used to manage the lifetime of temporary strings that will be  generated by custom formatters.....it was already removed as the custom formater si not needed rn

src/fmt-c.cc
soumik15630m: i added it for conveniance ...but sure - got the point

test/test_c.c
soumik15630m: sure ,,,, i will update the test file

CMakeLists.txt
soumik15630m: yuppp

CMakeLists.txt
soumik15630m: sure

include/fmt/fmt-c.h
soumik15630m: it won't be valid in c++ -- compile time 
if we use this 
```
return (fmt_arg){.type = fmt_int, .value.i64 = x};
```

instead of this

soumik15630m:
appologies for the force pushes i was working on someone else's laptop and forgot to change the commit authorship before pushing 

the changes reauested are implemented 
let me know if anything else needs to be updated

include/fmt/fmt-c.h
```diff
@@ -0,0 +1,200 @@
...
+  fmt_ptr,
+  fmt_bool,
+  fmt_char
+} fmt_type;
+
+typedef struct {
+  fmt_type type;
+  union {
+    int64_t i64;
+    uint64_t u64;
```
vitaut: Let's use more portable `long long` and `unsigned long long` and remove `stdint.h` include.

include/fmt/fmt-c.h
```diff
@@ -0,0 +1,200 @@
+#ifndef FMT_C_H
+#define FMT_C_H
+
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+
+#define FMT_C_MAX_ARGS 16
```
vitaut: Let's make it an enum.

include/fmt/fmt-c.h
```diff
@@ -0,0 +1,200 @@
...
+
+typedef enum {
+  fmt_ok = 0,
```
vitaut: `fmt_ok` is not used and the success is indicated by nonnegative numbers, let's remove.

include/fmt/fmt-c.h
```diff
@@ -0,0 +1,200 @@
...
+  fmt_type type;
+  union {
+    int64_t i64;
+    uint64_t u64;
+    float f32;
+    double f64;
+    long double f128;
+    const char* str;
+    const void* ptr;  // Used for FMT_PTR and custom data
+    int bool_val;
```
vitaut: Why `int` and not `_Bool`?

include/fmt/fmt-c.h
```diff
@@ -0,0 +1,200 @@
...
+    int char_val;
```
vitaut: Similarly, this should probably be `char`.

include/fmt/fmt-c.h
```diff
@@ -0,0 +1,200 @@
...
+        long long: fmt_from_int,           \
+        unsigned long long: fmt_from_uint, \
+        float: fmt_from_float,             \
+        double: fmt_from_double,           \
+        long double: fmt_from_long_double, \
+        char*: fmt_from_str,               \
+        const char*: fmt_from_str,         \
+        void*: fmt_from_ptr,               \
+        const void*: fmt_from_ptr,         \
+        default: fmt_from_ptr)(x)
```
vitaut: Why is the default `fmt_from_ptr`. I think we should make an unknown type an error.

include/fmt/fmt-c.h
```diff
@@ -0,0 +1,200 @@
...
+}
+
+static inline fmt_arg fmt_from_ptr(const void* x) {
+  fmt_arg arg;
+  arg.type = fmt_ptr;
+  arg.value.ptr = x;
+  return arg;
+}
+
+static inline fmt_arg fmt_from_bool(bool x) {
```
vitaut: I think we can use `_Bool` here and remove `stdbool.h` include.

include/fmt/fmt-c.h
```diff
@@ -0,0 +1,200 @@
+#ifndef FMT_C_H
+#define FMT_C_H
+
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
```
vitaut: This is not used, please remove.

src/fmt-c.cc
```diff
@@ -0,0 +1,92 @@
+#include "fmt/fmt-c.h"
+
+#include <fmt/core.h>
+
+#include <array>
```
vitaut: Unused include

src/fmt-c.cc
```diff
@@ -0,0 +1,92 @@
...
+#include <cassert>
+#include <exception>
+#include <string>
```
vitaut: same here

src/fmt-c.cc
```diff
@@ -0,0 +1,92 @@
...
+#include <vector>
```
vitaut: and here

src/fmt-c.cc
```diff
@@ -0,0 +1,92 @@
...
+
+extern "C" {
+
+using Context = fmt::format_context;
```
vitaut: Since it's only used in `fmt::basic_format_arg`, let's introduce an alias for the latter instead:

```
using format_arg = fmt::basic_format_arg<fmt::format_context>;
```

src/fmt-c.cc
```diff
@@ -0,0 +1,92 @@
...
+static bool populate_store(fmt::basic_format_arg<Context>* out,
+                           const fmt_arg* c_args, size_t arg_count) {
+  if (arg_count > FMT_C_MAX_ARGS) {
+    return false;
+  }
+
+  for (size_t i = 0; i < arg_count; ++i) {
+    switch (c_args[i].type) {
+    case fmt_int:
+      out[i] = fmt::basic_format_arg<Context>(c_args[i].value.i64);
```
vitaut: Is explicit ctor invocation needed here?

src/fmt-c.cc
```diff
@@ -0,0 +1,92 @@
...
+      break;
+    case fmt_char:
+      out[i] = fmt::basic_format_arg<Context>(
+          static_cast<char>(c_args[i].value.char_val));
+      break;
+    case fmt_bool:
+      out[i] = fmt::basic_format_arg<Context>(c_args[i].value.bool_val != 0);
+      break;
+    case fmt_string: {
+      const char* s = c_args[i].value.str ? c_args[i].value.str : "(null)";
```
vitaut: Let's not remap null strings to anything.

src/fmt-c.cc
```diff
@@ -0,0 +1,92 @@
...
+                const fmt_arg* args, size_t arg_count) {
+  assert(format_str);
+
+  fmt::basic_format_arg<Context> format_args[FMT_C_MAX_ARGS];
+
+  try {
+    if (arg_count > 0) {
+      assert(args);
+      if (!populate_store(format_args, args, arg_count)) {
+        return fmt_err_exception;
```
vitaut: I think it should be `fmt_err_invalid_arg` rather than `fmt_err_exception`.

src/fmt-c.cc
```diff
@@ -0,0 +1,92 @@
...
+    auto format_args_view = fmt::basic_format_args<Context>(
+        format_args, static_cast<int>(arg_count));
+
+    size_t write_capacity = (capacity > 0) ? capacity - 1 : 0;
+    auto result =
+        fmt::vformat_to_n(buffer, write_capacity, format_str, format_args_view);
+
+    if (capacity > 0) {
+      buffer[result.size] = '\0';
+    }
```
vitaut: Let's keep the behavior consistent with the rest of fmt and not add null termination here.

src/fmt-c.cc
```diff
@@ -0,0 +1,92 @@
...
+    return static_cast<int>(result.size);
+
+  } catch (const std::bad_alloc&) {
+    return fmt_err_memory;
```
vitaut: Since this is formatting to a fixed memory buffer and standard formatters normally don't allocate, I don't think we need to catch bad_alloc.

CMakeLists.txt
```diff
@@ -533,3 +532,54 @@ if (FMT_MASTER_PROJECT AND EXISTS ${gitignore})
   set(CPACK_RESOURCE_FILE_README ${PROJECT_SOURCE_DIR}/README.md)
   include(CPack)
 endif ()
+
+# C API Wrapper
+add_library(fmt_c STATIC src/fmt-c.cc)
+
+target_compile_features(fmt_c PUBLIC cxx_std_11)
+target_compile_definitions(fmt_c PUBLIC FMT_C_STATIC)
```
vitaut: This is no longer needed.

src/fmt-c.cc
soumik15630m: Sure

include/fmt/fmt-c.h
soumik15630m: sorry old habit .... in legacy c we normmally use int as bool and move on - so that's why ..
i updated it now 
thanks

include/fmt/fmt-c.h
soumik15630m: yeah - it was during the testing phase - later i forgot to change it ...i was testing and recompiling many times and it was genereting an error so i just defaulted it to ptr

i updated it now - thanks ...with this 

```
void fmt_error_unsupported_type_detected(void);

default: fmt_error_unsupported_type_detected)(x)
```

otherwise compiler can't resolve the _generic macro

include/fmt/fmt-c.h
soumik15630m: sure

src/fmt-c.cc
soumik15630m: yh

src/fmt-c.cc
soumik15630m: got it

src/fmt-c.cc
soumik15630m: nope , it was just from a safety standpoint 

updated ..

vitaut:
Merged, thank you!

soumik15630m:
Thanks 👍

include/fmt/fmt-c.h
```diff
@@ -0,0 +1,200 @@
...
+    f(a), f(b), f(c), f(d), f(e), f(g), f(h), f(i), f(j), f(k), f(l), f(m), \
+        f(n), f(o), f(p), f(q)
+
+#  define FMT_MAP(f, ...) \
+    FMT_CAT(FMT_MAP_, FMT_NARG(__VA_ARGS__))(f, ##__VA_ARGS__)
+
+#  define fmt_format(buf, cap, fmt, ...)                                  \
+    fmt_vformat(                                                          \
+        buf, cap, fmt,                                                    \
+        (fmt_arg[]){{fmt_int}, FMT_MAP(FMT_MAKE_ARG, ##__VA_ARGS__)} + 1, \
```
Ferdi265, line 195: Is the extra argument (the `{fmt_int}` before the `FMT_MAP()`) a workaround for MSVC?

The extra argument doesn't appear to be needed on GCC/Clang and (at least according to a quick check with godbolt) appears to cause worse codegen since the extra argument has to be written to the stack as well.

soumik15630m: Yes, exactly! It's a workaround for the zero-argument case. When `__VA_ARGS__` is empty, `FMT_MAP(...)` expands to nothing, which would leave an empty compound literal `(fmt_arg[]){}` — valid as a GCC/Clang extension (alongside `##__VA_ARGS__`), but a hard error in MSVC and technically non-conforming in standard C. The dummy `{fmt_int}` guarantees at least one element, and `+ 1` skips it. You're right that it costs a stack write on GCC/Clang;

Ferdi265: At least according to [cppreference](https://en.cppreference.com/w/c/language/compound_literal.html), empty compound literals (and `__VA_OPT__`) are part of C23, so at least for compilers that support that the variant without the initial argument could be conditionally used instead as an optimization.

We'd still need the workaround for MSVC of course.

(EDIT: nevermind, zero-size arrays are still an extension, but I think there might be a trick with `__VA_OPT__` to remove / replace the compund initializer with something else when there are no arguments; I'll experiment with that a bit)

Ferdi265: I found an ugly trick that could remove the need for the extra argument, even for MSVC. It uses the same trick used in `FMT_NARGS(...)` to implement something akin to a fake `__VA_OPT__` : `FMT_VA_SELECT` selects between two cases for empty (`e`) and non-empty (`n`) argument lists. This allows us to replace the problematic empty `(fmt_arg[]){}` with `NULL`.

```c
#  define FMT_VA_SELECT(e, n, ...)                                            \
    FMT_NARG_(dummy, ##__VA_ARGS__, n, n, n, n, n, n, n, n, n, n, n, n, n, n, \
              n, n, e)

#  define fmt_format(buffer, size, fmt, ...)                                   \
    fmt_vformat((buffer), (size), (fmt),                                       \
                FMT_VA_SELECT(                                                 \
                    NULL, ((fmt_arg[]){FMT_MAP(FMT_MAKE_ARG, ##__VA_ARGS__)}), \
                    ##__VA_ARGS__),                                            \
                FMT_NARG(__VA_ARGS__))
```

This expands to

```c
fmt_vformat((buffer), (size), (fmt), NULL, 0)
```
if `__VA_ARGS__` is empty and to
```c
fmt_vformat((buffer), (size), (fmt), ((fmt_arg[]){FMT_MAP(FMT_MAKE_ARG, ##__VA_ARGS__)}), FMT_NARG(__VA_ARGS__))
```

otherwise

EDIT: removed the need for `FMT_UNPACK()`/`FMT_EXPAND()`  
EDIT2: [godbolt link](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1AB9U8lJL6yAngGVG6AMKpaAVxYMJ0hwBk8DTAA5dwAjTGIQACZpAAdUBUJbBmc3Dy9SOISbAT8A4JYwiOiLTCtshiECJmICFPdPKRKypMrqglyg0PCo6QUqmrq0xr62jvzCnoBKC1RXYmR2DgB6JYBqADESFiYCGyNV2jwQ4mqAT1X%2BYlXHAFJIgCE7%2B9WAWlWCBExr1YBBZQBJG4aH4rIEgtbOGKnYh4YAIAirCDISarSIaLiRV6rGLETBKQSkVYANTw1hIqwAWgA3cKoKmkhBgla/Wi0VYwuEEBTsvHhGnoAB0TKWwo25I%2BX0OCwYSlW/ku23KPKo4XeqAuWx2AoQQuBYLuAGY8FQGFgqBsALIAFWMjmMAAljPrIgazf4vutrbaHU69XqXf5kG4sKsbgbHH10CFUC5tWHsKtVszoy5nUaGEHXCGwxGCOgzXGDQnE8yEgAvTDGAj6/1Gqhm1ZmGJuBTN1wKMGYVQEcIMUORSK3AehgDsj1rDmNNZ%2BBFOMUwDcY7lH45%2BiaoLAIxn8CLDABFVlxSGD15vjK4d8fgaetym2fvViOr2uLmfkAhqs%2Bb8YqLRUDsv1fLd0FmEJ6EAjctz/IxjBA1wwMwCC3z6GEjCQrc4h3VUHykfURwPSCqznTAw1XMFZ3nBsUNcawVxPICiPnd5iNI%2BjXAYJI6OvRMDgEYBeP2HdjCpMRXBIg1Vx41Z2ISYAAnQAT%2BOg/iL0EYTRPE54kzWRQFw1K5PRtZQrQAJVWQwFOQdsiBYVZ0B2Jh6MTO9VjvdS3E0pzVnfapvI/Yh3LE1juPXP8dguMKtxEjzgpfRM4IQuzQPoQLPJCxSkvg%2BhFNg5LK2ioKJK8tAZQRHziAAKm8lD/GAWKpJKvpVipVA8HQKrMMEcJ6tHA8CrSn4bnwhjqjqoq/R%2BJdbKG55CPCYhyQfF4jwY%2BaSG3Bhora4xRtDA0DxeTEhr3WKwR3BiqQVHYkX8qqQlcKgVTIVZy0rBFXsJRqytuoDPoEJrCNGqrRoUQCpPB8HXqrVYGHcHbiGABRJlO2tjQbJsWzbDs9XwtNJyoadDWNU1MHNDH2yx6cRhsZA5QYQ4AhGhGGKoBaWA2ggIGUjLVBRGb6NxAg5j7agz1G5GxwFCivgfQjL1WAV%2BoFIT%2Br2g9VGO07cYm6nSTphmvkB5nCNZ1B2dUzmZNheSMu53muJfQXhcRI3gAl%2B4peItWGItwlFY0gULdS72Ndx8bBu1wbgV12n/ANpn%2BJNtnjDvCAXPt/nuKd4gRdd93PaY2WzzvP2lbc1WH1Dk7w81nWqhp/X3QTlnk/Km7fIzscBcwIWc5dsWEfz6XvcI8rS4D8rg8rzWa8jsEY8bxnXZbs2f0i6hItWTvJJ5Xvc4Ht2ZoLmX9pZyLx48gVf3/KKNJDmeyLn6P671uOm%2BXpPV4S%2BgIG/r5t%2B7nvfuW5xZH2HkXYCeUL5iQFH/Kep8q5axOnXHYr96bvwPivdmylcpZUwFzPimVEoAKzj3Z2osQGDzAV7CBxgcF/2gZgAU9C8rwPVg/PCyCo4/AXm/JemDP7sxQkif631qhVRIY7MhfcKHw0PpLcBp9R41TQgrJWyAVH8WnmHR%2BXD54v1jug/hlDE5nlNuzGIBBiAiNKs1Vq7Ut58y7qQoBsjQEKJoUos8nUezPX9pfHx2EEEcJxno4ELU2o%2BxlK4GIcQagLkYvgiJ6BkY11rImYm6MLRCDtESbAJlyLESotbPS5UGKyXksYcqp0XSJlKEoQplFSarCmnRYahEKkJOqTXWpLTTRTmBMyEymAACOF5cSrBYBgXsqxslEkcKsAA7oQBA3kBBXUENiXEOJUALAUAoEgaYMnmjdPJCAxhsm5PyXzSIAA2O4tzERgDAKchc5y5l2lMj8Pc/wrT/AAPKBB%2BD4JxtwRwLIuUIeZxgvk/L%2BYC4FyN0k8TWlcO4g5fgAh5GMvAuJuQfNROiAAnE8JZKykwUmQCAHEmAdl7IOVcX8TBgBCgHEcvp%2BACYTUNPFUmTdDIXJ%2BAAaWwMYH4JkADiEB7YQ1lVJG4ABWW46VjASsYOEUkUBeaEjlbquVirlVxXBh0kp6Aqn%2BRAFgjmOqeIGq8lJK2clSkWqtb7CGdr0pSQUAgEgBBLWCOtXq8GHqjUOplKal6Pqaj%2BrMcnN18qlX2p4juGNW5zGBqDcGxNnqeKOptim118t3XZtDTxZSqafzJyLZmxMIa5V5r0uWwtBJi2GrldzJtAbq36pLfW8NTqFIdr4hW9Nbq62yrvCO5OJca21t7bK8qU7V5j1naGedENr47CXezTdBAbV6vHRDP%2B27cEIX3bKw94NubHqtSwvBhJL0NVuielC56D3rvBl9Py4iX1WLfbaj9UlkkVRPZY56s7H08S/cB0Df7dWQd5VQJgrhaB%2BqiQoGJcSexmulpMaVSLuE8rsnyxmArHA/CtBAJghIQgojIxR4wVGaMEaI68y0NpyM2iY65FEjkXSGhCITXpbGBVAslYx7c6BCTGBWsYSI0mDTSekI2BV0nbnSafI2AAHNJol0muAaH07JjED7AOrpk4pxsjQrOqas%2Bp1YgQ/YCgFCiVdNbL2BHZSJr0YmpXOZc25wLQWs1tp4qJ8VErGPoHcCwU4D7%2BMulMESH4YrJVCFMISLg9muC2es1wSzJnDwrQM4SPTqwdOPkJPZ2zCHZXKcs/Jw8hINAseRd5m0Fo/jGA0BAfzLHhMkY9F6TrygZPUEJEwFEVAqP9YNIh/lw2uuRHG%2BZZjFwZuEmm7Rrzg32OCtGwaFb1HXKfSmxt9btHNtIlm/N0ji3RuSCOzRz6hIUnrcm1dy763kRXZSTtmgd2OtdYVU9k73lXuEkwGdj7F3pjfbh9NlJV2of/YW0D0btzQchBe3ZSHhI3bvYR2nInP31tI/W1Dq7h82u7YFSN4wI4sc46ky0/HhIEB8w/dNmHW2SdE/J9Nyn623ZXY56jwH%2B3jBaaZ%2BD3HrPVjAHZ4SPAnPQvrnO7zq7pPEdE6F9NkX62OdXZV%2BLob6PjBEpl%2BYOXiEFdK7lISbQquvLc6J196b2vf666Jwb6bRv1sq6u0703e36cGatxD%2BXivVgIGV47wkABrZ36VXefb57973VOif%2B%2Bm4H9bTurtJ5D3TrrXAuAR5t2zmPcfVjaET4SWgyejWp9h1r/nmfhfZ6J3n6bBf1tJ6u434v92ZPLaoBN57suWe2%2Bj7Hh3tf68HEJCwJvUkW%2Ba/hxn5HPuu/G6J336bA/1uN6uyv4f5v8sV%2Bn1XufeB4%2BrATw35fhIGCr54uv4nbet8U536L7v%2B%2Bicj9psT91sV8rtX9z9JcpAr88c7dq95868H8n9JkX9CRUBXN4MucNdP9N8ycO99dd8A8ADC8icQDpswC101djUIBX8rt0DICw8Qdx9Vswdrdr84Db979H8l8UCYY0DCQYgMCQsXdsD3drtv9Bdf9Dd/8g9ADSCicKDatptaD1t0CrtBCGDS9MdmDjtscp9YDZ8a9EDuDaBn8%2BDVhUABDCQRkhDKCRCeccCPd29t8s8/899ZCSDB8FC4clCaCic1D1tBCrsbChM5tiMAczd9sVs%2Bs7D0p6NKMS9RtCRwtJVzljBktUsJV0tjBJg8NmDDRDQksUsItsjWtBoBsIi9sMihBsAfBsBHBKNbcGAnMAtgs2iE0qCUjItf4Ys4t%2BwDRCj0jii0sMtzDmixjUCJipjxiZjJjxjasIZZiWkyjWNdsTZNROZ7pHpwhCQPpfoFZnNbD2jMCqDCJLoNioAtino4cIBXobjII4dji3MFjZUBVqjaj6jKMnjvi50qC3NAg5AfAfBCQoBXZFVHgFU9x3ZEiIAS8RVMj4sBjEshjMjSjjofCzMIMEsDQijUTTBHifjAsXiIYui0iMiSj8SViXR8YSw1hyZWwWxQiaTtI9s7RHRpxkl3g8QCAesyk7oHonpdi8AKxoY7iHY18zELiridiXphTbd0V0VUldEOT7EuS%2Bhy8%2BTXIBSZSoZ3o5SnEd5vwrpNjtTno9iFSu5cYBxCRJBIglTOEVTIkew%2BhltNTpSzS5TRT9TxT39JTiBFQ05TShSKx4tIh%2BYrTGtbSbSDR7TQkOBphaBOAFVeBPAOAtBSBUBOAFkDk5gFh%2BieBSA/U0z4zpgE8QAFUuABQNBbktMDQiURxIgFUDQNAFURwRxVNEyOBJBeAWAJANBDNUz0zMyOBeAFAQBDMiytBpg4BYAkA0AWAYg6AdiKAREFylyIhkBgAct5MaBUNwgxy05NBeAQh/AzhOACyTzmBiBTh/kQhtBMBrBzzeB5y2BBB/l6ZTgjzSAsB7pgBHAxBaAxzuBeAsBtgjBxBizvzcUHybAaQgL0yuwHzXAewnzyAupOz0zDhjgzhnAsAvyrE8BezgLSAaRiBowlA9xMAwLgAGZQBizpgmVEYSRMBFl/l5xUyCz%2BBBARAxB2BGguL5AlA1AvzdAjwDAjAQBTAql9AjgxzIBphUBLEkggKXh/luQXhKJDAaZeA6R5o2o8R4BphLAYKkh7BTRBgGhSBfB3RxhugjxMhEgBALK9AHLygxgugIgjxjLrAWh%2BhagXB6g9BvLyhWgah3KCg7KLA/LnKvK/LwqJguAjLZh5g%2BKEykyUyvzhyt4tNbkXhblJAFdkBaYcsBRMQIBcBCBFoXREreBJz4zSAIA5yzZFz6AyAVz5yWrugirxLgABwStdzfEDyQgvzLyzziLRrrzbz7zHziKXzGACB3zaBPzIKfzXA/yAKgKCzQLDAaLFh0z8BcQfK4KvzEKrIULiKsIMLeAsKThrzcK9raqYQiKCzSLyLMBKLqLaKjyGKDAmK8AWK2LGBUKBKeLxB%2BLZBdJhLILdB5MerJKzAZKQg5KIAFKlL/pOBVL1LNLBA9YXgAB1BgVAF4OSVwF4CsBaF4ESGEJgfAZAF4bYZACm0adweahQHS0imELAZGoy0oEyuwCABwGKqy00eKyK1ypIIW8WnIGyjyoK3mnygQUK/y1ISy4K3y0YGWiKzyqKtoIWkYMKzWhKpK3M1K/QdK0gQcnSzgbK3K/K7yHq1ESIAUSsjQRECqogNFaqyYWq76hqpqtc1q8gSgDq9ckAKkZAYSLgIlOTEcdIoQLLBnYwVQfKvgOgQaygYayCia5ai808yau8mC1Cuat8j8r81a9a1kTakCqinaiC/a6Co6gyyC065CxYAsy6r8m6nC6ZfCp61C16%2BId6mu8C2qX2xihQZi1i9i4G2QUGvi6QASyG9QaGkARTOGqS8wLC7mjMtGmUTgJYf5LgJMSMFAMvClKlGlOlPEBldmvSrmwypoPmzwAW8ygKoYYW9AUW7WqW5IN%2Byyn%2Br%2BuW5oRW6Kv%2BoBp%2BpWwB2K3WsB6Bg2vIWWxKmYE2iQNKjgZMi2zK625OgqlgBQcO5qKO0qkcYkeO25AUEh8q/AD2/M72ws760skASQLTAUIlA0LLDQSIEcSQW5IlW0rhs2rsnsvsgcrBkciwcc%2Bh%2BitByIDKyC4cn26Rki/c0yyQIAA)

PR #4672, opened 2026-02-08
add include_guard (GLOBAL) to allow usage in multiple submodules

torsten48:
added `include_guard (GLOBAL)` to allow usage of this cmake file via add_subdirectory() from multiple submodules


CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 cmake_minimum_required(VERSION 3.8...3.28)
+include_guard (GLOBAL)
```
vitaut: nit: space is not needed before parenthesis

vitaut:
Merged, thank you

PR #4675, opened 2026-02-11
std.h: Allow formatting std::unexpected type.

17steen:
In some cases, people might want to format the std::unexpected type itself, independent of the value type, this commit makes it possible.


vitaut:
Thank you!

PR #4679, opened 2026-02-23
Speed up iterator_buffer

user202729:
Just a 1-line change to reuse the `fmt::detail::copy()` method instead of manually loop over each character. For that, some methods need to be moved around.

Demonstration of the speedup in https://github.com/fmtlib/format-benchmark/issues/35 . Arguably `vector<char>` is a rare type, but it might be useful for other things (custom string types...?)

To make tests pass, I need to  check whether the iterator type is move-assignable (`throwing_iterator` is not assignable), although I'd argue that in practice iterators ought to be move-assignable and that test is actually invalid.

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -1856,6 +1856,65 @@ class fixed_buffer_traits {
   }
 };
 
+template <typename OutputIt, typename InputIt, typename = void>
+struct has_back_insert_iterator_container_append : std::false_type {};
+
+template <typename OutputIt, typename InputIt>
+struct has_back_insert_iterator_container_append<
+    OutputIt, InputIt,
+    void_t<decltype(get_container(std::declval<OutputIt>())
+                        .append(std::declval<InputIt>(),
+                                std::declval<InputIt>()))>> : std::true_type {};
+
+template <typename OutputIt, typename InputIt, typename = void>
+struct has_back_insert_iterator_container_insert_at_end : std::false_type {};
+
+template <typename OutputIt, typename InputIt>
+struct has_back_insert_iterator_container_insert_at_end<
+    OutputIt, InputIt,
+    void_t<decltype(get_container(std::declval<OutputIt>())
+                        .insert(get_container(std::declval<OutputIt>()).end(),
+                                std::declval<InputIt>(),
+                                std::declval<InputIt>()))>> : std::true_type {};
+
+// An optimized version of std::copy with the output value type (T).
+template <typename T, typename InputIt, typename OutputIt,
+          FMT_ENABLE_IF(is_back_insert_iterator<OutputIt>::value&&
+                            has_back_insert_iterator_container_append<
+                                OutputIt, InputIt>::value)>
+FMT_CONSTEXPR auto copy(InputIt begin, InputIt end, OutputIt out) -> OutputIt {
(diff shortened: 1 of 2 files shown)


include/fmt/base.h
```diff
@@ -1873,7 +1932,12 @@ class iterator_buffer : public Traits, public buffer<T> {
...
     const T* begin = data_;
     const T* end = begin + this->limit(size);
-    while (begin != end) *out_++ = *begin++;
+#if defined(__cpp_if_constexpr)
+    if constexpr (std::is_move_assignable<OutputIt>::value)
+      out_ = copy<T>(begin, end, out_);
+    else
+#endif
+      while (begin != end) *out_++ = *begin++;
   }
```
vitaut, line 1941: Let's move the definition to `fmt/format.h` instead of bringing all the `copy` machinery here.

vitaut:
@user202729 do you still plan to update this PR?

user202729:
Updated.

That said, it leads to a bunch of other things such as `format_to` being moved to `format.h`, which is *technically* a breaking change since, for example, previously a program that only use `base.h` would be able to use `iterator_buffer`, now they can't. (See the failing tests changed in the latest commit. In particular, deleting the `error base-test includes format.h` is extraordinarily wrong. One option is to move them to `format-test.cc`.)

I can think of a few possibilities such as providing a faster overload in `format.h`, but that would probably be a ODR violation.

What do you think?

Side note, if it makes you feel better, you could just close the PR and comment like "if you want to work on this later you can reopen it", or make a [stale] label and set your pull requests page to by default filter them out. I appreciate that you are very active and keep the number of pending issues and pull requests extraordinarily low.

include/fmt/base.h
```diff
@@ -1858,136 +1858,6 @@ class fixed_buffer_traits {
...
 };
 
-// A buffer that writes to an output iterator when flushed.
-template <typename OutputIt, typename T, typename Traits = buffer_traits>
-class iterator_buffer : public Traits, public buffer<T> {
```
vitaut, line 1863: I was not suggesting to move `iterator_buffer`, just the implementation of the function you were modifying. Sorry if it wasn't clear enough.

user202729: If you move only the implementation, but not the interface, won't anything that include `base.h` but not `format.h`, get linker error? Besides, there is a header-only compilation option.

You can't just put it in the cpp, since the function is templated (depends on the template parameter `OutputIt`).

vitaut: No, only the consumers who need this will have to include `format.h` which is OK.

user202729:
I followed the suggestion, and sure enough, `base-test` can't link.

What then? Move the tests in `base-test.cc` that requires this method to `format-test`?

vitaut:
I see, the problems is that the definition of `flush` won't get instantiated with the output iterator type. I guess moving `copy` to `base.h` is indeed the simplest option, sorry for going back and forth on it.

user202729:
in retrospect, with https://github.com/fmtlib/fmt/pull/4716 , this wouldn't be too useful. (maybe `std::deque`, but for that one the append method is named `append_range`.) but anyway...?

vitaut:
Merged, thanks!

PR #4680, opened 2026-02-23
Mark static arrays as constexpr for consistency

Skylion007:
I realized from my a PR a few months ago.
1. These static arrays could be marked as constexpr
2. Doing so makes should decrease compile time and ~allow the linker to better de-duplicate across translation units (reducing binary size)~ (Modern compilers are pretty good, so they were able to auto-constexpr it anyway). No size difference on LLVM, approximately noise size different on gcc using LD linker).
3. It should be compatible with c++11, which is the minimum we target.
4. Allows sizeof to be constexpr when called on them, which will improve some constant folding in the library


Skylion007:
@vituat, good catch it doesn't for GCC/LLVM. Still good from a semantic point of view and matches the other static arrays in the file.

vitaut:
I guess it won't hurt and might even be beneficial in some build configurations. Merged.

PR #4681, opened 2026-02-23
wrap macro arguments in parentheses for robustness

berndpetrovitsch:

### DIFF
--- a/include/fmt/fmt-c.h
+++ b/include/fmt/fmt-c.h
@@ -170,7 +170,7 @@ typedef enum {} fmt_signed_char;
 
 #  define fmt_format(buffer, size, fmt, ...)                              \
     fmt_vformat(                                                          \
-        buffer, size, fmt,                                                \
+        (buffer), (size), (fmt),                                          \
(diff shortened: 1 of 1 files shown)


berndpetrovitsch:
Minor cleanup-style improvement - if only to avoid obscure errors/error messages in probably very seldom cases .-..

user202729:
The only case a macro expansion could contain a comma not protected by parentheses and not fail at the macro invocation is if someone do for example

```
#define COMMA ,
fmt_format(x COMMA y, z, t, ............)
```

sounds like one of these "the user gets the consequence of what they're doing" case. On the other hand one can imagine a macro provided by a third-party library that expands to some unparenthesized comma-expression—but in this case, the bug is arguably in that third-party library.

Maybe you can provide an example of such "very seldom" cases?

vitaut:
I think the change makes sense even if the examples that break are contrived.

Skylion007:
FYI: @vitaut there is a clang-tidy bugprone check you can enable to statically enforce this.

PR #4685, opened 2026-02-24
Add Separate CMake Target for C++20 Modules

### LINKED ISSUE #4684: No support for Installing C++20 module Target separate from Regular compiled Target
FMT_MODULE flag creates two different versions of the same CMake Target `fmt::fmt`

CMake, since version 3.28 supports installation of C++ Modules. 

It is indeed possible to install and use the module version of the library by setting `FMT_MODULE=ON` before installing.

For widespread support for C++ Modules, we likely need a new CMake Target  specific for modules, which can be installed alongside the regular `fmt::fmt` target and the `fmt::fmt-header-only` targets.

MathewBensonCode:
In the same vein as there is the `fmt::fmt-header-only`, `fmt::fmt` and `fmt::fmt_c` targets, I propose the addition of a new target `fmt::fmt-module` which will be for the compilation of the FMT_MODULE library option.

The new target will have the properties requried for Compiling, Installing and using the C++20 functionality

The `add_module_library` CMake function is factored out in favour of simply specifying the targets specifically in a cleaner and simpler CMake syntax.

The FMT_MODULE option is defaulted to true for versions of CMake newer than 3.28.

Closes #4684

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.8...4.3)
+cmake_minimum_required(VERSION 3.8...3.28)
 include_guard(GLOBAL)
 
 # Fallback for using newer policies on CMake <3.12.
@@ -42,22 +42,16 @@ function(enable_module target)
   endif ()
 endfunction()
 
-set(FMT_USE_CMAKE_MODULES FALSE)
-if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.28 AND
-    CMAKE_GENERATOR STREQUAL "Ninja")
-  set(FMT_USE_CMAKE_MODULES TRUE)
-endif ()
-
 # Adds a library compiled with C++20 module support.
 # `enabled` is a CMake variables that specifies if modules are enabled.
 # If modules are disabled `add_module_library` falls back to creating a
 # non-modular library.
 #
 # Usage:
-#   add_module_library(<name> [sources...] FALLBACK [sources...] [IF_MODULE enabled]
-# IF_MODERN enabled)
+#   add_module_library(<name> [sources...] FALLBACK [sources...] [IF_MODULE enabled] 
+#   [USE_CMAKE_MODULES true])
 function(add_module_library name)
-  cmake_parse_arguments(AML "" "IF_MODULE;IF_MODERN" "FALLBACK" ${ARGN})
+  cmake_parse_arguments(AML "" "IF_MODULE;USE_CMAKE_MODULES" "FALLBACK" ${ARGN})
   set(sources ${AML_UNPARSED_ARGUMENTS})
 
   add_library(${name})
@@ -73,7 +67,7 @@ function(add_module_library name)
   # Modules require C++20.
   target_compile_features(${name} PUBLIC cxx_std_20)
 
-  if (${AML_IF_MODERN})
+  if (${AML_USE_CMAKE_MODULES})
     target_sources(${name} PUBLIC FILE_SET fmt_module TYPE CXX_MODULES
                    FILES ${sources})
   else ()
@@ -154,26 +148,30 @@ if (FMT_MASTER_PROJECT AND NOT CMAKE_BUILD_TYPE)
 endif ()
 
 project(FMT CXX)
+
 # Determine Support for the C++ Module Scanning Features
-# Requires CMake>= 3.28 plus (Ninja >= 1.11 OR Visual Studio >=17.4)
+# Requires C++20, CMake>=3.28 and (Ninja >= 1.11 OR Visual Studio >=17.4)
 # The project() CMake Function sets several variables including those
 # needed for Compiler Versions
 set(FMT_USE_CMAKE_MODULES FALSE)
-if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.28)
+if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.28 
+    AND CMAKE_CXX_STANDARD GREATER_EQUAL 20)
   # Check Version of Ninja to determine CXX_MODULES support
   if (CMAKE_GENERATOR STREQUAL "Ninja")
-    execute_process(COMMAND "${CMAKE_MAKE_PROGRAM}" "--version" OUTPUT_VARIABLE MY_NINJA_VERSION)
-    message(STATUS "Ninja Version: ${MY_NINJA_VERSION}")
-    if (MY_NINJA_VERSION VERSION_GREATER_EQUAL 1.11)
+    execute_process(COMMAND "${CMAKE_MAKE_PROGRAM}" "--version" 
+      OUTPUT_VARIABLE NINJA_VERSION)
+    message(STATUS "Ninja Version: ${NINJA_VERSION}")
+    if (NINJA_VERSION VERSION_GREATER_EQUAL 1.11)
       set(FMT_USE_CMAKE_MODULES TRUE)
       message(STATUS "Using CXX Modules by Default with Ninja")
     endif ()
-  elseif (CMAKE_GENERATOR MATCHES "^Visual Studio"
-      AND MSVC_VERSION GREATER_EQUAL 1934)
-      set(FMT_USE_CMAKE_MODULES TRUE)
-      message(STATUS "Using CXX Modules by Default with Visual Studio")
+  elseif (CMAKE_GENERATOR MATCHES "^Visual Studio" 
+    AND MSVC_VERSION GREATER_EQUAL 1934)
+    set(FMT_USE_CMAKE_MODULES TRUE)
+    message(STATUS "Using CXX Modules by Default with Visual Studio")
   endif ()
 endif ()
+
 include(GNUInstallDirs)
 set_verbose(FMT_INC_DIR ${CMAKE_INSTALL_INCLUDEDIR} CACHE STRING
             "Installation directory for include files, a relative path that "
@@ -190,7 +188,7 @@ option(FMT_TEST "Generate the test target." ${FMT_MASTER_PROJECT})
 option(FMT_FUZZ "Generate the fuzz target." OFF)
 option(FMT_CUDA_TEST "Generate the cuda-test target." OFF)
 option(FMT_OS "Include OS-specific APIs." ON)
-option(FMT_MODULE "Build a module library in addition to the traditional library." ${CMAKE_CXX_SCAN_FOR_MODULES})
+option(FMT_MODULE "Build a module library in addition to the traditional library." ${FMT_USE_CMAKE_MODULES})
 option(FMT_SYSTEM_HEADERS "Expose headers with marking them as system." OFF)
 option(FMT_UNICODE "Enable Unicode support." ON)
 
@@ -321,9 +319,9 @@ add_headers(FMT_HEADERS args.h base.h chrono.h color.h compile.h core.h format.h
                         xchar.h)
 set(FMT_SOURCES src/format.cc)
 
-# Add Regular library By setting IF_MODULE=FALSE
-add_module_library(fmt src/fmt.cc
-    FALLBACK ${FMT_SOURCES} # TODO(CK): WHY? ${FMT_HEADERS} README.md ChangeLog.md
+# add regular library by setting IF_MODULE=FALSE
+add_module_library(fmt src/fmt.cc 
+                    FALLBACK ${FMT_SOURCES} ${FMT_HEADERS} README.md ChangeLog.md
                     IF_MODULE FALSE)
 add_library(fmt::fmt ALIAS fmt)
 
@@ -381,8 +379,8 @@ endif ()
 
 if (FMT_MODULE)
   add_module_library(fmt-module "src/fmt.cc"
-    IF_MODULE ${FMT_MODULE}
-    IF_MODERN ${FMT_USE_CMAKE_MODULES})
+    IF_MODULE TRUE
+    USE_CMAKE_MODULES ${FMT_USE_CMAKE_MODULES})
 
   add_library(fmt::fmt-module ALIAS fmt-module)
   enable_module(fmt-module)
@@ -394,37 +392,9 @@ if (FMT_MODULE)
     target_compile_options(fmt-module PRIVATE ${PEDANTIC_COMPILE_FLAGS})
   endif ()
 
-  if (FMT_IMPORT_STD)
-    target_compile_features(fmt-module PUBLIC cxx_std_23)
-    set_target_properties(fmt-module PROPERTIES CXX_MODULE_STD ON)
-    target_compile_definitions(fmt-module PUBLIC FMT_IMPORT_STD)
-    message(STATUS "Use FMT_IMPORT_STD=${FMT_IMPORT_STD}")
-  elseif (cxx_std_20 IN_LIST CMAKE_CXX_COMPILE_FEATURES)
-    target_compile_features(fmt-module PUBLIC cxx_std_20)
-  else ()
-    message(WARNING "Feature cxx_std_20 is unknown for the CXX compiler")
-  endif ()
-
   if (FMT_USE_CMAKE_MODULES)
-      target_sources(fmt-module PUBLIC FILE_SET fmt_module_headers TYPE HEADERS
-        BASE_DIRS include
-          FILES
-            include/fmt/ostream.h
-            include/fmt/format-inl.h
-            include/fmt/ranges.h
-            include/fmt/xchar.h
-            include/fmt/core.h
-            include/fmt/chrono.h
-            include/fmt/os.h
-            include/fmt/color.h
-            include/fmt/args.h
-            include/fmt/printf.h
-            include/fmt/compile.h
-            include/fmt/format.h
-            include/fmt/fmt-c.h
-            include/fmt/std.h
-            include/fmt/base.h
-      )
+    target_sources(fmt-module PRIVATE FILE_SET fmt_module_headers TYPE HEADERS
+      BASE_DIRS ${PROJECT_SOURCE_DIR}/include FILES ${FMT_HEADERS})
   else ()
     target_include_directories(fmt-module ${FMT_SYSTEM_HEADERS_ATTRIBUTE} BEFORE PUBLIC
       $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
@@ -434,8 +404,10 @@ if (FMT_MODULE)
   set(FMT_DEBUG_POSTFIX d CACHE STRING "Debug library postfix.")
 
   set_target_properties(fmt-module PROPERTIES
-    VERSION ${FMT_VERSION} SOVERSION ${CPACK_PACKAGE_VERSION_MAJOR}
-    DEBUG_POSTFIX "${FMT_DEBUG_POSTFIX}")
+    VERSION ${FMT_VERSION} 
+    SOVERSION ${CPACK_PACKAGE_VERSION_MAJOR}
+    DEBUG_POSTFIX "${FMT_DEBUG_POSTFIX}"
+    )
 endif ()
 
 add_library(fmt-header-only INTERFACE)
@@ -516,20 +488,22 @@ if (FMT_INSTALL)
   endif ()
 
   set(INSTALL_FILE_SET)
-  if (${CMAKE_VERSION} VERSION_GREATER 3.22)
+  if (${CMAKE_VERSION} VERSION_GREATER "3.22")
       list(APPEND INSTALL_FILE_SET FILE_SET fmt DESTINATION "${FMT_INC_DIR}/fmt")
-      # FIXME: missing? list(APPEND INSTALL_FILE_SET FILE_SET fmt_header_only DESTINATION "${FMT_INC_DIR}/fmt")
+      list(APPEND INSTALL_FILE_SET FILE_SET fmt_header_only DESTINATION "${FMT_INC_DIR}/fmt")
   endif ()
   if (FMT_MODULE AND FMT_USE_CMAKE_MODULES)
       list(APPEND INSTALL_FILE_SET FILE_SET fmt_module DESTINATION "${FMT_INC_DIR}/fmt")
-      list(APPEND INSTALL_FILE_SET FILE_SET fmt_module_headers)
   endif ()
 
   # Install the library and headers.
   install(TARGETS ${INSTALL_TARGETS}
           COMPONENT fmt_core
           EXPORT ${targets_export_name}
+          LIBRARY DESTINATION ${FMT_LIB_DIR}
+          ARCHIVE DESTINATION ${FMT_LIB_DIR}
           PUBLIC_HEADER DESTINATION "${FMT_INC_DIR}/fmt"
+          RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
           ${INSTALL_FILE_SET})
 
   # Use a namespace because CMake provides better diagnostics for namespaced
--- a/doc/get-started.md
+++ b/doc/get-started.md
@@ -8,8 +8,10 @@ with CMake, while the [Build Systems](#build-systems) section covers the rest.
 
 ## CMake
 
-{fmt} provides three CMake targets: `fmt::fmt` for the standard compiled library, `fmt::fmt-module` for the C++ modules library and `fmt::fmt-header-only` for the header-only library. It is recommended to use
-the compiled library or the module library for improved build times. 
+{fmt} provides three CMake targets: `fmt::fmt` for the standard compiled library,
+`fmt::fmt-module` for the C++ modules library and `fmt::fmt-header-only` for the 
+header-only library. It is recommended to use the compiled library or the module 
+library for improved build times. 
 
 There are three primary ways to use {fmt} with CMake:
 
@@ -41,7 +43,9 @@ There are three primary ways to use {fmt} with CMake:
 
 ### Alternative Targets
 
-In order to use the header-only target or the module target, simply substitute the `fmt::fmt` in the above steps with `fmt::fmt-header-only` or `fmt::fmt-module` accordingly.
+In order to use the header-only target or the module target, simply substitute the 
+`fmt::fmt` in the above steps with `fmt::fmt-header-only` or `fmt::fmt-module` 
+accordingly.
 
 ## Installation
 
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -81,25 +81,29 @@ target_compile_definitions(enforce-checks-test PRIVATE
 add_executable(perf-sanity perf-sanity.cc)
 target_link_libraries(perf-sanity fmt::fmt)
 
-# if (FMT_MODULE)
-#   # The tests need {fmt} to be compiled as traditional library
-#   # because of visibility of implementation details.
-#   # If module support is present the module tests require a
-#   # test-only module to be built from {fmt}
-#   add_library(test-module OBJECT ${CMAKE_SOURCE_DIR}/src/fmt.cc)
-#   target_compile_features(test-module PUBLIC cxx_std_11)
-#   target_include_directories(test-module PUBLIC
-#     $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>)
-#   enable_module(test-module)
-#
-#   add_fmt_test(module-test MODULE test-main.cc)
-#   if (MSVC)
-#     target_compile_options(test-module PRIVATE /utf-8 /Zc:__cplusplus
-#                            /Zc:externConstexpr /Zc:inline)
-#     target_compile_options(module-test PRIVATE /utf-8 /Zc:__cplusplus
-#                            /Zc:externConstexpr /Zc:inline)
-#   endif ()
-# endif ()
+if (FMT_MODULE)
+  # The module-test.cc needs some work and is not working yet. 
+  # For now We simply just return 
+  # so that the other tests are not affected.
+  return()
+  # The tests need {fmt} to be compiled as traditional library
+  # because of visibility of implementation details.
+  # If module support is present the module tests require a
+  # test-only module to be built from {fmt}
+  #add_library(test-module OBJECT ${CMAKE_SOURCE_DIR}/src/fmt.cc)
+  #target_compile_features(test-module PUBLIC cxx_std_11)
+  #target_include_directories(test-module PUBLIC
+  #  $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>)
+  #enable_module(test-module)
+
+  #add_fmt_test(module-test MODULE test-main.cc)
+  #if (MSVC)
+  #  target_compile_options(test-module PRIVATE /utf-8 /Zc:__cplusplus
+  #                         /Zc:externConstexpr /Zc:inline)
+  #  target_compile_options(module-test PRIVATE /utf-8 /Zc:__cplusplus
+  #                         /Zc:externConstexpr /Zc:inline)
+  #endif ()
+endif ()
 
 if (NOT DEFINED MSVC_STATIC_RUNTIME AND MSVC)
   foreach (flag_var


doc/api.md
```diff
@@ -20,6 +20,10 @@ The {fmt} library API consists of the following components:
 All functions and types provided by the library reside in namespace `fmt`
 and macros have prefix `FMT_`.
 
+## New C++ Modules API
```
vitaut: I don't think we need "New" here.

doc/api.md
```diff
@@ -20,6 +20,10 @@ The {fmt} library API consists of the following components:
...
+
+With the new C++ Modules API, all the headers listed above do not need to be explicitly #included. We can instead use the `import fmt;` statement instead. All other functionality, listed below remains the same.
```
vitaut: Same here and modules shouldn't be capitalized. Also please break the long line throughout.

doc/get-started.md
```diff
@@ -40,6 +39,10 @@ There are three primary ways to use {fmt} with CMake:
         add_subdirectory(fmt)
         target_link_libraries(<your-target> fmt::fmt)
 
+### Alternative Targets
+
+In order to use the Header Only Target or the Module Target, simply substitute the `fmt::fmt` in the above steps with `fmt::fmt-header-only` or `fmt::fmt-module` accordingly.
```
vitaut: Header Only Target -> header-only target
Module Target -> module target

CMakeLists.txt
```diff
@@ -169,7 +84,7 @@ option(FMT_TEST "Generate the test target." ${FMT_MASTER_PROJECT})
 option(FMT_FUZZ "Generate the fuzz target." OFF)
 option(FMT_CUDA_TEST "Generate the cuda-test target." OFF)
 option(FMT_OS "Include OS-specific APIs." ON)
-option(FMT_MODULE "Build a module instead of a traditional library." OFF)
+option(FMT_MODULE "Build a module instead of a traditional library." ${FMT_USE_CMAKE_MODULES})
```
vitaut: "instead" should probably be change to "in addition to" now.

CMakeLists.txt
```diff
@@ -27,102 +27,17 @@ function(join result_var)
   set(${result_var} "${result}" PARENT_SCOPE)
 endfunction()
 
-# DEPRECATED! Should be merged into add_module_library.
-function(enable_module target)
```
vitaut, line 31: Why is the block below removed?

MathewBensonCode: I'm glad you agree with the need for a separate target. I will try and break the changes down to separate pull requests as requested.

On the matter of the `add_module_library`, would you be comfortable with leaving it in, but calling the function twice, once for the ordinary library and once again for the module library?

I had tried this and found the structure a little strange as the function would be poorly named for this purpose, which is why I left it out.

We could rename it to something like `create_compiled_target` and use it to reduce the lines of code by placing in every repetitive aspect of creating the targets.

Kindly let me know your thoughts on the above.

MathewBensonCode: On the matter of the `enable_module` function, it was moved here:
https://github.com/fmtlib/fmt/blob/8c91cc2de25321f967a0254abd6ec12b08a14b3a/CMakeLists.txt#L309-L320

CMakeLists.txt
```diff
@@ -362,27 +280,161 @@ if (FMT_SAFE_DURATION_CAST)
...
+                # Use an absolute path to prevent target_link_libraries prepending -l.
+                set(sources ${sources} ${pcm} ${CMAKE_CURRENT_BINARY_DIR}/${obj})
+                add_custom_command(
+                    OUTPUT ${obj}
+                    COMMAND ${CMAKE_CXX_COMPILER} $<TARGET_PROPERTY:fmt-module,COMPILE_OPTIONS>
+                    -c -o ${obj} ${pcm}
+                    DEPENDS ${pcm})
+            endforeach ()
+        endif ()
+    endif()
```
ClausKlein: This block should be deleted, clang 22 is out and works as v21 fine

CMakeLists.txt
```diff
@@ -362,27 +280,161 @@ if (FMT_SAFE_DURATION_CAST)
...
+    add_library(fmt-module)
+    add_library(fmt::fmt-module ALIAS fmt-module)
+    # Modules require C++20.
+    target_compile_features(fmt-module PUBLIC cxx_std_20)
+
+    # Modern CMake Method
+    if (FMT_USE_CMAKE_MODULES)
+        target_sources(fmt-module 
+            PUBLIC FILE_SET fmt_module TYPE CXX_MODULES FILES src/fmt.cc
+            PRIVATE FILE_SET module_headers TYPE HEADERS BASE_DIRS ${PROJECT_DIR}/include)
```
ClausKlein: FILES ... are missing

CMakeLists.txt
vitaut: > On the matter of the add_module_library, would you be comfortable with leaving it in, but calling the function twice, once for the ordinary library and once again for the module library?

Why not just use `add_library` for the non-modular one?

CMakeLists.txt
```diff
@@ -432,19 +484,24 @@ if (FMT_INSTALL)
...
 
   # Install the library and headers.
   install(TARGETS ${INSTALL_TARGETS}
           COMPONENT fmt_core
           EXPORT ${targets_export_name}
-          LIBRARY DESTINATION ${FMT_LIB_DIR}
-          ARCHIVE DESTINATION ${FMT_LIB_DIR}
           PUBLIC_HEADER DESTINATION "${FMT_INC_DIR}/fmt"
-          RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
           ${INSTALL_FILE_SET})
```
ClausKlein, line 508: Installation is one part, but you should verify, that the installed package is complete and may be used!

see i.e. https://github.com/ClausKlein/fmt-module/blob/develop/.TODO.txt

CMakeLists.txt
MathewBensonCode: FILES in Header File_Sets are optional.. The `module_headers` file_set is private, I put this in for the event where the FMT_OS flag is set, and the os.cc file needs to know where to find the headers for the fmt library.

I don't know if the FMT_OS functionality is required for the fmt-module target, but I left it in an attempt to replicate the current functionality.

CMakeLists.txt
vitaut: We shouldn't require clang 22.

CMakeLists.txt
MathewBensonCode: I have only done some simple testing on my machine, where I installed the package and was able to use the fmt-header-only, fmt-module and fmt targets from an independent project.

I don't expect this to be conclusive, and I hope for more feedback and testing by anyone who is able to.

CMakeLists.txt
MathewBensonCode: As I understand it, the Version of CMake is the determinant for this path as opposed to the clang version. Most recent versions of CMake, from 3.28 should simply use the path in https://github.com/fmtlib/fmt/blob/8c91cc2de25321f967a0254abd6ec12b08a14b3a/CMakeLists.txt#L298-L302 and let CMake figure it all out.

The alternative to the above block, https://github.com/fmtlib/fmt/blob/8c91cc2de25321f967a0254abd6ec12b08a14b3a/CMakeLists.txt#L304-L365 which includes the block you have quoted, exists for situations where an older version of CMake is being used and doen't have this kind of logic baked in.

CMakeLists.txt
MathewBensonCode: This is actually what I had attempted to do. The regular library target is created here followed by its properties being set

https://github.com/fmtlib/fmt/blob/8c91cc2de25321f967a0254abd6ec12b08a14b3a/CMakeLists.txt#L224-L288

the `fmt-module` is optionally created based on the FMT_MODULE flag and its properties set just as the regular one here https://github.com/fmtlib/fmt/blob/8c91cc2de25321f967a0254abd6ec12b08a14b3a/CMakeLists.txt#L291-L413 which includes the section that was formerly in the add_module_library function.

There is a bit of code duplication on the CMake Code that depends on flags like FMT_OS and FMT_UNICODE between the two versions which can possibly be set up in a function later.

vitaut: But you removed `enable_module` which is not desirable because it is actually vendored from elsewhere and shouldn't be removed.

MathewBensonCode: > On the matter of the `enable_module` function, it was moved here:
> https://github.com/fmtlib/fmt/blob/8c91cc2de25321f967a0254abd6ec12b08a14b3a/CMakeLists.txt#L309-L320

I noted that it was marked as deprecated and was only being used in one place so I moved it and removed the indirection.

As you have indicated that this is vendored and it is not desirable to remove it, would you be happy with the indirection(function call) being reintroduced?

CMakeLists.txt
ClausKlein: **AFAIK** with this flag set and `import std;` does not build?
Every option has to be tested on CI, but the tests are disabled if `FMT_MODULE` is set!

ClausKlein: **A note by the way:** don't change the `cmake format style` if possible. It is not clean yet, but mostly indented by 2 spaces and space after controls: if, else, endif, ...

MathewBensonCode: thank you for the feedback and guidance. I will look into the formatting and do some more testing on the "tests"

CMakeLists.txt
vitaut: Deprecation comment only applies to `enable_module`, not `add_module_library`.

MathewBensonCode: yes, I saw that it was to be merged with the `add_module_library`. 

I was clarifying that the `enable_module` wasn't removed, but re-located.

I have tried to place the code in the same logical location that it was in and remove the need for the deprecated code. 

The `add_module_library` is actually at the heart of this pull request as it was a branching point determining the nature of the fmt::fmt library.

Creating a separate fmt::fmt-module library simply moves the branching logic out of the method, rendering the purpose of the function to be redundant(in my opinion).

Hence, my comment earlier asking about a re-name of the function, if really necessary, and to use the function in its original form but call it once with the IF value as `true` and again as `false`.

I realize that this is a breaking change and requires a lot of care when making drastic changes, especially because this library is used in many places.

As I have done a bit more testing with this, I think that having the separate module target would even enable tests to be written for the module version separate from the traditional library target.

It will also enable people to start experimenting with `import fmt;` within installed bases, but still have the original experience without having to recompile.

So I do hope to work out a way to get this done.

MathewBensonCode:
I have tried to make the requested changes. I have restored the Removed functions to the CMakeLists.txt file.

However, I marked the `add_module_library` as deprecated and unused.

Kindly advise if the change is agreeable.

test/CMakeLists.txt
```diff
@@ -2,48 +2,48 @@ add_subdirectory(gtest)
...
-target_include_directories(test-main PUBLIC
-  $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>)
-target_link_libraries(test-main gtest fmt)
+target_include_directories(
+  test-main PUBLIC $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>)
+target_link_libraries(test-main gtest fmt::fmt)
 
-# Adds a test.
-# Usage: add_fmt_test(name srcs...)
+# Adds a test. Usage: add_fmt_test(name srcs...)
```
vitaut: Please revert whitespace and unrelated changes.

CMakeLists.txt
```diff
@@ -55,6 +51,10 @@ endif ()
 #
 # Usage:
 #   add_module_library(<name> [sources...] FALLBACK [sources...] [IF enabled])
+
+# DEPRECATED! Not currently used as we create both the regular and modular
+# library with add_library below as opposed to selecting between the two as
+# was previously done with this function
```
vitaut: We should continue to use this function for the modular path. If any changes are necessary, they should be applied to this function, not duplicated elsewhere and ideally kept to a minimum.

MathewBensonCode: Ok. So we can have the regular library added outside of this function but move the module to this function. Thanks for the guidance.

MathewBensonCode: Do you oppose changing the function signature to remove the "IF" and "FALLBACK" clauses since we are only going to use the function for module library creation?

CMakeLists.txt
```diff
@@ -169,14 +190,10 @@ option(FMT_TEST "Generate the test target." ${FMT_MASTER_PROJECT})
...
 option(FMT_OS "Include OS-specific APIs." ON)
-option(FMT_MODULE "Build a module instead of a traditional library." OFF)
+option(FMT_MODULE "Build a module library in addition to the traditional library." ${FMT_USE_CMAKE_MODULES})
 option(FMT_SYSTEM_HEADERS "Expose headers with marking them as system." OFF)
 option(FMT_UNICODE "Enable Unicode support." ON)
 
-if (FMT_TEST AND FMT_MODULE)
-  # The tests require {fmt} to be compiled as traditional library
-  message(STATUS "Testing is incompatible with build mode 'module'.")
-endif ()
```
ClausKlein, line 179: **AFAIK**: the tests are still disabled if `FMT_MODULE` is set!

CMakeLists.txt
```diff
@@ -362,6 +379,43 @@ if (FMT_SAFE_DURATION_CAST)
...
+  endif ()
+  if (FMT_PEDANTIC)
+    target_compile_options(fmt-module PRIVATE ${PEDANTIC_COMPILE_FLAGS})
+  endif ()
+
+  if (cxx_std_11 IN_LIST CMAKE_CXX_COMPILE_FEATURES)
+    target_compile_features(fmt-module PUBLIC cxx_std_11)
+  else ()
+    message(WARNING "Feature cxx_std_11 is unknown for the CXX compiler")
+  endif ()
```
ClausKlein: `cxx_std_20` or newer must be set for CXX_MODULES!

CMakeLists.txt
```diff
@@ -362,6 +379,43 @@ if (FMT_SAFE_DURATION_CAST)
...
+
+  if(FMT_USE_CMAKE_MODULES)
+    target_sources(fmt-module PRIVATE FILE_SET fmt_module_headers TYPE HEADERS 
+      BASE_DIRS ${PROJECT_SOURCE_DIR}/include)
+  else()
+    target_include_directories(fmt-module ${FMT_SYSTEM_HEADERS_ATTRIBUTE} BEFORE PUBLIC
+      $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
+      $<INSTALL_INTERFACE:${FMT_INC_DIR}>)
+  endif()
```
ClausKlein: Why use a PRIVATE FILE_SET without files?

And what is the else part for?

In case of `import std;` you need an own header set with public definitions!

And with FILE_SET, you may use [VERIFY_INTERFACE_HEADER_SETS](https://cmake.org/cmake/help/v4.3/prop_tgt/VERIFY_INTERFACE_HEADER_SETS.html#verify-interface-header-sets)

CMakeLists.txt
```diff
@@ -432,19 +489,25 @@ if (FMT_INSTALL)
...
   set(INSTALL_FILE_SET)
-  if (FMT_USE_CMAKE_MODULES)
-    set(INSTALL_FILE_SET FILE_SET fmt DESTINATION "${FMT_INC_DIR}/fmt")
+  if(${CMAKE_VERSION} VERSION_GREATER "3.22")
+      list(APPEND INSTALL_FILE_SET FILE_SET fmt DESTINATION "${FMT_INC_DIR}/fmt")
+      list(APPEND INSTALL_FILE_SET FILE_SET fmt_header_only DESTINATION "${FMT_INC_DIR}/fmt")
+  endif()
+  if (FMT_MODULE AND FMT_USE_CMAKE_MODULES)
+      list(APPEND INSTALL_FILE_SET FILE_SET fmt_module DESTINATION "${FMT_INC_DIR}/fmt")
+      list(APPEND INSTALL_FILE_SET FILE_SET fmt_module_headers DESTINATION "${FMT_INC_DIR}/fmt")
```
ClausKlein: you install an empty `PRIVATE FILE_SET` here?

ClausKlein:
```bash
bash-5.3$ cmake -G Ninja -S . -B build --fresh -D CMAKE_CXX_STANDARD=17
-- CMake version: 4.2.1
-- The CXX compiler identification is AppleClang 16.0.0.16000026
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Ninja Version: 1.13.0.git.kitware.jobserver-pipe-1

-- Using CXX Modules by Default with Ninja                           <<<<<<<<<<<<<< he?
-- {fmt} version: 12.1.1
-- Build type: Release
-- Performing Test HAS_NULLPTR_WARNING - Success
-- Target 'doc' disabled because mkdocs not found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- FMT_PEDANTIC: OFF
-- The C compiler identification is AppleClang 16.0.0.16000026
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Configuring done (2.7s)
CMake Error in CMakeLists.txt:
  The target named "fmt-module" has C++ sources that may use modules, but the
  compiler does not provide a way to discover the import graph dependencies.
  See the cmake-cxxmodules(7) manual for details.  Use the
  CMAKE_CXX_SCAN_FOR_MODULES variable to enable or disable scanning.


-- Generating done (0.0s)
CMake Generate step failed.  Build files cannot be regenerated correctly.
bash-5.3$ 
```

ClausKlein:
you may have a look at https://github.com/ClausKlein/fmt/pull/1

CMakeLists.txt
MathewBensonCode: Before these lines were relevant because both the module and non-module versions of the library used the same CMake Target Name which would affect the testing.

However, by having the module version have its own identity, the current test framework will only affect the non-module version.

So the FMT_MODULE flag does not affect the target used in testing.

CMakeLists.txt
MathewBensonCode: The HEADER FILE_SET does not have to have FILES specified. It is a newer version of the `target_include_directories` command. It just indicates where headers will be searched for. 

The else clause is for the older c++20 module support, i.e. before the newer CXX_SCAN_FOR_MODULES error which we are indicating with the FMT_USE_CMAKE_MODULES.

MathewBensonCode: I think that the module target should not expose the header set to the consumer, so VERIFY_INTERFACE_HEADER_SETS will not be relevant here.

However, it would be useful to create a HEADER FILE_SET for the regular fmt::fmt target and specify that there. But that would probably have to be a separate pull request and issue.

MathewBensonCode:
> you may have a look at [ClausKlein#1](https://github.com/ClausKlein/fmt/pull/1)

@ClausKlein I have just seen this and noted the changes needed in the `test/CMakeLists.txt` file with regards to the FMT_MODULE flag. Thanks.

ClausKlein:
> > you may have a look at [ClausKlein#1](https://github.com/ClausKlein/fmt/pull/1)
> 
> @ClausKlein I have just seen this and noted the changes needed in the `test/CMakeLists.txt` file with regards to the FMT_MODULE flag. Thanks.

I was wondering, I could not open a PR against your repo?

MathewBensonCode:
> > > you may have a look at [ClausKlein#1](https://github.com/ClausKlein/fmt/pull/1)
> > 
> > 
> > @ClausKlein I have just seen this and noted the changes needed in the `test/CMakeLists.txt` file with regards to the FMT_MODULE flag. Thanks.
> 
> I was wondering, I could not open a PR against your repo?

I don't know why this didn't work, but I have incorporated the commit in the workflow. 

the tests for FMT_MODULE are still excluded, but the ordinary tests should now work. This can be solved in another pull request and issue.

doc/get-started.md
```diff
@@ -8,9 +8,8 @@ with CMake, while the [Build Systems](#build-systems) section covers the rest.
 
 ## CMake
 
-{fmt} provides two CMake targets: `fmt::fmt` for the compiled library and
-`fmt::fmt-header-only` for the header-only library. It is recommended to use
-the compiled library for improved build times.
+{fmt} provides three CMake targets: `fmt::fmt` for the standard compiled library, `fmt::fmt-module` for the C++ modules library and `fmt::fmt-header-only` for the header-only library. It is recommended to use
```
vitaut: Break the long line for consistency and readability.

doc/get-started.md
```diff
@@ -40,6 +39,10 @@ There are three primary ways to use {fmt} with CMake:
         add_subdirectory(fmt)
         target_link_libraries(<your-target> fmt::fmt)
 
+### Alternative Targets
+
+In order to use the header-only target or the module target, simply substitute the `fmt::fmt` in the above steps with `fmt::fmt-header-only` or `fmt::fmt-module` accordingly.
```
vitaut: Same here

test/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ set(TEST_MAIN_SRC test-main.cc gtest-extra.cc gtest-extra.h util.cc)
 add_library(test-main STATIC ${TEST_MAIN_SRC})
 target_include_directories(test-main PUBLIC
   $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>)
-target_link_libraries(test-main gtest fmt)
+target_link_libraries(test-main gtest fmt::fmt)
```
vitaut: Why change to "qualified" name?

test/CMakeLists.txt
```diff
@@ -85,25 +80,10 @@ target_compile_definitions(enforce-checks-test PRIVATE
...
-  if (MSVC)
-    target_compile_options(test-module PRIVATE /utf-8 /Zc:__cplusplus
-                           /Zc:externConstexpr /Zc:inline)
-    target_compile_options(module-test PRIVATE /utf-8 /Zc:__cplusplus
-                           /Zc:externConstexpr /Zc:inline)
-  endif ()
-endif ()
+ if (FMT_MODULE)
+   #  TODO: Fix Tests for modules
+   #   add_fmt_test(module-test MODULE test-main.cc)
```
vitaut: This looks like a regression. Can we keep the old test?

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.8...3.28)
+cmake_minimum_required(VERSION 3.8...4.3)
```
vitaut: Why bump the version?

CMakeLists.txt
```diff
@@ -30,39 +30,34 @@ endfunction()
...
-
 # Adds a library compiled with C++20 module support.
 # `enabled` is a CMake variables that specifies if modules are enabled.
 # If modules are disabled `add_module_library` falls back to creating a
 # non-modular library.
 #
 # Usage:
-#   add_module_library(<name> [sources...] FALLBACK [sources...] [IF enabled])
+#   add_module_library(<name> [sources...] FALLBACK [sources...] [IF_MODULE enabled] 
+#   [IF_MODERN enabled])
```
vitaut: Maybe rename `IF_MODERN` to a more descriptive and consistent with variable name `USE_CMAKE_MODULES`?

CMakeLists.txt
```diff
@@ -153,6 +148,29 @@ if (FMT_MASTER_PROJECT AND NOT CMAKE_BUILD_TYPE)
 endif ()
 
 project(FMT CXX)
```
vitaut, line 150: nit: Add an empty line to separate project command from the other stuff.

CMakeLists.txt
```diff
@@ -153,6 +148,29 @@ if (FMT_MASTER_PROJECT AND NOT CMAKE_BUILD_TYPE)
...
+# Requires C++20, CMake>=3.28 and (Ninja >= 1.11 OR Visual Studio >=17.4)
+# The project() CMake Function sets several variables including those
+# needed for Compiler Versions
+set(FMT_USE_CMAKE_MODULES FALSE)
+if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.28 
+    AND CMAKE_CXX_STANDARD GREATER_EQUAL 20)
+  # Check Version of Ninja to determine CXX_MODULES support
+  if (CMAKE_GENERATOR STREQUAL "Ninja")
+    execute_process(COMMAND "${CMAKE_MAKE_PROGRAM}" "--version" 
+      OUTPUT_VARIABLE MY_NINJA_VERSION)
```
vitaut: MY_ suffix is weird, let's drop it.

CMakeLists.txt
```diff
@@ -304,15 +318,15 @@ add_headers(FMT_HEADERS args.h base.h chrono.h color.h compile.h core.h format.h
                         xchar.h)
 set(FMT_SOURCES src/format.cc)
 
-add_module_library(fmt src/fmt.cc FALLBACK
-                   ${FMT_SOURCES} ${FMT_HEADERS} README.md ChangeLog.md
-                   IF FMT_MODULE)
+# Add Regular library By setting IF_MODULE=FALSE
```
vitaut: nit: unnecessary capitalization

CMakeLists.txt
```diff
@@ -362,6 +376,67 @@ if (FMT_SAFE_DURATION_CAST)
   target_compile_definitions(fmt PUBLIC FMT_SAFE_DURATION_CAST)
 endif ()
 
+if (FMT_MODULE)
+  add_module_library(fmt-module "src/fmt.cc"
+    IF_MODULE ${FMT_MODULE}
```
vitaut: `${FMT_MODULE} ` can be replaced with `TRUE`.

CMakeLists.txt
```diff
@@ -362,6 +376,67 @@ if (FMT_SAFE_DURATION_CAST)
...
+  add_module_library(fmt-module "src/fmt.cc"
+    IF_MODULE ${FMT_MODULE} 
+    IF_MODERN ${FMT_USE_CMAKE_MODULES})
+
+  add_library(fmt::fmt-module ALIAS fmt-module)
+  enable_module(fmt-module)
+
+  if (FMT_WERROR)
+    target_compile_options(fmt-module PRIVATE ${WERROR_FLAG})
+  endif ()
```
vitaut, line 390: I don't think we need this.

CMakeLists.txt
```diff
@@ -362,6 +376,67 @@ if (FMT_SAFE_DURATION_CAST)
...
+  if (FMT_IMPORT_STD)
+    target_compile_features(fmt-module PUBLIC cxx_std_23)
+    set_target_properties(fmt-module PROPERTIES CXX_MODULE_STD ON)
+    target_compile_definitions(fmt-module PUBLIC FMT_IMPORT_STD)
+    message(STATUS "Use FMT_IMPORT_STD=${FMT_IMPORT_STD}")
+  elseif (cxx_std_20 IN_LIST CMAKE_CXX_COMPILE_FEATURES)
+    target_compile_features(fmt-module PUBLIC cxx_std_20)
+  else ()
+    message(WARNING "Feature cxx_std_20 is unknown for the CXX compiler")
+  endif ()
```
vitaut: This looks unrelated to the current PR and better be split.

CMakeLists.txt
```diff
@@ -432,19 +510,24 @@ if (FMT_INSTALL)
...
 
+  if (FMT_MODULE)
+    list(APPEND INSTALL_TARGETS fmt-module)
+  endif ()
+
   set(INSTALL_FILE_SET)
-  if (FMT_USE_CMAKE_MODULES)
-    set(INSTALL_FILE_SET FILE_SET fmt DESTINATION "${FMT_INC_DIR}/fmt")
-  endif()
+  if (${CMAKE_VERSION} VERSION_GREATER "3.22")
```
vitaut, line 491: Why check cmake version?

CMakeLists.txt
```diff
@@ -432,19 +510,24 @@ if (FMT_INSTALL)
...
+      list(APPEND INSTALL_FILE_SET FILE_SET fmt DESTINATION "${FMT_INC_DIR}/fmt")
+      list(APPEND INSTALL_FILE_SET FILE_SET fmt_header_only DESTINATION "${FMT_INC_DIR}/fmt")
```
vitaut, line 493: Why do we need both fmt and fmt_header_only here?

CMakeLists.txt
```diff
@@ -432,19 +510,24 @@ if (FMT_INSTALL)
...
+  endif ()
 
   # Install the library and headers.
   install(TARGETS ${INSTALL_TARGETS}
           COMPONENT fmt_core
           EXPORT ${targets_export_name}
-          LIBRARY DESTINATION ${FMT_LIB_DIR}
-          ARCHIVE DESTINATION ${FMT_LIB_DIR}
           PUBLIC_HEADER DESTINATION "${FMT_INC_DIR}/fmt"
-          RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
```
vitaut, line 447: Why are there removed?

test/CMakeLists.txt
MathewBensonCode: CMake "Best Practices" Convention is to use the unqualified target name within the same CMakeLists.txt file(or directory), but to use the fully qualified name in different directories. 

Beyond "Best Practice" it also prevents bugs and error that come from a behavior in the `target_link_libraries` command of just adding the target name to the link arguments if it doesn't find a target with the name.

vitaut: Sure but I would suggest not doing such unrelated changes in this PR.

CMakeLists.txt
MathewBensonCode: @ClausKlein any response to this?

CMakeLists.txt
MathewBensonCode: Because FILE_SETs were introduced in version 3.23. Some newer versions of CMake raise Errors when FILE_SETs aren't used, and older code in some cases don't know about them.

This is why I have been using a pattern of checking for version support, and either using the older way and the newer way in a separate branch.

So for example for headers, if Header File_Sets are supported we use those, but if not the we use the target_include_directories with the BuildDir and installDir generator expressions.

test/CMakeLists.txt
MathewBensonCode: This is currently consistent with the "old" behavior that was gated with the FMT_MODULE flag.

Previously there was a return statement that would ignore the tests and exit if the FMT_MODULE was set.

The non-module tests now run fine whether this flag is set or not.

The module test needs some work, which can now be dealt with separately.

CMakeLists.txt
MathewBensonCode: Alright. I had left it in to be consistent with the previous behaviour if both FMT_MODULE and FMT_WERROR were set.

CMakeLists.txt
MathewBensonCode: When we declare the File_Set in the `target_sources` command for the `fmt::fmt` and the fmt::fmt-header-only` targets above we need to specify these File_Sets and where they are to be installed.

CMake actually Errors Out if they are not added to the `install(TARGET)` command.

CMakeLists.txt
MathewBensonCode: They are redundant as they resolve to the defaults.

CMakeLists.txt
ClausKlein: The current `cmake v4.3-rc1 `with official support for `import std;`

vitaut: Let's not try incorporate `import std` changes in this PR.

CMakeLists.txt
ClausKlein: The `FILE_SET` are only used if `FMT_MODULE` is set, and modules supports starts with `cmake v3.28 `first!

test/CMakeLists.txt
MathewBensonCode: I will restore it to what it looked like before, but still comment it out and just return() out of the block as we were doing here: https://github.com/fmtlib/fmt/blob/master/test/CMakeLists.txt#L44-L46.

Because we have changed the nature of FMT_MODULE, we only need to block the Module tests(until they are fixed) and let the rest of the tests just work.

CMakeLists.txt
MathewBensonCode: I will revert this, to reduce the changes in this commit.

CMakeLists.txt
MathewBensonCode: There are Two types of FILE_SETs, HEADER and CXX_MODULES. See doc for the [target_sources](https://cmake.org/cmake/help/latest/command/target_sources.html) command.

The HEADER File_Sets were introduced in 3.23, and the CXX_MODULES File_Sets came in v3.28 as you have pointed out. 

So when dealing with Header File_Sets, we are checking for version 3.23 to decide whether to introduce the Headers via File_Set(`target_source` command) or just the older `target_include_directories` command for older versions

When dealing with Modules, we gate it at version 3.28. Which is what we are doing with the FMT_USE_CMAKE_MODULES flag.

test/CMakeLists.txt
MathewBensonCode: alright. will revert

CMakeLists.txt
MathewBensonCode: I think if we remove this, then the meaning will change. 

Previously this was set here https://github.com/fmtlib/fmt/commit/fca04455651600dd4dbf4fc6d97cea103d1ae909/CMakeLists.txt#L319-L321

Meaning that whether FMT_MODULE was set or not. 

Now that we have two separate targets, to maintain the same behavior we need to set it on both the `fmt` target and the `fmt-module` target.

if `FMT_MODULE` isn't set then these lines simply won't be run.

Therefore, I propose leaving this in, in keeping with the spirit of not changing things unnecessarily.

ClausKlein:
## The CMakeFiles are still unreadable, the control flow is not clear, and the supported version range historic!

@MathewBensonCode I Ask to respect the cmake format style if possible. It is not clean yet, but mostly indented by 2 spaces and space after controls: if, else, endif, ...

### This Files should be formatted with `gersemi`; then the formatting battle may end!

@vitaut Check your **CI**, most often on `GitHub` the `most resent cmake version is available`, Today the range is `CMake version: 3.31.6...4.2.3`.
### Since 3.28 there are lot of fixes and improvements for CXX_MODULE handling in CMake

And starting with CMake v4.0, a lot of policies changed, you should notice this.
On CI no newer compiler and no newer OS runner is used.

# The world changes, don't stay back!

MathewBensonCode:
> ## The CMakeFiles are still unreadable, the control flow is not clear, and the supported version range historic!
> @MathewBensonCode I Ask to respect the cmake format style if possible. It is not clean yet, but mostly indented by 2 spaces and space after controls: if, else, endif, ...
> 
> ### This Files should be formatted with `gersemi`; then the formatting battle may end!
> @vitaut Check your **CI**, most often on `GitHub` the `most resent cmake version is available`, Today the range is `CMake version: 3.31.6...4.2.3`.
> 
9> ### Since 3.28 there are lot of fixes and improvements for CXX_MODULE handling in CMake
> And starting with CMake v4.0, a lot of policies changed, you should notice this. On CI no newer compiler and no newer OS runner is used.
> 
> # The world changes, don't stay back!

@ClausKlein, thank you for the time you have taken to provide your reviews and comments. I will try and ensure that the code I write has better formatting. It is not a battle, it is usually a lower priority as I want to focus on correctness more.

On this matter of the version range, I believe that you are wrong. kindly refer to the documentation for the [cmake_minimum_required](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html) for the use of the maximum policy.

We don't just write the latest version in the command and call it a day. It is meant to be a guarantee of the code we have written. This means we would have to write all the CMake code in this repo to that standard, test it everywhere and only after all that make that statement about the policies that our code supports. 

That is a big undertaking that will require many other issues and pull requests.

The scope and goal of this pull request is simply to add a "modern" target but also to ensure that it doesn't break any expected behavior. 

This repo is used in many varied environments, modern and legacy, so a lot of care is needed when making changes. 

I have had to revert several good and modern ideas in this pull request to keep to the scope. I intend to bring those in separate issues and pull requests later. 

So I appreciate and understand your desire to see modernization, I am pursuing the same goals and I'm certain @vitaut is too.

vitaut:
Merged, thank you!

vitaut:
@ClausKlein please stop spamming the repo or you will be blocked. Improvements to CMake configs are fine but they should be split into independent PRs that tackle one thing at a time.

PR #4687, opened 2026-02-27
Moved named_arg to public API

### LINKED ISSUE #4683: Feature request: Provide public interface to destruct a `"name"_a=value` argument
As in the title.

Reason: currently I have an asynchronous logger. If one call the macro

```
LOG("{a} {b}", "a"_a=x, "b"_a=y);
```

I'd like to be able to
1. compute `x1 = TransformOnCallingThread(x); y1 = TransformOnCallingThread(y);`
2. send the format string, `x1` and `y1` to the logging thread with an inter-thread queue
3. call `fmt::format("{a} {b}", "a"_a=x1, "b"_a=y1)`

Currently, as far as I can see, (1) requires reaching into the `fmt::detail` namespace.

TPPPP72:
I have moved `named_arg` to the public API and fixed the affected files. Local tests have passed.
Close #4683

### DIFF
--- a/include/fmt/args.h
+++ b/include/fmt/args.h
@@ -113,8 +113,7 @@ FMT_EXPORT template <typename Context> class dynamic_format_arg_store {
     data_.emplace_back(arg);
   }
 
-  template <typename T>
-  void emplace_arg(const detail::named_arg<T, char_type>& arg) {
+  template <typename T> void emplace_arg(const named_arg<T, char_type>& arg) {
     if (named_info_.empty())
       data_.insert(data_.begin(), basic_format_arg<Context>(nullptr, 0));
     data_.emplace_back(detail::unwrap(arg.value));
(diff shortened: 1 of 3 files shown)


TPPPP72:
Hi, I looked into the linux / build (g++-13) failure and checked the ppa:ubuntu-toolchain-r/test package history.
gcc-13 was updated 5 hours ago (version 13.4.0-6ubuntu1~22~ppa2), and it appears to have introduced a dependency conflict (libcc1-0) that is not compatible with the current Ubuntu 22.04 runner image.
Could you please take a look at the CI environment? @vitaut

include/fmt/args.h
```diff
@@ -114,7 +114,7 @@ FMT_EXPORT template <typename Context> class dynamic_format_arg_store {
   }
 
   template <typename T>
-  void emplace_arg(const detail::named_arg<T, char_type>& arg) {
+  void emplace_arg(const fmt::named_arg<T, char_type>& arg) {
```
vitaut: Here and elsewhere: `fmt::` is not needed since we are in the fmt namespace already.

include/fmt/base.h
```diff
@@ -1030,20 +1037,11 @@ struct is_view : std::false_type {};
...
 template <typename T, typename Char>
-struct is_named_arg<named_arg<T, Char>> : std::true_type {};
-
-template <typename T, typename Char = char> struct named_arg : view {
-  const Char* name;
-  const T& value;
-
-  named_arg(const Char* n, const T& v) : name(n), value(v) {}
-  static_assert(!is_named_arg<T>::value, "nested named arguments");
-};
```
vitaut, line 1046: Let's keep `named_arg` here and introduce a public alias in the `fmt` namespace to prevent any ABI impact, however unlikely.

TPPPP72:
> Thanks for the PR!
> 
> Mostly looks good but let's do it via a public alias per my comment inline.
> 
> (BTW the CI conflict was resolved.)

Okay, I have removed the modifications to fmt::detail.

vitaut:
Merged, thanks!

PR #4694, opened 2026-03-02
Fix fmt-c.h not being installed with fmt-c library

### LINKED ISSUE #4693: fmt-c.h header is not installed
Currently the `fmt-c.h` header will not be installed when running `cmake --install <BUILDDIR>`.

This means that the C API currently can not be used from an installed/packaged version of fmt, even if installing latest git master.

The reason is that the `fmt-c` CMake target is missing the `PUBLIC_HEADER` property.

Ferdi265:
This PR fixes the C API header not being installed on `cmake --install <BUILDDIR>` by setting the `PUBLIC_HEADER` target property on `fmt-c`.

This also adds the other properties set on the `fmt` target except for the VS2017 workaround since I don't know what needs to change for the C library. Feel free to add it or tell me what needs to change.

Fixes #4693.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -391,6 +391,9 @@ endif ()
 target_link_libraries(fmt-c PUBLIC fmt::fmt)
 add_library(fmt::fmt-c ALIAS fmt-c)
 
+set_target_properties(fmt-c PROPERTIES
+  PUBLIC_HEADER include/fmt/fmt-c.h)
+
 # Install targets.
 if (FMT_INSTALL)
   include(CMakePackageConfigHelpers)


CMakeLists.txt
```diff
@@ -391,6 +391,14 @@ endif ()
 target_link_libraries(fmt-c PUBLIC fmt::fmt)
 add_library(fmt::fmt-c ALIAS fmt-c)
 
+set(FMT_C_HEADERS)
+add_headers(FMT_C_HEADERS fmt-c.h)
```
vitaut: I don't think we need this, just inline `fmt-c.h` below.

Ferdi265: you're right, it's just one file, that makes sense.

Ferdi265: done

CMakeLists.txt
```diff
@@ -442,6 +391,11 @@ endif ()
 target_link_libraries(fmt-c PUBLIC fmt::fmt)
 add_library(fmt::fmt-c ALIAS fmt-c)
 
+set_target_properties(fmt-c PROPERTIES
+  VERSION ${FMT_VERSION} SOVERSION ${CPACK_PACKAGE_VERSION_MAJOR}
+  PUBLIC_HEADER include/fmt/fmt-c.h
+  DEBUG_POSTFIX "${FMT_DEBUG_POSTFIX}")
```
vitaut: This should only set `PUBLIC_HEADER`. The other properties are controlled elsewhere.

vitaut (changes requested):

CMakeLists.txt
Ferdi265: done

vitaut:
Merged, thanks!

Issue #4518, opened 2025-08-25
Optimize the default FP formatting


vitaut:
The default FP formatting can be slightly optimized, possibly using some ideas from `jkj::dragonbox::detail::to_chars`: https://github.com/fmtlib/fmt/issues/3675#issuecomment-3172953795.

vitaut:
Will use https://github.com/vitaut/zmij which is now feature-complete.

PR #4700, opened 2026-03-05
Optimize fast-float exponential presentation path

Neomelt:
## Summary
- Optimize the default fast-float exponential presentation path by avoiding the extra write-and-swap step in the pointer fast path.
- Reuse write_significand directly in that path to reduce redundant memory operations.

## Rationale
Issue: #4518

This change targets the post-Dragonbox presentation stage (not Dragonbox core generation), which is where #4518 suggests optimization opportunities.

## Validation
- format-test: pass
- format-impl-test: pass
- quick perf sanity check shows positive direction locally (non-rigorous sanity benchmark):
  - before: ~355.1 ms
  - after:  ~332.5 ms

## Notes
- Small, behavior-preserving change.
- No formatting semantics intentionally changed.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -2528,6 +2528,26 @@ inline auto write_exponent_significand(Char* out, UInt significand,
   return out;
 }
 
+template <typename Char, typename UInt,
+          FMT_ENABLE_IF(std::is_integral<UInt>::value)>
+inline auto write_default_fp(Char* out, UInt significand, int significand_size,
+                             int decimal_exponent, sign s) -> Char* {
+  if (s != sign::none) *out++ = Char('-');
+  int exp = decimal_exponent + significand_size;
+  if (decimal_exponent >= 0) {
+    out = write_significand<Char>(out, significand, significand_size);
+    return detail::fill_n(out, decimal_exponent, Char('0'));
+  }
+  if (exp > 0) {
+    return write_significand(out, significand, significand_size, exp,
+                             Char('.'));
+  }
+  *out++ = Char('0');
+  *out++ = Char('.');
+  out = detail::fill_n(out, -exp, Char('0'));
+  return write_significand<Char>(out, significand, significand_size);
+}
+
 template <typename OutputIt, typename Char>
 FMT_CONSTEXPR auto write_significand(OutputIt out, const char* significand,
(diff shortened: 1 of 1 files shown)


Neomelt:
I explored a second follow-up optimization in the same area and ran additional local microbenchmarks, but the gain was not stable across data patterns, so I intentionally did not include that change in this PR.

Keeping this PR focused on the small behavior-preserving improvement only.

If maintainers prefer a specific benchmark setup for #4518 (e.g. format-benchmark scenarios), I can run and post those numbers as a follow-up.

Neomelt:
Thanks for the suggestion — I ran dtoa-benchmark locally comparing base (f1f9223^) vs this PR (f1f9223) on the same machine.

Environment:
- CPU: Ryzen 7 H 260 (Radeon 780M)
- OS: Linux
- Compiler: GCC 14.3

Result (fmt, RandomDigit, digits 1..17 mean):
- Run 1: base 22.4145 ns, PR 22.7194 ns, delta +1.360% (slower)
- Run 2: base 23.2146 ns, PR 23.5357 ns, delta +1.383% (slower)

So this change appears to regress dtoa-benchmark slightly (~+1.37% on my setup).
Given this, would you prefer I close this PR, or continue with a different optimization direction and repost benchmark-proven improvements only?

Neomelt:
I took another pass at this and checked it with `fmtlib/dtoa-benchmark` on Ryzen 7 H 260 / Linux / GCC 14.3.

For `fmt` / `RandomDigit` (mean of digits 1..17, 10 trials), I got:
- base: `23.602888 ns`
- PR: `23.955345 ns` (`+1.493%` vs base)
- exponent fast-path fix: `23.479225 ns` (`-0.524%` vs base, `-1.988%` vs PR)
- current default-presentation fix: `23.051613 ns` (`-2.336%` vs base, `-3.773%` vs PR)

So the earlier regression is gone, and this also moves the default FP presentation path forward instead of just getting back to baseline.

The current change keeps the specialized exponential fast path, and adds a pointer fast path for the default fixed presentation in the fast-float writer.

vitaut:
Thanks for the PR but I don't think we need to change this because the whole implementation will be revamped and replaced with https://github.com/vitaut/zmij.

PR #4701, opened 2026-03-05
Add support for building and testing fmt on Windows ARM64

MugundanMCW:
**PR Description:**

- The PR adds CI support for building and testing fmt on Windows ARM64

**Changes Proposed:**

- Added Windows ARM64 build configurations in the workflow file[workflows/windows.yml].


vitaut:
Thanks for the PR but I don't think these configurations will bring a lot of value as we don't normally rely on arch-specific Windows capabilities.

PR #4702, opened 2026-03-05
Re-Enable fmt module tests, update module-test.cc

### LINKED ISSUE #4698: Enable module test
The test for the `fmt-module` target needs to be fixed and reenabled.

MathewBensonCode:
- Refactor test/CMakeLists.txt to enable testing for modules
- The tests in `module-test.cc` seem to not have been updated in some time despite changes in the main library.
- Wide String versions of several tests appear to be deprecated so have been commented out.
- Refactored tests related to `fmt::format_args` that now requires lvalue references as opposed to direct values.
- Tests for items that have since been removed from the library like `fmt::is_char` and `fmt::localtime` as far as I can tell
- testing for the fmt::detail namespace. The MSVC github workflow could not differentiate the between the fmt::detail and fmt::v12::detail. Removed windows testing for this
   
Should close #4698


MathewBensonCode:
@vitaut, what is the scope of testing that we require for testing the module target?

Do we just want to test the public API or the full API?

The tests for the regular library involve tests for the entire library as they are able to reach into the entire code base. 

With the module based library, it brings up the issue of visible and invisible APIs.

In this pull request, Ideally I have just remove the "inner" code(and some outdated code), which is being tested by the regular library anyway, and just test that the expected public functionality works. 

I have commented out the bits that would be deleted if we we're to take this approach.

The other approach would be to work out how to work with modules and macro based testing frameworks. 

I'm doing some experiments with this, but it may need another pull request as the changes are likely to be drastic, like moving tests into headers, or making the tests part of the module, etc.

vitaut:
We definitely don't need to test everything. A simple sanity test that exercises a small subset of the API should be sufficient. I would recommend splitting the changes into smaller PRs, e.g. focusing on one compiler (e.g. clang) first.

vitaut:
That said, we shouldn't delete code from the module test if it is broken with modules. On the contrary - we should keep it as a regression test for the fix. It is OK to conditionally compile it until the fix is available.

MathewBensonCode:
After further investigation, I have identified that the wide string versions of most tests in `module-test.cc` were not compiling due to the functionality being deprecated in the main library. 

I have just commented them out with explanations in the comments, however, these may be candidates for removal as I observed that they we're not present in the other `*-test.cc` files. 

Compiler Specific testing does not appear to be an issue in this case. The module-test.cc file was well written to only deal with module tests so likely just needed to be updated.

module-test run visible here:

https://github.com/fmtlib/fmt/actions/runs/22850169158/job/66276473337?pr=4702#step:7:1404

vitaut (changes requested):
Please apply cmake-format and revert unrelated changes.

test/CMakeLists.txt
```diff
@@ -19,9 +19,9 @@ function (add_fmt_test name)
       set(PEDANTIC_COMPILE_FLAGS ${PEDANTIC_COMPILE_FLAGS} -Wno-weak-vtables)
     endif ()
   elseif (ADD_FMT_TEST_MODULE)
-    set(libs test-main)
+    set(libs test-main fmt-module)
   else ()
-    set(libs test-main fmt)
+    set(libs test-main)
```
vitaut: Why is fmt removed here?

test/CMakeLists.txt
```diff
@@ -81,9 +81,11 @@ add_executable(perf-sanity perf-sanity.cc)
...
 
 if (FMT_MODULE)
-  # Module tests are currently disabled.
-  # add_fmt_test(module-test MODULE test-main.cc)
-  # target_link_libraries(module-test fmt-module)
+  add_fmt_test(module-test MODULE)
+  if (MSVC)
+    target_compile_options(module-test PRIVATE /utf-8 /Zc:__cplusplus
+                                               /Zc:externConstexpr /Zc:inline)
+  endif ()
```
vitaut: `/utf-8` should be inherited from the fmt target and shouldn't be added here. What are the other options for?

test/module-test.cc
```diff
@@ -60,8 +61,11 @@ TEST(module_test, namespace) {
   ASSERT_TRUE(true);
 }
 
-namespace detail {
-bool oops_detail_namespace_is_visible;
+#ifndef _WIN32 //windows msvc has issue with nested namespace
```
vitaut: Let's disable modules on MSVC for now instead of adding these hacks.

test/module-test.cc
```diff
@@ -170,7 +176,8 @@ TEST(module_test, format_args) {
   auto no_args = fmt::format_args();
   EXPECT_FALSE(no_args.get(1));
 
-  fmt::basic_format_args args = fmt::make_format_args(42);
+  int n = 42;
+  fmt::basic_format_args<fmt::format_context> args = fmt::make_format_args(n);
```
vitaut, line 151: Why is this changed?

test/CMakeLists.txt
MathewBensonCode: fmt is removed here because it is already included in the test-main target

test/module-test.cc
MathewBensonCode: When using `fmt::make_format_args(42)` as it was before, there was an error stating that a suitable function couldn't be found. 

Upon investigation I found that the tests we're written to use an lvalue, for example https://github.com/fmtlib/fmt/blob/4968433a6abecbbf7e72e97188926586a7fd0662/test/xchar-test.cc#L332-L333.

So I used the same pattern everywhere that the `fmt::make_format_args()` was used in the module-test.cc and the issue went away.

It appears that there was a change and since the module-test.cc wasn't being actively used it fell behind

test/module-test.cc
MathewBensonCode: I had taken inspiration from here https://github.com/fmtlib/fmt/blob/4968433a6abecbbf7e72e97188926586a7fd0662/test/module-test.cc#L509-L520

I think it is better to isolate the one area with an issue and have the rest of the tests active.

Could there perhaps be a more acceptable way of doing this? I chose to use comments, but we could also use the print message stating that the test is disabled

test/CMakeLists.txt
vitaut: I don't think we should rely on transitive dependencies here.

test/gtest-extra.cc
```diff
@@ -5,6 +5,11 @@
 //
 // For the license information refer to format.h.
 
+#ifdef FMT_MODULE_TEST
+  import fmt;
+#define FMT_OS_H_ //skip loading of fmt/os.h
```
vitaut: What is going on here?

test/module-test.cc
```diff
@@ -25,6 +25,7 @@
 #include <string>
 #include <string_view>
 #include <system_error>
+#include <gmock/gmock.h> //include this in global module so it isn't included in gtest-extra.h
```
vitaut: Let's not use gtest-extra in this test and avoid these hacks.

test/module-test.cc
vitaut: We could do it later. In this PR I suggest just making one configuration work, e.g. clang.

test/gtest-extra.cc
MathewBensonCode: accidental inclusion from attempt to add tests within the module. will revert

test/module-test.cc
MathewBensonCode: same as above

test/module-test.cc
MathewBensonCode: I have amended the current pull request to ensure that clang and gcc are working.

However, Upon further investigation, I have come up with an alternative to just using one configuration and take advantage of code that is already in the file.

I have implemented it successfully on a separate test branch. 

The core issue is that I believe Daniela who wrote this works primarily on windows and MSVC. So the `oops_detail_namespace_is_visible` variable needs to be specified like this. 

https://github.com/MathewBensonCode/fmt/blob/342b5fc6e10add6c2e8115a2dc7440ed1bef5484/test/module-test.cc#L70-L72

But GCC and Clang require the same to be specified like 

https://github.com/MathewBensonCode/fmt/blob/342b5fc6e10add6c2e8115a2dc7440ed1bef5484/test/module-test.cc#L64-L67

The other issue is the section of code 

https://github.com/MathewBensonCode/fmt/blob/342b5fc6e10add6c2e8115a2dc7440ed1bef5484/test/module-test.cc#L77-L80 needed to be updated to factor in newer versions of MSVC as the MSVC bug still hasn't been solved.

Doing this solves the problem of working in all the compilers. I may need some guidance on the acceptable way to implement the branch.

Would this alternative be better and/or acceptable to put into this pull request?

The final code would look something like this

https://github.com/MathewBensonCode/fmt/blob/342b5fc6e10add6c2e8115a2dc7440ed1bef5484/test/module-test.cc#L63-L81

MathewBensonCode: I have incorporated the above changes in the current commit. All compilers are now working. I had to bump up the version of GCC being tested to version 15 to make this work though as there was a linking issue with locale libraries in version 14. Version 16 is also fine.

I think this is a better approach as we now have a baseline on all compilers as the issues have been identified.

include/fmt/base.h
```diff
@@ -1872,10 +1872,10 @@ template <typename OutputIt, typename T, typename = void>
...
 
 template <typename OutputIt, typename T>
-struct has_insert<OutputIt, T,
-                  void_t<decltype(get_container(std::declval<OutputIt>())
-                                      .insert({}, std::declval<T>(),
-                                              std::declval<T>()))>>
+struct has_insert<
+    OutputIt, T,
+    void_t<decltype(get_container(std::declval<OutputIt>())
+                        .insert({}, std::declval<T>(), std::declval<T>()))>>
```
vitaut, line 1878: Unrelated change, please revert.

.github/workflows/linux.yml
```diff
@@ -37,10 +37,10 @@ jobs:
             std: 23
             install: sudo apt install g++-13
             shared: -DBUILD_SHARED_LIBS=ON
-          - cxx: g++-14
+          - cxx: g++-15
```
vitaut, line 40: Why do we need gcc 15 specifically?

test/module-test.cc
```diff
@@ -60,15 +63,24 @@ TEST(module_test, namespace) {
   ASSERT_TRUE(true);
 }
 
+#ifndef _MSC_FULL_VER
+namespace fmt {
```
vitaut: Wrapping this in the fmt namespace makes the test useless. What problem are you trying to solve here?

test/module-test.cc
```diff
@@ -60,15 +63,24 @@ TEST(module_test, namespace) {
...
 }
 
+#endif
+
 namespace fmt {
 bool namespace_detail_invisible() {
 #if defined(FMT_HIDE_MODULE_BUGS) && defined(_MSC_FULL_VER) && \
-    _MSC_FULL_VER <= 193700000
-  // bug in msvc up to at least 17.7:
+    _MSC_FULL_VER <= 195100000
```
vitaut: So this tests doesn't work on MSVC and judging from the change above other compilers as well? I think we should just remove it.

test/module-test.cc
```diff
@@ -182,41 +195,48 @@ TEST(module_test, format_args) {
...
 TEST(module_test, dynamic_format_args) {
   fmt::dynamic_format_arg_store<fmt::format_context> dyn_store;
   dyn_store.push_back(fmt::arg("a42", 42));
-  fmt::basic_format_args args = dyn_store;
+
+  fmt::basic_format_args<fmt::format_context> args = dyn_store;
   EXPECT_FALSE(args.get(3));
   EXPECT_TRUE(args.get(fmt::string_view("a42")));
 
+  /* wide format support?
```
vitaut: Why is this commented out?

test/module-test.cc
```diff
@@ -182,41 +195,48 @@ TEST(module_test, format_args) {
...
   fmt::vformat_to(std::back_inserter(s), "{}", store);
   EXPECT_EQ("42", s);
 
-  char buffer[4] = {0};
-  fmt::vformat_to(buffer, "{:}", store);
-  EXPECT_EQ("42", std::string_view(buffer));
+  // * Exlude Test Due to depracation
+  //  char buffer[4] = {0};
+  //  fmt::vformat_to(buffer, "{:}", store);
+  //  EXPECT_EQ("42", std::string_view(buffer));
```
vitaut: Let's remove this.

test/module-test.cc
```diff
@@ -227,16 +247,17 @@ TEST(module_test, vformat_to) {
 }
 
 TEST(module_test, vformat_to_n) {
-  auto store = fmt::make_format_args(12345);
+  int n = 12345;
+  auto store = fmt::make_format_args(n);
   std::string s;
-  auto result = fmt::vformat_to_n(std::back_inserter(s), 1, "{}", store);
+//  auto result = fmt::vformat_to_n(std::back_inserter(s), 1, "{}", store); //unused
```
vitaut: If the problem is that result is unused, then apply the usual `(void)result` trick.

test/module-test.cc
```diff
@@ -246,21 +267,25 @@ std::string as_string(std::wstring_view text) {
           text.size() * sizeof(text[0])};
 }
 
-TEST(module_test, print) {
+TEST(module_testint, print) {
```
vitaut: Stray "int", please remove.

test/module-test.cc
```diff
@@ -314,10 +341,13 @@ TEST(module_test, memory_buffer) {
   nbuffer.clear();
   EXPECT_EQ(0u, to_string(nbuffer).size());
 
-  fmt::wmemory_buffer wbuffer;
-  EXPECT_EQ(0u, to_string(wbuffer).size());
+  // #Excluded due to deleted
+  //  fmt::wmemory_buffer wbuffer;
+  //  EXPECT_EQ(0u, to_string(wbuffer).size());
```
vitaut: Let's remove this.

test/module-test.cc
```diff
@@ -326,21 +356,24 @@ TEST(module_test, is_char) {
   EXPECT_TRUE(fmt::is_char<char32_t>());
   EXPECT_FALSE(fmt::is_char<signed char>());
 }
+*/
```
vitaut: same here

test/module-test.cc
```diff
@@ -433,42 +470,63 @@ TEST(module_test, weekday) {
...
+    EXPECT_EQ("42", fmt::vsprintf("%d", fmt::make_printf_args(n)));
+    EXPECT_EQ(L"42", fmt::vsprintf(L"%d", fmt::make_wprintf_args(n)));
+    */
+  // replaced with version of the same in fprintf-test.cc
+  auto store = fmt::make_format_args<fmt::printf_context>(n);
+  auto args = fmt::basic_format_args<fmt::printf_context>(store);
+  EXPECT_EQ(fmt::vsprintf(fmt::string_view("%d"), args), "42");
+  EXPECT_WRITE(stdout, fmt::vfprintf(stdout, fmt::string_view("%d"), args),
+               "42");
 }
```
vitaut, line 403: Let's remove printf tests, we mostly care about format here.

.github/workflows/linux.yml
MathewBensonCode: I bumped up the version by one as there was a linking issue with locales when using version 14. The module support in version 14 was not very stable. So version 15 and the current 16 work without an issue. So in order to have a baseline for module testing I was proposing to have gcc15 as the starting version.

vitaut: Makes sense. Please address the remaining (minor) comments.

test/module-test.cc
MathewBensonCode: This didn't work and was creating warnings, which would fail the -WError check in the CI. I observed that it wasn't being used in any assertion so I removed it as well as all other code that was commented out as non-functioning

test/module-test.cc
MathewBensonCode: I wasn't sure if you meant just to remove the commented out code or all the tests with printf.

As I understand it, we are testing that the `exported` functions work, which I assume includes the print, vprintf and vsprintf.

I believe the tests are functioning as intended. But you can confirm if you want them completely removed.

test/module-test.cc
MathewBensonCode: removed as requested

test/module-test.cc
MathewBensonCode: the section below was not compiling due to missing supported functions for wide strings. It appears that it was removed at some point.

vitaut:
Merged, thanks!

PR #4703, opened 2026-03-06
Disable consteval on Apple clang 21

lapfelix:
## Summary
Disable the `FMT_CONSTEVAL` path on Apple clang 21.

## Context
While building Transit's vendored copy of `fmt` with Xcode 26.4 beta 2 (Apple clang 21.0.0 / clang-2100.0.121.2), clang rejects several `FMT_STRING(...)` call sites with errors like:

```
error: call to consteval function 'fmt::basic_format_string<...>' is not a constant expression
```

The first failure I hit was in `format-inl.h` while compiling a downstream translation unit that includes `fmt/core.h`.

## Change
This introduces `FMT_USE_CONSTEVAL` and disables it for Apple clang 21+, leaving the existing `FMT_CONSTEVAL` path unchanged for other toolchains.

## Verification
Built the affected downstream target successfully after this change:

```bash
xcodebuild -project Submodules/transitLib/xcode_build/transitLib.xcodeproj \
  -scheme bgtfs \
  -destination 'generic/platform=iOS Simulator' \
  build
```

### DIFF
--- a/include/fmt/core.h
+++ b/include/fmt/core.h
@@ -221,15 +221,26 @@
 #  define FMT_UNICODE !FMT_MSC_VERSION
 #endif
 
+#ifndef FMT_USE_CONSTEVAL
+#  if defined(__apple_build_version__) && FMT_CLANG_VERSION >= 2100
+// Apple clang 21 (Xcode 26 beta) rejects some FMT_STRING consteval paths.
+#    define FMT_USE_CONSTEVAL 0
+#  elif ((FMT_GCC_VERSION >= 1000 || FMT_CLANG_VERSION >= 1101) && \
+         (!defined(__apple_build_version__) ||                     \
+          __apple_build_version__ >= 14000029L) &&                \
+         FMT_CPLUSPLUS >= 202002L) ||                             \
+        (defined(__cpp_consteval) &&                              \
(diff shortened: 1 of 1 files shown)


lakshit-acv:
We can possibly give reason or some explanation as to why the PR has been closed. Helps a lot towards getting more context. Thanks !

turban:
We're hitting the same issue in a React Native / Expo project building with Xcode 16 (Apple clang 16, not quite clang 21 but the same failure mode). The errors are identical — `call to consteval function '...basic_format_string...' is not a constant expression` in `format-inl.h`.

We worked around it by patching `base.h` in a CocoaPods `post_install` hook, removing the `< 14000029L` version threshold so that all Apple clang versions take the `FMT_USE_CONSTEVAL 0` path:

```
content = content.gsub(
  'defined(__apple_build_version__) && __apple_build_version__ < 14000029L',
  'defined(__apple_build_version__)'
)
```

The approach in this PR (adding a new threshold for clang 21+) seems like the cleaner upstream fix. Would the maintainers consider reopening or a follow-up PR? The workaround burden is falling on every downstream project that builds fmt via CocoaPods with a recent Xcode.

vitaut:
The error "not a constant expression" usually indicates that the format string is invalid. You'll need to provide a repro to demonstrate that there is an actual issue with consteval in clang (which is unlikely).

PR #4704, opened 2026-03-06
Add missing moves to grouping std::string member

Skylion007:
This ensures is no extra copy string contents of grouping if it ever exceeds SSO size.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -1219,7 +1219,7 @@ FMT_API auto thousands_sep_impl(locale_ref loc) -> thousands_sep_result<Char>;
 template <typename Char>
 inline auto thousands_sep(locale_ref loc) -> thousands_sep_result<Char> {
   auto result = thousands_sep_impl<char>(loc);
-  return {result.grouping, Char(result.thousands_sep)};
+  return {std::move(result.grouping), Char(result.thousands_sep)};
 }
 template <>
 inline auto thousands_sep(locale_ref loc) -> thousands_sep_result<wchar_t> {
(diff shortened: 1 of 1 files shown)


include/fmt/format.h
```diff
@@ -1219,7 +1219,7 @@ FMT_API auto thousands_sep_impl(locale_ref loc) -> thousands_sep_result<Char>;
 template <typename Char>
 inline auto thousands_sep(locale_ref loc) -> thousands_sep_result<Char> {
   auto result = thousands_sep_impl<char>(loc);
-  return {result.grouping, Char(result.thousands_sep)};
+  return {std::move(result.grouping), Char(result.thousands_sep)};
```
vitaut, line 1222: Grouping is normally very small (0-2 chars) so not sure this is needed unless there are other benefits to move here.

Skylion007: Usually yeah, but with SSO it will dispatch to the same the codepath anyway. Good to have anyway. If for someone reasons anyone does every use a large grouping, it will be fast. WE already do it on line 1974/1975 in the ctors so might as well be consistent. It also makes performance linters happy. Technically, SSO is not mandated by the spec so for any STL implementations don't implement it this would be marginally faster (if exceedingly rare).

vitaut: > with SSO it will dispatch to the same the codepath anyway

Could you clarify this part?

Skylion007: https://pvs-studio.com/en/blog/terms/6658/

Skylion007: TLDR: for now it will likely dispatch to the same code path but if grouping is ever a string containing 16 or more bytes, it will be more efficient since it can move the allocated string.

Otherwise, most CPP implementations implement Small String Optimization and are therefore smart enough to try stash the std::string into some left over space on that stack inside the std::string object directly at least on 64bit systems.

This PR could also prove beneficial on embedded systems where the smaller pointer size is not enough to stash a string into via a union trick. In which case it will be stack allocated, which is slow, and then copied at all of these call sites. Adding the std::move allows the rvalue ctor to be used to steal ownership of the allocation.

vitaut: I know about SSO but thought there is more to the codepath comment. Anyway, it won't hurt so let's merge it.

PR #4708, opened 2026-03-09
Add C++-Modules Based Linux CI workflow

### LINKED ISSUE #4707: CMake check for CXX_MODULEs does not factor in all compilers supported
The check for modules support in the CMakeLists.txt file currently checks for the CMake Version, Ninja Version and Visual Studio Version before activating the use of the CXX_Module FILE_SET. 

However, when the above conditions are true, some compilers like g++-13 and clang-15, which have support for modules, throw an error as they do not have support for the CXX module scanning.

closes #4709

MathewBensonCode:
- This adds a set of CI tests for linux that include compilers that support modules
- This CI workflow uses ninja as the build system generator
- Can be integrated with the other linux workflow after successful testing.
- This will also enable us to add linux module testing to #4702, which currently only enables the module tests on windows.

Closes #4707

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -33,6 +33,10 @@ jobs:
             std: 23
             install: sudo apt install g++-13
             shared: -DBUILD_SHARED_LIBS=ON
+          - cxx: g++-14
+            build_type: Release
+            std: 23
+            install: sudo apt install g++-14
           - cxx: clang++-11
             build_type: Debug
             std: 17
@@ -53,6 +57,11 @@ jobs:
             std: 20
             cxxflags: -stdlib=libc++
             install: sudo apt install libc++-14-dev libc++abi-14-dev
+          - cxx: clang++-20
+            build_type: Debug
+            std: 20
+            cxxflags: -stdlib=libc++
+            install: sudo apt install clang-20 libc++-20-dev libc++abi-20-dev
 
     steps:
     - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
@@ -129,7 +138,14 @@ jobs:
     - name: Add repositories for newer GCC
       run: |
         sudo apt-add-repository ppa:ubuntu-toolchain-r/test
-      if: ${{ matrix.cxx == 'g++-13' }}
+      if: ${{ matrix.cxx == 'g++-13' || matrix.cxx == 'g++-14' }}
+
+    - name: Install LLVM-20
+      run: |
+          wget https://apt.llvm.org/llvm.sh
+          chmod +x llvm.sh
+          sudo ./llvm.sh 20
+      if: ${{ matrix.cxx == 'clang++-20' }}
 
     - name: Add Ubuntu mirrors
       run: |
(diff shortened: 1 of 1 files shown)


MathewBensonCode:
The main purpose for the new config was to use a newer version of Ubuntu(24.04) for easier access to newer compilers. Also the config needs to use Ninja.

The original config has easier access to older compilers.

Trying to mix the two, is bringing up another set of issues. For example is we use the ninja generator with the older config, the builds fail because its attempting to scan for modules even in the older compilers. This is the issue that would be solved by #4709 

Also, there may be a need to consider the range of compilers being supported for CI, perhaps the older ones can be dropped and we could move up to a different range like the one being supported in the newer config.

MathewBensonCode:
Thanks for merging #4709. Because of this, the single config for linux, both modern and older has been made possible.

.github/workflows/linux.yml
```diff
@@ -158,9 +176,11 @@ jobs:
       run: |
         cmake -DCMAKE_BUILD_TYPE=${{matrix.build_type}} \
               -DCMAKE_CXX_STANDARD=${{matrix.std}} \
+              -DCMAKE_CXX_EXTENSIONS=OFF \
+              -G Ninja \
               -DCMAKE_CXX_VISIBILITY_PRESET=hidden \
               -DCMAKE_VISIBILITY_INLINES_HIDDEN=ON \
-              -DFMT_DOC=OFF -DFMT_PEDANTIC=ON -DFMT_WERROR=ON \
+              -DFMT_DOC=OFF -DFMT_PEDANTIC=ON \
```
vitaut, line 195: Let's only enable Ninja for the clang-20 and gcc-14 since make is the default and it makes sense to test it more. Also please revert other option changes.

vitaut (changes requested):

.github/workflows/linux.yml
MathewBensonCode: I have made the requested changes. 

For the module version, I had to remove the FMT_WERROR flag to enable gcc to work. 

It throws a warning about linking to an internal module symbol.

There is also the option of silencing that specific warning.

.github/workflows/linux.yml
```diff
@@ -33,6 +33,11 @@ jobs:
             std: 23
             install: sudo apt install g++-13
             shared: -DBUILD_SHARED_LIBS=ON
+          - cxx: g++-14
+            build_type: Release
+            std: 23
+            install: sudo apt install g++-14
+            shared: -DBUILD_SHARED_LIBS=ON
```
vitaut: I don't think this needs shared libs

.github/workflows/linux.yml
```diff
@@ -48,11 +53,17 @@ jobs:
           - cxx: clang++-14
             build_type: Debug
             std: 20
+            install: sudo apt install libc++-14-dev libc++abi-14-dev
```
vitaut: Isn't libc++abi-14-dev installed automatically with libc++-14-dev?

.github/workflows/linux.yml
```diff
@@ -48,11 +53,17 @@ jobs:
...
           - cxx: clang++-14
             build_type: Debug
             std: 20
             cxxflags: -stdlib=libc++
-            install: sudo apt install libc++-14-dev libc++abi-14-dev
+            install: sudo apt install clang-14 libc++-14-dev libc++abi-14-dev
```
vitaut: This is existing config, let's not change it unnecessarily.

.github/workflows/linux.yml
vitaut: What is the warning? We should fix it.

vitaut:
Merged, thanks!

PR #4709, opened 2026-03-09
Update Check for module support to include compiler versions

### LINKED ISSUE #4707: CMake check for CXX_MODULEs does not factor in all compilers supported

MathewBensonCode:
- We need to confirm compiler support for module scanning before creating the CXX_MODULE FILE_SET using the `target_sources` command
- This is especially relevant for compilers that have module support but do not support the module scanning such as g++13, clang15. For these we can use the older way of creating the fmt-module target

closes #4707

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,16 +50,28 @@ endif ()
 
 project(FMT CXX)
 
-# Determine support for the C++ module scanning.
-# Requires C++20, CMake >= 3.28 and (Ninja >= 1.11 OR Visual Studio >= 17.4).
-set(FMT_USE_CMAKE_MODULES FALSE)
+# Determine support for the C++ module scanning and CXX_MODULE File_Sets
+# Requires C++20, CMake >= 3.28 and 
+# Generators(Ninja >= 1.11 OR Visual Studio >= 17.4).
+# Compilers GCC>=14, Clang>=16 or MSVC >= 17.4
+# Source: https://cmake.org/cmake/help/latest/manual/cmake-cxxmodules.7.html
+Set(FMT_USE_CMAKE_MODULES FALSE)
 if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.28 AND CMAKE_CXX_STANDARD
                                                  GREATER_EQUAL 20)
   if (CMAKE_GENERATOR STREQUAL "Ninja")
     execute_process(COMMAND "${CMAKE_MAKE_PROGRAM}" "--version"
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -50,16 +50,28 @@ endif ()
...
                     OUTPUT_VARIABLE NINJA_VERSION)
     if (NINJA_VERSION VERSION_GREATER_EQUAL 1.11)
+      if (
+          (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" 
+          AND CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 14)
+        OR(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" 
+          AND CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 16)
+        OR(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC"
+          AND MSVC_VERSION GREATER_EQUAL 1934)
+        )
```
vitaut, line 72: I wonder if there is a more robust way than listing all the compiler versions? CMake should have all this information already. Don't they provide any way to check if module scanning is supported?

MathewBensonCode: CMake does have this information, but it is only triggered automatically if we are using a minimum of CMake Version 3.28, i.e. set by the `cmake_minimum_required(VERSION 3.28[...4.2.3]) command. i.e. the first version number sets the policy.

As we have currently have the minimum set to 3.5 for backwards compatibility, we have to do the checks ourselves for the purpose of determining how we will be constructing the module target.

There is the CMAKE_CXX_SCAN_FOR_MODULES variable which has a related but different role. It simply triggers the scanning for modules and is not a reliable source of information for whether module support is available or not. 

This is based on my understanding of the documentation.
- [cmake-modules-docs](https://cmake.org/cmake/help/latest/manual/cmake-cxxmodules.7.html)
- [cmake policy introduced in 3.28](https://cmake.org/cmake/help/latest/policy/CMP0155.html)

Also based on testing, if you try g++-13 with the previous checks, i.e. cmake above 3.28 and Using Ninja as the generator, it throws an error stating that the compiler supports modules but not scanning.

vitaut:
Merged, thanks!

PR #4710, opened 2026-03-11
Clean up a couple of CMake includes

localspook:
CMakeParseArguments is a no-op since CMake 3.5 and the minimum version {fmt} supports is 3.8:
    https://cmake.org/cmake/help/latest/module/CMakeParseArguments.html

GNUInstallDirs is already unconditionally included earlier in the file:
https://github.com/fmtlib/fmt/blob/2b4eb7df478a0a2ec319cf192292d2fc7b6acab2/CMakeLists.txt#L100


vitaut:
Merged, thanks!

PR #4712, opened 2026-03-12
Make fmt-c.h compile with Clang on Windows

localspook:
`fmt-c.h` fails to compile with Clang when running in MSVC emulation mode (https://godbolt.org/z/jovMMbG95) because Clang doesn't define `_MSVC_TRADITIONAL` or support the `/Zc:preprocessor` option (its preprocessor is *always* in the standard-compliant mode). I confirmed that with this change the library builds and the C tests pass.


vitaut:
Thank you

PR #4713, opened 2026-03-12
Move a few classes into an anonymous namespace

localspook:


src/os.cc
```diff
@@ -59,6 +59,8 @@
 #  endif
 #endif
 
+FMT_BEGIN_NAMESPACE
```
vitaut: This is not needed, let's revert.

vitaut (changes requested):
Thanks for the PR.

src/os.cc
localspook: The way it is right now, these symbols are in the global namespace (at least, in an anonymous subnamespace of it). It would seem better to keep the library's symbols in the library's namespace, no? (if you disagree though I won't argue about it)

vitaut: We only put things in the fmt namespace if they need to be externally visible.

vitaut:
Thank you

Issue #4714, opened 2026-03-13
std::format_to_n() doesn't work with wchar_t and fmt::runtime()


n0lavar:
Basically title: https://godbolt.org/z/3sTn6MY4j
Uncomment check<wchar_t>() to reproduce it.

sunmy2019:
simplified repro.
```
wchar_t buffer[4];
fmt::format_to_n(buffer, 3, fmt::runtime(L"{}"), 12345);
```

I will open an PR fixing it.

n0lavar:
It would be cool to move tests to typed tests so you cover all at once. For instance for me it isn't clear if char16_t will also work (I just remembered UE uses it on Linux). But it's just my mumbling.

PR #4715, opened 2026-03-14
Support fmt::runtime with wchar_t in fmt::format_to_n

sunmy2019:
resolves https://github.com/fmtlib/fmt/issues/4714

### DIFF
--- a/test/xchar-test.cc
+++ b/test/xchar-test.cc
@@ -10,7 +10,6 @@
 #include <algorithm>
 #include <complex>
 #include <cwchar>
-#include <string>
 #include <vector>
 
 #include "fmt/chrono.h"
@@ -98,23 +97,11 @@ TEST(xchar_test, compile_time_string) {
 #endif
 }
 
-#define WLIT(x) L##x
-#define U16LIT(x) u##x
-#define U32LIT(x) U##x
-
-#define DEFINE_FORMAT_TO_TEST(SUITE, NAME, CHAR_TYPE, LIT)                 \
-  TEST(SUITE, NAME) {                                                      \
-    auto buf = std::vector<CHAR_TYPE>();                                   \
-    fmt::format_to(std::back_inserter(buf), LIT("{}{}"), 42, LIT('\0'));   \
-    auto expected = std::vector<CHAR_TYPE>{LIT('4'), LIT('2'), LIT('\0')}; \
-    EXPECT_EQ(buf, expected);                                              \
-  }
-
-DEFINE_FORMAT_TO_TEST(xchar_test, wchar_format_to, wchar_t, WLIT)
-DEFINE_FORMAT_TO_TEST(xchar_test, char16_format_to, char16_t, U16LIT)
-DEFINE_FORMAT_TO_TEST(xchar_test, char32_format_to, char32_t, U32LIT)
-
-#undef DEFINE_FORMAT_TO_TEST
+TEST(xchar_test, format_to) {
+  auto buf = std::vector<wchar_t>();
+  fmt::format_to(std::back_inserter(buf), L"{}{}", 42, L'\0');
+  EXPECT_STREQ(buf.data(), L"42");
+}
 
 TEST(xchar_test, compile_time_string_format_to) {
   std::wstring ws;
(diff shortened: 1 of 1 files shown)


sunmy2019:
> Thanks for the PR. We shouldn't modify base.h for this.

How about reimplementing similar functions in `xchar.h`?

vitaut:
I think what we need here as an overload of `format_to_n` that takes `wformat_string` in `xchar.h`. Existing runtime format string machinery should take care about the rest as it already does for other overloads.

n0lavar:
what about char16_t?

vitaut:
Let's focus on `wchar_t` in this PR.

sunmy2019:
> Let's focus on `wchar_t` in this PR.

I use wformat_string in the new implementation. This may remove previous support of char type not in `char`, `wchar`. Tell me if it is not desired.

include/fmt/xchar.h
```diff
@@ -279,14 +279,11 @@ inline auto vformat_to_n(OutputIt out, size_t n, basic_string_view<Char> fmt,
...
-inline auto format_to_n(OutputIt out, size_t n, const S& fmt, T&&... args)
-    -> format_to_n_result<OutputIt> {
-  return vformat_to_n(out, n, fmt::basic_string_view<Char>(fmt),
-                      fmt::make_format_args<buffered_context<Char>>(args...));
+template <typename OutputIt, typename... T,
+          FMT_ENABLE_IF(detail::is_output_iterator<OutputIt, wchar_t>::value)>
+FMT_INLINE auto format_to_n(OutputIt out, size_t n, wformat_string<T...> fmt,
+                            T&&... args) -> format_to_n_result<OutputIt> {
+  return vformat_to_n(out, n, fmt.get(), fmt::make_wformat_args(args...));
 }
```
vitaut, line 287: The new overload looks fine but let's keep the old one as well, just constrain it similarly to `format_to`.

sunmy2019: done

sunmy2019:
I modified the original overload. I also added new tests for `char16_t`, `char32_t` that calls the original overload.

Please give me feedback about the new test cases. This is new for this repository.

test/xchar-test.cc
```diff
@@ -97,11 +98,23 @@ TEST(xchar_test, compile_time_string) {
...
-TEST(xchar_test, format_to) {
-  auto buf = std::vector<wchar_t>();
-  fmt::format_to(std::back_inserter(buf), L"{}{}", 42, L'\0');
-  EXPECT_STREQ(buf.data(), L"42");
-}
+#define WLIT(x) L##x
+#define U16LIT(x) u##x
+#define U32LIT(x) U##x
+
+#define DEFINE_FORMAT_TO_TEST(SUITE, NAME, CHAR_TYPE, LIT)                 \
```
n0lavar: you can do it without macros using TYPED_TEST from gtest

sunmy2019: I tried in dab57dc. Note that we still need some macros to handle string literals.

n0lavar: If you want to improve it, you can use the following technique: 
https://github.com/n0lavar/qxLib/blob/master/include/qx/containers/string/string_utils.h#L253
And use it like this:
QX_STR_PREFIX(TypeParam, "MyString")
Note that this is nitpicking, you can ignore this message :)

vitaut (changes requested):
Let's avoid the macros in tests and just focus on `wchar_t` case.

test/xchar-test.cc
```diff
@@ -126,21 +167,47 @@ TEST(xchar_test, format_as) {
   EXPECT_EQ(fmt::format(L"{}", test::struct_as_wstring_view()), L"foo");
 }
 
-TEST(format_test, wide_format_to_n) {
+TYPED_TEST(xchar_format_test, format_to_n) {
```
vitaut: This looks like an overkill and unrelated to the main focus of the change (`wchar_t` support).

sunmy2019:
Ok, I will revert it to 1dcff42.

vitaut:
Merged, thanks!

Issue #4167, opened 2024-09-19
/utf-8 in nvcc + msvc


DWin233:
Sorry to interrupt.
I used fmt::format(...) for the host function in the .cu file, which should theoretically be compiled by the host compiler (i.e. msvc), and I also passed the /utf-8 flag via -Xcompiler. 
But I'm still getting this error output. `error: static assertion failed with "Unicode support requires compiling with /utf-8"`
How do I fix this?
Here is an example. 
```
// main.cu 
// utf-8
#include <fmt/core.h>
int main() { fmt::print("{} in cuda.\n", 42); }
```
And here is the command and the output.
```
(base) PS D:\Code\Mycode\code_demo_test\cuda>  &"C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.5\bin\nvcc.exe" `
>> --use-local-env -ccbin "C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.40.33807\bin\HostX64\x64"`
>> -x cu -rdc=true  -I"C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.5\include" -I"C:\dev\vcpkg\installed\x64-windows\include" `
>>  -I"C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.5\include"  -G   --keep-dir test_code_cuda1\x64\Debug  -maxrregcount=0  --machine 64 `
>> --compile -cudart shared --std=c++20  --expt-relaxed-constexpr -g  -DFMT_HEADER_ONLY -DWIN32 -DWIN64 -D_DEBUG -D_CONSOLE -D_UNICODE -DUNICODE `
>>  -Xcompiler "/EHsc /W3 /nologo /Od /FS /Zi /RTC1 /MDd " -Xcompiler "/Fdtest_code_cuda1\x64\Debug\vc143.pdb" `
>> -o D:\Code\Mycode\code_demo_test\cuda\test_code_cuda1\x64\Debug\main.cu.obj "D:\Code\Mycode\code_demo_test\cuda\src\main.cu" -Xcompiler "/utf-8"
main.cu
C:\dev\vcpkg\installed\x64-windows\include\fmt\base.h(458): error: static assertion failed with "Unicode support requires compiling with /utf-8"
  static_assert(!1 || use_utf8(),
  ^

C:\dev\vcpkg\installed\x64-windows\include\fmt\format.h(3222): warning #27-D: character value is out of range
[...]

PR #4719, opened 2026-03-18
converted fractional_part_rounding_thresholds to use an array of u32s

Cazadorro:
converted fractional_part_rounding_thresholds to use an array of u32s instead of a unicode literal character array for compatibility with CUDA and non CUDA platforms, based on discussions in https://github.com/fmtlib/fmt/issues/4167

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -3175,10 +3175,10 @@ constexpr auto fractional_part_rounding_thresholds(int index) -> uint32_t {
   // Then we split this up into two separate arrays of char16_ts, so they can
   // be properly recombined into uint32_t.
 
-  return static_cast<uint16_t>(
+  return static_cast<uint32_t>(
              u"\x9999\x828f\x8041\x8006\x8000\x8000\x8000\x8000"[index])
              << 16u |
-         static_cast<uint16_t>(
+         static_cast<uint32_t>(
              u"\x999a\x5c29\x8938\x8db9\xa7c6\x10c7\x01ae\x002b"[index]);
 }
 


Cazadorro:
Note that this works with vcpkg as well (tested with overlay port), ran tests locally all appeared to pass (note however I'm on MSVC 2026).   I did use the special CUDA compiler arguments for UTF8 in the separate project I used to test VCPKG compatibliity, dealing with that is a whole other can of worms I didn't want to include in this fix, I just wanted to make sure I could use the escape hatch with those compiler arguments as with out this change you can't use a fmt at all no matter what you do with vcpkg and cuda. 

below are the arguments I used in the separate project
```
target_compile_options(cuda_experiments PRIVATE
        $<$<COMPILE_LANGUAGE:CUDA>:-Xcompiler=-utf-8 -Xcudafe=--unicode_source_kind=UTF-8 -Xcicc=--unicode_source_kind=UTF-8>)
```

soumik15630m:
The PR appears clean to me - Please run ` clang-format` to fix the failing lint check...

include/fmt/format.h
```diff
@@ -3149,15 +3149,21 @@ FMT_CONSTEXPR20 void format_hexfloat(Float value, format_specs specs,
   format_hexfloat(static_cast<double>(value), specs, buf);
 }
 
+// pre C++23 does not allow static constexpr inside of constexpr functions
+// so we must declare this outside of fractional_part_rounding_thresholds
+static constexpr uint32_t utf8_raw_rounding_thresholds[8] = {
+    0x9999999au, 0x828f5c29u, 0x80418938u, 0x80068db9,
+    0x8000a7c6u, 0x800010c7u, 0x800001aeu, 0x8000002bu};
```
vitaut: This will be duplicated in every TU which is undesirable. Another workaround is to use a UTF-16 literal and reconstruct a constant from two 16-bit halves.

Also `utf8_raw_` prefix is not needed (and incorrect) since these constants have nothing to do with UTF-8 or Unicode in general. We only use Unicode literals to workaround language limitations here.

Cazadorro: Okay, I had to look up rules on C++ on how this works, I didn't realize static constexpr outside of a function basically has the exact opposite effect in terms of linkage that static constexpr inside a constexpr function has.  However, I believe if I change this to an *inline* constexpr variable, it will fix this issue: https://stackoverflow.com/a/57407675  The address will be the same across translation units if I understand this correctly.    Inline is implied with constexpr functions, but *not* with constexpr variables.

vitaut: `inline` won't work because it's C++17.

Cazadorro: For some reason I thought the library required at least c++17, not c++11 I'll see about trying utf16 but if there's any weird platform limitations with that too I won't have the bandwidth to track that down.  Also won't there be endian-ness concerns with using utf16 litterals? And at that point why even bother with those if you could just use a char array? Though I guess the previous version must have also had endian-ness issues as well if I read the documentation correctly.

include/fmt/format.h
```diff
@@ -3161,8 +3161,32 @@ constexpr auto fractional_part_rounding_thresholds(int index) -> uint32_t {
...
+  const uint32_t byte_2 =
+      static_cast<uint8_t>("\x99\x8f\x41\x06\x00\x00\x00\x00"[index]);
+  const uint32_t byte_1 =
+      static_cast<uint8_t>("\x99\x5c\x89\x8d\xa7\x10\x01\x00"[index]);
+  const uint32_t byte_0 =
+      static_cast<uint8_t>("\x9a\x29\x38\xb9\xc6\xc7\xae\x2b"[index]);
+
+  // recombine as uint32, this should eliminate endian issues, as now we are
+  // shifting bytes as uint32 which should match platform endian.
+  return byte_3 << 24u | byte_2 << 16u | byte_1 << 8u | byte_0;
```
vitaut: I suggest using 16-bit (`u`) literals to simplify this.

vitaut (changes requested):

include/fmt/format.h
Cazadorro: I'm really hesitant to do that, won't it actually make things *more* complicated?  I'm still going to have to recombine byte-wise in the final step, and I'm also going to have to *extract* the bytes of the uint16s in order to do this.    There's no way to guarantee that utf16 would match platform endian right?  The standard does not appear guarantee utf16 big endian or little endian, nor does it appear to guarantee it matches platform endianness/ endianness of std::uint16_t. 

It would look something like:
```

const uint32_t first_bytes =
      static_cast<uint16_t>(u"..."
                            u"..."[index]);
const uint32_t second_bytes =      
        static_cast<uint16_t>(u"..."
                              u"..."[index]);
                              
const uint32_t byte_3 = first_bytes >> 8; 
const uint32_t byte_2 = first_bytes & 0xFF; 

const uint32_t byte_1 = second_bytes >> 8; 
const uint32_t byte_0 = second_bytes & 0xFF; 

// recombine as uint32, this should eliminate endian issues, as now we are
// shifting bytes as uint32 which should match platform endian.
return byte_3 << 24u | byte_2 << 16u | byte_1 << 8u | byte_0;
```

vitaut: The endianness doesn't matter, you just need to split the constants into two parts instead of four and recombine as you do now but with fewer operations. You don't nee to extract bytes from the 16-bit constants for the same reason we don't do it now for 32-bit constants.

Cazadorro:
> Mostly looks good but the body of the function must be a single expression for C++11 compatibility.

I'm confused, what isn't compatible with C++11 there?  Or is it like an optimization that's not guaranteed to happen or something in c++11 unless you make it a single line statement?

vitaut:
> I'm confused, what isn't compatible with C++11 there? Or is it like an optimization that's not guaranteed to happen or something in c++11 unless you make it a single line statement?

https://stackoverflow.com/q/8308812

vitaut:
Merged, thanks

PR #4731, opened 2026-04-06
Provide reasonable default definition for is_contiguous

user202729:
Another attempt at https://github.com/fmtlib/fmt/pull/4716 . Compiling `base-test.cc` takes around 10 seconds (both before and after), probably is fine. (not sure what else to measure.)

Instead of using [`contiguous_iterator`](https://en.cppreference.com/w/cpp/iterator/contiguous_iterator.html), I just check whether `.data()` can be called on it; if it can be called, the container is likely contiguous. Opinion?

[Some code](https://github.com/fmtlib/fmt/blob/602df7dab8c795777c82d30dc5c3b3a60364134c/include/fmt/format.h#L468-L482) assumes that if it's contiguous, `.resize()` exists, so it won't work for e.g. `std::array` even though that one is contiguous. But then `back_insert_iterator` doesn't work on `std::array` either.

Directly define a partial specialization of `is_contiguous` with `typename = void` leads to ambiguous overload errors, which is backwards incompatible if someone else special-case `is_contiguous`. See for example https://github.com/user202729/fmt/actions/runs/24015836992/job/70035092887

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -587,8 +587,21 @@ using string_view = basic_string_view<char>;
 template <typename T> class basic_appender;
 using appender = basic_appender<char>;
 
+namespace detail {
+
 // Checks whether T is a container with contiguous storage.
-template <typename T> struct is_contiguous : std::false_type {};
+template <typename T>
+auto is_contiguous_helper(int)
+    -> decltype(std::declval<T&>().data(), std::declval<T&>().size(),
+                std::declval<T&>().resize(size_t{}),
+                std::declval<T&>()[size_t{}], std::true_type{});
+
+template <typename T> auto is_contiguous_helper(...) -> std::false_type;
+
+}  // namespace detail
+
+template <typename T>
+struct is_contiguous : decltype(detail::is_contiguous_helper<T>(0)) {};
 
 class context;
 template <typename OutputIt, typename Char> class generic_context;


vitaut (changes requested):
Thanks for the PR! The idea makes sense but let's move SFINAE from a helper function and varargs hack to `is_contiguous` using `void_t`.

afurm:
Requiring `.resize(size_t{})` as a contiguity predicate seems unusual — `.resize()` is a mutating operation rather than a storage property. `std::array` has no `.resize()`, and `std::string`/`std::vector` can be used contiguously without resizing. Is this intentional, or could `.resize()` be removed from the check? What types would incorrectly match without it?

vitaut:
I agree that we should drop `resize` check. Maybe a better check would be to see if iterator is a pointer?

afurm:
Two technical concerns with this approach:

**1. `.resize()` requirement is too strict**

`std::array`, `std::string_view`, and `std::span` are all contiguous containers but lack `.resize()`. The current implementation will incorrectly classify them as non-contiguous. Consider splitting the check: one SFINAE expression for `.data() + .size() + []` (the universal contiguous requirement), and a separate trait for `.resize()` availability for code paths that genuinely need it.

**2. Varargs overload is unprincipled**

`is_contiguous_helper(...)` as a fallback is a fragile SFINAE pattern. `std::void_t` is the idiomatic C++ standard approach for this — it is the technique used throughout the rest of the fmt codebase and is more maintainable.

Your suggestion to use `void_t` addresses both issues and would produce a cleaner diff overall.

user202729:
about resize:

> [Some code](https://github.com/fmtlib/fmt/blob/602df7dab8c795777c82d30dc5c3b3a60364134c/include/fmt/format.h#L468-L482) assumes that if it's contiguous, `.resize()` exists, so it won't work for e.g. `std::array` even though that one is contiguous. But then `back_insert_iterator` doesn't work on `std::array` either.

about void:

> Directly define a partial specialization of `is_contiguous` with `typename = void` leads to ambiguous overload errors, which is backwards incompatible if someone else special-case `is_contiguous`. See for example https://github.com/user202729/fmt/actions/runs/24015836992/job/70035092887

what do you suggest? use another `detail::is_contiguous_helper` and `void`?

vitaut:
OK, let's keep the helper but I still think that `resize` should go since it's unrelated to contiguity. The case that uses `resize` assumes that it exists mostly based on the presence of `back_insert_iterator`, not because the container is contiguous.

vitaut:
@user202729 do you still plan to update this PR or shall we close it for now?

PR #4732, opened 2026-04-06
Avoid dropping null byte in format string

user202729:
Consider

```cpp
fmt::format("a\0{}");
fmt::format("{}\0a");
```
Both ought to raise a compile-time error because of missing format argument, but before this change, only the second one will. This is because previously `s` instead of `str` was passed in the checking function, which uses the following constructor overload

```cpp
#if FMT_GCC_VERSION
  FMT_ALWAYS_INLINE
#endif
  FMT_CONSTEXPR basic_string_view(const Char* s) : data_(s) {
#if FMT_HAS_BUILTIN(__builtin_strlen) || FMT_GCC_VERSION || FMT_CLANG_VERSION
    if (std::is_same<Char, char>::value && !detail::is_constant_evaluated()) {
      size_ = __builtin_strlen(detail::narrow(s));  // strlen is not constexpr.
      return;
    }
#endif
    size_t len = 0;
    while (*s++) ++len;
    size_ = len;
  }
```

and drop everything after the null byte.

Also simplify the other constructor since `str`'s type is already `string_view`.

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -2690,16 +2690,16 @@ template <typename... T> struct fstring {
     static_assert(count<(is_view<remove_cvref_t<T>>::value &&
                          std::is_reference<T>::value)...>() == 0,
                   "passing views as lvalues is disallowed");
-    if (FMT_USE_CONSTEVAL) parse_format_string<char>(s, checker(s, arg_pack()));
+    if (FMT_USE_CONSTEVAL)
+      parse_format_string<char>(str, checker(str, arg_pack()));
     constexpr bool unused = detail::enforce_compile_checks<sizeof(s) != 0>();
     (void)unused;
   }
   template <typename S,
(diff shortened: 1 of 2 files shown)


sunmy2019:
Interesting find.

Besides that, fmt's behavior differs from libstdc++ and libc++.

https://godbolt.org/z/cvMocMoKh

sunmy2019:
I think we need to figure out which behavior is desired.

Possible relevant issue:
Content after "\0" in FMT_COMPILE would cause a compile failure now.
https://godbolt.org/z/1jrcx8z9n

vitaut (changes requested):
Thanks for the fix! LGTM but please add a test case.

user202729:
test case added. Confirm that if I revert the commit before, the test fail.

vitaut:
Merged, thanks!

PR #4739, opened 2026-04-09
Fix bug re: return type of `f(un)lockfile` wrappers for Windows

mvastola:
Just a tiny bugfix I spotted: The `f(un)lockfile` wrappers in `format-inl.h` that wrap Windows's `_(un)lock_file` methods are defined with a trailing return type derived by using decltype on a hypothetical call to the underlying functions.

The wrappers don't contain a `return` in their bodies, however, so if the return type of the underlying functions were to ever change from `void`, there would be a compile error.

~~IIRC, the ability to write `return [other function returning void]();` in a function returning `void` is a new-ish addition to the standard (no idea when). Since these functions aren't public-facing, and no library code ever attempts to use their return values, I opted to just declare them `void` instead of adding `return`s.~~

This just adds `return` to each function.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -1477,10 +1477,10 @@ template <typename T> struct span {
 };
 
 template <typename F> auto flockfile(F* f) -> decltype(_lock_file(f)) {
-  _lock_file(f);
+  return _lock_file(f);
 }
 template <typename F> auto funlockfile(F* f) -> decltype(_unlock_file(f)) {
-  _unlock_file(f);
+  return _unlock_file(f);
 }
 
 #ifndef getc_unlocked


mwinterb:
I'm pretty sure that the trailing return type is for SFINAE.
And "return void-expr" in a function that returns void is a C++-ism from C++98.

mvastola:
> I'm pretty sure that the trailing return type is for SFINAE. 

Oops. I remember thinking of adding `requires requires(_unlock_file(f))` and then realized that was probably too new. Will fix.

> And "return void-expr" in a function that returns void is a C++-ism from C++98.

I wonder what language I was thinking of then. Maybe plain `C`? (Or `rust` or `go`?) I've run into that issue before somewhere.

mvastola:
@mwinterb , FYI just reverted and added in the `return`s since I hallucinated the compatibility issue.  :stuck_out_tongue:

vitaut:
It's unlikely that the return type for these functions will change but I guess there is no harm to being a bit more defensive here. Thanks for the PR, merged.

mvastola:
> It's unlikely that the return type for these functions will change but I guess there is no harm to being a bit more defensive here. Thanks for the PR, merged.

Agreed. It was pretty trivial. Never would have noticed if I hadn't been taking such a deep dive into that section of code.

PR #4748, opened 2026-04-16
Update link anchor

Powerbyte7:
I've updated a link anchor in the README to match the actual anchor written [here](https://github.com/fmtlib/fmt/blob/main/doc/api.md?plain=1#L671).


vitaut:
thank you

PR #4750, opened 2026-04-19
Forward outer context in nested_formatter for dynamic precision

### LINKED ISSUE #3860: Dynamic precision for `nested_formatter` does not work
When changing the `struct Point` example explaining `nested_formatter`-usage in the API reference to use dynamic precision, execution fails with an error

```
Execution build compiler returned: 0
Program returned: 139
terminate called after throwing an instance of 'fmt::v10::format_error'
  what():  precision is not integer
Program terminated with signal: SIGSEGV
```

Reproducer on godbolt: https://godbolt.org/z/srYqdvnEb

barry3406:
Dynamic precision like `{:.{}}` on a `nested_formatter` previously failed with `argument not found`, because the inner formatter stored an `arg_ref` that was being resolved against the intermediate `format_to` context rather than the original outer format call.

`nested()` now takes the outer `format_context`, and `nested_view` plumbs the outer args and locale through to `formatter<nested_view>::format`, so dynamic width/precision references resolve against the user's original argument list. The one-argument `nested(value)` overload is kept for backward compatibility; existing static-precision uses are unaffected.

Fixes #3860.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -3885,6 +3885,9 @@ template <typename OutputIt, typename Char> class generic_context {
   constexpr auto arg_id(basic_string_view<Char> name) const -> int {
     return args_.get_id(name);
   }
+  auto args() const -> const basic_format_args<generic_context>& {
+    return args_;
+  }
 
   constexpr auto out() const -> iterator { return out_; }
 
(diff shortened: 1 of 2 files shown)


include/fmt/format.h
```diff
@@ -4085,20 +4088,31 @@ template <typename T> struct formatter<group_digits_view<T>> : formatter<T> {
...
   template <typename FormatContext>
-  auto format(nested_view<T, Char> view, FormatContext& ctx) const
+  auto format(nested_view<T, Char, OuterContext> view, FormatContext& ctx) const
       -> decltype(ctx.out()) {
-    return view.fmt->format(*view.value, ctx);
+    // Delegate to the inner formatter using a context that carries the outer
+    // format args so dynamic specs like "{:.{}}" can be resolved.
+    auto outer_ctx = OuterContext(ctx.out(), view.outer_args, view.outer_loc);
+    auto it = view.fmt->format(*view.value, outer_ctx);
+    ctx.advance_to(it);
```
vitaut, line 4114: This will only work if iterators are compatible (same context types) so there shouldn't be parameterization on `OuterContext`.

test/format-test.cc
```diff
@@ -1995,6 +1995,12 @@ FMT_END_NAMESPACE
 TEST(format_test, nested_formatter) {
   EXPECT_EQ(fmt::format("{:>16.2f}", point{1, 2}), "    (1.00, 2.00)");
 }
+
+TEST(format_test, nested_formatter_dynamic_precision) {
+  // https://github.com/fmtlib/fmt/issues/3860
```
vitaut, line 2000: This is not needed, let's remove.

vitaut (changes requested):
Thanks for the PR.

vitaut:
Do you still plan to update this PR or shall we close it for now?

PR #4751, opened 2026-04-19
Replace `try` and `catch` with the `FMT_TRY` and `FMT_CATCH` macros

st0rmbtw:


vitaut (changes requested):
LGTM but please fix lint.

vitaut:
Merged, thanks!

PR #4752, opened 2026-04-20
Fix locale_ref should report invalid when locale is disabled

FredeHoey:
Fixes this error when calling `fmt::format("{:%F %T}", std::chrono::system_clock::now())`

```
fmt/include/fmt/chrono.h:2150:28: error: invalid 'static_cast' from type 'fmt::v12::locale_ref' to type 'bool'
 2150 |     detail::get_locale loc(static_cast<bool>(loc_ref), loc_ref);
      |
```


FredeHoey:
Sorry, didn't see that it was already fixed upstream.

PR #4753, opened 2026-04-20
Use FMT_TRY and FMT_CATCH in C api

FredeHoey:
If exceptions are disabled, try/catch don't compile


FredeHoey:
Duplicate of #4751

PR #4758, opened 2026-04-25
chore(IWYU): Remove some redundant includes

KaganCanSit:
Checked header dependencies using `IWYU` and removed includes identified as redundant. I check potential false positives and some warnings were addressed here — not all. 

`fmt` project has platform-specific guards and backward-compatibility build paths that's why I tried to approach it as minimally as possible.

I am marking this PR draft. I want analyze CI results and check again this changes.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -13,10 +13,7 @@
 
 #  include <algorithm>
 #  include <cerrno>  // errno
-#  include <climits>
-#  include <cmath>
-#  include <exception>
-#  include <new>  // std::bad_alloc
+#  include <new>     // std::bad_alloc
 #endif
 
 #if defined(_WIN32) && !defined(FMT_USE_WRITE_CONSOLE)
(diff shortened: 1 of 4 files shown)


KaganCanSit:
```cpp
#include "fmt/os.h"  // fmt::system_category
#include "fmt/ranges.h"
#include "gtest-extra.h"  // StartsWith

#ifdef __cpp_lib_filesystem
TEST(std_test, path) {
  using std::filesystem::path;
// ...
// ...
// ...
// ...
  EXPECT_EQ(fmt::format("{:?}", path(L"\xD800")), "\"\\ud800\"");
#  endif
}

// Intentionally delayed include to test #4303
#  include "fmt/ranges.h"
```

`fmt/test/std-test.cc` have two same include `fmt/ranges.h`. But If I change compile failed.  I'll take a look at it when I have time. (I need to find the test purpose of the second include.)

vitaut:
LGTM. Do you still plan to make changes to this PR or shall we merge?

KaganCanSit:
> LGTM. Do you still plan to make changes to this PR or shall we merge?

Thank you for your review. We can mergw. At the moment, I haven’t been able to find the time to look into the situation I mentioned in the comment above.

vitaut:
Merged, thanks!

PR #4759, opened 2026-04-29
Fix compile time format for ranges, style and std

j4niwzis:
Fix compile time format for ranges, style and std

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -133,8 +133,8 @@ void write_escaped_path(basic_memory_buffer<Char>& quoted,
 #if defined(__cpp_lib_expected) || FMT_CPP_LIB_VARIANT
 
 template <typename Char, typename OutputIt, typename T, typename FormatContext>
-auto write_escaped_alternative(OutputIt out, const T& v, FormatContext& ctx)
-    -> OutputIt {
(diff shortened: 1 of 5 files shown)


vitaut (changes requested):
Thanks for the PR. Overall looks good but please fix CI errors.

j4niwzis:
> Thanks for the PR. Overall looks good but please fix CI errors.

It should work now

vitaut:
Lint error is unrelated (please rebase) but there are still errors that do look related.

vitaut:
@j4niwzis do you plan to update this PR?

j4niwzis:
There seem to be no errors now

include/fmt/std.h
```diff
@@ -635,7 +635,7 @@ struct formatter<
   }
 
   template <typename Context>
-  auto format(const std::exception& ex, Context& ctx) const
+  FMT_CONSTEXPR auto format(const std::exception& ex, Context& ctx) const
```
vitaut, line 638: I don't think this will work because `what()` will only be `constexpr` in C++26.

vitaut: Anyway, this is a minor thing that can be addressed separately.

vitaut:
Merged, thanks!

Issue #4760, opened 2026-04-30
Ambiguous formatter<std::optional<T>> between fmt/std.h and fmt/ranges.h on C++26 (P3168R2)


indiosmo:
With C++26 `std::optional` range support, including both `<fmt/std.h>` and `<fmt/ranges.h>` fails with `ambiguous template instantiation`.

  ### Environment

  - fmt 12.1.0 (also reproduces on current `master` at the time of writing)
  - GCC 16.1.0, libstdc++ (ships P3168R2)
  - `-std=c++26`
  - Linux x86_64

  ### Reproduction

  ```cpp
  #include <optional>
  #include <string>

  #include <fmt/format.h>
  // commenting either works.
  // ranges prints: ["hi"] []
  // std prints: optional("hi") none
  #include <fmt/ranges.h>
  #include <fmt/std.h>

  int main() {
    std::optional<std::string> some{"hi"};
    std::optional<std::string> none{};
    fmt::print("{} {}\n", some, none);
  }
```

**Comment by indiosmo** [comment]
The way I solved it at least for now is by explicitly forcing `range_format::disabled` for std::optional, as I don't have a use case for formatting it as a range.

```
FMT_BEGIN_NAMESPACE

template <typename T, typename Char>
struct range_format_kind<std::optional<T>, Char> : std::integral_constant<range_format, range_format::disabled>
{
};

FMT_END_NAMESPACE
```

I suppose the solution at the library level first needs a design decision as to whether you want to support optional through ranges at all.

I would argue that it should stay in `fmt/std.h` and disabling it in `fmt/ranges.h` specifically keeps the behavior unchanged from previous versions and is the semantics users would expect from optional vs `["value"]` and `[]`.

indiosmo:
I think there's a stronger argument for picking one,.
I'm going to try this now, but I suppose we could construct code in a way where we silently get the same struct serializing in two different ways if we have two separate TUs each including either fmt/std.h or fmt/ranges.h and returning a string with the serialized struct.

indiosmo:
Ok, so not two different ways, but we now get link order dependent behavior.
If we have two TUs, each including one of the fmt headers, the one that gets linked first is the one that gets used, so still a very surprising behavior as the TU including fmt/std.h will serialize as range and vice-versa.

### CMakeLists.txt

```cmake
add_executable(sandbox
  main.cpp
  # flip the order here and you get std output
  serialize_with_ranges.cpp
  serialize_with_std.cpp
)

[...]

PR #4761, opened 2026-04-30
Fix ambiguous formatter<std::optional<T>> between fmt/std.h and fmt/ranges.h on C++26 (P3168R2)

phprus:
Issue: #4760 

@indiosmo, please test this PR.


phprus:
The clang-format error is not in the changes in the PR.

vitaut:
Thanks for the fix!

PR #4762, opened 2026-04-30
Fix std.h/ranges.h ambiguity for optional in c++26 (fmtlib#4760)

indiosmo:
C++26 adds range support for `std::optional` which causes ambiguity when including both `fmt/std.h` and `fmt/ranges.h`.

This PR specifically disables optional in `fmt/ranges.h`

See #4760 .


vitaut:
Closing in favor of #4761 but thanks for the PR!

PR #4770, opened 2026-05-13
Provide reasonable default definition for is_contiguous

phprus:
Based on PR #4731 with applied review suggestions.


vitaut (changes requested):
LGTM but could you add a test that exercises this check to base-test?

phprus:
> LGTM but could you add a test that exercises this check to base-test?

Tests added.

vitaut:
Merged, thank you!

PR #4771, opened 2026-05-14
docs: document official C++20 module usage

### LINKED ISSUE #4237: What is the proper way to use fmt as a C++ 20 module?

Reshi12:
TITLE: docs: document official C++20 module usage

## What this fixes
This PR addresses the ambiguity regarding C++20 module support in {fmt}. Users have been attempting various workarounds—such as manual module wrapping or complex subdirectory inclusions—due to a lack of explicit documentation, leading to build failures like missing `__fwd/string` headers.

## Why it was happening
There is no official support for "wrapping" {fmt} in a custom module. The library provides its own module interface, but users were often bypassing the standard CMake target configuration. This resulted in incorrect include paths and module scanning errors when trying to integrate {fmt} into modern C++23 projects.

## What changed
- `README.md`: Added a dedicated section "Using {fmt} as a C++20 Module." This clarifies that users should rely on the provided CMake target rather than manual wrapping and outlines the necessary CMake configuration to enable module scanning.

## How this was tested
Verified against the internal `cmake` test repository to ensure the documented configuration correctly triggers module scanning and builds without header resolution errors.

## Impact
Without this documentation, users will continue to struggle with build-time errors when attempting to adopt C++20 modules, likely leading to further issues with dependency management and incorrect library integration.

Closes #4237


Shrinidhibijapur (changes requested):
The Python cleanup/refactor looks good overall, but I noticed a few concerns:

1. Most of the README was removed — was that intentional?
2. `curl` dependency in `printable.py` may break portability on systems without curl (especially Windows).
3. Dict merge operator (`|`) requires Python 3.9+, is that guaranteed for this repo?
4. `FMT_TOKEN` should probably be validated before use to avoid runtime failure.

Other than that the refactor structure looks cleaner.

PR #4772, opened 2026-05-14
Add support for C++23 <stdfloat> types and std::complex<T>

Reshi12:
TITLE: Add support for C++23 <stdfloat> types and std::complex<T>

## What this fixes
This PR adds support for `std::float16_t`, `std::float32_t`, and `std::float64_t` in `{fmt}`. It also enables formatting for `std::complex<T>` for any type `T` that is already supported by the library.

## Why it was happening
The library's type traits for floating-point detection did not account for the new C++23 extended floating-point types defined in `<stdfloat>`. Consequently, `fmt::format` failed to recognize these types as valid floating-point numbers. Additionally, `std::complex<T>` lacked a dedicated formatter specialization, preventing it from being printed even when the underlying scalar type was supported.

## What changed
- `include/fmt/core.h`: Updated `is_floating_point` to include the `std::floatN_t` types when `__STDCPP_FLOAT16_T__` is defined.
- `include/fmt/std.h`: Added a `formatter` specialization for `std::complex<T>`. This implementation delegates to the existing formatter for the real and imaginary components, ensuring consistency with existing floating-point formatting logic.
- `test/std-test.cc`: Added unit tests to verify formatting for `std::float16_t` and `std::complex<std::float16_t>`.

## How this was tested
Verified with `cmake` build and execution of the updated `std-test.cc` suite. All tests passed, confirming that the new types are correctly identified and formatted.

## Impact
Without this change, users attempting to format C++23 extended floating-point types or complex numbers containing them encounter compilation errors. This change aligns `{fmt}` with the C++23 standard library additions.

Closes #4725


vitaut:
Please don't submit AI slop and don't spam on the issues or you will be banned.

PR #4773, opened 2026-05-17
docs: note std::format hex float 0x prefix difference

### LINKED ISSUE #4657: Difference in hexadecimal floating-point (`a`) format between `fmt` and `std` isn't documented
https://godbolt.org/z/TaK4s5YTM

When formatting floating-point numbers to hexadecimal (`a`) format, `fmt` includes the `0x` prefix (similar to `printf` `%a`), but `std` doesn't (specified in terms of `to_chars`). ~~Perhaps `std` should be fixed.~~

wyf027:
## Summary

Documents a behavioral difference between {fmt} and `std::format` for hexadecimal floating-point formatting (`'a'`/`'A'`): {fmt} includes a `0x`/`0X` prefix (like `printf` `%a`), while `std::format` (via `std::to_chars`) does not.

Fixes #4657.

## Test plan

- [x] Documentation-only change; verified wording against Godbolt example in the issue.

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -715,6 +715,10 @@ following differences:
   (ignoring redundant digits and sign in exponent) and may produce more
   decimal digits than necessary.
 
+- For the hexadecimal floating-point presentation types (`'a'`/`'A'`),
+  {fmt} includes a `0x`/`0X` prefix (similar to `printf`'s `%a`), while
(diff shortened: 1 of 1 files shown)


vitaut:
This is an LWG issue which should be resolved before documenting.

PR #4774, opened 2026-05-17
fix: convert std::error_code messages to UTF-8 on Windows

### LINKED ISSUE #4436: error code `:s` fail on windows due to encoding error
```cxx
#include <exception>
#include <filesystem>

#include "fmt/std.h"

int main() {

  try {
    fs::copy_file(from, to);
    try {
    } catch (std::exception e) {
    }
  }

  fmt::println("exit");
  return 0;
}
```

wyf027:
## Summary

On Windows, `std::error_code::message()` may return text in the active ANSI code page (e.g. GBK) rather than UTF-8. Formatting with `{:s}` then failed with "invalid utf8".

Convert the message from ACP to UTF-8 (via UTF-16) before appending when the string presentation is used.

Fixes #4436.

## Test plan

- [ ] Windows CI / manual repro from the issue (filesystem_error + `fmt::println("{:s}", ex.code())`).
- [x] Non-Windows builds unchanged (`#ifdef _WIN32` guard).

### DIFF
--- a/include/fmt/os.h
+++ b/include/fmt/os.h
@@ -117,6 +117,8 @@ FMT_API const std::error_category& system_category() noexcept;
 namespace detail {
 FMT_API void format_windows_error(buffer<char>& out, int error_code,
                                   const char* message) noexcept;
+FMT_API bool append_system_error_message_as_utf8(buffer<char>& buf,
+                                                 string_view message) noexcept;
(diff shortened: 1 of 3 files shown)


vitaut:
This is also an LWG issue what should be addressed first.

PR #4775, opened 2026-05-17
docs: add C++20 module usage to get-started guide

### LINKED ISSUE #4237: What is the proper way to use fmt as a C++ 20 module?

wyf027:
## Summary

Adds a **C++20 Modules** section to `doc/get-started.md` describing:

- Use `fmt::fmt-module` (do not wrap {fmt} in a custom module)
- Toolchain / CMake version requirements
- Minimal `import fmt;` consumer example
- `-DFMT_MODULE=ON` when building {fmt} from source

Fixes #4237.

## Test plan

- [x] Documentation-only change; examples match existing `fmt-module` CMake target and `test/module-test.cc`.


vitaut:
Please don't submit AI slop or you will be banned.

PR #4782, opened 2026-05-18
Allow to use fmt::println with colors and formatting

ahoarau:
Hi,
This PR simply adds the possibility to add colors and formatting while using `fmt::println`.
Example:
```c++
fmt::println(fmt::emphasis::bold | fg(fmt::color::red), "Hi!");
```

godbolt: https://godbolt.org/z/Ws511KPTz


vitaut:
Merged, thank you

PR #4789, opened 2026-05-31
Add `fmt_print` to C API

Ferdi265:
This PR introduces a `fmt_print()` macro to the C API, which print to a `FILE*` stream using `fmt::vprint(stream, fmt, format_args)`.

Additionally, this PR introduces the following helpers in `fmt-c.h`:

- a helper macro `FMT_FORMAT_ARGS(fmt, ...)` that abstracts away the macro magic necessary to produce the `fmt_arg` array, useful for avoiding duplicated logic when declaring additional wrapper macros that expand to calls to `fmt_vformat` or similar
- a `fmt_vprint()` function used to implement the `fmt_print()` macro

### DIFF
--- a/src/fmt-c.cc
+++ b/src/fmt-c.cc
@@ -9,12 +9,12 @@
 
 #include <fmt/base.h>
 
-extern "C" int fmt_vformat(char* buffer, size_t size, const char* fmt,
-                           const fmt_arg* args, size_t num_args) {
-  constexpr size_t max_args = 16;
-  if (num_args > max_args) return fmt_error_invalid_arg;
+constexpr size_t max_c_format_args = 16;
+static int convert_c_format_args(
+    fmt::basic_format_arg<fmt::format_context>* format_args,
+    const fmt_arg* args, size_t num_args) {
+  if (num_args > max_c_format_args) return fmt_error_invalid_arg;
 
-  fmt::basic_format_arg<fmt::format_context> format_args[max_args];
   for (size_t i = 0; i < num_args; ++i) {
     switch (args[i].type) {
     case fmt_int:    format_args[i] = args[i].value.int_value; break;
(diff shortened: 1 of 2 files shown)


vitaut (changes requested):
Thanks for the PR. I think adding `fmt_print` makes sense as a safe replacement for `printf` but I don't think we need `fmt_println` - let's keep the API small for now.

src/fmt-c.cc
```diff
@@ -8,13 +8,14 @@
...
 
 #include <fmt/base.h>
+#include <utility>
 
-extern "C" int fmt_vformat(char* buffer, size_t size, const char* fmt,
-                           const fmt_arg* args, size_t num_args) {
-  constexpr size_t max_args = 16;
-  if (num_args > max_args) return fmt_error_invalid_arg;
+constexpr size_t max_c_format_args = 16;
+static int parse_c_format_args(
```
vitaut: It's not parsing, maybe rename to convert?

src/fmt-c.cc
```diff
@@ -31,12 +32,44 @@ extern "C" int fmt_vformat(char* buffer, size_t size, const char* fmt,
     default:          return fmt_error_invalid_arg;
     }
   }
+  return 0;
+}
+
+template <typename F>
+static int fmt_c_wrapper(const char* fmt, const fmt_arg* args, size_t num_args,
+                         F&& f) {
```
vitaut: I don't think this is needed.

src/fmt-c.cc
Ferdi265: makes sense, done

src/fmt-c.cc
Ferdi265: ok, removed

Ferdi265:
> Thanks for the PR. I think adding `fmt_print` makes sense as a safe replacement for `printf` but I don't think we need `fmt_println` - let's keep the API small for now.

Thanks for the review! I removed `fmt_println()` and applied the other two suggestions you made.

vitaut (approved):
LGTM

PR #4790, opened 2026-06-01
Fix formatting std::tm with null tm_zone

Carmel0:
This fixes formatting `%Z` for a `std::tm` whose platform `tm_zone` field exists but is null. The syntax documentation says that unavailable timezone information should result in `format_error`; currently this path passes `tm_zone` directly to string formatting and can dereference a null C string.

The change reports the existing `no timezone` error for a null `tm_zone` and adds a chrono regression test.

Tests:
- `/tmp/fmt-pr-build/bin/chrono-test`
- `ctest --test-dir /tmp/fmt-pr-build --output-on-failure -j$(nproc)`


vitaut (approved):
Thanks for the PR!

PR #4795, opened 2026-06-03
Add CMake option to disable XCHAR support

ibookstein:
This is useful in some embedded scenarios where the libc's wchar_t support is disabled.


ibookstein:
I thought this would be sufficient in order to do https://github.com/fmtlib/fmt/issues/1077, but the uses of `std::wstring` and `wchar_t` are not isolated to that header so it may require additional work. So it seems that `FMT_XCHAR` needs to become a macro for this to be a full solution.

PR #4799, opened 2026-06-06
Add .gradle directory to .gitignore

EXtremeExploit:
When building the project with java/gradle, it generates a .gradle folder in directories, these are used by gradle to keep track of file hashes, cache and other stuff, so not needed to track and should be ignored by git


vitaut (approved):
Thanks

PR #4801, opened 2026-06-09
Use logical operators for bool checks

Kielek:
These floating-point formatting checks combine bool values, but they use bitwise '&' and '|' operators. CodeQL flags the expression that applies '!' next to bitwise '&' as incorrect not-operator usage. Use logical '&&' and '||' for the related bool checks so the code expresses boolean logic directly while preserving the same truth table.

Found by CodeQL analysis in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/5079 which using this project as transitive dependency through spdlog.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -1338,7 +1338,7 @@ template <typename T> auto to_decimal(T x) noexcept -> decimal_fp<T> {
 
   if (r < deltai) {
     // Exclude the right endpoint if necessary.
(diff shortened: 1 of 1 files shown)


sunmy2019:
I think the bitwise operator here is used for performance optimization. It reduces branches and avoids branch mispredictions.

vitaut:
> I think the bitwise operator here is used for performance optimization. It reduces branches and avoids branch mispredictions.

That's correct and also this code is soon going away in favor of https://github.com/vitaut/zmij. But thanks for the PR anyway.

PR #4809, opened 2026-06-15
fix: allow zero-padding for pointer formatting (fixes #2025)

### LINKED ISSUE #2025: Support zero-padding with pointers
```c++
#include <cstdio>

#include "fmt/format.h"

int main()
{
    void* ptr = (void*)0x2993;
    printf("%010p\n", ptr);
    fmt::print("{:010}\n", ptr);
    return 0;
}
```

### Expected
```
0x00002993
0x00002993
```
### Actual
```
terminate called after throwing an instance of 'fmt::v6::format_error'
  what():  format specifier requires numeric argument
```

I can understand not wanting to open the whole set of numeric formatting options to pointers, but the lack of even a very basic "zero-pad to the width of a pointer" kind of odd.  Would adding this be in the cards?

Muptezel58:
## Problem
Using the `0` (zero-pad) flag with pointer arguments currently raises:
  format specifier requires numeric argument

## Root Cause
In `include/fmt/base.h`, the `case '0':` handler only allows 
`is_arithmetic_type`, which excludes `pointer_type`.

## Fix
Extended the check to also permit `pointer_type`:

```cpp
// Before:
if (!is_arithmetic_type(arg_type))

// After:
if (!is_arithmetic_type(arg_type) && !in(arg_type, pointer_set))
```

## Test
```cpp
void* ptr = (void*)0x2993;
fmt::format("{:018}", ptr); // Previously threw, now returns: 000000000000x2993
```

Closes #2025

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -1501,7 +1501,7 @@ FMT_CONSTEXPR auto parse_format_specs(const Char* begin, const Char* end,
       break;
     case '0':
       enter_state(state::zero);
-      if (!is_arithmetic_type(arg_type))
+      if (!is_arithmetic_type(arg_type) && !in(arg_type, pointer_set))
         report_error("format specifier requires numeric argument");
       if (specs.align() == align::none) {
         // Ignore 0 if align is specified for compatibility with std::format.


vitaut:
Thanks for the PR but the format specs are pretty much fixed at this point for compatibility with `std::format`.

PR #4812, opened 2026-06-16
Turn libfmt-c into shared

jengelh:
Linux distros don't like having static archives. Set fmt-c to shared just like fmt.

### DIFF
--- a/include/fmt/fmt-c.h
+++ b/include/fmt/fmt-c.h
@@ -12,6 +12,25 @@
 #include <stddef.h>   // size_t
 #include <stdio.h>    // FILE
 
+#if defined(__GNUC__)
+#  define FMT_CVISIBILITY(value) __attribute__((visibility(value)))
+#else
+#  define FMT_CVISIBILITY(value)
+#endif
+
+#if !defined(FMT_HEADER_ONLY) && defined(_WIN32)
+#  if defined(FMT_LIB_EXPORT)
+#    define FMT_CAPI __declspec(dllexport)
+#  elif defined(FMT_SHARED)
+#    define FMT_CAPI __declspec(dllimport)
+#  endif
+#elif defined(FMT_LIB_EXPORT) || defined(FMT_SHARED)
+#  define FMT_CAPI FMT_CVISIBILITY("default")
+#endif
+#ifndef FMT_CAPI
+#  define FMT_CAPI
+#endif
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -45,10 +64,10 @@ typedef struct {
 
 enum { fmt_error = -1, fmt_error_invalid_arg = -2 };
 
-int fmt_vformat(char* buffer, size_t size, const char* fmt, const fmt_arg* args,
-                size_t num_args);
-int fmt_vprint(FILE* stream, const char* fmt, const fmt_arg* args,
-               size_t num_args);
+int FMT_CAPI fmt_vformat(char* buffer, size_t size, const char* fmt,
+                         const fmt_arg* args, size_t num_args);
+int FMT_CAPI fmt_vprint(FILE* stream, const char* fmt, const fmt_arg* args,
+                        size_t num_args);
 
 #ifdef __cplusplus
 }
@@ -92,7 +111,7 @@ static inline fmt_arg fmt_from_ptr(const void* x) {
   return (fmt_arg){.type = fmt_pointer, .value.pointer = x};
 }
 
-void fmt_unsupported_type(void);
+void FMT_CAPI fmt_unsupported_type(void);
 
 #  if !defined(_MSC_VER) || defined(__clang__)
 typedef signed char fmt_signed_char;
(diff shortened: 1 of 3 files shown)


include/fmt/fmt-c.h
```diff
@@ -12,6 +12,48 @@
...
+#endif
+#if FMT_GCC_VERSION || FMT_CLANG_VERSION
+#  define CFMT_VISIBILITY(value) __attribute__((visibility(value)))
+#else
+#  define CFMT_VISIBILITY(value)
+#endif
+
+#if !defined(FMT_HEADER_ONLY) && FMT_WIN32
+#  if defined(FMT_LIB_EXPORT)
+#    define CFMT_API __declspec(dllexport)
```
vitaut: I think we can use the same macro, `FMT_API`, here.

include/fmt/fmt-c.h
```diff
@@ -12,6 +12,48 @@
...
+#  define FMT_ICC_VERSION __INTEL_COMPILER
+#else
+#  define FMT_ICC_VERSION 0
+#endif
+#if defined(_MSC_VER)
+#  define FMT_MSC_VERSION _MSC_VER
+#else
+#  define FMT_MSC_VERSION 0
+#endif
+#if FMT_GCC_VERSION || FMT_CLANG_VERSION
```
vitaut: We could probably just check `__GNUC__` here and remove all of the above.

vitaut (changes requested):

include/fmt/fmt-c.h
jengelh: I agree it's ugly, but `#including <fmt/base.h>` pulls in C++ code, so I deferred doing that :-/

vitaut: I meant reusing the name without pulling in `fmt/base.h`. The macros should be defined to the same thing in both places.

jengelh: Yeah.
Or, maybe, should we make a new separate header file?

jengelh:
```
In file included from ~/fmt/src/fmt-c.cc:10:
~/fmt/include/fmt/base.h:288:11: warning: ‘FMT_API’ redefined
```

Because fmt-c.cc includes both fmt/base.h and fmt-c.h, FMT_API would get defined twice.

vitaut (changes requested):
Please don't add new headers.

include/fmt/fmt-c.h
vitaut: This comment still applies.

include/fmt/fmt-c.h
```diff
@@ -12,6 +12,48 @@
...
+#elif defined(__INTEL_COMPILER)
+#  define FMT_ICC_VERSION __INTEL_COMPILER
+#else
+#  define FMT_ICC_VERSION 0
+#endif
+#if defined(_MSC_VER)
+#  define FMT_MSC_VERSION _MSC_VER
+#else
+#  define FMT_MSC_VERSION 0
+#endif
```
vitaut: Further to my earlier comment: this is not needed.

vitaut (changes requested):
I think it's OK to have a separate macro for now, but let's rename it from `CFMT_API` to `FMT_CAPI` to avoid reserving a new prefix. And similarly for visibitlity.

tom--pollard:
We're looking to distribute 12.2.0 in the freedesktop-sdk project and also agree that we would prefer not to include the static lib:

https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/merge_requests/32873

include/fmt/fmt-c.h
```diff
@@ -12,6 +12,26 @@
 #include <stddef.h>   // size_t
 #include <stdio.h>    // FILE
 
+// Detect compiler versions.
```
vitaut: This comment is no longer correct, please remove.

vitaut (changes requested):
There are some CI failures.

include/fmt/fmt-c.h
```diff
@@ -12,6 +12,25 @@
 #include <stddef.h>   // size_t
 #include <stdio.h>    // FILE
 
+#if defined(__GNUC__)
+#  define FMT_CVISIBILITY(value) __attribute__((visibility(value)))
+#else
+#  define FMT_CVISIBILITY(value)
+#endif
+
+#if !defined(FMT_HEADER_ONLY) && FMT_WIN32
```
vitaut: FMT_WIN32 -> defined(_WIN32) since FMT_WIN32 is not defined here.

jengelh:
> There are some CI failures.
>
>`LINK : fatal error LNK1104: cannot open file '..\Debug\fmt-cd.lib' [D:\a\fmt\build\test\c-test.vcxproj]`

I do not know how to fix this. On a whim I had previously tried switching `PRIVATE fmt::fmt-c` in tests/CMakeLists.txt for PUBLIC, but that error stayed. (so the change is backed out again)

vitaut (approved):

tom--pollard:
Thanks for this!

PR #4819, opened 2026-06-18
feat: add formatter for std::exception_ptr (#4808)

### LINKED ISSUE #4808: Formatting std::exception_ptr:s
It would be nice if the library (and the Standard) supported formatting std::exception_ptr:s as part of fmt/std.h or \<format>. Of course, anyone can implement a formatter, but if third-party libraries do this, they will eventually clash.

Atlas-Zero:
### Description
This PR adds a `fmt::formatter` specialization for `std::exception_ptr` to `fmt/std.h`, resolving the feature request in #4808. 

It safely handles `nullptr` cases, extracts the message via `e.what()` for standard exceptions, and provides a fallback for unknown types.

### Related Issue
Closes #4808

### Checklist
- [x] Formatter implemented in `include/fmt/std.h`
- [x] Unit tests added to `test/std-test.cc`
- [x] All tests passing locally

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -649,6 +649,25 @@ struct formatter<
   }
 };
 
+template <> struct formatter<std::exception_ptr> : formatter<std::exception> {
+  template <typename FormatContext>
+  auto format(const std::exception_ptr& ex_ptr, FormatContext& ctx) const
+      -> decltype(ctx.out()) {
+    if (!ex_ptr) {
+      return detail::write(ctx.out(), string_view("nullptr"));
+    }
+
+    try {
+      std::rethrow_exception(ex_ptr);
+    } catch (const std::exception& e) {
+      // Reuses the base formatter<std::exception>::format behavior perfectly
+      return formatter<std::exception>::format(e, ctx);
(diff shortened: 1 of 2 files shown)


phprus:
Maybe it makes sense to reuse `formatter<std::exception>` or its behavior?

https://github.com/fmtlib/fmt/blob/588b3a0f8f6a8bcf2a959cae882d5b2703e86737/include/fmt/std.h#L618-L650

vitaut (changes requested):
As @phprus correctly pointed out the new formatter should reuse the existing one for `std::exception`.

Atlas-Zero:
Thanks for the guidance @vitaut and @phprus! 

I have updated the implementation to explicitly inherit from and reuse `formatter<std::exception>`. This cleanly bypasses duplicating the logic and automatically brings support for the `{:t}` flag to `std::exception_ptr`. 

I also updated the unit tests in `test/std-test.cc` to use the internal `my_ns1::my_ns2::my_exception` type to verify both standard formatting and type-demangling output layouts.

vitaut (approved):

PR #4822, opened 2026-06-23
avoid MSC warning C4702: unreachable code

torsten48:
if FMT_CONSTEXPR is resolved to constexpr MSC warns about dead code after 'return' instruction. Explicitly using 'else' in these cases avoids the warning.


Skylion007 (approved):

vitaut (approved):
In general we don't care about level 4 MSVC warnings because of their very low quality but I think this particular change is relatively small scope so I'm OK merging it.

PR #4823, opened 2026-06-23
fix: use logical AND instead of bitwise AND in endpoint check

ComixHe:
While configuring CodeQL for my own project, it flagged a warning about mixing a logical NOT with a bitwise operator in the endpoint check: if (r == 0 && (z_mul.is_integer & !include_right_endpoint)). 

After looking into the code, I verified that both operands are bool types, which means the current logic happens to work fine due to implicit type promotion. 

I've changed it to a logical && to restore standard boolean behavior, future-proof the code, and clear the alert.

warning: https://github.com/OpenAtom-Linyaps/linyaps-box/security/code-scanning/1


mwinterb:
See https://github.com/fmtlib/fmt/pull/4801#issuecomment-4659669692

PR #4827, opened 2026-06-26
Move export() directive outside of FMT_INSTALL scope

### LINKED ISSUE #4806: `export(` should not be scoped by `FMT_INSTALL`
Currently the `export(...)` cmake directive  is scoped inside `if (FMT_INSTALL)`:
https://github.com/fmtlib/fmt/blob/11ddbcb7898d2d3445d431a54814367b21dee6ad/CMakeLists.txt#L496-L499

However, there are cases in which downstream projects might not want to install `fmt` but might still want to provide a `export(...)` themselves. However, this is impossible to do when depending on another target that doesn't `export` itself. CMake will error with something like
```
CMake Error: export("MyLibTargets" ...) includes target "mylib" which requires target "fmt" that is not in any export set.
```

Since having the exported target files in the build tree don't harm anyone, I suggest just unconditionally exporting regardless of whether `FMT_INSTALL` is true or not.

See also e.g. https://github.com/Kistler-Group/sdbus-cpp/pull/523

RealSteel99-codes:
Closes #4806

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -506,10 +506,7 @@ if (FMT_INSTALL)
 
   # Use a namespace because CMake provides better diagnostics for namespaced
   # imported targets.
-  export(
-    TARGETS ${INSTALL_TARGETS}
-    NAMESPACE fmt::
-    FILE ${PROJECT_BINARY_DIR}/${targets_export_name}.cmake)
+ 
 
   # Install version, config and target files.
   install(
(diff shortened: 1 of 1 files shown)


vitaut (changes requested):
Mostly looks good but please fix the formatting (see CI error).

CMakeLists.txt
```diff
@@ -506,10 +506,7 @@ if (FMT_INSTALL)
 
   # Use a namespace because CMake provides better diagnostics for namespaced
   # imported targets.
```
vitaut, line 508: This should be moved to the new `export` location too.

vitaut (changes requested):
Please fix formatting (see lint) and address the inline comment

vitaut:
@RealSteel99-codes do you plan to update this PR or shall we close it?

vitaut:
Closing due to inactivity.

PR #4830, opened 2026-06-28
Skip VERSION/SOVERSION/DEBUG_POSTFIX on the header-only INTERFACE target

### LINKED ISSUE #4818: 12.2.0 fails to build with old CMake 3.15.1
Hello,

According to CMakeLists.txt supported version range it should work with 3.8 and newer versions. But on old 3.15.1 (default in OpenWRT 19.07) it broke in setup_target() function when specifying VERSION, SOVERSION and DEBUG_POSTFIX properties for INTERFACE type fmt-header-only target. 
```
-- {fmt} version: 12.2.0
-- Build type: Release
CMake Error at CMakeLists.txt:259 (set_target_properties):
  INTERFACE_LIBRARY targets may only have whitelisted properties.  The
  property "VERSION" is not allowed.
Call Stack (most recent call first):
  CMakeLists.txt:418 (setup_target)


CMake Error at CMakeLists.txt:259 (set_target_properties):
  INTERFACE_LIBRARY targets may only have whitelisted properties.  The
  property "SOVERSION" is not allowed.
Call Stack (most recent call first):
  CMakeLists.txt:418 (setup_target)


CMake Error at CMakeLists.txt:259 (set_target_properties):
  INTERFACE_LIBRARY targets may only have whitelisted properties.  The
  property "DEBUG_POSTFIX" is not allowed.
Call Stack (most recent call first):
  CMakeLists.txt:418 (setup_target)
```

This worked fine on 12.1.0. Not sure if it is best way, but can be fixed by not setting shared library specific properties for INTERFACE header-only targets. Patch in attachment

[001-fix-cmake.patch](https://github.com/user-attachments/files/29093843/001-fix-cmake.patch)

soumik15630m:
Closes #4818 

Skip that `set_target_properties()` call when `kind` is `INTERFACE`. These properties have no effect on a header-only target on *any* CMake version (there's no binary artifact to version or suffix), so this is a no-op change for the actual compiled `fmt`/`fmt-c` targets and only removes a call that was never doing anything useful for `fmt-header-only`.

Originating manily from ....
`setup_target(target kind)` is called for both compiled targets (`PUBLIC`/`PRIVATE` kind) and the header-only `fmt-header-only` target (`INTERFACE` kind, line 418). It unconditionally calls `set_target_properties(... VERSION ... SOVERSION ... DEBUG_POSTFIX ...)`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -258,10 +258,10 @@ function (setup_target target kind)
 
   if (NOT "${kind}" STREQUAL "INTERFACE")
     set_target_properties(
-            ${target}
-            PROPERTIES VERSION ${FMT_VERSION}
-            SOVERSION ${CPACK_PACKAGE_VERSION_MAJOR}
-            DEBUG_POSTFIX "${FMT_DEBUG_POSTFIX}")
+      ${target}
+      PROPERTIES VERSION ${FMT_VERSION}
+                 SOVERSION ${CPACK_PACKAGE_VERSION_MAJOR}
+                 DEBUG_POSTFIX "${FMT_DEBUG_POSTFIX}")
   endif ()
 endfunction ()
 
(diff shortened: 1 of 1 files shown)


vitaut (approved):
Thanks for the fix!

PR #4837, opened 2026-07-05
Add formatter for std::exception_ptr

### LINKED ISSUE #4808: Formatting std::exception_ptr:s

avikivity:
The formatter prints std::exception::what() (if that's what the
exception is) and prints the type name for exceptions
that are not derived from std::exception.

Fixes #4808

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -506,6 +506,8 @@ chrono-format-specifications).
 - [`std::bitset`](https://en.cppreference.com/w/cpp/utility/bitset)
 - [`std::error_code`](https://en.cppreference.com/w/cpp/error/error_code)
 - [`std::exception`](https://en.cppreference.com/w/cpp/error/exception)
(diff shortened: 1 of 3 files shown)


phprus:
It makes sense to reuse `formatter<std::exception>` format behavior:
https://github.com/fmtlib/fmt/pull/4819#issuecomment-4746083478

avikivity:
Update:
 - rebase
 - add comment in api.md
 - reuse formatter\<std::exception>

vitaut:
Thanks for the PR but the `std::exception_ptr` formatter has already been added in #4819. If you would like to propose any extra functionality please build on top of that.

PR #4838, opened 2026-07-05
Name namespace-scope unscoped enums

avikivity:
Unnamed unscoped enums at namespace scope are problematic
(https://github.com/llvm/llvm-project/issues/207581). Giving each such
enum a name (derived from the constant it declares) is enough to avoid
the issue, so do that for the header-level integral constants instead of
leaving them anonymous.

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -445,8 +445,8 @@ struct is_std_string_like<T, void_t<decltype(std::declval<T>().find_first_of(
                           const typename T::value_type*> {};
 
 // Check if the literal encoding is UTF-8.
-enum { is_utf8_enabled = "\u00A7"[1] == '\xA7' };
-enum { use_utf8 = !FMT_WIN32 || is_utf8_enabled };
+enum is_utf8_enabled_t { is_utf8_enabled = "\u00A7"[1] == '\xA7' };
+enum use_utf8_t { use_utf8 = !FMT_WIN32 || is_utf8_enabled };
 
 #ifndef FMT_UNICODE
 #  define FMT_UNICODE 1
@@ -1030,7 +1030,7 @@ constexpr auto in(type t, int set) -> bool {
 }
 
 // Bitsets of types.
-enum {
+enum type_set {
   sint_set =
       set(type::int_type) | set(type::long_long_type) | set(type::int128_type),
   uint_set = set(type::uint_type) | set(type::ulong_long_type) |
@@ -1121,7 +1121,9 @@ FMT_CONSTEXPR void init_static_named_arg(named_arg_info<Char>* named_args,
 
 // To minimize the number of types we need to deal with, long is translated
 // either to int or to long long depending on its size.
-enum { long_short = sizeof(long) == sizeof(int) && FMT_BUILTIN_TYPES };
+enum long_short_t {
+  long_short = sizeof(long) == sizeof(int) && FMT_BUILTIN_TYPES
+};
 using long_type = conditional_t<long_short, int, long long>;
 using ulong_type = conditional_t<long_short, unsigned, ullong>;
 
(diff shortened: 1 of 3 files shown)


avikivity:
Update:
 - rebased
 - simplified: it's enough to name the unscoped enum to make it module-friendly

avikivity:
@vitaut please review

vitaut:
Thanks for the PR.

The linked issue says

> Two conditions are both necessary (and, together, sufficient):
> 
> 1. ...
> 2. A module **re-exports that enumerator** with a using-declaration. Re-exporting
>    a sibling (a type, or the dependent alias itself) does not trigger it.

Could you elaborate where these enumerators are re-exported?

avikivity:
The first module to export the enum is Seastar. It doesn't name the enumerator directly. Rather it uses fmt via headers, (Seastar cannnot assume all its users consume it as a module), and some of Seastar's exported symbols depend on this enum (max_packed_args, easy to see how it would become a dependency).

The second module is my fmt.cppm (I should probably switch to fmt.cc, but it isn't packaged by Fedora). Whether it is named there or not, it is referred to.

When the two modules are imported by ScyllaDB, the two declarations clash.

Naming those enums makes the problem go away. It was also fixed in clang, but I ask that you accept this patch as a band-aid for people who use ancient compilers like clang 22.1.8 (clang 22 isn't backporting any more, as well).

gcc is also vulnerable, I'll file a bug report there too.

avikivity:
> 
> gcc is also vulnerable, I'll file a bug report there too.

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=126192

vitaut:
But fmt's module doesn't re-export the `max_packed_args` enumerator with a using-declaration, so I still don't understand where the issue is coming from.

avikivity:
Modules will stealthily export enough stuff to be able to build the exported inline functions.

IIRC my fmt.cppm did not export max_packed_args using a using statement, and I only added it later trying to workaround the failures. I will verify and report.

vitaut (changes requested):
> IIRC my fmt.cppm did not export max_packed_args using a using statement, and I only added it later trying to workaround the failures. I will verify and report.

OK, I'll wait until a more precise repro then.

avikivity:
I tried to switch to the built-in module support.

I get errors like

```
In module 'fmt' imported from ./seastarx.hh:26:
/home/avi/scylla/fmt/include/fmt/base.h:904:7: error: declaration 'locale_ref' attached to named module 'fmt' cannot be attached to other modules
  904 | class locale_ref {
      |       ^
/home/avi/scylla/build/dev/fmt/install/include/fmt/base.h:904:7: note: also found
  904 | class locale_ref {
      |       ^
```

fmt is both imported and included, and in a large project that has many dependencies, this often cannot be avoided.

My home spun modules were using using statement, e.g.

```c++
+export module fmt;
+
+// Re-export every fmt:: declaration that ScyllaDB uses.
+export namespace fmt {
+    // core formatting
+    using fmt::format;
+    using fmt::format_to;
+    using fmt::print;
+    using fmt::to_string;
+    using fmt::join;
+    using fmt::ptr;
+    using fmt::runtime;
+    using fmt::underlying;
+    using fmt::streamed;
+    using fmt::gmtime;
...
```

I expect (but don't know for sure) that exporting using using statements is more forgiving wrt multipath include/import.

avikivity:
I was able to work around the problem by adjusting both the library and the application that consumed it to switch to consuming fmt as a module simultaeously.

I still think this is better in than out. It doesn't pose any maintenance burden, and might save the next one to switch to modules using fmt some pain.

vitaut:
Thanks for the update. I am going to close it for now - we can always revisit this workaround if the issue turns out to be more common.

PR #4840, opened 2026-07-11
Negative integral to char overflow fix

### LINKED ISSUE #4839: :c may overflow with integral values
## Prerequisite

char is signed

 

## Description

Using `fmt::format("{:c}", int{-104})")` ends up with overflowing instead of casting to char.

 

[repro on godbolt](https://www.godbolt.org/z/oeePGoaG1).

 

## Expected Behavior:

c: Copies the character static_cast<CharT>(value) to the output, where CharT is the character type of the format string. Throws std::format_error if value is not in the range of representable values for CharT ([cppref](https://en.cppreference.com/cpp/utility/format/spec)).

dxchel:
Closes https://github.com/fmtlib/fmt/issues/4839
Fixed the negative integral overflow issue and added a test to check for 'x' in two's complement


vitaut:
Closing in favor of https://github.com/fmtlib/fmt/commit/7852fc384ce4232284fc685f66f70ed99b0c2cd4 but thanks for the PR.

PR #4844, opened 2026-07-12
Format nested exceptions in std::exception formatter

avikivity:
Detect when a std::exception also derives from std::nested_exception (e.g. via std::throw_with_nested) and unwind the whole chain, joining each level with ": ". The type-name flag ({:t}) is honored at every level, and the std::exception_ptr formatter inherits this behavior.


vitaut (approved):

vitaut:
Thank you!

PR #4845, opened 2026-07-12
Add a note on value category of static_named_arg

user202729:
Add an explanation on the value category of `static_named_arg::value`, which turns out to be quite subtle. In particular, I don't think the mutable lvalue case is intended.

I think most user shouldn't use things in `detail`, since they may break at any time.
The motivation behind this write-up is I demonstrate how `static_named_arg` can be used to simulate [named parameters in C++](https://stackoverflow.com/a/79848344/5267751), and while doing that, I notice that perfect forwarding is not satisfied.

This newly added comment block will not appear in the HTML. (It's possible to, but currently `cxx/__init__.py` does not handle HTML table, I believe. And this lives in `detail` namespace anyway, probably it shouldn't be.)

That said, `named_arg` is public API, and kind of have the same issue.

I suggest changing `const T&` to `T&` so that the idiomatic `std::forward<T>(value)` can be used instead of `const_cast<T&&>(value)`. Note that it cannot be `T&&`, since a rvalue reference is not copy-constructible, so unfortunately you need to deduce the `T` instead of doing `std::forward<decltype(value)>(value)`.

(Idea about how to deal with clang-format? It's messing up the table. Line wrapping for other parts is fine.)

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -3750,6 +3750,46 @@ FMT_CONSTEXPR void handle_dynamic_spec(
 }
 
 #if FMT_USE_NONTYPE_TEMPLATE_ARGS
+/**
+ * `fmt::detail::static_named_arg<...>` is the type of expressions like
+ * `"argname"_a = value` when `FMT_USE_NONTYPE_TEMPLATE_ARGS` is defined.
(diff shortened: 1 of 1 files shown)


vitaut:
Thanks for the PR but these are internal APIs that shouldn't be used directly by users.

Issue #4695, opened 2026-03-02
libfmt-module.a build with g++15 on OSX  generate unresolved symbols


ClausKlein:
```bash
bash-5.3$ CXX=g++-15 cmake -S . -B build -G Ninja --fresh --log-level=VERBOSE
-- The CXX compiler identification is GNU 15.2.0
-- Checking whether CXX compiler has -isysroot
-- Checking whether CXX compiler has -isysroot - yes
-- Checking whether CXX compiler supports OSX deployment target flag
-- Checking whether CXX compiler supports OSX deployment target flag - yes
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/local/bin/g++-15 - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Configuring done (1.2s)
-- Generating done (0.0s)
-- Build files have been written to: /Users/clausklein/Workspace/cpp/cxx20/fmt-module/tests/build
bash-5.3$ ninja -C build
ninja: Entering directory `build'
[6/6] Linking CXX executable header-only-test
FAILED: [code=1] header-only-test 
: && /usr/local/bin/g++-15  -Wl,-search_paths_first -Wl,-headerpad_max_install_names CMakeFiles/header-only-test.dir/header-only-test.cc.o -o header-only-test  /Users/clausklein/.local/lib/libfmt-module.a && :
Undefined symbols for architecture x86_64:
  "fmt::v12::vprint(fmt::v12::basic_string_view<char>, fmt::v12::basic_format_args<fmt::v12::context>)", referenced from:
      void fmt::v12::print<>(fmt::v12::fstring<>::t) in header-only-test.cc.o
ld: symbol(s) not found for architecture x86_64
collect2: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.
bash-5.3$ 
```

my test file:
```c++
// Header-only configuration test

#ifdef FMT_MODULE
#  ifdef FMT_IMPORT_STD
import std;
import std.compat;
#  else
#    include <string_view>
#  endif
import fmt;
#else
#  include "fmt/base.h"
#  include "fmt/ostream.h"
#endif

// NOLINTNEXTLINE(bugprone-exception-escape)
auto main() -> int
{
  constexpr std::string_view text {"constexpr"};
  fmt::print(text);
}
```

vitaut:
Is `header-only-test` your test of fmt's test?

[...]

PR #4848, opened 2026-07-13
Prepare use of import std;

ClausKlein:
## Tested only on OSX with g++-16 and CMake v4.4

use GNUmakefile as usage hint.

This would fix following issues:

- [x]  https://github.com/fmtlib/fmt/issues/4695
- [x] https://github.com/fmtlib/fmt/issues/4858
- [x] https://github.com/fmtlib/fmt/issues/4849
- [x] https://github.com/fmtlib/fmt/issues/4237


ClausKlein:
@vitaut is clang++-3.6 really needed on CI?

.github/workflows/linux.yml
```diff
@@ -177,9 +177,19 @@ jobs:
...
       run: |
+        case "$CXX" in
+          g++-*)
+            export CC="${CXX/g++/gcc}"
+            ;;
+          clang++-*)
+            export CC="${CXX/clang++/clang}"
+            ;;
+        esac
+        echo "CC=$CC"
```
ClausKlein, line 191: CI used the default $CC compiler until now?

.github/workflows/linux.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ${{ matrix.os || 'ubuntu-22.04' }}
     strategy:
       matrix:
-        cxx: [g++-4.9, g++-11, clang++-3.6, clang++-11]
+        cxx: [g++-4.9, g++-11, clang++-11]
```
ClausKlein, line 18: with `clang-11` compiler fails the ci build job!

test/find-package-test/CMakeLists.txt
```diff
@@ -1,8 +1,9 @@
 cmake_minimum_required(VERSION 3.8...4.4)
 
-project(fmt-test)
+project(fmt-test CXX)
 
-find_package(FMT REQUIRED)
```
ClausKlein, line 5: ### how dit this work?
the config file is: fmt-config.cmake

see https://github.com/fmtlib/fmt/issues/4849

vitaut:
Thanks, but this PR mixes several unrelated changes and breaks supported configurations, including older CMake versions. It also removes required component validation and does not add CI coverage for the new mode.

PR #4850, opened 2026-07-15
Generate CMake export set regardless of FMT_INSTALL

### LINKED ISSUE #4806: `export(` should not be scoped by `FMT_INSTALL`

hexonal:
Closes #4806

`export(TARGETS ...)` is currently scoped inside `if (FMT_INSTALL)`, so a project that pulls in fmt via `add_subdirectory()`/`FetchContent` without also installing it never gets the `fmt::*` import targets in its build tree. If that project then tries to `export()` one of its own targets that depends on fmt, CMake fails with:

```
CMake Error: export("MyLibTargets" ...) includes target "mylib" which requires target "fmt" that is not in any export set.
```

This moves the `export()` call, along with the `INSTALL_TARGETS`/`targets_export_name` it needs, outside the `FMT_INSTALL` guard so it always runs. Everything that's actually install-specific (config/version files, pkg-config, `install(EXPORT ...)`) stays behind the guard, so installed behavior is unchanged — confirmed that `fmt-config.cmake`, `fmt-config-version.cmake` and `fmt.pc` are still only produced with `FMT_INSTALL=ON`.

Extended `test/add-subdirectory-test` with a small static library that publicly links `fmt::fmt` and exports itself. This reproduces the exact failure from the issue when `FMT_INSTALL` is off (the default when fmt is a subproject) and passes with this fix.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -441,6 +441,16 @@ add_library(fmt::fmt-c ALIAS fmt-c)
 set_target_properties(fmt-c PROPERTIES PUBLIC_HEADER include/fmt/fmt-c.h)
 
 # Install targets.
+set(targets_export_name fmt-targets)
+set(INSTALL_TARGETS fmt fmt-header-only fmt-c)
+
+if (FMT_MODULE)
+  list(APPEND INSTALL_TARGETS fmt-module)
+  if (FMT_USE_CMAKE_MODULES)
+    set(INSTALL_FILE_SET FILE_SET fmt DESTINATION ${FMT_INC_DIR}/fmt)
+  endif ()
+endif ()
+
 if (FMT_INSTALL)
   include(CMakePackageConfigHelpers)
   set_verbose(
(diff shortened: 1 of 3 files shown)


vitaut (approved):

vitaut:
Thank you!

PR #4854, opened 2026-07-16
add support for more primitive types in scan.h

KareemOtoum:


vitaut:
Thanks for the PR but this is just a test/prototype that doesn't need to be extended.

KareemOtoum:
hi thanks for the feedback, is there anything that needs work that you can point me to?

vitaut:
You could check issues with the "help wanted" tag.

PR #4855, opened 2026-07-16
Libcxx transitive include change fix

mccakit:
LLVM 23 depracated libc++ transitive includes, as so <atomic> needs to be included directly at fmt module interface file rather than coming from a another libc++ header.


vitaut (changes requested):
Looks good but please fix the lint.

Issue #4852, opened 2026-07-15
Should the "Formatting User-Defined Types" API doc mention the potential interaction of format_as and e.g. ranges.h?


rimathia:
Feel free to close the issue if this has already been discussed.

The case where defining `format_as` for a user-defined type conflicts with a formatting coming for example from ranges.h and the type having a begin() and end() function is probably rare, but at least for me it was difficult to diagnose (it didn't help that I didn't see the inclusion of ranges.h because it was transitive).

Might there be a short note in the API doc as a hint? My first draft:
Note that `format_as` is not suitable for user-defined types which are already formattable due to one of the general implementations (e.g. as provided in ranges.h). Use `formatter` specializations for these cases.

vitaut:
`format_as` is a way to provide a `formatter` specialization, so the proposed change is not quite right but I'm open to a doc PR with better suggestions. The opt out mechanism for ranges that does address it has already been discussed elsewhere.

PR #4857, opened 2026-07-17
doc extension concerning format_as and general implementations

rimathia:
add a sentence about the possible interaction of format_as and general implementations


rimathia:
Follow up on https://github.com/fmtlib/fmt/issues/4852.
My point here is only to have a hint somewhere close to format_as about the possible interaction with general implementations. I'm not very confident I've found the best wording, I trust that you will adjust as you see fit.

vitaut (changes requested):
I think something like this would be more precise:

`format_as` cannot be used when a type also matches another `formatter` specialization, such as the range formatter, because the specializations would be ambiguous. Disable the conflicting specialization, if possible, or provide an explicit `formatter` specialization instead.

vitaut (approved):

vitaut:
Merged, thanks

PR #4860, opened 2026-07-20
Support width and alignment in std::exception formatter

s3yato:
The formatter for std::exception (and std::exception_ptr) previously accepted only the optional 't' type-name specifier, so standard fill, align and width specifiers were rejected with "unknown format specifier". This made it impossible to pad or align exception messages, e.g. when laying them out in a log column.

Parse the standard fill/align/width specifiers (as the std::filesystem:: path formatter already does) and apply them when writing the message. Dynamic width ({:{}}) and the 'none'/'unknown exception' cases for exception_ptr are handled too. Existing behavior ({}, {:t}, nested exception unwinding) is unchanged.

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -320,7 +320,7 @@ template <typename Char> struct formatter<std::filesystem::path, Char> {
     if (it == end) return it;
 
     Char c = *it;
-    if ((c >= '0' && c <= '9') || c == '{')
+    if ((c >= '1' && c <= '9') || c == '{')
       it = detail::parse_width(it, end, specs_, width_ref_, ctx);
     if (it != end && *it == '?') {
       debug_ = true;
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -620,6 +620,8 @@ struct formatter<
     T, char,
     typename std::enable_if<std::is_base_of<std::exception, T>::value>::type> {
  private:
+  format_specs specs_;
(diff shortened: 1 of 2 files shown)


vitaut (changes requested):
Thanks for the PR. Mostly looks good but let's avoid writing to an intermediate buffer for the common case of no width.

s3yato:
Good catch — added a fast path that writes directly to the output when no width is requested; only the padded case allocates an intermediate buffer now. Dynamic width ({:{}}) still goes through the buffered path. Thanks!

include/fmt/std.h
```diff
@@ -620,14 +620,23 @@ struct formatter<
...
     auto it = ctx.begin();
     auto end = ctx.end();
     if (it == end || *it == '}') return it;
-    if (*it == 't') {
+
+    it = detail::parse_align(it, end, specs_);
+    if (it == end) return it;
+
+    char c = *it;
+    if ((c >= '0' && c <= '9') || c == '{')
```
vitaut, line 637: One more thing: zero shouldn't be allowed to start a width, see e.g. https://godbolt.org/z/jxfjMhaT8

vitaut (changes requested):

s3yato (commented):
Fix

vitaut (approved):

vitaut:
Merged, thanks!

PR #4861, opened 2026-07-20
Export ostream formatters and streamed() from the module

avikivity:
Make basic_ostream_formatter, ostream_formatter and streamed() accessible to module users.

### DIFF
--- a/include/fmt/ostream.h
+++ b/include/fmt/ostream.h
@@ -70,6 +70,8 @@ template <typename T> struct streamed_view {
 };
 }  // namespace detail
 
+FMT_BEGIN_EXPORT
+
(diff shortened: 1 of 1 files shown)


vitaut (changes requested):
Thanks for the PR! Exporting makes sense but I think we should wrap the whole public API block in `FMT_BEGIN_EXPORT` / `FMT_END_EXPORT` instead of doing it one declaration by declaration.

avikivity:
v2: use FMT_BEGIN_EXPORT, drop now-unneeded FMT_EXPORT

vitaut (approved):

vitaut:
Lint issue is preexisting - will fix it separately.

vitaut:
Merged, thanks!

PR #4865, opened 2026-07-23
Fix inconsistent display width for emoji outside East Asian Wide ranges

### LINKED ISSUE #4851: Inconsistent widths when UTF8 characters are used
Hi,

I find the widths are inconsistent when UTF8 characters are used. I'm not sure whether it's caused by the character display of the terminal or the library implementation.

Example:

```cpp
#include <fmt/format.h>

auto main() -> int {
    fmt::println("|{:^20}|", 12345);
    fmt::println("|{:^20}|", "normal string");
    fmt::println("|{:^20}|", "❌");
    fmt::println("|{:^20}|", "✅");
    fmt::println("|{:^20}|", "Müller");
    fmt::println("|{:^20}|", "我");
    return 0;
}
```
Printouts:

```text
|       12345        |
|   normal string    |
|         ❌          |
|         ✅          |
|       Müller       |
|         我         |
```

[Godbolt link](https://compiler-explorer.com/z/cEWdv8Eob)

edugomz:
Fixes #4851

## Problem

- `display_width_of()` only counted East Asian Wide/Fullwidth code points and two hand-picked emoji ranges (U+1F300-1F64F, U+1F900-1F9FF) as 2 columns wide
- Emoji outside those ranges (e.g. ❌ U+274C, ✅ U+2705, in the Dingbats block) were measured as 1 column, even though most terminals render them double-width
- Result: inconsistent `{:^N}` padding vs. CJK text

```cpp
fmt::println("|{:^20}|", "❌");
fmt::println("|{:^20}|", "我");
```
```
|         ❌          |
|         我         |
```

## Fix

- Replaced the boolean-expression range check with a sorted range table, looked up via binary search
- Table = original East Asian Wide/Fullwidth blocks + full `Emoji_Presentation` set from Unicode's `emoji-data.txt` (fetched from unicode.org, not hand-transcribed)

## Testing

Added regression tests (`test/format-test.cc`) for:
- exact example from #4851
- multiple emoji back to back
- mixed emoji/CJK/ASCII content
- precision truncation at a display-width boundary
- alignment/fill variants
- newly-covered ranges (Transport, Symbols Extended-A)
- regional indicator (flag) pairs — documents current per-code-point behavior since fmt has no grapheme clustering

Full suite passes: 147/147 `format-test`, 21/21 `ctest` binaries.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -655,26 +655,92 @@ FMT_CONSTEXPR void for_each_codepoint(string_view s, F f) {
   } while (buf_ptr < buf + num_chars_left);
 }
 
+struct wide_cp_range {
+  uint32_t first;
+  uint32_t last;
+};
+
+// Code points with display width 2: East Asian Wide/Fullwidth
+// (https://www.unicode.org/reports/tr11/) and Emoji_Presentation
+// (https://www.unicode.org/reports/tr51/) ranges, sorted and merged.
+inline constexpr wide_cp_range wide_cp_ranges[] = {
+    {0x1100, 0x115f},    // Hangul Jamo init. consonants
+    {0x231a, 0x231b},    // Watch, hourglass done
+    {0x2329, 0x232a},    // Angle brackets
+    {0x23e9, 0x23ec},    // Fast-forward/-backward, next/last track buttons
+    {0x23f0, 0x23f0},    // Alarm clock
+    {0x23f3, 0x23f3},    // Hourglass not done
+    {0x25fd, 0x25fe},    // White/black medium-small square
+    {0x2614, 0x2615},    // Umbrella with rain drops, hot beverage
+    {0x2648, 0x2653},    // Zodiac signs
+    {0x267f, 0x267f},    // Wheelchair symbol
+    {0x2693, 0x2693},    // Anchor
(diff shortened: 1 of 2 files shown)


vitaut:
To fix timezone issues, please rebase.

vitaut (changes requested):
Thanks for the PR. I think this needs a couple of changes before merging:

* Regional indicators shouldn’t be treated as width 2 individually; that makes flags like 🇺🇸 width 4 instead of 2.
* `display_width_of` should remain `FMT_CONSTEXPR`; the current change appears to drop constexpr support.
* I think we should update the existing table from the current East Asian Width data rather than add `Emoji_Presentation`. The [standard width rules](https://eel.is/c++draft/format.string.std) are based on `East_Asian_Width=W/F` (plus a few explicit ranges), and `Emoji_Presentation` includes regional indicators that shouldn’t individually have width 2.

vitaut:
@edugomz do you still plan to update this PR or should we close it for now?

vitaut:
Closing for now but feel free to reopen with feedback addressed.

edugomz:
Hey, sorry I've been out for a while, I will update the PR soon

PR #4875, opened 2026-08-01
Ignore run-msbuild.bat

illusionyy:
Only for visual studio generator, when adding as subdir, this file is created and breaks submodule tracking until this file is ignored.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -231,7 +231,7 @@ if (FMT_MASTER_PROJECT AND CMAKE_GENERATOR MATCHES "Visual Studio")
   join(netfxpath
        "C:\\Program Files\\Reference Assemblies\\Microsoft\\Framework\\"
        ".NETFramework\\v4.0")
-  file(WRITE run-msbuild.bat "${MSBUILD_SETUP}
+  file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/run-msbuild.bat "${MSBUILD_SETUP}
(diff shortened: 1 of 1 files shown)


vitaut (changes requested):
Thanks for the PR.

I think it would be better to fix the source of the problem rather than ignore the generated file. `run-msbuild.bat` is created by `file(WRITE run-msbuild.bat ...)`, which writes it into the source directory. I suggest writing it to `${CMAKE_CURRENT_BINARY_DIR}` instead.

Also, could you elaborate on what you mean by "breaks submodule tracking"?

illusionyy:
> Also, could you elaborate on what you mean by "breaks submodule tracking"?

When I add this as submodule, cmake generates this file and diff will show in git.
```diff
diff --git a/external/formatter/fmt/fmtlib b/external/formatter/fmt/fmtlib
--- a/external/formatter/fmt/fmtlib
+++ b/external/formatter/fmt/fmtlib
@@ -1 +1 @@
-Subproject commit abb66c1f3576489ac3bdcee47c19a3fba6dddb07
+Subproject commit abb66c1f3576489ac3bdcee47c19a3fba6dddb07-dirty
```

illusionyy:
> I suggest writing it to `${CMAKE_CURRENT_BINARY_DIR}` instead.

This is a better fix and does work. I updated PR with this change

vitaut (approved):

vitaut:
Merged, thanks!

PR #4876, opened 2026-08-02
Avoid -Wdangling-reference false positives in locale helpers

viordash:
GCC 13/14 warns in two locale helpers when a precompiled header is used:

    include/fmt/format-inl.h:132: warning: possibly dangling reference to a temporary
      [-Wdangling-reference]
    include/fmt/xchar.h:49: same

It is a false positive: `locale_ref::get<std::locale>()` returns a copy, and the
reference returned by `std::use_facet` stays valid as long as any copy of that locale
exists. GCC only warns when the declaration comes from a precompiled header; with
`<locale>` included normally it stays quiet. Reproducible with plain `<locale>`,
no fmt involved:

    // t.cpp
    #include <locale>
    template <typename Char> auto f(const std::locale& l) -> Char {
      auto&& facet = std::use_facet<std::numpunct<Char>>(std::locale(l));
      return facet.decimal_point();
    }
    template char f<char>(const std::locale&);

    // pch.h
    #include <locale>

    g++ -Wdangling-reference -c t.cpp                    # quiet
    g++ -x c++-header pch.h -o pch.h.gch
    g++ -Wdangling-reference -include pch.h -c t.cpp     # warns

    (GCC 13.4; -Wdangling-reference comes with -Wextra there)

This patch binds the locale copy to a named variable at both call sites, so no reference
is bound to a temporary. Same single copy, no behaviour change, and the lifetime becomes
explicit instead of relying on the "valid while any copy exists" guarantee.

Checked with GCC 13.4: warnings gone when a PCH is in use, ctest passes 21/21, and
locale-aware output is unchanged

Originally filed against spdlog, which vendors these files (gabime/spdlog#3638), and
closed there with a pointer to this repository.

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -129,7 +129,8 @@ inline void fwrite_all(const void* ptr, size_t count, FILE* stream) {
 
 template <typename Char>
 FMT_FUNC auto thousands_sep_impl(locale_ref loc) -> thousands_sep_result<Char> {
-  auto&& facet = use_facet<numpunct<Char>>(loc.get<locale>());
+  auto loc_copy = loc.get<locale>();
+  auto&& facet = use_facet<numpunct<Char>>(loc_copy);
   auto grouping = facet.grouping();
   auto thousands_sep = grouping.empty() ? Char() : facet.thousands_sep();
   return {std::move(grouping), thousands_sep};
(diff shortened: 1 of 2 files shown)


vitaut:
Could you provide a godbolt CMake-project (since it requires PCH) repro?

viordash:
Here is a CMake project repro: https://godbolt.org/z/vfcfWThT7

    /opt/compiler-explorer/libs/fmt/trunk/include/fmt/format-inl.h:132:10: warning:
      possibly dangling reference to a temporary [-Wdangling-reference]

Removing the target_precompile_headers() line makes it quiet.

vitaut (approved):
Thanks for the repro. The workaround LGTM.

PR #4877, opened 2026-08-02
revert clang tidy suppression

### LINKED ISSUE #4871: Commit removing a clang-tidy false positive also introduces one
The following simple program:

```c++
#include <fmt/compile.h>

#include <iostream>

int main() {
    constexpr auto fmtstr = FMT_COMPILE(std::string_view("Hello {}"));
    constexpr auto a = 42;
    constexpr auto sz = fmt::formatted_size(fmtstr, a);

    std::cout << sz << "\n";
}
```

when run under clang-tidy (v19) gives the following error message:

`error: constexpr variable 'sz' must be initialized by a constant expression [clang-diagnostic-error]`

when compiled with clang++ there are no errors.

Reverting [this fmt commit](https://github.com/fmtlib/fmt/commit/619b3a5aa031a25e8d06fec1e03992e5e727c5a5) ("Suppress a false positive in clang-tidy") causes the clang-tidy error to go away.

Is it possible this change needs tweaking?

aouxwoux:
## what changed

revert commit 619b3a5, which added the FMT_CLANG_ANALYZER special cases to the compile-string macro and apple file buffer initialization.

## why

issue #4871 shows that the suppression removes one clang-tidy false positive but introduces another for FMT_COMPILE and ormatted_size. reverting the suppression restores the original code paths and matches the maintainer's suggested fix.

## checks

- git diff --check passes.
- the change is an exact reverse of 619b3a5 with no unrelated edits.
- this checkout does not have cmake or a c++ compiler installed, so i could not run the local test suite.

fixes #4871

### DIFF
--- a/include/fmt/format-inl.h
+++ b/include/fmt/format-inl.h
@@ -1599,7 +1599,7 @@ template <typename F> class apple_file : public file_base<F> {
   void init_buffer() {
     if (this->file_->_p) return;
     // Force buffer initialization by placing and removing a char in a buffer.
-    if (!FMT_CLANG_ANALYZER) putc_unlocked(0, this->file_);
+    putc_unlocked(0, this->file_);
     --this->file_->_p;
     ++this->file_->_w;
   }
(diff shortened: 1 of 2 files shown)


vitaut (approved):

vitaut:
Merged, thanks

PR #4879, opened 2026-08-03
Revert "Suppress a false positive in clang-tidy"

jefftrull:
This reverts commit 619b3a5aa031a25e8d06fec1e03992e5e727c5a5.

The original commit may have addressed a clang-tidy false positive, but it also introduced one, as described in #4871.
Merging this PR will fix that issue.

Builds, passes all tests, produces the expected output in gcc and clang, and does not error in clang-tidy.


vitaut:
This is already reverted in #4877 but thanks anyway.

PR #4882, opened 2026-08-09
Cleanup unusual MSVC warnings.

IRainman:
Small cleanup because many places in code produce unusual warnings.

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -188,8 +188,15 @@
 #  define FMT_FALLTHROUGH
 #endif
 
-// Disable [[noreturn]] on MSVC/NVCC because of bogus unreachable code warnings.
-#if FMT_HAS_CPP_ATTRIBUTE(noreturn) && !FMT_MSC_VERSION && !defined(__NVCC__)
+#if FMT_MSC_VERSION || defined(__NVCC__)
+// Disable unusual warning on MSVC/NVCC because of bogus unreachable code
+// warnings in many places.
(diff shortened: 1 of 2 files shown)


include/fmt/base.h
```diff
@@ -188,8 +188,15 @@
 #  define FMT_FALLTHROUGH
 #endif
 
-// Disable [[noreturn]] on MSVC/NVCC because of bogus unreachable code warnings.
-#if FMT_HAS_CPP_ATTRIBUTE(noreturn) && !FMT_MSC_VERSION && !defined(__NVCC__)
+#if FMT_MSC_VERSION || defined(__NVCC__)
+// Disable unusual warning on MSVC/NVCC because of bogus unreachable code
+// warnings in many places.
+#  pragma warning(disable : 4702)
```
vitaut, line 194: This won't have any effect on `format-inl.h`, let's keep the local workaround as before.

include/fmt/base.h
```diff
@@ -188,8 +188,15 @@
...
+// Disable because don't need to check
+#  pragma warning(disable : 26495)
```
vitaut, line 196: What's the motivation for this change?

vitaut (changes requested):

include/fmt/base.h
IRainman: Ok, this is not working properly, as FMT_PRAGMA_MSVC too. I close the request.

PR #4884, opened 2026-08-11
fix(windows): UTF-8 for std::error_code {:s} (#4436)

### LINKED ISSUE #4436: error code `:s` fail on windows due to encoding error

hannahgreendesk:
## Summary

Fixes #4436.

On Windows, `std::system_category().message()` is often encoded in the ANSI code page (e.g. GBK), not UTF-8 (LWG 4156). Formatting that string with `{:s}` then fails with `invalid utf8`.

For `std::error_code` with the `s` presentation type on Windows:
1. Prefer `FormatMessageW` + UTF-8 conversion (same approach as `fmt::system_category()`).
2. If that fails, fall back to ACP → UTF-8 for `ec.message()`.
3. Non-system categories keep the existing `ec.message()` path.

No `windows.h` include in `fmt/std.h` — only the few Win32 imports needed.

## Test plan

- [ ] Existing `std_test.error_code` still passes
- [ ] New Windows check: `fmt::format("{:s}", std::error_code(3, std::system_category()))` does not throw
- [ ] Non-Windows builds unchanged

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,9 @@
 # 12.2.1 - TBD
 
+- Fixed `std::error_code` `{:s}` formatting on Windows when
+  `std::system_category().message()` returns a non-UTF-8 (ACP) string, which
+  previously could throw `invalid utf8` (https://github.com/fmtlib/fmt/issues/4436).
+
 - Added a formatter for `std::exception_ptr` that formats the referenced
   exception (or `none` when null)
   (https://github.com/fmtlib/fmt/issues/4808,
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -546,6 +546,90 @@ struct formatter<Variant, Char,
 
 #endif  // FMT_CPP_LIB_VARIANT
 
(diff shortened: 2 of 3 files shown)


vitaut:
Thanks for the PR. Using ACP here is not standard compliant - `error_category::message` should be using execution character set (basically C locale encoding) per https://eel.is/c++draft/syserr.errcat.virtuals, please report to Microsoft/STL.

hannahgreendesk:
Thanks @vitaut — that makes sense. I hadn't clocked that `error_category::message` is required to use the execution character set, so converting ACP→UTF-8 inside fmt isn't the right layer.

I'll close this PR and follow up with Microsoft/STL about the UTF-8-ness of `std::system_category().message()` on Windows instead.

PR #4885, opened 2026-08-13
Format enums annotated with fmt::as_identifiers (requires C++26 reflection)

avikivity:
Format an enum as the identifier of the matching enumerator if the enum is
annotated with fmt::as_identifiers:

  enum class [[=fmt::as_identifiers]] color { red, green, blue };
  fmt::format("{}", color::green);  // "green"

Values that don't match any enumerator are represented as their underlying
value in decimal before applying string formatting.

Identifiers are retrieved via C++26 reflection (P2996) and the annotation
via P3394. FMT_USE_REFLECTION is autodetected and can be overridden by the
user; without reflection the header is empty.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -9,6 +9,7 @@ The {fmt} library API consists of the following components:
 - [`fmt/ranges.h`](#ranges-api): formatting of ranges and tuples
 - [`fmt/chrono.h`](#chrono-api): date and time formatting
 - [`fmt/std.h`](#std-api): formatters for standard library types
+- [`fmt/enum.h`](#enum-api): formatting of annotated enums
 - [`fmt/compile.h`](#compile-api): format string compilation
 - [`fmt/color.h`](#color-api): terminal colors and text styles
 - [`fmt/os.h`](#os-api): system APIs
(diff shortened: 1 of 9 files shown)


avikivity:
v2: apply cmake-format

avikivity:
@vitaut please review

doc/api.md
```diff
@@ -556,6 +557,46 @@ fmt::print("{}", +s.bit);
...
+
+    fmt::print("[{:>7}]", color::red);
+    // Output: [    red]
+
+Identifiers are only available as `char` strings so annotated enums are not
+formattable with other character types.
+
+If several enumerators have the same value, the first one in the order of
+declaration is used. A value that doesn't match any enumerator is formatted as
+the corresponding value of the underlying type:
```
vitaut: "formatted as underlying type" is not quite correct because it is still uses string formatter applied to the decimal representation represented as a string which is different from using a formatter for the underlying type, e.g. you cannot use integer format specs. I suggest rephrasing it as something like

> unknown values are represented as their underlying value in decimal before applying string formatting

include/fmt/reflect.h
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * An annotation that makes an enum format as identifiers of its enumerators.
+ *
+ * **Example**:
+ *
+ *     enum class [[=fmt::as_identifiers]] color { red, green, blue };
+ *     auto s = fmt::format("{}", color::green);  // s == "green"
+ *
+ * A value that doesn't match any enumerator is formatted as the corresponding
+ * value of the underlying type.
```
vitaut: same issue with underlying here

vitaut (changes requested):
This looks great, very cool use of C++26 annotations for enum formatting. Thanks for putting this together!

One thing I noticed: it looks like the reflection tests are currently skipped everywhere because none of the CI compilers support reflection. It would be good to have at least one GCC 16 + -freflection job that actually builds and runs this code.

include/fmt/enum.h
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * **Example**:
+ *
+ *     enum class [[=fmt::as_identifiers]] color { red, green, blue };
+ *     auto s = fmt::format("{}", color::green);  // s == "green"
+ *
+ * A value that doesn't match any enumerator is formatted as the corresponding
+ * value of the underlying type.
+ */
+inline constexpr auto as_identifiers = as_identifiers_t();
```
vitaut: I wonder if this should be singular `as_identifier` because the result is a single identifier?

avikivity:
Update:
- renamed reflect.h to enum.h and made it more about enums rather than reflection
- improved wording around how enum values that don't match any identifier are formatted
- added gcc 16 to CI

doc/api.md
avikivity: Updated

include/fmt/reflect.h
avikivity: Updated

avikivity:
> This looks great, very cool use of C++26 annotations for enum formatting. Thanks for putting this together!
> 
> One thing I noticed: it looks like the reflection tests are currently skipped everywhere because none of the CI compilers support reflection. It would be good to have at least one GCC 16 + -freflection job that actually builds and runs this code.

Done.

avikivity:
Update: support C++20 modules too

include/fmt/enum.h
```diff
@@ -0,0 +1,121 @@
...
+}
+
+template <typename E, size_t N = count_enumerators<E>()>
+consteval auto make_identifiers() -> std::array<std::pair<E, string_view>, N> {
+  auto ids = std::array<std::pair<E, string_view>, N>();
+  auto i = size_t();
+  for (std::meta::info e : std::meta::enumerators_of(^^E)) {
+    auto id = std::meta::identifier_of(e);
+    ids[i++] = {std::meta::extract<E>(e),
+                string_view(std::define_static_string(id), id.size())};
```
vitaut: I don't think `define_static_string` is necessary. `identifier_of` should already provide the required lifetime guarantee.

vitaut (changes requested):
Mostly looks good but I think unknown values of `enum : char` will be formatted as `char` rather than decimal and also `defined_static_string` seems unnecessary.

include/fmt/enum.h
avikivity: I read it as "format this enum type as its identifiers" rather than "format this enum value as the corresponding identifier", since the type is annotated.

include/fmt/enum.h
avikivity: Right.

avikivity:
> Mostly looks good but I think unknown values of `enum : char` will be formatted as `char` rather than decimal 

Good catch, fixed.

> and also `defined_static_string` seems unnecessary.

Yes

avikivity:
Update:
 - removed unneeded define_static_string
 - fixed formatting of enum class with underlying type of char when no identifier is found
 - optimized the case where the enum is contiguous or near-contiguous to use array indexing rather than linear search (new commit)
 
Unchanged:
 - still using fmt::as_identifiers, happy to adjust if needed

vitaut (approved):

vitaut:
Merged, thanks!

PR #4887, opened 2026-08-16
Add pkgconfig file for fmt-c library

Ferdi265:
This PR adds a pkgconfig file for the `fmt-c` library, allowing the C API to be found and linked via pkgconfig. The CMake package already exports `fmt::fmt-c`, so that doesn't need changes.

The only thing I was unsure of was if `fmt-c.pc` should explicitly link `-lfmt` as well, or if the implicit transitive linkage is enough. For now I have added it, since the `fmt::fmt-c` CMake target explicitly lists `fmt::fmt` as a linked library as well, but it can be removed if not needed.

Also, currently the `fmt::fmt-c` target is part of the `fmt_core` CMake package component, so I added the `fmt_c.pc` file to `fmt_core` as well. Maybe that could be changed to a `fmt_c` or `fmt_c_api` component.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -422,6 +422,14 @@ setup_target(fmt-header-only INTERFACE)
 
 add_library(fmt-c src/fmt-c.cc)
 target_compile_features(fmt-c INTERFACE c_std_11)
+
+# Set FMT_C_LIB_NAME for pkg-config fmt-c.pc. We cannot use the OUTPUT_NAME
+# target property because it's not set by default.
+set(FMT_C_LIB_NAME fmt-c)
+if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+  set(FMT_C_LIB_NAME ${FMT_C_LIB_NAME}${FMT_DEBUG_POSTFIX})
+endif ()
+
 if (BUILD_SHARED_LIBS)
   target_compile_definitions(
     fmt-c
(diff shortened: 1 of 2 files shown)


vitaut (changes requested):
Thanks for the PR. 

fmt is an implementation dependency of fmt-c, so I think this should use Requires.private: fmt = @FMT_VERSION@ rather than putting -lfmt in Libs.

Ferdi265:
> Thanks for the PR.
> 
> fmt is an implementation dependency of fmt-c, so I think this should use Requires.private: fmt = @FMT_VERSION@ rather than putting -lfmt in Libs.

I didn't know about `Requires.private`. Will use that instead, thanks!

Ferdi265:
Added `Requires.private` instead of directly linking `-lfmt`.

CMakeLists.txt
```diff
@@ -520,6 +531,10 @@ if (FMT_INSTALL)
     FILES "${pkgconfig}"
     DESTINATION "${FMT_PKGCONFIG_DIR}"
     COMPONENT fmt_core)
+  install(
+    FILES "${pkgconfig_c}"
+    DESTINATION "${FMT_PKGCONFIG_DIR}"
+    COMPONENT fmt_core)
```
vitaut: Let's merge this into the above `install` to avoid repetition.

Ferdi265: done

vitaut (approved):

vitaut:
Merged, thanks

PR #4888, opened 2026-08-19
Fixed producing unusual warnings from MSVC like compilers, enable support for noreturn for them, and cleanup crutches.

IRainman:
I use my pet project https://github.com/IRainman/Calc with the warning level on MSVC 4+ and it shows many unusual warnings on fmt.

### DIFF
--- a/include/fmt/base.h
+++ b/include/fmt/base.h
@@ -188,14 +188,6 @@
 #  define FMT_FALLTHROUGH
 #endif
 
-#if FMT_MSC_VERSION || defined(__NVCC__)
-// Disable unusual warning on MSVC/NVCC because of bogus unreachable code
-// warnings in many places.
-#  pragma warning(disable : 4702)
-// Disable because don't need to check
-#  pragma warning(disable : 26495)
-#endif
-
 #if FMT_HAS_CPP_ATTRIBUTE(noreturn)
 #  define FMT_NORETURN [[noreturn]]
 #else
@@ -1773,9 +1765,13 @@ template <typename T> class buffer {
   grow_fun grow_;
 
  protected:
+FMT_PRAGMA_MSVC(warning(push))
+FMT_PRAGMA_MSVC(warning(disable : 26495))
   // Don't initialize ptr_ since it is not accessed to save a few cycles.
   FMT_CONSTEXPR buffer(grow_fun grow, size_t sz) noexcept
-      : size_(sz), capacity_(sz), grow_(grow) {}
+      : size_(sz), capacity_(sz), grow_(grow) {
+  }
+FMT_PRAGMA_MSVC(warning(pop))
 
   constexpr buffer(grow_fun grow, T* p = nullptr, size_t sz = 0,
                    size_t cap = 0) noexcept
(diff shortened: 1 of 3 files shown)


include/fmt/base.h
```diff
@@ -1766,11 +1765,13 @@ template <typename T> class buffer {
   grow_fun grow_;
 
  protected:
+FMT_PRAGMA_MSVC(warning(push))
+FMT_PRAGMA_MSVC(warning(disable : 26495))
```
mwinterberg-dtek, line 1769: I don't know what vitaut would prefer here, but since this warning is a coreguideline warning, and since clang-tidy seems to cover it https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/pro-type-member-init.html, it would seem preferable to use the gsl suppression syntax if it works with clang-tidy and msvc and also doesn't cause other problems.

IRainman: Hmm, that's reasonable, but FMT_PRAGMA_MSVC already exists in the project, and this warning is related only to MSVC-like compilers. Moreover, removing crutches from the library would also be nice, and it's also related only to MSVC. Neither GCC nor Clang produced those warnings.

vitaut (changes requested):
Please provide github repros of the actual warnings that you are trying to suppress.

include/fmt/base.h
vitaut: Could you give an example how such suppression would look like?

mwinterb: Like with @IRainman , I wasn't able to get clang-tidy to issue the warning in this particular example. Which is weird, since clang-analyzer-optin.cplusplus.UninitializedObject seemingly should do the same thing and does report the issue. But the guidelines suppression mechanism doesn't suppress the UninitializedObject warning.

Seemingly even using an example from the docs didn't suppress:
https://releases.llvm.org/22.1.0/tools/clang/docs/AttributeReference.html#suppress

https://gcc.godbolt.org/z/9ha5Pdhvq

But, this does work in MSVC. I just don't know if it's fully correct and would work with other(?) Core Guidelines analyzers.

```c++
#if FMT_HAS_CPP_ATTRIBUTE(gsl::suppress)
#  define FMT_GSL_SUPPRESS(x) [[gsl::suppress(x)]]
#else
#  define FMT_GSL_SUPPRESS(x)
#endif
```

```c++
    // Don't initialize ptr_ since it is not accessed to save a few cycles.
    FMT_GSL_SUPPRESS("type.6")
    constexpr buffer(grow_fun grow, size_t sz) noexcept
        : size_(sz), capacity_(sz), grow_(grow) {}
```

IRainman:
@vitaut I use my pet project https://github.com/IRainman/Calc with a fork of fmt and this suppresses warnings in MSVC for level 4 and more.

vitaut:
Thanks for the PR but this doesn't look like an improvement. I suggest suppressing esoteric warnings in your project.

IRainman:
OK, fine, I simply remove these changes and exclude library path from diagnostic completely.

PR #4890, opened 2026-08-20
Fix utf8 display width

Satyakam-Mishra:
## Description

Fixes incorrect display width calculation for UTF-8 characters in formatted
strings.

Characters with East Asian Width properties `W` and `F` are treated as
double-width when calculating display width. The implementation uses a binary
search over a generated Unicode range table.

This fixes alignment issues for characters such as emoji and CJK characters.

Regional Indicator symbols are not included as double-width ranges, so flag
sequences such as 🇺🇸 are handled correctly.

## Testing

Added regression tests covering:

- ❌
- ✅
- 😀
- 🇺🇸
- Müller
- 我

All tests pass:

- 21/21 tests passed
- `git diff --check` passes


vitaut (changes requested):
Please don't add new headers.

Satyakam-Mishra:
Thank you, I have removed the separate unicode_width_table.h header and moved the unicode with table to format.h as suggested. I have also rerun the tests and it still passes. Pushing the update now.

vitaut:
Closing in favor of #4896 but thanks anyway.

PR #4891, opened 2026-08-23
Document C++20 module usage with CMake

### LINKED ISSUE #4237: What is the proper way to use fmt as a C++ 20 module?

way4sahil:
Adds CMake guidance for using {fmt} as a C++20 module.

- Enable `FMT_MODULE` before adding {fmt} as a subdirectory
- Link against `fmt::fmt`
- Use `import fmt;` in consumer source files
- Document the CMake 3.28+ and Ninja requirement for GCC module dependency scanning

Fixes #4237.

### DIFF
--- a/doc/get-started.md
+++ b/doc/get-started.md
@@ -47,6 +47,22 @@ In order to use the header-only target or the module target, simply substitute t
 `fmt::fmt` in the above steps with `fmt::fmt-header-only` or `fmt::fmt-module` 
 accordingly.
 
+### Using the C++20 Module
+
+Link your target to `fmt::fmt-module` and import `fmt` instead of including a
+{fmt} header:
+
+    target_link_libraries(<your-target> PRIVATE fmt::fmt-module)
+
+    import fmt;
+
+    int main() {
+      fmt::print("Hello, world!\\n");
+    }
+
+The module target requires C++20 module support in CMake and the compiler. For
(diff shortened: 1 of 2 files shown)


vitaut (changes requested):
Thanks for the PR.

The documented example can fail because `fmt::fmt-module` is only defined when `FMT_MODULE` is enabled. Could you mention that `FMT_MODULE` needs to be enabled before adding fmt (or otherwise ensure modules are enabled via `CMAKE_CXX_STANDARD`)? This is already mentioned in the PR description but not in the documentation itself.

Also, the GCC/CMake/Ninja versions listed here seem to be requirements for CMake’s native module support rather than for `fmt::fmt-module` in general, since fmt also has the fallback path.

doc/get-started.md
```diff
@@ -47,6 +47,22 @@ In order to use the header-only target or the module target, simply substitute t
...
+
+Link your target to `fmt::fmt-module` and import `fmt` instead of including a
+{fmt} header:
+
+    target_link_libraries(<your-target> PRIVATE fmt::fmt-module)
+
+    import fmt;
+
+    int main() {
+      fmt::print("Hello, world!\\n");
```
vitaut: There is extra `\` here.

vitaut (approved):

vitaut:
Merged, thank you!

PR #4893, opened 2026-08-23
Suppress GCC 16 incomplete type warning

TheNonConformist:
GCC 16 emits `-Wsfinae-incomplete` when the incomplete type used by
  `incomplete_type_test` is later defined.

  The pattern is intentional because the test verifies formatting a type while
  it is still incomplete. Suppress the GCC 16 diagnostic narrowly around the
  definition so the test retains its coverage without producing a warning.

  Tested with GCC 16.2.1:

  - Full build succeeded
  - 21/21 tests passed
  - `git diff --check` passed

### DIFF
--- a/test/format-test.cc
+++ b/test/format-test.cc
@@ -2649,6 +2649,9 @@ TEST(incomplete_type_test, format) {
   EXPECT_EQ(fmt::format("{}", external_instance), "42");
 }
 
+#if FMT_GCC_VERSION >= 1600
+FMT_PRAGMA_GCC(diagnostic ignored "-Wsfinae-incomplete")
+#endif
 struct incomplete_type {};
 const incomplete_type& external_instance = {};
 


vitaut:
Could you post the full warning with context and the repro command (or godbolt link) for future reference?

test/format-test.cc
```diff
@@ -2649,7 +2649,14 @@ TEST(incomplete_type_test, format) {
...
 }
 
+#if FMT_GCC_VERSION >= 1600
+FMT_PRAGMA_GCC(diagnostic push)
+FMT_PRAGMA_GCC(diagnostic ignored "-Wsfinae-incomplete")
+#endif
 struct incomplete_type {};
+#if FMT_GCC_VERSION >= 1600
+FMT_PRAGMA_GCC(diagnostic pop)
+#endif
```
vitaut: Since this is just a test I don't think we need push/pop.

vitaut (changes requested):

TheNonConformist:
> Could you post the full warning with context and the repro command (or godbolt link) for future reference?





Sure. I reproduced it with GCC 16.2.1 using:

  cmake -S . -B build -DFMT_TEST=ON -DFMT_DOC=OFF -DFMT_INSTALL=OFF
  cmake --build build --target format-test --clean-first --verbose

  The complete diagnostic is:

image

vitaut (approved):

PR #4895, opened 2026-08-24
Right-align inf and nan with a bare width

### LINKED ISSUE #4894: inf and nan are left-aligned with a bare width
`inf` and `nan` are left-aligned with a bare width, while finite values are right-aligned:

```cpp
fmt::print("[{:10}]\n", 1.5);   // [       1.5]
fmt::print("[{:10}]\n", inf);   // [inf       ]
```

`std::format` right-aligns both, as does `printf("%10f")`, and the docs give right as the default for numeric types.

In `format.h` the nonfinite path takes the string default rather than the numeric one.

I have a patch for it, +7/-6 with a test, but it changes output for anyone formatting a possibly-nonfinite double with a bare width. Is the current behaviour intended?

advitrocks9:
Fixes #4894.

The nonfinite path in `write_float` takes the string default alignment rather than the numeric one, so `inf` and `nan` come out left-aligned while finite values with the same spec come out right-aligned.

```cpp
fmt::print("[{:10}]\n", 1.5);   // [       1.5]
fmt::print("[{:10}]\n", inf);   // [inf       ]
```

After the change both are right-aligned, matching `std::format` and the documented default. An explicit alignment is unaffected.

`format-test.cc` gains cases for `inf` and `nan` under a bare width. The assertion next to them already expected the new string for a signed nonfinite, so it was passing for the wrong reason.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -2402,12 +2402,11 @@ FMT_CONSTEXPR20 auto write_nonfinite(OutputIt out, bool isnan,
   const bool is_zero_fill =
       specs.fill_size() == 1 && specs.fill_unit<Char>() == '0';
   if (is_zero_fill) specs.set_fill(' ');
-  return write_padded<Char>(out, specs, size,
-                            [=](reserve_iterator<OutputIt> it) {
-                              if (s != sign::none)
-                                *it++ = detail::getsign<Char>(s);
-                              return copy<Char>(str, str + str_size, it);
-                            });
+  return write_padded<Char, align::right>(
+      out, specs, size, [=](reserve_iterator<OutputIt> it) {
+        if (s != sign::none) *it++ = detail::getsign<Char>(s);
+        return copy<Char>(str, str + str_size, it);
+      });
 }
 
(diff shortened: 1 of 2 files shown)


vitaut:
Overall looks good but some CI jobs failed to start and are in a weird state. Could you rebase to rerun them?

advitrocks9:
Rebased.

fallenmi (approved):
Approved at exact head `7449a2b` after independent validation.

I reproduced the bug on base/current main `e27cc20`: a bare width left-aligns `nan` and `inf`, unlike finite numeric values. The same seven-case oracle passes on this head, covering NaN, positive and negative infinity, explicit left and center alignment, zero-fill normalization, and finite-value parity. The full `format-test` binary passes (139 tests) with AppleClang, and `git diff --check` is clean. The `write_padded<Char, align::right>` default is correctly overridden by explicit alignment.

Reviewed with OpenAI Codex.

vitaut:
@fallenmi, please don't spam the issues with AI-generated comments.

vitaut (approved):

vitaut:
Thanks for the fix!

PR #4896, opened 2026-08-24
Fix inconsistent display width for emoji outside East Asian Wide ranges

### LINKED ISSUE #4851: Inconsistent widths when UTF8 characters are used

edugomz:
Fixes #4851

Continues #4865 with the requested changes addressed:

- Table is now derived from Unicode's `East_Asian_Width=W/F` data instead of `Emoji_Presentation`, so regional indicators (flag halves) aren't individually double-counted - 🇺🇸 still measures 2 columns, not 4
- `display_width_of()` remains `FMT_CONSTEXPR`

## Testing

Regression tests added to `test/format-test.cc`: the #4851 example, multiple/mixed emoji, precision truncation, alignment, newly-covered ranges, and a regional indicator (flag) pair.


vitaut (changes requested):
The standard width rules are not just East_Asian_Width=W/F; [format.string.std] also explicitly includes 1F300–1F5FF and 1F900–1F9FF.

include/fmt/format.h
```diff
@@ -649,26 +649,197 @@ FMT_CONSTEXPR void for_each_codepoint(string_view s, F f) {
...
-            (cp >= 0xfe30 && cp <= 0xfe6f) ||    // CJK Compatibility Forms
-            (cp >= 0xff00 && cp <= 0xff60) ||    // Fullwidth Forms
-            (cp >= 0xffe0 && cp <= 0xffe6) ||    // Fullwidth Forms
-            (cp >= 0x20000 && cp <= 0x2fffd) ||  // CJK
-            (cp >= 0x30000 && cp <= 0x3fffd) ||
-            // Miscellaneous Symbols and Pictographs + Emoticons:
-            (cp >= 0x1f300 && cp <= 0x1f64f) ||
-            // Supplemental Symbols and Pictographs:
-            (cp >= 0x1f900 && cp <= 0x1f9ff))));
+FMT_CONSTEXPR auto display_width_of(uint32_t cp) noexcept -> size_t {
```
vitaut: I think `inline` is needed here.

edugomz: true!

vitaut:
Looks like this PR was affected by the recent Actions outage, let me try to close/reopen to see if it fixes stuck CI jobs.

vitaut (approved):

PR #4897, opened 2026-08-25
Fix default alignment for inf/nan

### LINKED ISSUE #4894: inf and nan are left-aligned with a bare width

TheRodzz:
Fixes #4894

As per #4894, `inf` and `nan` are left-aligned with a bare width, while finite values are right-aligned. This PR fixes the default alignment for nonfinite values to be right-aligned.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -2402,12 +2402,11 @@ FMT_CONSTEXPR20 auto write_nonfinite(OutputIt out, bool isnan,
   const bool is_zero_fill =
       specs.fill_size() == 1 && specs.fill_unit<Char>() == '0';
   if (is_zero_fill) specs.set_fill(' ');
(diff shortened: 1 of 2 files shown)


vitaut:
Thanks for the PR but this has already been fixed by #4895.

PR #4899, opened 2026-08-25
Look up sparse enum identifiers in a hash table instead of a linear search

avikivity:
Enums that are too sparse for the index table were formatted by scanning all enumerators. Replace the scan with an open-addressed hash table with linear probing, sized to the smallest power of two that keeps the load factor at or below 0.5. This guarantees a free slot, which terminates the probe sequence and doubles as the not-found result because identifiers are never empty.

### DIFF
--- a/test/enum-test.cc
+++ b/test/enum-test.cc
@@ -37,7 +37,7 @@ enum class [[=fmt::as_identifiers]] dense { d0, d1, d2, d3, d4 };
 enum class [[=fmt::as_identifiers]] holey {
   h0, h1, h2, h3, h4, h5, h6 = 9
 };
-// 4 holes out of 11: just too sparse, formatted via a linear search.
+// 4 holes out of 11: just too sparse, formatted via a hash table.
 enum class [[=fmt::as_identifiers]] sparse {
   s0, s1, s2, s3, s4, s5, s6 = 10
 };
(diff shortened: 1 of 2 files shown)


avikivity:
Note: https://github.com/Kronuz/constexpr-phf exists, but I consider it overkill for this purpose.

test/enum-test.cc
```diff
@@ -47,6 +47,15 @@ enum class [[=fmt::as_identifiers]] signed_enum {
   minus_one = -1,
   one = 1
 };
+// Many scattered values, exercising collisions in the hash table.
+enum class [[=fmt::as_identifiers]] scattered {
+  a = 1, b = 17, c = 33, d = 49, e = 65, f = 81,
+  g = 97, h = 113, i = 129, j = 145, k = 161, l = 177
+};
```
vitaut, line 54: The scattered test doesn't actually exercise collisions: all of its values hash to distinct slots. Could we use values that force at least one probe chain?

vitaut (changes requested):
Looks reasonable but could you test collision per inline comment?

test/enum-test.cc
avikivity: Done.

avikivity:
Update:
 - add tests for correct behavior when hash collisions occur

vitaut (approved):

vitaut:
Merged, thanks!

PR #4902, opened 2026-08-28
Fix debug format width calculation

### LINKED ISSUE #4901: {:N?} miscomputes the width of debug-escaped output
```c++
fmt::print("[{:1?}]\n", std::string(""));  // ["]        missing the closing quote
fmt::print("[{:6?}]\n", 'a');              // ['a'     ] padded to 8, not 6
```

The width is computed from the size before escaping, so debug output pads wrong, and for an
empty string the closing quote gets truncated. `std::format` prints `[""]` and `['a'   ]`.

Happy to put up a PR.

Satyakam-Mishra:
Fixes #4901.

The width calculation for debug-formatted characters and strings was based on the unescaped output size. This caused incorrect padding for debug characters and could truncate the closing quote for empty debug strings.

This change:

* Calculates the formatted size of debug characters using the escaped representation.
* Correctly accounts for the closing quote of an empty debug string when width/precision permits it.
* Adds regression tests for both cases.

All tests pass locally.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -1907,11 +1907,22 @@ template <typename Char, typename OutputIt>
 FMT_CONSTEXPR auto write_char(OutputIt out, Char value,
                               const format_specs& specs) -> OutputIt {
   bool is_debug = specs.type() == presentation_type::debug;
-  return write_padded<Char>(out, specs, 1, [=](reserve_iterator<OutputIt> it) {
-    if (is_debug) return write_escaped_char(it, value);
-    *it++ = value;
-    return it;
-  });
+  Char buf[12];
+  auto* begin = buf;
+  auto* end = begin;
+  size_t size = 1;
+
+  if (is_debug) {
+    end = write_escaped_char(begin, value);
+    size = to_unsigned(end - begin);
+  }
+
+  return write_padded<Char>(out, specs, size,
+                            [=](reserve_iterator<OutputIt> it) {
+                              if (is_debug) return copy<Char>(begin, end, it);
+                              *it++ = value;
+                              return it;
+                            });
 }
 
 template <typename Char> class digit_grouping {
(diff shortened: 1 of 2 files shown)


include/fmt/format.h
```diff
@@ -1907,11 +1907,19 @@ template <typename Char, typename OutputIt>
...
-  return write_padded<Char>(out, specs, 1, [=](reserve_iterator<OutputIt> it) {
-    if (is_debug) return write_escaped_char(it, value);
-    *it++ = value;
-    return it;
-  });
+  size_t size = 1;
+  if (is_debug) {
+    counting_buffer<Char> buf;
+    write_escaped_char(basic_appender<Char>(buf), value);
+    size = buf.count();
```
vitaut: counting_buffer seems pretty heavyweight here for an output that is at most 12 characters. Could we format once into a small local buffer and use its size for padding, then copy it to the output? Something along the lines of

```c++
Char buf[12];
auto* begin = buf;
auto* end = begin;
size_t size = 1;

if (is_debug) {
  end = write_escaped_char(begin, value);
  size = to_unsigned(end - begin);
}

return write_padded<Char>(out, specs, size, [=](auto it) {
  if (is_debug) return copy<Char>(begin, end, it);
  *it++ = value;
  return it;
});
```

vitaut (changes requested):

include/fmt/format.h
Satyakam-Mishra: Thanks! I have updated the implementation to use a small local buffer as suggested, and the tests and clang-format-21 checks passes.

Satyakam-Mishra:
Updated the implementation to use reserve_iterator<OutputIt> instead of generic auto lambda parameter, which was causing the CI build failure with the older compiler. The changes are pushed and the local tests are passing

vitaut (approved):

vitaut:
Thanks for the fix!

PR #4904, opened 2026-08-29
Fix exception formatters with exceptions disabled

cjacek:
Fixes building with Clang 20 and older with -fno-exceptions, which treats try/catch usage as an error.

Fixes FEX builds: https://github.com/FEX-Emu/FEX/pull/5870

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -665,7 +665,7 @@ struct formatter<
     }
 #endif  // FMT_USE_RTTI
     out = detail::write_bytes<char>(out, string_view(ex.what()));
-#if FMT_USE_RTTI
+#if FMT_USE_RTTI && FMT_USE_EXCEPTIONS
     // If the exception carries a nested exception (e.g. via
     // std::throw_with_nested), format the whole chain.
     if (auto* nested = dynamic_cast<const std::nested_exception*>(&ex)) {
@@ -680,7 +680,7 @@ struct formatter<
         }
       }
     }
-#endif  // FMT_USE_RTTI
+#endif  // FMT_USE_RTTI && FMT_USE_EXCEPTIONS
     return out;
   }
 };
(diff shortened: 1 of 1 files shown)


vitaut (changes requested):
Thanks for the PR, but I don’t think we should disable the `std::exception` formatter itself. Instead, please disable only the minimal part that relies on `try`/`catch`. Also, could you post the full compiler error message?

cjacek:
Thanks for the review. The new version disables only minimal part of `std::exception` formatter. for `exception_ptr`, it seems that there would be nothing left if I did that. I also considered using `FMT_TRY`/`FMT_CATCH`, but that would just be broken with disabled exceptions.

The error message is:
```
fmt/include/fmt/std.h:685:9: error: cannot use 'try' with exceptions disabled
fmt/include/fmt/std.h:704:5: error: cannot use 'try' with exceptions disabled
```
Here is a compiler explorer repro: https://godbolt.org/z/7jKncahEK

It's specific to usage in templates that's now allowed. Outside templates it's always an error.

include/fmt/std.h
```diff
@@ -680,11 +691,12 @@ struct formatter<
         }
       }
     }
-#endif  // FMT_USE_RTTI
+#endif  // FMT_USE_RTTI && FMT_USE_EXCEPTIONS
     return out;
   }
 };
 
+#if FMT_USE_EXCEPTIONS
```
vitaut, line 699: Presence of `std::exception_ptr` doesn't depend on whether exceptions are enabled or disabled and in general we prefer not to make formatter availability depend on configuration. Let's move this to the `format` function and only guard `try`/`catch` block there giving "unknown exception" if exceptions are disabled.

vitaut (changes requested):

include/fmt/std.h
cjacek: I moved it as suggested, thanks!

vitaut (approved):

vitaut:
Merged, thanks

PR #4912, opened 2026-09-03
Fix debug format width calculation by clarifying maximum buffer size.

### LINKED ISSUE #4901: {:N?} miscomputes the width of debug-escaped output

ilike314272:
Width for debug-escaped chars/strings was computed from the unescaped size, causing incorrect padding and, for empty strings, a dropped closing quote. Removes magic number by assigning 12 to `max_buffer_size`. Fixes #4901 and adds additional test coverage for non-printing characters that exceed the maximum.

### DIFF
--- a/include/fmt/format.h
+++ b/include/fmt/format.h
@@ -2061,11 +2061,19 @@ template <typename Char, typename OutputIt>
 FMT_CONSTEXPR auto write_char(OutputIt out, Char value,
                               const format_specs& specs) -> OutputIt {
   bool is_debug = specs.type() == presentation_type::debug;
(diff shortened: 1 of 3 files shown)


vitaut:
Thanks for the PR but this is already being addressed in #4902.

Issue #4913, opened 2026-09-03
Add Dark Mode Toggle Option


sparshsingh12312:
## Problem Statement

There is no option to switch to dark mode in the Appearance section.

## Proposed Solution

I want to add an option to switch to dark mode in the appearance section.

## Alternatives Considered

We can also add this to the upper right corner of the website next to the new task.

## Additional Context

<img width="1470" height="956" alt="Image" src="https://github.com/user-attachments/assets/098e0e63-9874-4a9b-b927-5075648c7a99" />

sparshsingh12312:
May I lead on addressing this issue? @vitaut

vitaut:
A PR is welcome and it looks like someone submitted one already: #4914

PR #4914, opened 2026-09-03
Feature: dark mode docs

ilike314272:
Added dark mode by editing mkdocs.yml at request of #4913.


vitaut (changes requested):
Thanks for the PR.

Could you make the syntax highlighting follow the selected color scheme as well? We currently load Highlight.js's `default.min.css` unconditionally, which is a light theme, so code blocks will remain light in dark mode.

It would be good to add dark-mode highlighting styles scoped to `[data-md-color-scheme="slate"]` (and keep the current styles for the default scheme).

ilike314272:
New dark.css commit adds syntax highlighting that looks more pleasing than copying light mode css
image
image

vitaut (changes requested):
Thanks. Could we avoid adding a separate `dark.css`? The current file is essentially a copy of Highlight.js's Atom One Dark theme, which also brings in third-party licensing concerns.

I think a cleaner approach would be to keep the palette change in `mkdocs.yml` and add a small set of Highlight.js overrides to `fmt.css` that use MkDocs Material's existing code color variables. That way the highlighting follows the selected light/dark scheme without maintaining a separate hard-coded theme.

ilike314272:
I noticed that using material theme would have kept the banner from being blue like the Ukraine flag so I also provided override for theme on banner. `dark.css` removed and replaced it with `.hjls` overrides at bottom of `fmt.css`. Thank you for reviewing, I am learning from it.

vitaut (changes requested):
Thanks, this is much cleaner. Two remaining issues:

* `dark.css` is still listed in `extra_css`, but the file has been removed. Please remove that entry as well.
* Could you make the Highlight.js overrides cover all of the color rules from `default.min.css`? A few selectors such as `.hljs-subst`, `.hljs-template-tag`, `.hljs-section`, `.hljs-deletion` and `.hljs-code` still inherit hard-coded colors from the light theme, so they can have poor contrast in dark mode.

Using the Material code color variables for those as well should make the highlighting fully scheme-independent.

ilike314272:
Thank you for your patience, I have learned about hljs and done both.

vitaut (approved):

vitaut:
Merged, thanks

PR #4918, opened 2026-09-04
fix(error-code): sanitize non-UTF-8 bytes from ec.message() (fixes #4436)

### LINKED ISSUE #4436: error code `:s` fail on windows due to encoding error

yunaremaia:
On Windows, std::error_code::message() may return a string in a system code page (e.g. CP-1252) containing bytes invalid in UTF-8. When formatted with {:s}, this caused encoding errors/crashes (issue #4436).

Adds detail::append_utf8_sanitized() which validates UTF-8 input and replaces invalid byte sequences with U+FFFD (replacement character). The std::error_code formatter now uses this helper instead of appending ec.message() directly.

Test: std_test.error_code includes a direct test of the sanitize helper with non-UTF-8 input. All existing tests pass.

### DIFF
--- a/include/fmt/std.h
+++ b/include/fmt/std.h
@@ -290,6 +290,63 @@ struct has_format_as_member<
     T, void_t<decltype(formatter<T>::format_as(std::declval<const T&>()))>>
     : std::true_type {};
 
+// On Windows, std::error_code::message() may return a string in a system
+// code page (e.g. CP-1252) that contains bytes invalid in UTF-8. This helper
+// validates the input as UTF-8 and replaces any invalid byte sequences with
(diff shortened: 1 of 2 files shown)


yunaremaia:
CI hasn't been triggered yet on this PR. Could you re-run the workflow or let me know if there's anything I need to do to trigger CI checks?

vitaut:
Thanks for the PR but this would just mask a bug in Microsoft STL, please see #4884 for details.

PR #4923, opened 2026-09-05
Correct locale support in size optimization docs

dajiaohuang:
Fixes #4922.

Move locale disabling from the level 1 description to level 2, matching the default `FMT_USE_LOCALE` definition in `base.h` and the existing macro documentation. No runtime behavior changes.

Validation:

- MSVC compile-time probes for `FMT_OPTIMIZE_SIZE=0/1/2` confirm default `FMT_USE_LOCALE=1/1/0`.
- Built the patched MkDocs site with the pinned documentation requirements, Doxygen 1.18.0 and `PYTHONUTF8=1`; checked the rendered configuration table. One existing `chrono-format-specifications` anchor warning remains. Without Python UTF-8 mode, the existing C++ handler cannot read Doxygen's UTF-8 XML under Windows cp1252.
- `git diff --check` passes.

PR #4925, opened 2026-09-05
Read Doxygen XML independently of the system text encoding

### LINKED ISSUE #4924: C++ docs handler reads UTF-8 Doxygen XML using the system text encoding
The C++ mkdocstrings handler opens Doxygen XML in text mode without an encoding at three sites in `support/python/mkdocstrings_handlers/cxx/__init__.py` (header, compound and namespace loading).

On current main `6226bc70b8e15060a21a281c5ec5f4b91be8582f`, building the docs with Python's Windows cp1252 locale fails at the first header-loading site:

```text
UnicodeDecodeError: 'charmap' codec can't decode byte 0x8d in position 4724
```

The input `chrono_8h.xml` declares UTF-8 and contains valid UTF-8 Chinese text. The text-mode stream decodes it using the system locale before ElementTree can process the XML declaration. The same build completes with `PYTHONUTF8=1`.

Reproduction environment: Windows x64, repository-pinned `support/doc-requirements.txt` in an isolated Python environment, official Doxygen 1.18.0 on PATH. From the source root, run the environment's Python with UTF-8 mode disabled:

```text
python -X utf8=0 support/mkdocs build --site-dir <output-directory>
```

Expected: generated XML is decoded according to its encoding declaration, independently of the user's default text codec. Opening each XML stream in binary mode before `ET.parse` would let the XML parser handle that decoding without changing the document content or dependencies.

This concerns local documentation generation, not C++ runtime formatting or the hosted Ubuntu deployment. The separate locale-description PR #4923 only changes `doc/api.md`; it does not fix these reads.

dajiaohuang:
Fixes #4924.

Open the three Doxygen XML streams in binary mode so ElementTree decodes their declared encoding instead of Python's default text codec. This fixes local documentation generation under Windows cp1252 without requiring `PYTHONUTF8=1`.

Adds three unittest regressions for header, compound and namespace loading. They exercise the handler with UTF-8 XML and a simulated cp1252 default, without running Doxygen or depending on the host locale. All three fail with `UnicodeDecodeError` before the fix and pass afterward.

Validation:

- `python -B -X utf8=0 -m unittest discover -s support/python/tests -v`: 3 passed, using the pinned documentation environment.
- `python -X utf8=0 support/mkdocs build --site-dir <output>`: passed on Windows with Doxygen 1.18.0 and `PYTHONUTF8=0`. The existing unrelated chrono-anchor warning remains.
- Ruff check on both changed Python files, Ruff format check on the new test, and `git diff --check` pass.

No dependency, CI, C++ API, or release changes. The separate locale-description PR #4923 is unaffected.

### DIFF
--- a/support/python/mkdocstrings_handlers/cxx/__init__.py
+++ b/support/python/mkdocstrings_handlers/cxx/__init__.py
@@ -302,7 +302,7 @@ class CxxHandler(BaseHandler):
         self._file_doxyxml: "ET.ElementTree[ET.Element[str]] | None" = None
         for h in headers:
             filename = h.replace(".h", "_8h.xml")
-            with open(os.path.join(self._doxyxml_dir, filename)) as f:
+            with open(os.path.join(self._doxyxml_dir, filename), "rb") as f:
                 doxyxml = ET.parse(f)
                 if self._file_doxyxml is None:
                     self._file_doxyxml = doxyxml
@@ -317,7 +317,7 @@ class CxxHandler(BaseHandler):
         if refid is None:
             raise ValueError
         path = os.path.join(self._doxyxml_dir, refid + ".xml")
-        with open(path) as f:
+        with open(path, "rb") as f:
             xml = ET.parse(f)
             node = xml.find("compounddef")
             if node is None:
@@ -371,7 +371,7 @@ class CxxHandler(BaseHandler):
         doxyxml = self._ns2doxyxml.get(namespace)
         if doxyxml is None:
             path = f"namespace{namespace.replace('::', '_1_1')}.xml"
-            with open(os.path.join(self._doxyxml_dir, path)) as f:
+            with open(os.path.join(self._doxyxml_dir, path), "rb") as f:
                 doxyxml = ET.parse(f)
                 self._ns2doxyxml[namespace] = doxyxml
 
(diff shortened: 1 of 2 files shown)


support/python/mkdocstrings_handlers/cxx/__init__.py
```diff
@@ -302,7 +302,7 @@ def __init__(
         self._file_doxyxml: "ET.ElementTree[ET.Element[str]] | None" = None
         for h in headers:
             filename = h.replace(".h", "_8h.xml")
-            with open(os.path.join(self._doxyxml_dir, filename)) as f:
+            with open(os.path.join(self._doxyxml_dir, filename), "rb") as f:
                 doxyxml = ET.parse(f)
```
vitaut: We don't actually need open, let's just pass `path` to `parse` which handles encoding correctly:
```
doxyxml = ET.parse(os.path.join(self._doxyxml_dir, filename))
```
and similarly in other cases.

vitaut (changes requested):
The fix looks good, but the regression tests seem like overkill for this. Let's drop them together with `open` calls (see the inline comment).

vitaut (approved):

vitaut:
Merged, thank you!

PR #4927, opened 2026-09-05
Fix the chrono format specification link

### LINKED ISSUE #4926: Chrono API reference links to a missing syntax anchor
On current main `6226bc70b8e15060a21a281c5ec5f4b91be8582f`, the chrono API
reference links to `syntax.md#chrono-format-specifications`:

https://github.com/fmtlib/fmt/blob/6226bc70b8e15060a21a281c5ec5f4b91be8582f/doc/api.md#L475-L477

That fragment is absent from the current syntax page. Its explicit anchor is
`chrono-format-spec`, and the generated heading anchor is
`chrono-format-specification` (singular):

https://github.com/fmtlib/fmt/blob/6226bc70b8e15060a21a281c5ec5f4b91be8582f/doc/syntax.md#L292-L295

To reproduce, build the docs with the declared documentation environment and
`python support/mkdocs build`, then follow the chrono API's format-specification
link. MkDocs also reports the missing `chrono-format-specifications` anchor.
Expected: navigate directly to the chrono grammar. Actual: the syntax page
loads, but the fragment has no target.

Using the existing explicit `#chrono-format-spec` anchor would fix the link
without changing the syntax documentation or API. This is separate from the
padding-modifier correction in #4917 and the locale-description fix in #4923.

dajiaohuang:
Fixes #4926.

Point the chrono API reference to the existing explicit `chrono-format-spec`
anchor in `syntax.md`. The old plural fragment has no target.

Validation:

- Documentation build passes using the pinned docs environment and Doxygen
  1.18.0; the missing-anchor warning is gone.
- Generated `api/index.html` links to `../syntax/#chrono-format-spec`, and
  generated `syntax/index.html` contains that ID. The old fragment is absent.
- `git diff --check` passes.

One documentation line only; no API, syntax, dependency or CI change. On this
Windows host the build uses process-local UTF-8 mode; the separate default-codec
fix in #4925 is not included in this branch.

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -474,7 +474,7 @@ Using `fmt::join`, you can separate tuple elements with a custom separator:
 - [`std::tm`](https://en.cppreference.com/w/cpp/chrono/c/tm)
 
 The format syntax is described in [Chrono Format Specifications](syntax.md#
-chrono-format-specifications).
+chrono-format-spec).
 
 **Example**:
 


vitaut (approved):

vitaut:
Thanks for the fix. In the future, for simple cases like this I suggest skipping the issue and just opening the PR.

PR #4931, opened 2026-09-06
Add Quill to the list of notable users

solbjorn:
One of its unique features is that it serializes the format string and all format arguments to the thread's SPSQ queue to minimize processing on hotpath and then deserializes them on the backend thread which performs formatting using {fmt} and writing to the log sinks.


vitaut (approved):

vitaut:
thanks

PR #4932, opened 2026-09-06
Fix unreachable code warning in range formatter

### LINKED ISSUE #4930: Unreachable code warning in ranges.h for msvc with /W4
msvc complains about unreachable code in `ranges.h`

I'm unable to repo on godbolt (with 12.0.0 and "trunk" doesn't seem to work) so I assume it has been introduces sometime between 12.0.0 and 12.2.0

### How to reproduce
Compile the following code with `/W4`
```cpp
#include <cstdio>

#include <fmt/ranges.h>
#include <fmt/format.h>

void example()
{
    int arr[] = {1, 2, 3, 5, 7, 11};
    const auto test = fmt::format("{}", arr);
    std::puts(test.c_str());
}
```

Something like this
```
>cl fmt-test.cpp /W4 /WX /std:c++20 /I"C:\libs\fmt-12.2.0\include" /utf-8
Microsoft (R) C/C++ Optimizing Compiler Version 19.51.36243 for x86
Copyright (C) Microsoft Corporation.  All rights reserved.

fmt-test.cpp
C:\libs\fmt-12.2.0\include\fmt\ranges.h(521) : error C2220: the following warning is treated as an error
C:\libs\fmt-12.2.0\include\fmt\ranges.h(521) : warning C4702: unreachable code
```

### Solution
I've been able to solve it by adding a `else` clause to the `if` statment here https://github.com/fmtlib/fmt/blob/0f3c5fd596f27108e0c338b52a5aae9e4a491e26/include/fmt/ranges.h#L517-L522
```cpp
FMT_CONSTEXPR formatter() {
  if FMT_CONSTEXPR20 (range_format_kind<R, Char>::value != range_format::set)
    return;
  
  else range_formatter_.set_brackets(detail::string_literal<Char, '{'>{},
                                detail::string_literal<Char, '}'>{});
}
```

Zhao0335:
Fixes #4930.

The range formatter constructor returned from a compile-time branch for every
non-set range, leaving the bracket customization after the return. New MSVC
versions diagnose that statement as unreachable when sequence range formatters
are instantiated.

Invert the condition and only customize brackets for set ranges. This preserves
the behavior for both pre-C++20 and C++20 builds while allowing `if constexpr`
to discard the unused branch.

Testing:

- MSVC 19.44 `/W4`: the `ranges.h` C4702 warning reproduced before the change
  and is absent after it
- `ranges-test` passed
- full MSVC Release build passed
- full CTest: 20/22 passed; `chrono-test.locale` and
  `unicode-test.legacy_locale` fail under this machine's zh-CN locale

### DIFF
--- a/include/fmt/ranges.h
+++ b/include/fmt/ranges.h
@@ -515,10 +515,9 @@ struct formatter<
   using nonlocking = void;
 
   FMT_CONSTEXPR formatter() {
-    if FMT_CONSTEXPR20 (range_format_kind<R, Char>::value != range_format::set)
-      return;
-    range_formatter_.set_brackets(detail::string_literal<Char, '{'>{},
-                                  detail::string_literal<Char, '}'>{});
+    if FMT_CONSTEXPR20 (range_format_kind<R, Char>::value == range_format::set)
+      range_formatter_.set_brackets(detail::string_literal<Char, '{'>{},
+                                    detail::string_literal<Char, '}'>{});
   }
 
   FMT_CONSTEXPR auto parse(parse_context<Char>& ctx) -> const Char* {


include/fmt/ranges.h
```diff
@@ -581,10 +515,9 @@ struct formatter<
   using nonlocking = void;
 
   FMT_CONSTEXPR formatter() {
-    if FMT_CONSTEXPR20 (range_format_kind<R, Char>::value != range_format::set)
-      return;
-    range_formatter_.set_brackets(detail::string_literal<Char, '{'>{},
-                                  detail::string_literal<Char, '}'>{});
+    if FMT_CONSTEXPR20 (range_format_kind<R, Char>::value == range_format::set)
+      range_formatter_.set_brackets(detail::string_literal<Char, '{'>{},
+                                    detail::string_literal<Char, '}'>{});
```
vitaut: Since this is now a multiline-statement, please add braces around the if body.

vitaut (changes requested):

include/fmt/ranges.h
Zhao0335: Added braces around the if body in af21761.

vitaut (approved):

vitaut:
Merged, thanks

PR #4933, opened 2026-09-06
Fix argument count in dynamic stores with named arguments

IfkumRfnl:
## Summary

Code counted formatting arguments incorrectly when named arguments were present. It took the internal metadata slot as a formatting argument. Change made the math correct with

```cpp
 return data_.size() - (named_info_.empty() ? 0 : 1);
```

i.e. subtracting 1 from the arguments vector count when named arguments are present which makes the counting correct. The corrected count is also used when constructing `basic_format_args`. Also regression tests were added.


vitaut (approved):

vitaut:
Thanks for the fix!

PR #4934, opened 2026-09-07
Fix out-of-bounds read in error code format parsing

IfkumRfnl:
## Summary

The `std::error_code` formatter read past the end of a format string when the string ended after an alignment specifier, for example, `"{:>"`.

Added a check after alignment parsing to prevent this read:

```cpp
if (it == end) return it;
```

Removed the redundant check before width parsing. Added two regression tests. Both tests failed with AddressSanitizer before the fix. All 28 `std-test` tests pass with AddressSanitizer after the fix.


vitaut (changes requested):
Thanks for the fix. Overall looks good but one regression test should be sufficient, let's remove the redundant `error_code_truncated_fill_alignment` test.

IfkumRfnl:
Thanks for looking into it, removed the test.

vitaut (approved):

PR #4935, opened 2026-09-07
Fix width and alignment for calendar types

IfkumRfnl:
## Summary

Calendar formatters used a full timestamp when the format specified only width or alignment. For example, `fmt::format("{:5}", fmt::day(5))` returned `1900-01-05 00:00:00`. Formatting a year with width caused an assertion failure.

Set the default format for each calendar type before parsing. This keeps the calendar output when width or alignment is specified.

Added regression tests for all five calendar types, including fill, dynamic width, explicit chrono formats, and localization. The C++17 chrono and wide-character test suites pass.


vitaut (approved):

vitaut:
Thanks for the fix!

PR #4936, opened 2026-09-07
Fix debug string formatting for range formatters

cuishuang:
`range_format::debug_string` manually added quotes around the result of the underlying string formatter. This bypassed debug string escaping and caused width specifications to exclude the quotes.

For example, newline, quote, and backslash characters were emitted without the expected escaping, and `{:8}` could produce a result wider than eight columns.

Parse the underlying formatter first, enable its debug format for `range_format::debug_string`, and delegate formatting directly to it. This ensures that escaping, alignment, and width are handled consistently by the standard string formatter.


vitaut (approved):

vitaut:
Merged, thanks!

PR #4939, opened 2026-09-10
Make nolocale-test actually disable locale

tiagomacarios:
`nolocale-test` defines `FMT_STATIC_THOUSANDS_SEPARATOR`, which stopped doing anything in b90b4bc9 ("Remove FMT_STATIC_THOUSANDS_SEPARATOR in favor of FMT_USE_LOCALE"). That macro no longer appears anywhere under `include/`, so since then the target has compiled `src/format.cc` with locale support **enabled**.

It is not a dead target: the pedantic CI jobs build it (`linux.yml` and `macos.yml` both pass `-DFMT_PEDANTIC=ON`), and it passes. So the configuration looks covered while nothing actually tests it. #4627, a locale-off build break, landed during that window.

This defines `FMT_USE_LOCALE=0` instead. Verified the target still builds clean in that configuration (MSVC 19.51, Ninja, `-DFMT_PEDANTIC=ON`).

Two things this deliberately does not do, happy to follow up on either:

- It does not add `chrono.h`/`xchar.h` to the target. `chrono.h` includes `<locale>` unconditionally, and this target looks like it exists for platforms that have no `<locale>` at all. Instantiating the formatters in that configuration - which is what would actually have caught #4627 - probably wants a separate target.
- `FMT_PEDANTIC` is off in the Windows workflow, so this configuration still never builds on MSVC.


vitaut (approved):

vitaut:
Nice catch, thanks

PR #4940, opened 2026-09-10
Don't construct a locale when formatting year_month_day

tiagomacarios:
`formatter<year_month_day>::format` passes `true` to `get_locale`, so every default format constructs a `std::locale`, and `tm_writer` then compares it against the classic locale (`is_classic_(loc_ == get_classic_locale())`). The only thing written is `on_iso_date()`, which is pure arithmetic and reads neither `loc_` nor `is_classic_`.

It is the odd one out among its siblings: `formatter<day>` and `formatter<year>` pass `false`; `formatter<weekday>` and `formatter<month>` pass their `localized()` flag. Only `year_month_day` hardcodes `true`, and its `parse()` never sets localized in the first place.

With `FMT_USE_LOCALE=0` this costs more than a copy: the `loc.get<std::locale>()` arm of `get_locale` is compiled out, so it default-constructs instead, which captures the current **global** locale in a build that asked for no locale support.

No output change: `chrono_test.year_month_day` already sets a non-classic global locale and expects `"2024-01-01"`. Verified 33/33 chrono tests pass (MSVC 19.51, `-DCMAKE_CXX_STANDARD=20`).


vitaut (approved):

vitaut:
#4921 subsumes this but since it's not ready to merge let's merge this small change anyway. Thanks.

PR #4945, opened 2026-09-13
Fix format_as for tuple-like types with ranges.h

### LINKED ISSUE #4944: Tuple-like type becomes unformattable if `<fmt/ranges.h>` is included
I have a library that provides a type that for convenience can be destructed like so:
```cpp
auto [key, value] = obj;
```
To make the type formattable without having a dependency on fmtlib itself I provide a `format_as` function in the same namespace
```cpp
std::string format_as(FooType const& value) {
   // ...
}
```

This works fine, except if `<fmt/ranges.h>` has been included somewhere, where the following error appears:
```cpp
/opt/compiler-explorer/libs/fmt/12.0.0/include/fmt/ranges.h:172:21: error: no matching function for call to 'fmt::v12::detail::is_tuple_formattable_<lib::FooType, char, true>::check(fmt::v12::detail::tuple_index_sequence<lib::FooType>)'
  172 |       decltype(check(tuple_index_sequence<T>{}))::value;
```
I've tested on 12.2.0 and trunk and got the same issue on both versions.
Full code: https://godbolt.org/z/GPEWoe3Y5

dajiaohuang:
Fixes #4944

A tuple-like type can also provide an ADL `format_as` function. The tuple formattability probe should defer to that formatter instead of recursively checking tuple elements.

This change detects `format_as` before probing tuple elements, so the existing `format_as` formatter takes precedence.

Tests:
- Reproducer compile with MSVC /std:c++20
- Full CMake/CTest suite (22 tests)


vitaut (changes requested):
Thanks for the PR. I think `format_as` should indeed take precedence over tuple formatting, but I’d prefer not to change the semantics of `is_tuple_formattable`.

`fmt::is_tuple_formattable` is an exported trait and currently answers whether the tuple elements are formattable. With this change it becomes false for every tuple-like type that has `format_as`, even if all of its elements are formattable. Could we instead keep `is_tuple_formattable_` unchanged and exclude `format_as` when selecting the tuple formatter, while still making sure the tuple formattability check is not instantiated in that case?

Nit: could you also replace the `\n` escapes in the PR description with actual newlines? The current description is a bit odd to read/edit.

dajiaohuang:
Updated in 4d5286b4 (amended after CI showed the public tuple-formattability value is intentionally unchanged for this recursive tuple-like test type).

- Restored `fmt::is_tuple_formattable` to its existing tuple-element semantics.
- Added a formatter-only trait that skips the tuple probe when ADL `format_as` is available, so the dedicated `format_as` formatter still takes precedence without instantiating the recursive check.
- Kept the existing runtime regression test for a tuple-like `format_as` type.
- Rewrote the PR description with actual newlines.

Validation: `git diff --check` passed. A local CMake/compiler toolchain is not installed in this workspace; the repository CI will provide the full build/test validation.

dajiaohuang:
Implemented and pushed commit 57dc0903cfe92be0d1f4b8d66103b0388e9dbb05. This preserves the public `fmt::is_tuple_formattable` trait semantics while keeping formatter dispatch excluded for `format_as` tuple-like types, and adds a regression assertion covering the distinction. `git diff --check` passes. Local CMake/CTest are unavailable in this environment, so please rely on the remote CI for the full test suite.

dajiaohuang:
Follow-up: corrected the regression assertion to pass the required `char` parameter to the exported `fmt::is_tuple_formattable<T, C>` trait (commit 13e3aa74). The preceding CI failure was only this test compile error; the fix is now pushed. Remote CI should rerun.

dajiaohuang:
Follow-up: the exported trait assertion now uses `static_assert` (commit c462af2d), avoiding C++11/14 ODR-use linker errors while still verifying `fmt::is_tuple_formattable<format_as_tuple_like, char>::value` remains true. The prior CI failures were limited to that assertion form.

include/fmt/ranges.h
```diff
@@ -164,6 +164,11 @@ using make_index_sequence = make_integer_sequence<size_t, N>;
 template <typename T>
 using tuple_index_sequence = make_index_sequence<std::tuple_size<T>::value>;
 
+template <typename T, typename = void>
+struct has_format_as_ : std::false_type {};
+template <typename T>
+struct has_format_as_<T, void_t<format_as_result<T>>> : std::true_type {};
```
vitaut: This can be folded into `is_tuple_formattable_for_formatter`.

vitaut (changes requested):
Mostly looks good, just one more inline comment.

include/fmt/ranges.h
dajiaohuang: Folded the `format_as` detection into `is_tuple_formattable_for_formatter` so the helper directly defaults to the exported tuple trait and specializes to false only when `format_as_result<T>` is valid. Pushed as 83c869a0; `git diff --check` passes.

vitaut (approved):

vitaut:
Merged, thanks

Issue #4941, opened 2026-09-10
L specifier has three different behaviours when FMT_USE_LOCALE is 0


tiagomacarios:
With `FMT_USE_LOCALE=0`, `L` does three different things depending on the argument type, and one of them still reads the global locale. I think two of the three are regressions, but the right contract is your call, so this is an issue rather than a PR.

### Repro

Head (8dc5d3f6), MSVC 19.51, `/std:c++20`, header-only, global locale set to `de-DE`:

```cpp
#define FMT_USE_LOCALE 0
#include <fmt/chrono.h>

std::locale::global(std::locale("de-DE"));

fmt::format("{:L}", 1234567)             // "1234567"   L ignored
fmt::format("{:L}", 1234.5)              // "1,234.5"   stub numpunct applied
fmt::format("{:L%a}", tm)                // "Mon"       inert
fmt::format("{:L}", std::chrono::Monday) // "Mo"        global locale
```

| Path | Behaviour | Why |
| --- | --- | --- |
| int | inert | the locale-off `write_loc` returns false (`format.h:2225`), so it falls to `write_int` (`format.h:2371`) |
| float | stub grouping `"\03"` / `','` | `do_write_float` always runs `Grouping(loc, specs.localized())` (`format.h:2743`, `2759`, `2793`) |
| `std::tm` | inert | `get_locale(static_cast<bool>(loc_ref), ...)`, and `operator bool` is false (`chrono.h:2148`, `core.h:919`) |
| weekday / month / duration | **global locale** | `get_locale(localized(), ...)` (`chrono.h:1705`, `1925`, `1977`). With the `loc.get<std::locale>()` arm compiled out, `get_locale` default-constructs `std::locale()`, which is the global locale, not the classic one. That then reaches `std::use_facet<std::time_put>` at `chrono.h:389`. |

### On the numeric split

b98ffb7d ("Improve locale handling", 2022-09-04) moved the integer localized path onto `write_loc`, whose locale-off form returns false, so integers silently stopped grouping. Floats never used `write_loc` for grouping and kept theirs.

Two years later b90b4bc9 ("Remove FMT_STATIC_THOUSANDS_SEPARATOR in favor of FMT_USE_LOCALE") hard-coded `','` into the stub `numpunct` rather than dropping it, which reads like the static-separator behaviour was deliberately retained. If that is right, the integer path is the regression, not the float path. I did not want to assume it either way.

### On the chrono path

[...]

PR #4947, opened 2026-09-15
Use classic locale grouping without locale support

tiagomacarios:
Addresses #4941.

When `FMT_USE_LOCALE=0`, keep accepting the `L` specifier but use classic/C locale semantics. In particular, localized floating-point formatting no longer applies the fallback comma grouping; integer behavior is unchanged.

The second commit converts `nolocale-test` from a compile-only target into a registered executable and verifies localized integer and floating-point output at runtime. Chrono locale handling remains separate in #4921.

Tested with MSVC 19.51 using `FMT_PEDANTIC=ON`:

```
ctest -R ^nolocale-test$ --output-on-failure
```


vitaut (changes requested):
Thanks for the PR. I think we should skip the grouping completely and avoid unnecessary work.

PR #4948, opened 2026-09-16
Limit Clang diagnostic suppressions to fmt declarations

yqtian-se:
Clang diagnostic pragmas remain in effect until another pragma changes or
restores them. Two public fmt headers suppress warnings for fmt declarations
but do not restore the diagnostic state before returning to consumer code:

- `format.h` ignores `-Wweak-vtables` before declaring `format_error`;
- `std.h` ignores `-Wbit-int-extension` before declaring the `bitint` and
  `ubitint` aliases.

Because neither suppression has a matching push and pop, declarations after
the include inherit it. For example, this consumer class normally triggers
`-Wweak-vtables`, but the warning disappears after including `format.h`:

```cpp
#include <fmt/format.h>

class ConsumerClass {
 public:
  virtual void method() {}
};
```

The same occurs for a consumer `_BitInt` alias after including `std.h`:

```cpp
#include <fmt/std.h>

template <int N>
using consumer_bitint = _BitInt(N);
```

The requested `-Wbit-int-extension` diagnostic is suppressed along with fmt's
own aliases. This branch is enabled by default with Clang 15 and later; I used
`FMT_USE_BITINT=1` to exercise the same branch with the available Clang 14
compiler.

The matched probes produced the following results:

| Consumer declaration | Without the fmt header | Current `main` | This change |
| --- | --- | --- | --- |
| Weak-vtable class with `-Wweak-vtables` | warning | no diagnostic | warning |
| `_BitInt` alias with `-Wbit-int-extension` | warning | no diagnostic | warning |

Thus, including either header currently disables a warning that the consumer
explicitly requested for its own code. In a build that promotes the warning to
an error, the header also changes whether that consumer code is accepted.

This change brackets each suppression with fmt's existing Clang diagnostic
push/pop macros. The warnings remain suppressed for the fmt declarations that
need them, while the consumer's previous diagnostic state is restored
immediately afterward. The macros remain no-ops for non-Clang compilers.

Validation:

- a Clang 14 build with `FMT_PEDANTIC=ON` and `FMT_WERROR=ON` succeeds;
- the configured test suite passes 26 of 26 tests.


vitaut (approved):

vitaut:
Merged, thanks

yqtian-se:
Thanks!

PR #4950, opened 2026-09-18
Add support for single-argument print/println without format string

mikomikotaishi:
Currently, `print()` and `println()` have no support for simply printing an object without a format string, for example `println(123);` or `println(s);`. Introducing support for these is not difficult; we just make these call the usual function with the format string of `"{}"`.

To prevent ambiguity with format strings, we use `is_format_string_arg` to ensure that if `T` is a format string or a literal, the single-argument overload is disabled instead of being treated as a value to be formatted.

Introducing support would simplify the API and have parity with iostreams (`std::cout << 42 << std::endl;`), as well as other languages (such as Python's `print(x)`, Java's `System.out.println(x);`, C#'s `Console.WriteLine(x);`, etc.); similar quality-of-life changes have been made in the past, such as `println()` for printing an empty line.

(I originally also had a `printf()` single-argument overload, but removed it because no such thing existed in the original C `printf()` nor in Java `PrintStream.printf()`, and I highly doubt anyone would even use it rather than just `print(42);`/`println(42);` anyway.)


vitaut:
Thanks for the proposal, but single-argument `print`/`println` without a format string is intentionally not supported. Like `printf`, `std::print`, and similar formatting APIs, `print` treats the first argument as a format string. Keeping that convention also avoids overload-resolution complexity and potential compatibility issues with existing and future format-string types.

mikomikotaishi:
> potential compatibility issues with existing and future format-string types.

Couldn't this be solved by simply consistently having tests against other format string types?

PR #4951, opened 2026-09-18
Fix FMT_COMPILE formatting of ranges, tuples and nested formatters

### LINKED ISSUE #4949: FMT_COMPILE fails to format std::array
On `main` (e4751cc70), this fails to compile with GCC 15.2 and `-std=c++20`. It compiles and runs correctly with fmt 12.2.0.

Compiler Explorer: https://godbolt.org/z/hxMdoozx8

```cpp
#include <array>
#include <fmt/compile.h>
#include <fmt/ranges.h>

int main() {
  std::array<int, 3> values{1, 2, 3};
  auto s = fmt::format(FMT_COMPILE("array={}"), values);
  fmt::print("{}\n", s);
}
```

Expected output:
`array=[1, 2, 3]`

It works without `FMT_COMPILE`.

This issue seems to have been introduced in 61e3b923.

A branch in `nested_format_specs::write` attempts to construct an iterator-specific `FormatContext` from a `basic_appender<char>`, which cannot be converted to a `std::back_insert_iterator<std::string>` as required by the `generic_context<std::back_insert_iterator<std::string>, char>` constructor. This example never actually tries to construct the `FormatContext` (as the branch is for padding, and we don't specify a width), but it still needs to be instantiated.

linsen458-spec:
Fixes #4949.

Formatting a range, tuple or `nested_formatter` through the compile-time API failed to compile: the padding path assumed the context iterator is `basic_appender<Char>`, which only holds for `format_context`.

This makes `nested_format_specs::write` work with arbitrary context iterators instead. For back insert iterators (what `FMT_COMPILE` and `format_to` use) the body is formatted into a temporary container of the matching type; iterators that can't provide such storage get a runtime `format_error` when padding is requested, rather than a template error.

Testing: added compile-time tests for ranges and tuples with width; full suite passes.

---

This is not my native language, I'm using an LLM to restate and fix grammar in my replies. All the code was reviewed by me personally, and the description above was reviewed and approved by me as well. Looking forward to feedback.


vitaut (changes requested):
There are CI failures.

PR #4952, opened 2026-09-18
Avoid false positive from `-Wmaybe-uninitialized` in `named_arg_store` constructor

ilazaric:
Since GCC 16, in Debug builds, `-Wmaybe-uninitialized` falsely warns in certain situations where \
you convert a pointer to a not-yet-initialized variable to pointer-to-const.

Minimized version of what happens: https://godbolt.org/z/chrW4rGaP

Associated bug report: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=127470

Alternative fixup is to wrap the `named_arg_store` constructor with GCC pragmas disabling the warning.


ilazaric:
This can be observed from building module-test, which doesn't seem to be built in https://github.com/fmtlib/fmt/actions/runs/35233312141/job/105242745568

Locally I build with cmake command identical from the job, except with `-DCMAKE_COLOR_DIAGNOSTICS=ON -DCMAKE_CXX_STANDARD_REQUIRED=ON -G Ninja` , with cmake version 4.4.3, ninja version 1.11.1, and gcc trunk.

vitaut (changes requested):
Could you provide a godbolt repro of the actual (not minimized) warning?

ilazaric:
Not easily, only way I observed this is when going through modules.
Where it shows up: https://github.com/fmtlib/fmt/blob/main/test/module-test.cc#L174
This line causes the false positive warning, in Debug build, on recent gcc

ilazaric:
chatgpt butchered up fmt module build in godbolt: https://godbolt.org/z/666WrjaY6

PR #4953, opened 2026-09-18
Remove `fmt::path`

ilazaric:
The class was a temporary experiment,
and is just emitting warnings now,
because `std::filesystem::path::string()` was deprecated in C++26.


vitaut (changes requested):
Thanks for the PR. This was an experiment for the paper and we should just remove it.

ilazaric:
Updated to just kill `fmt::path` , built and ran `std-test` , works

PR #4954, opened 2026-09-18
Touchups to allow for `fmt::formatter<std::meta::info>`

ilazaric:
C++26 added reflection, and with it a new fundamental type `std::meta::info`. \
Trying to implement a formatter for it would run into a couple of constraints / enable-ifs \
that don't understand it. We touch them up accordingly, and add a test confirming it works.


include/fmt/core.h
```diff
@@ -18,6 +18,7 @@
 #  include <string.h>  // memcmp
 
 #  include <type_traits>  // std::enable_if
+#  include <version>      // __cpp_lib_reflection
```
ilazaric: We care about this to see if `std::is_reflection` from `<type_traits>` exists.
The macro is only in `<meta>` and `<version>` currently though.
I've pinged LWG on this, to add it to `<type_traits>`.

test/CMakeLists.txt
```diff
@@ -56,6 +56,13 @@ if (NOT (MSVC AND BUILD_SHARED_LIBS))
 endif ()
 add_fmt_test(ostream-test)
 add_fmt_test(compile-test)
+if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  check_cxx_compiler_flag(-freflection HAS_REFLECTION)
+  if (HAS_REFLECTION)
+    add_fmt_test(reflection-test)
```
ilazaric, line 62: Intentionally added a separate test, instead of rolling it into `compile-test.cc`.
This way there's clearer visibility into this being tested in CI, and we don't mess with `compile-test` compilation flags.

ilazaric:
Guarding the `<version>` include with `__has_include()` for the old clang build. \
Also noticed added test was slightly broken, was meant to format via `std::meta::identifier_of()`, \
this got lost during some offline debugging though. Fixed. \
Also also removed an unnecessary `remove_cvref_t<>`

vitaut:
Thanks for the PR but users should only implement formatters for types they own. Therefore, a compile error is expected here.

PR #4956, opened 2026-09-20
Make range formatting constexpr friendly

ilazaric:
Added test was not working previously.
Reason was two-fold, missing constexpr on a lot of ranges stuff, \
and compile.h using `std::back_inserter(std::string)`, \
while format.h was using `back_appender(basic_memory_buffer)`.


ilazaric:
During experimentation related to this I ran into a compiler bug, not sure atm if it affected this PR, so marking as draft until I think harder.

Issue #4767, opened 2026-05-02
unusual warnings in gcc 16 with LTO


oschonrock:
The crux of the warning is:
```
/home/oliver/psmem/build/gcc/release/_deps/fmt-src/include/fmt/format.h:1209:11: warning: writing 2 bytes into a region of size 0 [-Wstringop-overflow=]
 1209 |     memcpy(out, digits2(value), 2);
      |           ^
/home/oliver/psmem/build/gcc/release/_deps/fmt-src/include/fmt/format.h: In function ‘write_fixed’:
/home/oliver/psmem/build/gcc/release/_deps/fmt-src/include/fmt/format.h:2433:8: note: at offset [-11, -1] into destination object ‘buffer’ of size 11
 2433 |   Char buffer[digits10<UInt>() + 2];
      |        ^
```

Note this only happens since 

```
$ gcc --version
gcc (GCC) 16.1.1 20260430
````
which has just been pushed into the repos on Arch Linux. 

compiling fmtlib as static lib, and then linking with LTO enabled... 

```cmake

set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)

FetchContent_Declare(
  fmt
  GIT_REPOSITORY https://github.com/fmtlib/fmt.git
  GIT_TAG 12.1.0
)
FetchContent_MakeAvailable(fmt)

add_executable(app ...)
target_compile_features(app PRIVATE cxx_std_26)
target_link_libraries(app PRIVATE fmt )
```

and during link stage, I get these odd warnings

```bash
cmake -GNinja -S . -B ./build/gcc/release -DCMAKE_COLOR_DIAGNOSTICS=ON -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DCMAKE_BUILD_TYPE=release
....
[2/11] /usr/bin/g++  -I/home/oliver/psmem/build/gcc/release/_deps/fmt-src/include -O3 -DNDEBUG -std=c++20 -flto=auto -fno-fat-lto-objects -fdiagnostics-color=always -MD -MT _deps/fmt-build/CMakeFiles/fmt.dir/src/os.cc.o -MF _deps/fmt-build/CMakeFiles/fmt.dir/src/os.cc.o.d -o _deps/fmt-build/CMakeFiles/fmt.dir/src/os.cc.o -c /home/oliver/psmem/build/gcc/release/_deps/fmt-src/src/os.cc
[...]

PR #4957, opened 2026-09-21
Fix write_significand buffer size to avoid GCC 16 LTO -Wstringop-overflow (#4767)

MuskanKhatwani:
Description
- Increase the temporary buffer in write_significand to ensure there is enough headroom for backward writes performed by the digit-writing path under LTO.
- Change: digits10<UInt>() + 2 -> digits10<UInt>() + 3 in include/fmt/format.h.

Rationale
- GCC 16 with LTO reports -Wstringop-overflow in the floating-point formatting path because the backward-writing routine can touch significand_size + 1 bytes while the previous buffer allocation was tight in one edge case.
- Increasing the buffer by one element removes the off-by-one risk without changing observable behavior.

Verification
- Built on Windows (MSYS2 UCRT64) with:
  - gcc 16.2.0, g++ 16.2.0, cmake 4.4.3
  - cmake configured with -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON
- Example link-time warning excerpt observed during build (for reviewer context):

warning: writing 2 bytes into a region of size 0 [-Wstringop-overflow=]
include/fmt/format.h:1375:11: memcpy(out, digits2(value), 2);
note: at offset [-13, -2] into destination object 'buffer' of size 12
include/fmt/format.h:2659:8: Char buffer[digits10<UInt>() + 3];

Compatibility & Risk
- No API or behavioral changes.
- Minimal internal change to buffer sizing; low risk.

Suggested testing
- Build with GCC 16 + LTO (-flto) on CI (MSYS2/Ubuntu with GCC 16) to confirm no false-positive -Wstringop-overflow remains.
- Run unit tests with and without LTO to ensure no regressions.

Notes
- If maintainers prefer not to change allocation, an alternate approach is to gate a compiler warning suppression for affected toolchains. The buffer-size fix is simpler and safer.


MuskanKhatwani:
Hi, attaching the full verbose LTO build log (`lto-build-log.txt`) I captured locally.
Summary:
- Toolchain: MSYS2 UCRT64, gcc/g++ 16.2.0, CMake 4.4.3
- Built with `-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON`
- The log contains the link-time `-Wstringop-overflow` warnings observed during LTO.
Thanks for taking a look. I can also:
- produce a tiny standalone repro (CMake + source) to reproduce the warning, or
- run additional builds/tests and attach their logs — tell me which and I’ll add it.
[lto-build-log.txt](https://github.com/user-attachments/files/32520622/lto-build-log.txt)

